PR #2, opened 2016-02-02
Fixes asynchronous error handling when chrome.webstore.install throws…

iloire:
Before the fix, if you execute this code in Chrome without the extension installed (and without a trusted domain for an inline installation)

```
          JitsiMeetJS.createLocalTracks({
            resolution: '720',
            devices: ['video', 'desktop']
          }).then(tracks => {
            resolve();
          }).catch(err => {
            // this never gets called
          });
```
after the fix:
```
          JitsiMeetJS.createLocalTracks({
            resolution: '720',
            devices: ['video', 'desktop']
          }).then(tracks => {
            resolve();
          }).catch(err => {
              alert('this gets called and you can run your error handling code');
          });
```

### DIFF
--- a/JitsiTrackErrors.js
+++ b/JitsiTrackErrors.js
@@ -23,5 +23,7 @@ module.exports = {
     },
     UNSUPPORTED_RESOLUTION: "gum.unsupported_resolution",
     FIREFOX_EXTENSION_NEEDED: "gum.firefox_extension_needed",
+    CHROME_EXTENSION_INSTALLATION_ERROR:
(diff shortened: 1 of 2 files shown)


damencho:
I'm not able to reproduce the scenario mentioned with latest lib versions. I've uninstalled from chrome all my desktop sharing extensions and when clicking desktop sharing the code in the catch is always called. (in my config I have extension id which is not existing in webstore).
Can you give more details and whether you still reproduce the problem?

PR #12, opened 2016-02-11
Fix logic inversion in Settings.prototype.save

champtar:
This fixes "machine UID mismatch or empty" when using authenticated conference.
(machineUID come from Settings.userId)

Fixing this hide the fact that there is 2 different Settings object
coming from 2 calls to initJitsiConference (when using authenticated conference)

There is also another Settings module in jitsi-meet, with a userId ...

Signed-off-by: Etienne CHAMPETIER <champetier.etienne@gmail.com>


champtar:
duplicate of #15

PR #15, opened 2016-02-16
fix Settings.save to actually save data to localstrage

isymchych:


champtar:
hi @isymchych, @paweldomas 
I've provided this fix last week and even sent a mail on dev ML :(
https://github.com/jitsi/lib-jitsi-meet/pull/12

As noted in my commit message with authenticated conference there are 2 Settings object, so if localStorage doesn't work you're dead. Maybe we should refuse to work without localStorage?

paweldomas:
Hi @champtar ! I'm sorry for missing your PR and thanks for suggestions - we're going to fix those issues in the following weeks.

PR #25, opened 2016-02-24
Moves the code that trigger TRACK_REMOVED events for remote track to use peerconnection event instead of relying on TRACK_STOPPED event.

hristoterezov:


damencho:
Everything passes and seems ok in the context of jitsi-meet and torture tests. What bothers me if you do:
room = APP.conference._room.getLocalTracks()[1].stream.getTracks()[0].stop() (this is just stopping the local video track)
on the side of the second participant, then on the first one side we see again stopped and removed event. 
This is because in addTrack(https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L343) we add stop listener to tracks and remove them once they are stopped, shouldn't we leave this to the users of the library to decide whether they should remove the track? This way calling stop on a track in the library will result only stopped event.

hristoterezov:
We have added this for convenience for the users of the library a long time ago. I can't think of use case that the track is stopped and the user is going to use it in the conference. That way you can execute only the "stop" method instead of "stop" and then "removeTrack". My personal opinion is that we should keep this code otherwise it will be possible to have stopped streams in the conference which doesn't seems good.

However I haven't changed that code with my PR and I think the behavior is the same as before the PR. Have you noticed different behavior after the PR? If not I think this issue is not related to the PR and we can discuss it separately and change the code with another PR. WDYT?

PR #27, opened 2016-03-02
Share start muted policy if user became moderator

isymchych:
Modarator shares start muted policy with other participants. 
So when moderator left the room and new user became moderator he should also start sharing start muted policy.


damencho:
Too old :)

PR #28, opened 2016-03-04
Display local audio level on Safari using statistics provided by Temasys

isymchych:


damencho:
Let's add to whitelist @isymchych for auto testing.

damencho:
I've fixed the testing scripts and it is ok to test now.

damencho:
Still having problems with scripts. Fixed it and it is ok to test now by jenkins.

PR #31, opened 2016-03-08
Syncs mute/unmute calls.

damencho:
Returns Promise from mute/unmute and ignores requests while mute/unmuting is in progress. Moves mute/unmute methods to JitsiLocalTrack as said in API docs.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -2,6 +2,7 @@
 var JitsiTrack = require("./JitsiTrack");
 var RTCBrowserType = require("./RTCBrowserType");
 var JitsiTrackEvents = require('../../JitsiTrackEvents');
+var JitsiTrackErrors = require("../../JitsiTrackErrors");
 var RTCUtils = require("./RTCUtils");
 
 /**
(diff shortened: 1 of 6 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -28,26 +29,92 @@ function JitsiLocalTrack(stream, videoType,
...
 
 JitsiLocalTrack.prototype = Object.create(JitsiTrack.prototype);
 JitsiLocalTrack.prototype.constructor = JitsiLocalTrack;
 
 /**
+ * Mutes the track. Will reject the Promise if there is mute/unmute operation
+ * in progress.
+ * @returns {Promise}
+ */
+JitsiLocalTrack.prototype.mute = function () {
```
iloire: I see what you did this, but since ``mute`` and ``unmute`` look identical (except for a parameter), I was wondering if it is worthy to extract another function for the common logic?

damencho: Is it ok now? Thanks

iloire: Looks good. You could have made it part of the prototype to save you form having to do the context binding.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -98,8 +156,10 @@ JitsiLocalTrack.prototype._setMute = function (mute) {
                         }
                     }
 
-                    if(!stream)
+                    if(!stream) {
```
iloire: I have a concern about the call to

```
RTCUtils.obtainAudioAndVideoPermissions(streamOptions)
```
if it throws an error, you will have an uncaught error that won't neither resolve or reject your main promise. You can avoid that by adding a catch handler:

```
RTCUtils.obtainAudioAndVideoPermissions(streamOptions).then(function (streams) {
  // all the code here
}).catch(reject);
```

modules/RTC/RTC.js
```diff
@@ -188,15 +188,20 @@ RTC.prototype.getLocalVideoStream = function () {
...
         var stream = this.localStreams[i];
         if(stream.getType() !== "audio") {
             continue;
         }
-        stream._setMute(value);
+        // this is a Promise
+        mutes.push(stream.mute(value));
     }
+    // we return a Promise from all Promises so we can wait for their execution
+    return Promise.all(mutes);
```
iloire: I wish we could do:

```
const onlyAudioTracks = t => track.getType() === 'audio';
return Promise.all(this.localStreams.filter(onlyAudioTracks).map(t => track.mute()));
```
:)

damencho:
Jenkins, it is ok to test again.

PR #36, opened 2016-03-15
Monitor device list changes

isymchych:
* monitor device list changes and trigger DEVICES_LIST_CHANGED on `JitsiMeetJS`
* so now `JitsiMeetJS` is event bus
* provide single method `JitsiMeetJS.getDevicesList()` which returns available devices. If browser do not support `navigator.mediaDevices.enumerateDevices` or `MediaStreamTrack.getSource` it always returns empty lists.


tsareg:
This can be closed now, as another code for the same functionality was already merged in

PR #37, opened 2016-03-15
Adds CONFERENCE_MAX_USERS conference error.

hristoterezov:
Implements CONFERENCE_MAX_USERS conference error to be fired when the room already reached the maximum users limit.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -129,7 +129,7 @@ JitsiMeetJS.setLogLevel(JitsiMeetJS.logLevels.ERROR);
         - CHAT_ERROR - chat error happened
         - FOCUS_DISCONNECTED - focus error happened
         - FOCUS_DISCONNECTED - focus left the conference
-        - CONFERENCE_MAX_USERS - The maximum users limit is already reached
+        - CONFERENCE_MAX_USERS - The maximum users limit has been reached
     2. connection
         - PASSWORD_REQUIRED - passed when the connection to the server failed. You should try to authenticate with password.
         - CONNECTION_ERROR - indicates connection failures.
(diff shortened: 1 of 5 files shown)


doc/API.md
```diff
@@ -129,6 +129,7 @@ JitsiMeetJS.setLogLevel(JitsiMeetJS.logLevels.ERROR);
         - CHAT_ERROR - chat error happened
         - FOCUS_DISCONNECTED - focus error happened
         - FOCUS_DISCONNECTED - focus left the conference
+        - CONFERENCE_MAX_USERS - The maximum users limit is already reached
```
paweldomas: Probably it should be "The maximum users limit has been reached"

paweldomas: Please also have a look for other places where you wrote "limit is reached" or "have been reached"(should be "has been reached")

service/xmpp/XMPPEvents.js
```diff
@@ -120,6 +120,10 @@ var XMPPEvents = {
     /**
      * Indicates that the local sendrecv streams in local SDP are changed.
      */
-    SENDRECV_STREAMS_CHANGED: "xmpp.sendrecv_streams_changed"
+    SENDRECV_STREAMS_CHANGED: "xmpp.sendrecv_streams_changed",
+    /**
+     * Indicates that max users limit have been reached.
+     */
+    ROOM_MAX_USERS_ERROR: "xmpp.room_max_users_error"
```
paweldomas: This commit adds only one event, but on a side note all of them should be either sorted in alphabetical order or grouped in some meaningful way, cause now they are just scattered around randomly

PR #38, opened 2016-03-17
add VideoOperator utility which can save frames from HTMLVideoElement

isymchych:


gpolitis:
@isymchych I'm not sure if we want this in the lib-jitsi-meet because it's specific to PSNR testng. Do you think it would be possible to move it in jitsi-meet-torture (in its own VideoOperator.js file) and inject it when we run the test? @bgrozev @paweldomas @hristoterezov @yanas do you think we might use this functionality anywhere else in meet?

isymchych:
@gpolitis I've added that file to lib-jitsi-meet because we already have some helpers for torture there (`RTCUIHelper`). But its a good idea to move it directly to torture repository.

gpolitis:
Great! I'm going to wait for you to move the VideoOperator in jitsi-meet-torture and then close this PR.

isymchych:
moved script to https://github.com/jitsi/jitsi-meet-torture/pull/10

PR #40, opened 2016-03-18
Strophe cleanup

paweldomas:
This PR removes scenarios and methods which are not used in jitsi-meet.
Does not accept session-initiate from non-focus users.
Cleanups error handling logic.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -578,7 +578,7 @@ ChatRoom.prototype.getMemberRole = function (peerJid) {
 
 ChatRoom.prototype.setJingleSession = function(session){
     this.session = session;
-    this.session.room = this;
+    this.session.initialize(false /* initiator */, this /* room */);
 };
 
 
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -578,7 +578,6 @@ ChatRoom.prototype.getMemberRole = function (peerJid) {
 
 ChatRoom.prototype.setJingleSession = function(session){
     this.session = session;
-    this.session.initialize(false /* initiator */, this /* room */);
 };
 
 
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -28,8 +28,6 @@ function JingleSessionPC(me, sid, connection, service) {
     this.hadturncandidate = false;
     this.lasticecandidate = false;
 
-    this.statsinterval = null;
-
     this.reason = null;
 
     this.addssrc = [];
@@ -313,10 +311,6 @@ JingleSessionPC.prototype.terminate = function (reason) {
     this.state = 'ended';
     this.reason = reason;
     this.peerconnection.close();
-    if (this.statsinterval !== null) {
-        window.clearInterval(this.statsinterval);
-        this.statsinterval = null;
-    }
 };
 
 JingleSessionPC.prototype.active = function () {
@@ -791,10 +785,6 @@ JingleSessionPC.prototype.sendTerminate = function (reason, text) {
             $(document).trigger('ack.jingle', [self.sid, error]);
         },
         10000);
-    if (this.statsinterval !== null) {
-        window.clearInterval(this.statsinterval);
-        this.statsinterval = null;
-    }
 };
 
 /**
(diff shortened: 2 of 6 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -1186,29 +1134,45 @@ JingleSessionPC.prototype.notifyMySSRCUpdate = function (old_sdp, new_sdp) {
...
         logger.log('addition not necessary');
     }
 };
 
-JingleSessionPC.onJingleError = function (session, error)
-{
-    logger.error("Jingle error", error);
-}
+JingleSessionPC.prototype.newJingleErrorHandler = function(request, failureCb) {
+    return function (errResponse) {
```
hristoterezov: Why do you have to return new function every time? Couldn't we implement that code directly in newJingleErrorHandler? I think that way it's going to be simpler and faster.

paweldomas: Because what we want to have as a handler is a callback function(error). Apart from "error" it also needs "request" as a variable which will not be provided as callback argument. Also it optionally calls failureCb which would not be provided as one of callback arguments.

paweldomas: I will add a comment there which will explain how this works

modules/xmpp/JingleSessionPC.js
```diff
@@ -11,6 +11,13 @@ var XMPPEvents = require("../../service/xmpp/XMPPEvents");
 var RTCBrowserType = require("../RTC/RTCBrowserType");
 var RTC = require("../RTC/RTC");
 
+/**
+ * Constant tells how long we're going to wait for IQ response, before timeout
+ * error is  triggered.
+ * @type {number}
+ */
+const IQ_TIMEOUT = 10000;
```
hristoterezov: I'm not sure if "const" should work on all the browsers but maybe it's better to change it to "var" for code consistency.

modules/xmpp/JingleSessionPC.js
```diff
@@ -501,205 +302,46 @@ JingleSessionPC.prototype.setRemoteDescription = function (elem, desctype) {
...
+        }.bind(this),
         this.media_constraints
     );
 };
 
-JingleSessionPC.prototype.createdAnswer = function (sdp, provisional) {
+JingleSessionPC.prototype.createdAnswer = function (sdp, success, failure) {
     //logger.log('createAnswer callback');
-    var self = this;
+    const self = this;
```
hristoterezov: I'm not sure if "const" should work on all the browsers but maybe it's better to change it to "var" for code consistency.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1252,111 +873,45 @@ JingleSessionPC.prototype.notifyMySSRCUpdate = function (old_sdp, new_sdp) {
...
-                loss.video = (delta.video > 0) ? Math.ceil(100 * (lost.video - lastlost.video) / delta.video) : 0;
-                $(document).trigger('packetloss.jingle', [self.sid, loss]);
-            });
+JingleSessionPC.prototype.newJingleErrorHandler = function(request, failureCb) {
+    return function (errResponse) {
+
+        var error = { };
+
+        // Get XMPP error code and condition(reason)
+        const errorElSel = $(errResponse).find('error');
```
hristoterezov: I'm not sure if "const" should work on all the browsers but maybe it's better to change it to "var" for code consistency.

modules/xmpp/JingleSessionPC.js
hristoterezov: But if you use ".bind" you can provide those arguments, am I missing something?

modules/xmpp/ChatRoom.js
```diff
@@ -578,7 +578,7 @@ ChatRoom.prototype.getMemberRole = function (peerJid) {
 
 ChatRoom.prototype.setJingleSession = function(session){
     this.session = session;
-    this.session.room = this;
+    this.session.initialize(false /* initiator */, this /* room */);
```
hristoterezov: Maybe we should move "session.initialize" call somewhere else. I don't think it's related to the ChatRoom instance. Maybe in JitsiConference or in JingleSession?

paweldomas: Yes, will move that, thanks !

modules/xmpp/JingleSessionPC.js
bgrozev: Babel takes care of this (via the babel-plugin-transform-es2015-block-scoping plugin which is part of the ecma2015 preset). There is a difference between "var" and "const" (with the latter we will get "compile"-time errors and so detect bugs earlier), so it is not just a matter of consistency.

modules/xmpp/JingleSessionPC.js
paweldomas: Thanks for the suggestion ! But I'd rather to stay with newJingleErrorHandler which does the same thing as bind, but is slightly easier to use in my opinion as the optional argument can be skipped. Also I feel it more readable, but that's subjective.

PR #44, opened 2016-03-27
Fix early DTLS

lyubomir:
Commit 7cf5191 gives a comprehensive explanation (in the form of a source code comment) of the need for this PR. While the modifications should work with current revisions of Videobridge, the fix they represent will be activated by a Videobridge running https://github.com/jitsi/jitsi-videobridge/pull/191.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -388,6 +388,10 @@ JingleSessionPC.prototype.createdAnswer = function (sdp, success, failure) {
                 success,
                 self.newJingleErrorHandler(accept, failure),
                 IQ_TIMEOUT);
+        // XXX Videobridge needs WebRTC's answer (ICE ufrag and pwd, DTLS
+        // fingerprint and setup) ASAP in order to start the connection
+        // establishment.
+        self.connection.flush();
     };
     sdp.sdp = this.localSDP.raw;
     this.peerconnection.setLocalDescription(sdp,
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -366,28 +366,28 @@ JingleSessionPC.prototype.createdAnswer = function (sdp, success, failure) {
     var self = this;
     this.localSDP = new SDP(sdp.sdp);
     var sendJingle = function (ssrcs) {
-                var accept = $iq({to: self.peerjid,
-                    type: 'set'})
-                    .c('jingle', {xmlns: 'urn:xmpp:jingle:1',
-                        action: 'session-accept',
-                        initiator: self.initiator,
-                        responder: self.responder,
-                        sid: self.sid });
-                if (self.webrtcIceTcpDisable) {
-                    self.localSDP.removeTcpCandidates = true;
-                }
-                if (self.webrtcIceUdpDisable) {
-                    self.localSDP.removeUdpCandidates = true;
-                }
-                self.localSDP.toJingle(
-                    accept,
-                    self.initiator == self.me ? 'initiator' : 'responder',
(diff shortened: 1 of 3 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -339,31 +339,41 @@ JingleSessionPC.prototype.sendAnswer = function (success, failure) {
 
 JingleSessionPC.prototype.createdAnswer = function (sdp, success, failure) {
     //logger.log('createAnswer callback');
```
iloire: Should we delete unused comments?

modules/xmpp/SDP.js
```diff
@@ -293,8 +292,7 @@ SDP.prototype.toJingle = function (elem, thecreator) {
             this.rtcpFbToJingle(i, elem, '*');
 
             // XEP-0294
-            if (SDPUtil.find_line(this.media[i], 'a=extmap:')) {
-                lines = SDPUtil.find_lines(this.media[i], 'a=extmap:');
+            if ((lines = SDPUtil.find_lines(this.media[i], 'a=extmap:')).length) {
```
iloire: Would 
```
var lines = SDPUtil.find_lines(this.media[i], 'a=extmap:');
if (lines.length) {
```
be easier to reader?

modules/xmpp/SDP.js
```diff
@@ -200,9 +199,9 @@ SDP.prototype.toJingle = function (elem, thecreator) {
 
                 elem.up();
             }
-            if (SDPUtil.find_line(this.media[i], 'a=crypto:', this.session)) {
+            var crypto = SDPUtil.find_lines(this.media[i], 'a=crypto:', this.session);
+            if (crypto.length) {
```
iloire: Would this be a good candidate for a function extraction in the path for avoiding having functions with hundreds of lines?

modules/xmpp/SDP.js
```diff
@@ -166,16 +165,16 @@ SDP.prototype.toJingle = function (elem, thecreator) {
...
-            ssrc = SDPUtil.find_line(this.media[i], 'a=ssrc:').substring(7).split(' ')[0]; // take the first
+        if ((tmp = SDPUtil.find_line(this.media[i], 'a=ssrc:'))) {
+            ssrc = tmp.substring(7).split(' ')[0]; // take the first
         } else {
             ssrc = false;
         }
 
         elem.c('content', {creator: thecreator, name: mline.media});
-        if (SDPUtil.find_line(this.media[i], 'a=mid:')) {
+        if ((tmp = SDPUtil.find_line(this.media[i], 'a=mid:'))) {
```
iloire: Maybe we can find a more descriptive name for this variable?

modules/xmpp/JingleSessionPC.js
lyubomir: Thank you, @iloire! +1 from me on deleting unused comments in general. In the case, I'd rather not act on this comment of yours because:

- There are multiple "unused comments" throughout this and other files. If we are to remove them, I'd prefer to have that done in a separate dedicated PR.

- I'm not even 100% sure that this particular comment is truly unused. I guess there's a chance it was left for easy uncommenting.

modules/xmpp/SDP.js
lyubomir: I'm afraid I don't get what you mean here.

I'd rather have this merged as soon as possible because it fixes a particular problem: searching for one and the same substring in a string.

damencho:
Jenkins: it is ok to test now.

modules/xmpp/JingleSessionPC.js
iloire: (thumbsup)

PR #45, opened 2016-03-28
Implements attach method in JitsiConnection. Fixes token auth.

hristoterezov:


connection_optimization/external_connect.js
```diff
@@ -0,0 +1,50 @@
...
+    if (!window.XMLHttpRequest) {
+        error_callback(new Error("XMLHttpRequest is not supported!"));
+        return;
+    }
+
+    var xhttp = new XMLHttpRequest();
+
+    xhttp.onreadystatechange = function() {
+        if (xhttp.readyState == 4) {
+            if (xhttp.status == 200) {
```
damencho: Can we create some constants with documentation what does 4 or 200 mean.

PR #47, opened 2016-03-29
Fixes issue with muc join

hristoterezov:


damencho:
Jenkins: it is ok to test again

PR #48, opened 2016-03-29
Jibri recording modifications - work in progress.

yanas:


modules/xmpp/recording.js
```diff
@@ -1,19 +1,26 @@
...
     this.type = type;
-    this._isSupported = ((type === Recording.types.JIBRI)
-        || (type === Recording.types.JIRECON && !this.jirecon))? false : true;
+    this._isSupported
+        = ( type === Recording.types.JIRECON && !this.jirecon
+            || type === Recording.types.JIBRI && !this._isServiceAvailable)
+            ? false : true;
+
+    this._isServiceAvailable = false;
+
```
damencho: checking for type === Recording.types.JIBRI && !this._isServiceAvailable will always be the same as _isServiceAvailable is defined after it.

PR #49, opened 2016-03-30
Reduce MUC joined time

lyubomir:
Reduces the (TIME) from "strophe CONNECTED" to "muc joined" by explicitly flushing Strophe's connection.

Additionally, attempts to improve the performance of the JavaScript source code.

### DIFF
--- a/modules/xmpp/SDP.js
+++ b/modules/xmpp/SDP.js
@@ -107,13 +107,10 @@ SDP.prototype.mangle = function () {
                 if (rtpmap.name == 'CN' || rtpmap.name == 'ISAC')
                     continue;
                 mline.fmt.push(rtpmap.id);
-                newdesc += lines[j] + '\r\n';
-            } else {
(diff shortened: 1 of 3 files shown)


modules/xmpp/SDP.js
```diff
@@ -71,23 +71,20 @@ SDP.prototype.getMediaSsrcMap = function() {
...
-    Object.keys(medias).forEach(function (mediaindex) {
-        if (result)
-            return;
-        if (medias[mediaindex].ssrcs[ssrc]) {
-            result = true;
-        }
-    });
-    return result;
+    var media = this.getMediaSsrcMap();
+    for (var i = 0, length = media.length; i < length; ++i) {
```
hristoterezov: "media" is JS object and it seems that length property is undefined. Maybe we should use and return array in getMediaSsrcMap method. That way length won't be undefined.

PR #54, opened 2016-03-31
Recompile browserify bundle on file changes (watch mode)

iloire:


damencho:
Jenkins: add to whitelist

PR #57, opened 2016-04-04
Enable adaptive simulcast by default

bgrozev:


paweldomas:
checks passed here https://github.com/jitsi/jicofo/pull/63

PR #58, opened 2016-04-05
Global error handler and reporting those errors.

damencho:
Adds global error handler function and option to auto enable it. Reports unhandled exceptions catched by error handler through callstats.


damencho:
Jenkins: it is ok to test again

PR #60, opened 2016-04-06
Fixes callstats issue: Cannot send feedback to callstats after JitsiConference.leave

hristoterezov:
Now when the conference is left we are not sending callstats terminate event and we are clearing the callstats property of statistics module. That way all the checks are going to pass when sendFeedback method is called and since we haven’t send callstats terminate event the feedback will be sent. The terminate event should be sent automatically on window.unload event.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -102,7 +102,6 @@ JitsiMeetJS.setLogLevel(JitsiMeetJS.logLevels.ERROR);
         - AVAILABLE_DEVICES_CHANGED - notifies that available participant devices changed (camera or microphone was added or removed) (parameters - id(string), devices(JS object with 2 properties - audio(boolean), video(boolean)))
         - CONNECTION_STATS - New local connection statistics are received. (parameters - stats(object))
         - AUTH_STATUS_CHANGED - notifies that authentication is enabled or disabled, or local user authenticated (logged in). (parameters - isAuthEnabled(boolean), authIdentity(string))
-        - CONNECTION_CLOSED - Indicates that the connection to the conference has been closed. (Should be received after JitsiConference.leave())
 
     2. connection
         - CONNECTION_FAILED - indicates that the server connection failed.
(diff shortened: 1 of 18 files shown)


doc/API.md
```diff
@@ -102,6 +102,7 @@ JitsiMeetJS.setLogLevel(JitsiMeetJS.logLevels.ERROR);
         - AVAILABLE_DEVICES_CHANGED - notifies that available participant devices changed (camera or microphone was added or removed) (parameters - id(string), devices(JS object with 2 properties - audio(boolean), video(boolean)))
         - CONNECTION_STATS - New local connection statistics are received. (parameters - stats(object))
         - AUTH_STATUS_CHANGED - notifies that authentication is enabled or disabled, or local user authenticated (logged in). (parameters - isAuthEnabled(boolean), authIdentity(string))
+        - CONNECTION_CLOSED - Indicates that the connection to the conference has been closed. (Should be received after JitsiConference.leave())
```
iloire: What's the difference with ``CONNECTION_DISCONNECTED`` and when should we use one or the other?

modules/RTC/DataChannels.js
```diff
@@ -197,7 +197,13 @@ DataChannels.prototype._onXXXEndpointChanged = function (xxx, userResource) {
             jsonObject.colibriClass = (upper + 'EndpointChangedEvent');
             jsonObject[lower + "Endpoint"]
                 = (userResource ? userResource : null);
-            dataChannel.send(JSON.stringify(jsonObject));
+            //Tries to fix errors on peer connection close.
+            try {
+                dataChannel.send(JSON.stringify(jsonObject));
```
iloire: In which scenarios this would fail?

modules/RTC/DataChannels.js
```diff
@@ -197,7 +197,13 @@ DataChannels.prototype._onXXXEndpointChanged = function (xxx, userResource) {
...
+            } catch (e) {
+                // FIXME: Handle properly
```
iloire: How would the actual fix look like?
I am trying to understand why are we delaying the actual implementation with a temporary hack.

modules/RTC/DataChannels.js
hristoterezov: If you try to use the data channels and peer connection is closed.

modules/RTC/DataChannels.js
hristoterezov: If the peer connection is closed the user shouldn't use the data channels. But even if they are used that way they won't brake anything by stopping the execution. 

I was just thinking that we can maybe find more gentle solution in future. But this seems to me not directly related to the current PR.

doc/API.md
hristoterezov: CONNECTION_DISCONNECTED is fired when the status of JitsiConnection is changed to disconnected (normally that happens when you call JitsiConnection.disconnect). CONNECTION_CLOSED was related to JitsiConference object. But I removed it because when I thought again about it, I realized that it shouldn't be useful outside the library since it indicates that only specific part of JitsiConference.leave process was finished successfully.

damencho:
Jenkins, test this please

hristoterezov:
Jenkins, test this please

hristoterezov:
Jenkins, test this please

PR #65, opened 2016-04-08
Use roomname as callstats conference id

dwilson6:
This change allows for the conference id to match across multiple platforms (including a cordova app).  It also removes the assumption that the path contains the room name.


hristoterezov:
The current version uses the room name and the hostname. I think it's very similar to this PR. Does the current version works for you?

dwilson6:
It doesn't. Lib-jitsi-meet makes the assumption that your URL will contain the environment name and room name (and it includes the whole path). In a Cordova app there is no hostname so it doesn't match a browser client conference ID. I thought there was talk about switching to just the room name to match the videobridge conference ID? I could make the value for the conference ID be configurable and default to how it works now

bgrozev:
Can we keep the current naming scheme? I.e. something like
confID = config.hosts.domain + '/' + this.session.getName()

dwilson6:
Yeah, something like that would be fine.  It might be nice to be able to override the hostname part since the config.hosts.domain may be different than the server or environment name but defaulting it to config.hosts.domain seems like a good idea.  I will update this PR to work like you specified.

dwilson6:
The callstats conf id will now default to the xmpp domain from the config plus the room jid separated by a slash.

dwilson6:
Is there any reason for this to not be merged? It makes it easier to use lib-jitsi-meet in a packaged app form and maintain consistency of the callstats conference id.

bgrozev:
Sorry, I missed the latest changes on this one. Looks good to me as it it now, I see no reason not to merge.

Jenkins: ok to test.

PR #66, opened 2016-04-08
Add maxFps option for requesting video

dwilson6:
This PR makes it possible to set a maximum frame rate constraint when requesting a video stream.


bgrozev:
Jenkins: ok to test

PR #74, opened 2016-04-18
Fixes using options from correct locations

damencho:


modules/xmpp/moderator.js
```diff
@@ -109,7 +111,6 @@ Moderator.prototype.createConferenceIq =  function () {
     // Session Id used for authentication
     var sessionId = this.settings.getSessionId();
     var machineUID = this.settings.getUserId();
-    var options = this.xmppService.options;
```
paweldomas: Not a big deal, but we could have redirected 'options' var to 'this.options.connection'. I guess Boris aimed to simplify this previously: https://github.com/jitsi/lib-jitsi-meet/commit/1f0e71a842c310376c3c382bfd21861594b88cb3

PR #80, opened 2016-04-25
Fixes callstats wrtcFuncNames signalingError.

damencho:
Fixes the function name was signallingError now it is signalingError (single l).


bgrozev:
The callstats API documentation [uses a double-l spelling](http://www.callstats.io/api/#enumeration-of-wrtcfuncnames). If that's inaccurate we should ping them.
@vr000m ?

damencho:
When I initially introduced this logging I was testing it and seeing it in the callstats UI. So it was working with double l. But they seem to changed the API without changing the doc, I found it when started testing. The change is on the their test deployment and in production.

vr000m:
oops, we should have changed the documentation. it is now changed to signaling. @karthikbr82

PR #83, opened 2016-04-27
Lipsync

paweldomas:
Required to enable lipsync hack in Jicofo from [PR 77]. The goal of the changes here is to handle WebRTC media on MediaStreamTrack level instead of MediaStream. We were kind of doing that anyway because we were creating separate streams for audio and video. Now with lipsync we need to handle multiple tracks in one media stream which requires "track added"/"track removed" events handling. We found these events working as expected with Chrome only, so the lipsync is enabled only for Chrome clients. Other clients will still receive each audio and video tracks in separate streams.

[PR 77]: https://github.com/jitsi/jicofo/pull/77

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -705,7 +705,7 @@ ChatRoom.prototype.remoteStreamAdded = function(data, sid, thessrc) {
             && audiomuted[0]["value"] === "true")? true : false);
     }
 
-    this.eventEmitter.emit(XMPPEvents.REMOTE_STREAM_RECEIVED, data, sid, thessrc);
+    this.eventEmitter.emit(XMPPEvents.REMOTE_TRACK_ADDED, data, sid, thessrc);
 };
 
 /**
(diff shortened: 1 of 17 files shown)


modules/xmpp/ChatRoom.js
```diff
@@ -691,21 +692,28 @@ ChatRoom.prototype.removeListener = function (type, listener) {
...
+            mutedNode = filterNodeFromPresenceJSON(pres, "audiomuted");
+        } else if (mediaType === MediaType.VIDEO) {
+            mutedNode = filterNodeFromPresenceJSON(pres, "videomuted");
+        } else {
+            logger.warn("Unsupported media type: " + mediaType);
+            data.muted= null;
+        }
+
+        if (mutedNode) {
+            data.muted = !!(mutedNode.length > 0 &&
```
bgrozev: Why do we need to "!!" here? The expression in the parens already evaluates to a boolean.

paweldomas: It was a suggestion from IDE and I accepted since not JS expert, but yes it should be a boolean anyway

PR #88, opened 2016-05-02
Prepare for React Native

lyubomir:
This is a collection of commits which fix issues that I've stumbled upon while porting lib-jitsi-meet to React Native. They should not contain anything that is exclusively specific to and/or mentions React Native but are rather issues that may cause problems when executing in a JavaScript environment that's outside of a Web browser.

### DIFF
--- a/modules/xmpp/SDPUtil.js
+++ b/modules/xmpp/SDPUtil.js
@@ -4,7 +4,9 @@ var RTCBrowserType = require("../RTC/RTCBrowserType");
 
 SDPUtil = {
     filter_special_chars: function (text) {
-        return text.replace(/[\\\/\{,\}\+]/g, "");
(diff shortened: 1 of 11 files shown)


modules/xmpp/SDPUtil.js
```diff
@@ -4,7 +4,11 @@ var RTCBrowserType = require("../RTC/RTCBrowserType");
 
 SDPUtil = {
     filter_special_chars: function (text) {
-        return text.replace(/[\\\/\{,\}\+]/g, "");
+        // XXX Neither null, nor undefined "contain" special chars.
+        return
+            (typeof text === 'undefined')
```
hristoterezov: If text is `null`, `(typeof text === 'undefined')` will be false. Maybe we should add check for `null` too.

PR #90, opened 2016-05-03
Fix jibri presence initial state

yanas:


aaronkvanmeerten:
This all looks good to me.

PR #91, opened 2016-05-04
Fix off recording state handling

yanas:


modules/xmpp/recording.js
```diff
@@ -60,7 +63,12 @@ Recording.prototype.handleJibriPresence = function (jibri) {
         this.state = Recording.status.UNAVAILABLE;
     }
     else if (newState === "off") {
-        this.state = Recording.status.AVAILABLE;
+        if (!this.state
+            || this.state === "undefined"
```
paweldomas: by looking at the code I do not see the case where this.state could be "undefined" as it is translated into "Recording.status.UNAVAILABLE" at line 62

paweldomas: oh I see it could be set from setRecording callback, sorry

PR #93, opened 2016-05-05
Fixes issue with not completing promise for remove/add track and conference leave

hristoterezov:


modules/xmpp/JingleSessionPC.js
```diff
@@ -943,6 +943,11 @@ JingleSessionPC.prototype.addStream = function (stream, callback, ssrcInfo,
         return;
     }
 
+    if(this.modifySourcesQueue.paused) {
+        throw new Error("modifySourcesQueue paused");
```
gpolitis: We could add a `throw` to the JS doc of the method.

PR #94, opened 2016-05-06
Added ability to change output audio device for browsers that support it

tsareg:
1. Added new methods to JitsiMeetJS (and RTC and RTCUtils respectively)
 - isAudioOutputDeviceChangeAvailable
 - getAudioOutputDevice
 - setAudioOutputDevice

2. Added new events:
 - JitsiTrackEvents.TRACK_AUDIO_OUTPUT_CHANGED
 - RTCEvents.AUDIO_OUTPUT_DEVICE_CHANGED

3. Update sample from docs to give ability to select output devices

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -147,6 +147,31 @@ var LibJitsiMeet = {
     isDeviceChangeAvailable: function () {
         return RTC.isDeviceChangeAvailable();
     },
+    /**
+     * Returns true if changing the audio output of media elements is supported
+     * and false if not.
+     */
+    isAudioOutputDeviceChangeAvailable: function () {
+        return RTC.isAudioOutputDeviceChangeAvailable();
+    },
+    /**
+     * Returns currently used audio output device id, '' stands for default
+     * device
+     * @returns {string}
+     */
+    getAudioOutputDevice: function () {
+        return RTC.getAudioOutputDevice();
+    },
+    /**
+     * Sets current audio output device.
+     * @param {string} deviceId - id of 'audiooutput' device from
+     *      navigator.mediaDevices.enumerateDevices()
+     * @returns {Promise} - resolves when audio output is changed, is rejected
+     *      otherwise
+     */
+    setAudioOutputDevice: function (deviceId) {
+        return RTC.setAudioOutputDevice(deviceId);
+    },
     enumerateDevices: function (callback) {
         RTC.enumerateDevices(callback);
     },
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -140,19 +140,14 @@ var LibJitsiMeet = {
         return RTC.isDeviceListAvailable();
     },
     /**
-     * Returns true if changing the camera / microphone device is supported and
-     * false if not.
+     * Returns true if changing the input (camera / microphone) or output
+     * (audio) device is supported and false if not.
+     * @params {string} [deviceType] - type of device to change. Default is
+     *      undefined or 'input', 'output' - for audio output device change.
      * @returns {boolean} true if available, false otherwise.
      */
-    isDeviceChangeAvailable: function () {
-        return RTC.isDeviceChangeAvailable();
-    },
-    /**
-     * Returns true if changing the audio output of media elements is supported
-     * and false if not.
-     */
-    isAudioOutputDeviceChangeAvailable: function () {
-        return RTC.isAudioOutputDeviceChangeAvailable();
+    isDeviceChangeAvailable: function (deviceType) {
+        return RTC.isDeviceChangeAvailable(deviceType);
     },
     /**
      * Returns currently used audio output device id, '' stands for default
@@ -165,7 +160,7 @@ var LibJitsiMeet = {
     /**
      * Sets current audio output device.
      * @param {string} deviceId - id of 'audiooutput' device from
-     *      navigator.mediaDevices.enumerateDevices()
+     *      navigator.mediaDevices.enumerateDevices(), '' is for default device
      * @returns {Promise} - resolves when audio output is changed, is rejected
      *      otherwise
      */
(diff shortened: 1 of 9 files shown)


modules/RTC/JitsiTrack.js
```diff
@@ -310,4 +312,46 @@ JitsiTrack.prototype.getMSID = function () {
...
+            new Error('Audio output device change is not supported'));
+    }
+
+    return Promise.all(this.containers.map(function(element) {
+        return element.setSinkId(audioOutputDeviceId)
+            .catch(function (error) {
+                console.error('Failed to change audio output device on element',
+                    element, error);
+
+                // TODO: for some reason 'AbortError' is raised on video
```
tsareg: Actually found the reason why this happens!
The error happens when we assign MediaStream from webcam to video element with
```
element.src = URL.createObjectURL(stream);
```
if we do this through srcObject, everything works OK:
```
element.srcObject = stream;
```
Going to fix this next week.

tsareg: Actually found the reason why this happens - when media stream from GUM has only video tracks and no audio tracks. In RTCUtils handleLocalStream function we actually split stream obtained from GUM into separate audio and video streams. So I suggest just to ignore elements for local tracks at all in this case. And maybe even all 'video' elements can be ignored as audio communication is done via 'audio' elements?

modules/RTC/JitsiTrack.js
```diff
@@ -310,4 +312,46 @@ JitsiTrack.prototype.getMSID = function () {
...
+                // elements with local track blobs. Maybe this is something
+                // similar to https://goo.gl/TKLiqx. Ignoring this error for
+                // now. Spec says that "If the device identified by the given
+                // sinkId cannot be used due to a unspecified error, throw a
+                // DOMException whose name is AbortError."
+                // In any case, all audio communication is done via separate
+                // audio elements, so maybe it doesn't make sense to change
+                // sinkId for <video> elements at all.
+                if (!(self.isVideoTrack() && self.isLocal && self.isLocal() &&
```
tsareg: Actually found the reason why this happens - when media stream from GUM has only video tracks and no audio tracks. In RTCUtils handleLocalStream function we actually split stream obtained from GUM into separate audio and video streams. So I suggest just to ignore elements for local tracks at all in this case. And maybe even all 'video' elements can be ignored as audio communication is done via 'audio' elements?

modules/RTC/JitsiTrack.js
```diff
@@ -190,6 +190,36 @@ JitsiTrack.prototype.attach = function (container) {
     return container;
 };
 
+JitsiTrack.prototype.changeAudioOutput = function (audioOutputDeviceId) {
+    return Promise.all(this.containers.map(function(element) {
+        if (typeof element.setSinkId !== 'undefined') {
+            try {
+                return element.setSinkId(audioOutputDeviceId)
+                    .then(function () {
+                        console.log('Audio output device changed on element ' + element);
```
yanas: @paweldomas reminded me recently that we should try using the internal logger and not the console log in the library, so I'm just echoing his advice here.

modules/RTC/RTCUtils.js
```diff
@@ -24,6 +24,10 @@ var devices = {
     video: true
 };
 
+var audioOuputDeviceId = ''; // default device
+
+var featureDetectionVideoEl = document.createElement('video');
```
hristoterezov: Isn't it better to test on audio element since we are going to use the setSinkId method only on audio elements?

JitsiMeetJS.js
```diff
@@ -147,6 +147,31 @@ var LibJitsiMeet = {
     isDeviceChangeAvailable: function () {
         return RTC.isDeviceChangeAvailable();
     },
+    /**
+     * Returns true if changing the audio output of media elements is supported
+     * and false if not.
+     */
+    isAudioOutputDeviceChangeAvailable: function () {
```
hristoterezov: I think the names of the methods about input devices are confusing now. I think we should change them to "*InputDevice*" instead of "*Device*" or we should use them without changing the name for the audio output devices too by adding a parameter type ("output"/"input"). 

I think using the existing methods for output devices too is better solution. That way we won't change the existing interfaces. WDYT?

hristoterezov:
I think it's going to be useful for audio elements that are not handled by lib-jitsi-meet to expose a utility method for changing the output device of given element. WDYT?

tsareg:
@hristoterezov can you please explain

> I think it's going to be useful for audio elements that are not handled by lib-jitsi-meet to expose a utility method for changing the output device of given element. WDYT?

HTMLMediaElement.setSinkId() (https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/setSinkId) does actual that - change the output device of given element. Do you propose to create some kind of a wrapper around it? What should it do then?

modules/RTC/RTCUtils.js
tsareg: Yes, makes sense. I've did this before I discovered the problems with video elements.

JitsiMeetJS.js
tsareg: Yep, I will remove 'isAudioOutputDeviceChangeAvailable' from the JitsiMeetJS object.
Instead will make 'isDeviceChangeAvailable' to accept optional argument 'deviceType'. If value of this argument will be 'input' or undefined, it will work as before, in case of 'output' value - will use internal RTC.isAudioOutputDeviceChangeAvailable method.

tsareg:
Updated the PR

hristoterezov:
@tsareg My proposal is to implement a wrapper. Now the wrapper can do what setSinkId is doing but in future we can handle the different browser implementations of setSinkId there. But that's not very important since we have only the chromes implementation for now.

gpolitis:
Add to whitelist.

gpolitis:
Jenkins: it's ok to test.

PR #106, opened 2016-05-19
Solves the issue with not reported JS errors

hristoterezov:


modules/RTC/RTCUtils.js
```diff
@@ -753,10 +759,12 @@ var RTCUtils = {
                 });
             } else {
                 try {
-                    logger.error('Browser does not appear to be WebRTC-capable');
+                    logger.error(
+                        'Browser does not appear to be WebRTC-capable');
                 } catch (e) {
                 }
```
lyubomir: @hristoterezov, what is the purpose of the empty catch around logger.error?

Is the idea to ensure reject gets invoked in case logger.error throws? If that's the case, wouldn't a finally be more appropriate?

modules/util/GlobalOnErrorHandler.js
```diff
@@ -0,0 +1,82 @@
...
+    },
+    /**
+     * Calls the global error handler if there is one.
+     * @param error the error that is going to be passed to the error handler
+     */
+    callErrorHandler: function (error) {
+        var errHandler = window.onerror;
+        if(!errHandler)
+            return;
+        errHandler.call(null, null, null, null, null, error);
```
lyubomir: OK, that's:
errorHandler = window.onerror; errHandler.onerror.call(null, ...)
and above there ways
oldOnErrorHandler = window.onerror; oldOnErrorHandler(...).

Given that you're providing null as the first argument of Function.prototype.call(), is there are reason for the inconsistency in style?

modules/xmpp/JingleSessionPC.js
```diff
@@ -405,7 +412,8 @@ JingleSessionPC.prototype.setLocalDescription = function (sdp, success,
             logger.error('setLocalDescription failed', error);
             if (failure)
                 failure(error);
-            self.room.eventEmitter.emit(XMPPEvents.CONFERENCE_SETUP_FAILED);
+            self.room.eventEmitter.emit(XMPPEvents.CONFERENCE_SETUP_FAILED,
+                error);
```
lyubomir: So in this file there are now 2 room.eventEmitter.emit(XMPPEvents.CONFERENCE_SETUP_FAILED with error and 1 without error? Is that by design?

lyubomir:
@hristoterezov, I'll rebase this to the latest master because at least one of the merges didn't go right and a part of the source code is messed up.

lyubomir:
Jenkins, test this please.

modules/RTC/RTCUtils.js
hristoterezov: @lyubomir I haven't added the try catch in this commit. Maybe I have added it a long time ago in the code but I don't remember why. I think logger.error shouldn't throw anything and we should remove the try catch but I don't think it's related to the current PR. WDYT?

hristoterezov:
@lyubomir Thanks for the rebase! I'm sorry I haven't noticed it!

modules/util/GlobalOnErrorHandler.js
hristoterezov: Fixed!

modules/xmpp/JingleSessionPC.js
hristoterezov: Fixed! Now we are always emitting error for that event.

hristoterezov:
Jenkins: it's ok to test.

modules/xmpp/JingleSessionPC.js
```diff
@@ -859,14 +859,30 @@ JingleSessionPC.prototype._modifySources = function (successCallback, queueCallb
...
-    this.peerconnection.setRemoteDescription(new RTCSessionDescription({type: 'offer', sdp: sdp.raw}),
+    /**
+     * Implements a failure callback which reports an error message and an
+     * optional error through (1) logger, (2) GlobalOnErrorHandler, and (3)
+     * queueCallback.
+     *
+     * @param {string} errmsg the error messsage to report
+     * @param {*} error an optional error to report in addition to errmsg
+     */
+    function reportError(errmsg, err) {
```
hristoterezov: @lyubomir I can move that function to Utils.js and use it for every error. WDYT?

PR #112, opened 2016-05-24
Traceable PeerConnection improvements.

gpolitis:
- Adds a transformation history that keeps track of the SDP modifications.
- Makes it possible to configure the number of simulcast layers through config.js

### DIFF
--- a/modules/xmpp/TraceablePeerConnection.js
+++ b/modules/xmpp/TraceablePeerConnection.js
@@ -6,7 +6,7 @@ var XMPPEvents = require("../../service/xmpp/XMPPEvents");
 var transform = require('sdp-transform');
 var RandomUtil = require('../util/RandomUtil');
 
-var SIMULCAST_LAYERS = 3;
+var DEFAULT_NUM_OF_SIMULCAST_LAYERS = 3;
 
 function TraceablePeerConnection(ice_config, constraints, session) {
     var self = this;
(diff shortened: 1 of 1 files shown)


modules/xmpp/TraceablePeerConnection.js
```diff
@@ -593,26 +658,62 @@ TraceablePeerConnection.prototype.createAnswer
...
 
             if (!RTCBrowserType.isFirefox())
             {
                 answer = self.ssrcReplacement(answer);
-                self.trace('createAnswerOnSuccess::mungeLocalVideoSSRC',
-                    dumpSDP(answer));
+
+                if (self.history) {
+                    // Add a transformation in the point.
+                    point.transformations.push({
```
hristoterezov: Shouldn't you define "point" here?

bgrozev: Or rather define it once -- all these "var point"s are in the same scope.

gpolitis: Thanks @hristoterezov, @bgrozev ! I'll fix this with a few more changes that I'm preparing.

lyubomir: Yup, as @bgrozev said, the multiple var declarations are automatically hoisted by JS into one to the top so it doesn't matter (to JS) where the 'var point's  are and a single one anywhere in the function is enough.

PR #113, opened 2016-05-24
Removes fake video tracks from the code

hristoterezov:
We don’t need the fake video tracks anymore. We are always receiving tracks even if the track is muted.


tsareg:
Should FAKE_VIDEO_TRACK_CREATED event be removed from RTCEvents then?

hristoterezov:
@tsareg Sure! Thanks for noticing that!

PR #114, opened 2016-05-25
Remove repetitions

lyubomir:
There were at least 3 kinds of repetitions that caused inefficiency and
burdened debugging:

1. Complex if conditions in which the value of one expression was
practically dependent upon the other so it was enough to check 1 of the
2 exressions.

2. getStatValue was called multiple times within the same function to
retrieve the value with 1 and the same key. That was wasteful, harder to
read and hampered debugging (on React Native).

3. getStatValue which is called multiple times within 1 second was
checking the RTCBrowserType of the executing browser. Since the latter
is very unlikely to change during a session i.e. at runtime, I
refactored the source code to perform as many of the checks as possible
just once.

Additionally, getStatValue used to throw an error if the executing
browser is unsupported but that throw would happen in the callback of
RTCPeerConnection#getStats. Hence, (a) it would be repeated multiple
times within 1 second which was unnecessarily inefficient and
repetitive and (b) it would come much later after the startup of the
application. It makes sense to discover the lack of support for the
executing browser as soon as possible and log an error once.

### DIFF
--- a/modules/statistics/RTPStatsCollector.js
+++ b/modules/statistics/RTPStatsCollector.js
@@ -503,29 +503,24 @@ StatsCollector.prototype.processStatsReport = function () {
 
         if(now.type == 'googCandidatePair')
         {
-            var ip, type, localIP, active;
+            var ip, type, localip, active;
             try {
                 ip = getStatValue(now, 'remoteAddress');
                 type = getStatValue(now, "transportType");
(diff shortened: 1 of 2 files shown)


lyubomir:
Jenkins, test this please.

lyubomir:
Jenkins, test this please.

PR #115, opened 2016-05-25
Correctly subclass JingleSession

legastero:
The previous method was clobbering the JingleSession prototype,
preventing the creation of additional subclasses.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -1,6 +1,7 @@
 /* jshint -W117 */
 
 var logger = require("jitsi-meet-logger").getLogger(__filename);
+var util = require("util");
 var JingleSession = require("./JingleSession");
 var TraceablePeerConnection = require("./TraceablePeerConnection");
 var MediaType = require("../../service/RTC/MediaType");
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -56,9 +57,8 @@ function JingleSessionPC(me, sid, peerjid, connection,
     // stable and the ice connection state is connected.
     this.modifySourcesQueue.pause();
 }
-//XXX this is badly broken...
-JingleSessionPC.prototype = JingleSession.prototype;
-JingleSessionPC.prototype.constructor = JingleSessionPC;
+
+util.inherits(JingleSessionPC, JingleSession);
```
lyubomir: Is that the first use of util.inherits in lib-jitsi-meet? If that's the case why do it for JingleSession(PC) only and not do it everywhere we extend classes in order to not introduce inconsistencies?

tsareg:
@lyubomir, you have already added babel to lib-jitsi-meet. So maybe we should use ES6 classes and extends instead?

modules/xmpp/JingleSessionPC.js
tsareg: @lyubomir, you have already added babel to lib-jitsi-meet. So maybe we should use ES6 classes and extends instead?

hristoterezov: I don't see babel in the current version of library. Am I missing something?

legastero:
Thanks for the review feedback @lyubomir! I've changed things to be consistent with the other subclassing instances that I've seen in the lib.

modules/xmpp/JingleSessionPC.js
lyubomir: @tsareg and @hristoterezov, I added Babel + ES2015 in a PR while trying to fix a failing automatic tests. Eventually, I manually rewrote my modifications to ES5 (because it's hard for me to engage in a discussion with the team on the subject of ES6 from Sydney).

PR #116, opened 2016-05-25
Add detection support for Edge

legastero:
Datachannels and statistics are disabled until the underlying support has been fully shimmed.

### DIFF
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -84,9 +84,10 @@ RTC.obtainAudioAndVideoPermissions = function (options) {
 };
 
 RTC.prototype.onIncommingCall = function(event) {
-    if(this.options.config.openSctp)
+    if(this.options.config.openSctp && !RTCBrowserType.usesORTC()) {
         this.dataChannels = new DataChannels(event.peerconnection,
             this.eventEmitter);
+    }
     // Add local Tracks to the ChatRoom
     this.localTracks.forEach(function(localTrack) {
         var ssrcInfo = null;
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -61,6 +61,9 @@ var rawEnumerateDevicesWithCallback = navigator.mediaDevices
 // http://w3c.github.io/mediacapture-main/#event-mediadevices-devicechange
 // TODO: check MS Edge
 var isDeviceChangeEventSupported = false;
+if (RTCBrowserType.isEdge()) {
+    isDeviceChangeEventSupported = true;
+}
 
 var rtcReady = false;
 
@@ -389,6 +392,23 @@ function wrapGetUserMedia(getUserMedia) {
   };
 }
 
+function wrapGetUserMediaPromise(getUserMedia) {
+    return function (constraints, successCallback, errorCallback) {
+        getUserMedia(constraints).then(function (stream) {
+            maybeApply(successCallback, [stream]);
+            if (!getUserMediaStatus.initialized) {
+                getUserMediaStatus.initialized = true;
+                getUserMediaStatus.callbacks.forEach(function (callback) {
+                    callback();
+                });
+                getUserMediaStatus.callbacks.length = 0;
+            }
+        }).catch(function (error) {
+            maybeApply(errorCallback, [error]);
+        });
+    };
+}
+
 /**
  * Execute function after getUserMedia was executed at least once.
  * @param {Function} callback function to execute after getUserMedia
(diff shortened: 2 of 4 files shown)


bgrozev:
jenkins: add to whitelist

modules/RTC/RTCUtils.js
```diff
@@ -61,6 +61,9 @@ var rawEnumerateDevicesWithCallback = navigator.mediaDevices
 // http://w3c.github.io/mediacapture-main/#event-mediadevices-devicechange
 // TODO: check MS Edge
 var isDeviceChangeEventSupported = false;
+if (RTCBrowserType.isEdge()) {
+    isDeviceChangeEventSupported = true;
```
lyubomir: Aren't though 3 lines like var isDeviceChangeEventSupported = RTCBrowserType.isEdge(); ?

modules/RTC/RTCUtils.js
```diff
@@ -61,6 +61,9 @@ var rawEnumerateDevicesWithCallback = navigator.mediaDevices
 // http://w3c.github.io/mediacapture-main/#event-mediadevices-devicechange
 // TODO: check MS Edge
```
lyubomir: Check MS Edge for what? MS Edge is ifed right bellow.

modules/RTC/RTCUtils.js
```diff
@@ -389,6 +392,23 @@ function wrapGetUserMedia(getUserMedia) {
   };
 }
 
+function wrapGetUserMediaPromise(getUserMedia) {
```
lyubomir: Wrap getUserMedia Promise why? Wouldn't a jsdoc help here?

modules/RTC/RTC.js
```diff
@@ -84,9 +84,10 @@ RTC.obtainAudioAndVideoPermissions = function (options) {
 };
 
 RTC.prototype.onIncommingCall = function(event) {
-    if(this.options.config.openSctp)
+    if(this.options.config.openSctp && !RTCBrowserType.usesORTC()) {
```
lyubomir: If the executing browser uses ORTC what? Something's missing? Something doesn't work? I expect a comment explaining the exclusion to come handy.

Can we detect it by feature instead of by checking the browser type?

modules/RTC/RTCUtils.js
```diff
@@ -690,6 +711,48 @@ var RTCUtils = {
...
+                        }
+                        id = tracks[0].id;
+                    }
+                    return SDPUtil.filter_special_chars(id);
+                };
+                this.getVideoSrc = function (element) {
+                    if (!element) {
+                        return null;
+                    }
+                    return element.srcObject;
```
lyubomir: Aren't these 4 lines like return element ? element.srcObject : null; ?

modules/RTC/RTCUtils.js
```diff
@@ -690,6 +711,48 @@ var RTCUtils = {
...
+                    var id = stream.id;
+                    if (!id) {
+                        var tracks = stream.getVideoTracks();
+                        if (!tracks || tracks.length === 0) {
+                            tracks = stream.getAudioTracks();
+                        }
+                        id = tracks[0].id;
+                    }
+                    return SDPUtil.filter_special_chars(id);
+                };
```
lyubomir: That getStreamID is a 100% copy of the Firefox case. It could be extracted into a function of its own to be reused by both Firefox and Edge.

modules/statistics/statistics.js
```diff
@@ -54,6 +55,10 @@ function Statistics(xmpp, options) {
 Statistics.audioLevelsEnabled = false;
 
 Statistics.prototype.startRemoteStats = function (peerconnection) {
+    if (RTCBrowserType.usesORTC()) {
```
lyubomir: Again, what's missing or not functioning as expected? Can it be feature detected?

lyubomir:
Jenkins, test this please.

tsareg:
```javascript
navigator.mediaDevices.addEventListener('devicechange', function () {
    RTCUtils.enumerateDevices(onMediaDevicesListChanged);
});
```

As it's now possible to test, I think that instead of calling ```RTCUtils.enumerateDevices``` we should call ```rawEnumerateDevicesWithCallback``` as in ```pollForAvailableMediaDevices()```. This is because ```RTCUtils.enumerateDevices``` will wait for ```afterUserMediaInitialized``` which not may happen at all, while ```rawEnumerateDevicesWithCallback``` will be executed in any case.

saghul:
We landed (some) Edge support in a different manner. Closing. Please holler if this still applies.

PR #103, opened 2016-05-17
Switch local audio and video tracks when list of available devices changes

tsareg:
1. New methods added
 - JitsiMediaDevices.prototype.isDevicePermissionGranted(type)
 - JitsiLocalTrack.prototype.isEnded()
 - JitsiLocalTrack.prototype.getDeviceId()

2. Fixed potential memory leaks with not clearing event handlers

3. Updated API document

PR #117, opened 2016-05-26
Do not subscribe each JitsiTrack to RTCEvents.AUDIO_OUTPUT_DEVICE_CHANGED event

tsareg:
Instead we subscribe once for this event in RTC, and then go through all our local and remote audio tracks.

Also restored "dispose" functionality as it was before changes done in https://github.com/jitsi/lib-jitsi-meet/pull/103

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -36,6 +36,7 @@ function JitsiLocalTrack(stream, track, mediaType, videoType, resolution,
     this.resolution = resolution;
     this.deviceId = deviceId;
     this.startMuted = false;
+    this.disposed = false;
     //FIXME: This dependacy is not necessary.
     this.conference = null;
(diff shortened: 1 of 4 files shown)


hristoterezov:
@tsareg I think that way we are going to miss the tracks that are not added to the conference. Maybe we should think about another solution that is not adding every track as a listener but it sets the output for every track or maybe we can increase the limit.

tsareg:
@hristoterezov Good point! Now I remember why I did it like I did.
But am I right if I say that "tracks that are not added to the conference" - are only local tracks? If yes, then we can subscribe to event only JitsiLocalTrack instances, while handle JitsiRemoteTrack instances like I did in this commit.

hristoterezov:
@tsareg Yes. I think that will work. And that way we should reduce the listeners a lot.

tsareg:
@hristoterezov I've updated the code according to solution above

PR #118, opened 2016-05-26
Handle jibri errors and busy status

yanas:


hristoterezov:
Why we are not reporting the errors as Error objects anymore?

yanas:
Because I think we need to either throw them as error if they're Error objects or treat them as state value. We currently handle them as a special state, so I thought in this context it makes more sense to pass them as objects without the Error class wrapper.

PR #119, opened 2016-05-27
Handle create local tracks errors better

tsareg:
Following changes are introduced:

- JitsiMeetJS.createLocalTracks is now rejected with JitsiTrackError instance, not a string
- Introduced JitsiTrackError class that is inherited from base JavaScript Error class, so ```"name"```, ```"message"``` and ```"stack"``` properties are available. For GUM-related errors, exposes additional ```"gum"``` property, which is an object with following properties:
  - error - original GUM error
  - constraints - GUM constraints object used for the call
  - devices - array of devices requested in GUM call (possible values - "audio", "video", "screen", "desktop")
- Changed the format of how JitsiTrackError instances are logged to callstats
- JitsiTrackError is exposed to global scope as ```JitsiMeetJS.JitsiTrackError```
- callstats are now initialized as soon as we have JingleSessionPC instance, do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never happen in case if user doesn't have or denied permission to both camera and microphone.
-  new JitsiTrackErrors types introduced:
  - JitsiTrackErrors.CHROME_EXTENSION_GENERIC_ERROR
  - JitsiTrackErrors.PERMISSION_DENIED
  - JitsiTrackErrors.NOT_FOUND
  - JitsiTrackErrors.CONSTRAINT_FAILED
- API.md document updated with changes

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -210,30 +210,23 @@ var ScreenObtainer = {
                                 streamCallback, failCallback);
                         }, 500);
                     },
-                    function (arg) {
-                        var msg = "Failed to install the extension from "
-                            + getWebStoreInstallUrl(self.options);
-
-                        logger.log(msg, arg);
-
-                        failCallback(new JitsiTrackError(
-                            JitsiTrackErrors
-                                .CHROME_EXTENSION_INSTALLATION_ERROR,
-                            msg
-                        ));
-                    }
+                    handleExtensionInstallationError
                 );
             } catch(e) {
-                var msg = "Failed to install the extension from "
-                    + getWebStoreInstallUrl(self.options);
-
-                logger.log(msg, e);
-
-                failCallback(new JitsiTrackError(
-                    JitsiTrackErrors.CHROME_EXTENSION_INSTALLATION_ERROR,
-                    msg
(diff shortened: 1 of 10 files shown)


modules/RTC/ScreenObtainer.js
```diff
@@ -208,23 +211,28 @@ var ScreenObtainer = {
...
-                });
+                var msg = "Failed to install the extension from "
+                    + getWebStoreInstallUrl(self.options);
+
+                logger.log(msg, e);
+
+                failCallback(new JitsiTrackError(
+                    JitsiTrackErrors.CHROME_EXTENSION_INSTALLATION_ERROR,
+                    msg
+                ));
```
lyubomir: Aren't these 9 lines the same as the 9 lines that are 3 lines above them and that are conveniently wrapped into an anonymous function?

tsareg: Thanks for pointing this out, will fix.

tsareg:
@lyubomir did changes according to your review

lyubomir:
I've 'cloned' this PR into https://github.com/jitsi/lib-jitsi-meet/pull/124 so that I have (1) push access and (2) automatic PR testing.

PR #121, opened 2016-05-31
Adds all SRD/SLD to modifySourcesQueue and removes modifySourcesQueue.pause calls

hristoterezov:


modules/xmpp/JingleSessionPC.js
```diff
@@ -307,118 +278,15 @@ JingleSessionPC.prototype.acceptOffer = function(jingleOffer,
...
-            self.room.eventEmitter.emit(XMPPEvents.CONFERENCE_SETUP_FAILED);
-        }
-    );
-    // Some checks for STUN and TURN candiates present in local SDP
-    // Eventually could be removed as we don't really care
-    var cands = SDPUtil.find_lines(this.localSDP.raw, 'a=candidate:');
-    for (var j = 0; j < cands.length; j++) {
-        var cand = SDPUtil.parse_icecandidate(cands[j]);
-        if (cand.type == 'srflx') {
-            this.hadstuncandidate = true;
```
paweldomas: is it related to the problem which this PR fixes or do you remove it because it's not used ?

modules/xmpp/JingleSessionPC.js
```diff
@@ -307,118 +278,15 @@ JingleSessionPC.prototype.acceptOffer = function(jingleOffer,
...
-        } else if (cand.type == 'relay') {
-            this.hadturncandidate = true;
-        }
-    }
+    this.jingleOfferIq = jingleOfferIq;
```
paweldomas: I don't like how offer IQ processing is moved to modify sources method which will looks even more hacky - a combination of actions submitted from different locations. It makes the code unreadable. Maybe we should refactor "modify sources queue" to queue tasks and the actual logic would be passed to the queue as a function here. Same should be done for add/remove sources/mute etc. But I will have a deeper look at that later if that makes sense

modules/xmpp/JingleSessionPC.js
hristoterezov: Because it was not used mainly and this method doesn't exist anymore.

modules/xmpp/JingleSessionPC.js
hristoterezov: I don't think moving everything to the queue is more hacky. I think it's even more reliable than before. Now we have included the the session initiate IQs there. Before this change add source and remove source IQs where already handled by the queue and it was strange that only the session initiate isn't included there. Also if we keep everything in the queue we don't have to care about the signaling state, etc and we don't have to pause/resume the queue. So I think it's more readable this way.

Although I agree with you that the code in _modifySources method is not the best one. It is a good idea to refactor the  _modifySources method. I don't like how we are passing the operations to it by setting a property. When we are passing arguments to this.modifySourcesQueue.push we can actually pass object as first argument and after that we can pass callback( like the code in the next line but success callback can be replaced with object and we can call success callback in the function passed as second argument). Actually this should be the proper way of using that queue according  to their documentation. But we can implement that in another PR.

modules/xmpp/JingleSessionPC.js
```diff
@@ -307,118 +278,15 @@ JingleSessionPC.prototype.acceptOffer = function(jingleOffer,
...
+    this.modifySourcesQueue.push(success, function (error) {
+        if(!error)
+            return;
+        if (failure)
+            failure(error);
+        JingleSessionPC.onJingleFatalError(this, error);
+        failure(error);
```
paweldomas: Looks like "failure" is called 2nd time here

paweldomas:
Can you have a look at this PR https://github.com/jitsi/lib-jitsi-meet/pull/123 and see if I've merged it correctly ?

modules/xmpp/JingleSessionPC.js
paweldomas: I agree that having SRD/SLD cycle on the queue is a good thing, but just don't like how it is and was implemented. In my opinion that is not very readable, because mixes many different operations into one method which come from different places around the code. Agree that we can fix that another time, but I'm just not a big fan of piling onto hacky solutions instead of fixing them asap.

modules/xmpp/JingleSessionPC.js
paweldomas: Fixed in rebased PR #123

modules/xmpp/JingleSessionPC.js
paweldomas: Please put this kind of changes in separate commit in the future, cause it's confusing

paweldomas:
merged through https://github.com/jitsi/lib-jitsi-meet/pull/123

PR #127, opened 2016-06-07
getUserMedia: prefer the "user" camera if we have not specified the exact device id for mobile devices

tsareg:
This should make us to select "front" camera for mobile apps by default. To really work with react-native, needs support for "facingMode" constraint in react-native-webrtc, which will be done by @paweldomas and @lyubomir

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -119,6 +119,12 @@ function getConstraints(um, options) {
             constraints.video.optional.push({
                 sourceId: options.cameraDeviceId
             });
+        } else {
+            // by default prefer the "user" camera if we have not specified the
+            // exact device id for mobile devices
+            if (RTCBrowserType.isReactNative() || RTCBrowserType.isAndroid()) {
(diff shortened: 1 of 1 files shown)


modules/RTC/RTCUtils.js
```diff
@@ -119,6 +119,12 @@ function getConstraints(um, options) {
             constraints.video.optional.push({
                 sourceId: options.cameraDeviceId
             });
+        } else {
+            // by default prefer the "user" camera if we have not specified the
+            // exact device id for mobile devices
+            if (RTCBrowserType.isReactNative() || RTCBrowserType.isAndroid()) {
+                constraints.video.facingMode = "user";
```
lyubomir: Why is the facingMode constraint applied on React Native and Android only? Don't we want a user-facing camera otherwise? Or maybe the constraint breaks Chrome or Firefox?

tsareg: Well, it's not yet supported in Chrome, but it doesn't break anything. So probably you are right, let's  make it for all browsers.

PR #128, opened 2016-06-08
Reduce the dependency on HTML elements

lyubomir:
Since React Native doesn't have HTML elements, depend on HTML element properties and don't necessarily depend on them being properties on actual HTML elements but rather assume plain JavaScript objects.

(This PR deals with the dependencies on HTML elements only in the context of JitsiTrack#attach and #detach because that's what we need now in React Native. Of course, in the future we'll very likely want to deal with other dependencies on HTML elements as well.)

### DIFF
--- a/modules/RTC/JitsiTrack.js
+++ b/modules/RTC/JitsiTrack.js
@@ -184,25 +184,25 @@ JitsiTrack.prototype.attach = function (container) {
 };
 
 /**
- * Removes the track from the passed HTML container.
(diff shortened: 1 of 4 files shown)


lyubomir:
This PR will very likely break the automatic PR testing because of the value returned by RTCUIHelper#getVideoId. The very tests i.e. jitsi-meet-torture will (very likely) need modifications to deal with the new return value.

lyubomir:
Jenkins, test this please.

PR #130, opened 2016-06-09
Report DOMException when audio output device change fails as gUM error to callstats

tsareg:
Currently if browser fails to set audiooutput device on tracks, this error was logged as SignalingError to callstats. This is actually not very obvious, so instead log it as gUM error.

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -197,7 +197,7 @@ var LibJitsiMeet = {
               handler(error);
           });
         } else {
-            Statistics.reportGlobalError.call(Statistics, error);
+            Statistics.reportGlobalError(error);
         }
     },
 
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -197,11 +197,7 @@ var LibJitsiMeet = {
               handler(error);
           });
         } else {
-            if (error instanceof JitsiTrackError && error.gum) {
-                Statistics.sendGetUserMediaFailed(error);
-            } else {
-                Statistics.sendUnhandledError(error);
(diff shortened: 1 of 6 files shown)


JitsiMeetJS.js
```diff
@@ -197,7 +197,11 @@ var LibJitsiMeet = {
               handler(error);
           });
         } else {
-            Statistics.sendUnhandledError(error);
+            if (error instanceof JitsiTrackError && error.gum) {
+                Statistics.sendGetUserMediaFailed(error);
+            } else {
+                Statistics.sendUnhandledError(error);
+            }
```
lyubomir: This piece looks suspiciously close to a duplicate of the piece in JitsiConference.js. Is it possible to eliminate this duplication?

tsareg: Actually, I don't know why it's duplicated at all. I mean, why do we do the same in JitsiConference and in JitsiMeetJS.createLocalTracks. I'll ask in the chat probably.

tsareg: Extracted code duplication

JitsiMeetJS.js
```diff
@@ -197,7 +197,7 @@ var LibJitsiMeet = {
               handler(error);
           });
         } else {
-            Statistics.sendUnhandledError(error);
+            Statistics.reportGlobalError.call(Statistics, error);
```
lyubomir: Isn't that the same as Statistics.reportGlobalError(error)?

tsareg: My bad. Fixed

PR #132, opened 2016-06-10
Add ability to pass facingMode as option to JitsiMeetJS.createLocalTracks()

tsareg:


modules/RTC/RTCUtils.js
```diff
@@ -122,7 +123,10 @@ function getConstraints(um, options) {
         } else {
             // Prefer the front i.e. user-facing camera (to the back i.e.
             // environment-facing camera, for example).
-            constraints.video.facingMode = 'user';
+            // TODO: maybe use "exact" syntax if options.facingMode is defined
+            // but this probably needs to be updated with other constraints,
+            // as we currently don't use "exact" syntax anywhere
```
lyubomir: 1. I don't understand your comment because I don't know what "exact" means. Could you please provide the standard which defines what that is?

2. Could I please ask you to convert this comment to a proper sentence i.e. capitalization, full stop, etc?

tsareg: @lyubomir "exact" constraint means that we require exactly this constraint. If it's not satisfied, gUM will fail with something like "ConstraintNotSatisfiedError". It can be applied to various parameters, like deviceId, width, height, facingMode etc. Here is an example from https://w3c.github.io/mediacapture-main/getusermedia.html.
```javascript
// Here's an example of "I want a front-facing camera and it must be VGA.":
navigator.mediaDevices.getUserMedia({
  video: {
    facingMode: {exact: "user"},
    width: {exact: 640},
    height: {exact: 480}
  }
});
```
If "exact" is not used, that means that parameter is optional: try to get me "user" camera, but it's OK if you give me any other. Because currently we are loose on such constraints and don't require, for example, exact deviceId, I left that TODO in order to revise this approach when we will work on updating our constraints (because we still will need to do this at some point).

And here is example from https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia:

> For example, on mobile devices, the following will prefer the front camera (if one is available) over the rear one:
> ```{ audio: true, video: { facingMode: "user" } }```
> To require the rear camera, use:
> ```{ audio: true, video: { facingMode: { exact: "environment" } } }```

Also I fixed the comment

lyubomir:
Jenkins, test this please.

tsareg:
@lyubomir do you have any other questions or comments?

PR #134, opened 2016-06-15
Sends additional logs to callstats.

damencho:
Logs versions of components and library version. Logs shard and region if available.

### DIFF
--- a/connection_optimization/external_connect.js
+++ b/connection_optimization/external_connect.js
@@ -39,6 +39,13 @@ function createConnectionExternally(webserviceUrl, success_callback,
             if (xhttp.status == HTTP_STATUS_OK) {
                 try {
                     var data = JSON.parse(xhttp.responseText);
+                    var headers = xhttp.getAllResponseHeaders();
+                    window.jitsiProxyRegion
+                        = xhttp.getResponseHeader('X-Proxy-Region');
+                    window.jitsiRegion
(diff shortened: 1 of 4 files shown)


connection_optimization/external_connect.js
```diff
@@ -39,6 +39,13 @@ function createConnectionExternally(webserviceUrl, success_callback,
             if (xhttp.status == HTTP_STATUS_OK) {
                 try {
                     var data = JSON.parse(xhttp.responseText);
+                    var headers = xhttp.getAllResponseHeaders();
+                    window.jitsiProxyRegion
```
tsareg: Maybe use a one object to store those values? e.g. window.jitsiRegionInfo...

connection_optimization/external_connect.js
```diff
@@ -39,6 +39,13 @@ function createConnectionExternally(webserviceUrl, success_callback,
             if (xhttp.status == HTTP_STATUS_OK) {
                 try {
                     var data = JSON.parse(xhttp.responseText);
+                    var headers = xhttp.getAllResponseHeaders();
```
hristoterezov: I think that variable is not used. Can you remove it?

connection_optimization/external_connect.js
hristoterezov: +1

PR #136, opened 2016-06-16
Switch to new style gUM constraints according to W3C spec for new browsers

tsareg:
So here I tried to switch to new style for constraints according to W3C spec (https://www.w3.org/TR/mediacapture-streams/#constrainable-interface). You can try it in Chrome 53 (Canary) or Firefox 47.
Though it works, there are number of issues with it:
- non-standard Chrome-specific props seems not to work with these new styles constraints (https://bugs.chromium.org/p/chromium/issues/detail?id=605673). This means that for screen sharing in Chrome we still need to use old-style constraint.
- "exact" constraint for video dimensions is not working in Chrome - https://bugs.chromium.org/p/chromium/issues/detail?id=620665
- Firefox distorts 4:3 videos - https://bugzilla.mozilla.org/show_bug.cgi?id=1280449 - ALREADY FIXED IN 50.0a1 (2016-06-17)

So actually I don't know if it makes sense to switch to this approach now.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -40,6 +40,12 @@ var featureDetectionAudioEl = document.createElement('audio');
 var isAudioOutputDeviceChangeAvailable =
     typeof featureDetectionAudioEl.setSinkId !== 'undefined';
 
+var constraintsNewFormatSupported = !!(navigator.mediaDevices &&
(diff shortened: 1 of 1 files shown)


saghul:
This has been worked on by @virtuacoplenny lately. Closing.

PR #137, opened 2016-06-16
Refactor callstats removing improve static calls to Statistics.

damencho:
You can call static Statistics methods which will internally handle callstats instances to send stats. Drops handlers and JitsiMeetJS references.


tsareg:
Excellent!
That what I was actually asking about a week ago - why do we need to duplicate code in JitsiMeetJS and in JitsiConference.

PR #138, opened 2016-06-17
Don't mix old and new-style for gUM constraints for Chrome, as it fails with TypeError in new Chrome versions

tsareg:
For more details see https://bugs.chromium.org/p/chromium/issues/detail?id=614716.
In Chrome 51 this error happens when "enable-experimental-web-platform-feature" flag is enabled, while in Chrome Canary 53 it fails always.

This is extracted code from https://github.com/jitsi/lib-jitsi-meet/pull/136 as it seems that currently we're not able to fully switch to new style constraints for new browsers due to bugs.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -114,10 +114,10 @@ function getConstraints(um, options) {
         constraints.video = { mandatory: {}, optional: [] };
 
         if (options.cameraDeviceId) {
-            // Don't mix new and old style settings for Chrome as this leads to
-            // TypeError in new Chrome versions. @see
+            // Don't mix new and old style settings for Chromium as this leads
+            // to TypeError in new Chromium versions. @see
             // https://bugs.chromium.org/p/chromium/issues/detail?id=614716
-            if (!RTCBrowserType.isChrome()) {
+            if (!RTCBrowserType.isChrome() && !RTCBrowserType.isOpera()) {
                 // New style of setting device id.
                 constraints.video.deviceId = options.cameraDeviceId;
             }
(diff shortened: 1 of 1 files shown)


tsareg:
Jenkins, test this please.

modules/RTC/RTCUtils.js
```diff
@@ -114,9 +114,14 @@ function getConstraints(um, options) {
         constraints.video = { mandatory: {}, optional: [] };
 
         if (options.cameraDeviceId) {
-            // new style of settings device id (FF only)
-            constraints.video.deviceId = options.cameraDeviceId;
-            // old style
+            // Don't mix new and old style settings for Chromium as this leads
+            // to TypeError in new Chromium versions. @see
+            // https://bugs.chromium.org/p/chromium/issues/detail?id=614716
+            if (!RTCBrowserType.isChrome() && !RTCBrowserType.isOpera()) {
```
lyubomir: - So the fix applies to Chrome and Opera only? Not React Native?

- Since you have the check 3 times, could I please ask you to have it once stored in a variable with a descriptive name which makes it clear that it stands for a new style of constraints or something like that?

modules/RTC/RTCUtils.js
```diff
@@ -114,9 +114,14 @@ function getConstraints(um, options) {
...
+                // New style of setting device id.
+                constraints.video.deviceId = options.cameraDeviceId;
+            }
+            // Old style.
             constraints.video.optional.push({
```
lyubomir: In the 3 cases that you've modified, the if affects the new style only. Why isn't it an if/else that switches between old and new style?

modules/RTC/RTCUtils.js
tsareg: > So the fix applies to Chrome and Opera only? Not React Native?
It applies for Chromium-based devices only.

> Since you have the check 3 times, could I please ask you to have it once stored in a variable with a descriptive name which makes it clear that it stands for a new style of constraints or something like that?

Fixed

modules/RTC/RTCUtils.js
tsareg: This is a temporary solution, in future we will fully split old and new style constraints when new versions of Chromium and Firefox will have stable support of new constraints format. For more information see https://github.com/jitsi/lib-jitsi-meet/pull/136

tsareg:
Jenkins: it's ok to test

modules/RTC/RTCUtils.js
lyubomir: To start on a humorous note, the Internet says: There's nothing more permanent than a temporary fix. (You've pointed out that there's a PR but, as you said, it's not ready yet.)

Do I understand you right that you don't want to use an if/else in the 3 cases of this PR?

tsareg: For the sake of compatibility, I would still leave mix of both old and new constraints for now. Is that what you're asking me?

tsareg:
Jenkins, test this please.

PR #139, opened 2016-06-17
Don't fire RTCEvents.DEVICE_LIST_CHANGED for the first time we get devices from enumerateDevices()

tsareg:
This is how we're preventing unnecessary JitsiMediaDevicesEvents.DEVICE_LIST_CHANGED on start and gives us possibility to better mimic original 'devicechange' event that is currently not available in browsers.


damencho:
Please do not merge for now. A second PR is coming that is using this event, but still implements the changes from this one.

PR #142, opened 2016-06-21
Fire an optional JitsiMediaDevices.PERMISSION_PROMPT_IS_SHOWN event when JitsiMeetJS.createLocalTracks()

tsareg:
Sometimes we need to do some browser/environment detection in code that uses lib-jitsi-meet, so we can re-use this info from lib.

This is required for https://github.com/jitsi/jitsi-meet/pull/701

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -18,7 +18,6 @@ var Statistics = require("./modules/statistics/statistics");
 var Resolutions = require("./service/RTC/Resolutions");
 var ScriptUtil = require("./modules/util/ScriptUtil");
 var GlobalOnErrorHandler = require("./modules/util/GlobalOnErrorHandler");
-var RTCBrowserType = require("./modules/RTC/RTCBrowserType");
 
 function getLowerResolution(resolution) {
     if(!Resolutions[resolution])
@@ -43,9 +42,6 @@ var LibJitsiMeet = {
 
     version: '{#COMMIT_HASH#}',
 
-    // expose JitsiTrackError this way to give library consumers to do checks
-    // like if (error instanceof JitsiMeetJS.JitsiTrackError) { }
-    JitsiTrackError: JitsiTrackError,
     JitsiConnection: JitsiConnection,
     events: {
         conference: JitsiConferenceEvents,
@@ -61,7 +57,6 @@ var LibJitsiMeet = {
     },
     logLevels: Logger.levels,
     mediaDevices: JitsiMediaDevices,
-    environment: RTCBrowserType,
     init: function (options) {
         Statistics.audioLevelsEnabled = !options.disableAudioLevels;
 
@@ -206,6 +201,10 @@ var LibJitsiMeet = {
         RTCUIHelper: RTCUIHelper
     }
 };
+
+// expose JitsiTrackError this way to give library consumers to do checks like
+// if (error instanceof JitsiMeetJS.JitsiTrackError) { }
+LibJitsiMeet.JitsiTrackError = JitsiTrackError;
 
 //Setups the promise object.
 window.Promise = window.Promise || require("es6-promise").Promise;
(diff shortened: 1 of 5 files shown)


JitsiMeetJS.js
```diff
@@ -42,6 +43,9 @@ var LibJitsiMeet = {
 
     version: '{#COMMIT_HASH#}',
 
+    // expose JitsiTrackError this way to give library consumers to do checks
+    // like if (error instanceof JitsiMeetJS.JitsiTrackError) { }
+    JitsiTrackError: JitsiTrackError,
```
yanas: There are a couple of things that are confusing here: 1) We have JitsiTrackError and JitsiTrackErrors. I don't understand why we need 2 different files dealing with JitsiTrackErrors or at least we need to name them differently, so that it's clear what's the exact purpose behind if it's different. 2) They're both exposed in different ways. We're already exporting errors.track.JitsiTrackErrors (see below code).

tsareg: JitsiTrackError is constructor, JitsiTrackErrors is enum.

We already expose JitsiConnection, JitsiConnectionErrors and JitsiConnectionEvents the same way: https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiMeetJS.js#L45, https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiMeetJS.js#L54, https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiMeetJS.js#L48

I can suggest to rename JitsiTrackErrors to JitsiTrackErrorNames or JitsiTrackErrorsEnum. But we still need to expose JitsiTrackError constructor to give library users opportunity to do checks like:
```javascript
if (error instanceof JitsiMeetJS.JitsiTrackError) {
    switch (error.name) {
        case JitsiMeetJS.errors.track.PERMISSION_DENIED:
            // do something
    }
}
```


So I can suggest renaming

JitsiMeetJS.js
```diff
@@ -57,6 +61,7 @@ var LibJitsiMeet = {
     },
     logLevels: Logger.levels,
     mediaDevices: JitsiMediaDevices,
+    environment: RTCBrowserType,
```
hristoterezov: @tsareg Why do you need this? We were trying to avoid this since long time.

Personally I prefer to keep simple the API of the library and I'm trying to avoid adding a lot of new methods to the interface.

Even if we need this I don't think that the browser detection is functionality that is related to the functionality of the lib-jitsi-meet. But if you absolutely need this we can extract it in separate npm package instead of adding it to the interface.

tsareg: I need to construct different message depending on browser type for overlay with guidance for allowing camera and microphone access.
We can re-implement browser detection in jitsi-meet if that's OK.

JitsiMeetJS.js
yanas: JitsiConnection is corresponding more to JitsiTrack as a type of object we expose, not so much to JitsiTrackError. I've had a closer look to the code and I suggest that we expose JitsiTrackError and expose the possible error types from JitsiTrackError.

tsareg: OK, as this is already an existing code, I suggest to refactor it in a separate PR.

hristoterezov: @tsareg I think we can add `LibJitsiMeet.errorTypes` map where you can expose the constructor. That way the users of the library are going to access it with `JitsiMeetJS.errorTypes.JitsiTrackError`. WDYT?

If you decide to do that trough different PR could you please remove this change from this PR?

tsareg: Yes, I will do this in separate PR.

tsareg:
I've completely refactored this PR. Now a separate JitsiMediaDevices.PERMISSION_PROMPT_IS_SHOWN event is (optionally) fired when we call JitsiMeetJS.createLocalTrack(). It passes "environmentType" as an argument to know which browser or execution environment is currently used.

tsareg:
Jenkins, test this please.

tsareg:
Jenkins, test this please.

JitsiMeetJS.js
tsareg: @hristoterezov @yanas PRs about JitsiTrackError refactoring - https://github.com/jitsi/lib-jitsi-meet/pull/149 and https://github.com/jitsi/jitsi-meet/pull/707

PR #147, opened 2016-06-22
Stores jitsiObjectURL when creating url.

damencho:
Fixes test failing when using chrome-beta. Stores jitsiObjectURL otherwise on every call of getVideoSrc we create new different url.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -668,6 +668,8 @@ function defaultGetVideoSrc(element) {
         if (URL) {
             url = URL.createObjectURL(srcObject);
             try {
+                // store the url to avoid creating different url for every call
+                srcObject.jitsiObjectURL = url;
                 return url.toString();
             } finally {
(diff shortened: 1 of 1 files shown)


modules/RTC/RTCUtils.js
```diff
@@ -668,6 +668,8 @@ function defaultGetVideoSrc(element) {
         if (URL) {
             url = URL.createObjectURL(srcObject);
             try {
+                // store the url to avoid creating different url for every call
+                srcObject.jitsiObjectURL = url;
```
lyubomir: There's a URL.revokeObjectURL(url) which disposes of the value assigned to srcObject.jitsiObjectURL!

damencho:
@lyubomir is it possible to just use the MediaStream.id? Is that property available everywhere including react? This way we will eliminate creating/destroying the url objects.

lyubomir:
Obsoleted by https://github.com/jitsi/lib-jitsi-meet/pull/157

PR #148, opened 2016-06-23
Logs active devices to callstats.

damencho:
Also fixes callstats initialization, changes a deprecated fabric event. Fixes logging recorder states.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1082,6 +1082,16 @@ function setupListeners(conference) {
 
     conference.room.addListener(XMPPEvents.LOCAL_ROLE_CHANGED, function (role) {
         conference.eventEmitter.emit(JitsiConferenceEvents.USER_ROLE_CHANGED, conference.myUserId(), role);
+
+        // log all events for the recorder operated by the moderator
+        if (conference.statistics && conference.isModerator()) {
+            conference.on(JitsiConferenceEvents.RECORDER_STATE_CHANGED,
+                function (status, error) {
+                    Statistics.sendLog(
+                        "[Recorder] status: " + status
+                            + (error? " error: " + error : ""));
+                });
+        }
     });
     conference.room.addListener(XMPPEvents.MUC_ROLE_CHANGED, conference.onUserRoleChanged.bind(conference));
 
@@ -1334,14 +1344,6 @@ function setupListeners(conference) {
                 var type = (track.getType() === "audio")? "audio" : "video";
                 conference.statistics.sendMuteEvent(track.isMuted(), type);
             });
-
-        // log all events for the recorder operated by the moderator
-        if (conference.isModerator()) {
-            conference.on(JitsiConferenceEvents.RECORDER_STATE_CHANGED,
-                function (status, error) {
-                    Statistics.sendLog("Recorder: " + status);
-                });
-        }
 
         conference.room.addListener(XMPPEvents.CREATE_OFFER_FAILED, function (e, pc) {
             conference.statistics.sendCreateOfferFailed(e, pc);
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1339,7 +1339,7 @@ function setupListeners(conference) {
         if (conference.isModerator()) {
             conference.on(JitsiConferenceEvents.RECORDER_STATE_CHANGED,
                 function (status, error) {
-                    conference.statistics.sendLog("Recorder: " + status);
+                    Statistics.sendLog("Recorder: " + status);
                 });
         }
 
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -332,6 +332,16 @@ JitsiConference.prototype.addTrack = function (track) {
     this.room.addListener(XMPPEvents.SENDRECV_STREAMS_CHANGED,
         track.ssrcHandler);
 
+    // Report active device to statistics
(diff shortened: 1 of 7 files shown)


damencho:
Jenkins ok to test

modules/RTC/RTCUtils.js
```diff
@@ -321,6 +321,7 @@ function pollForAvailableMediaDevices() {
  * @emits RTCEvents.DEVICE_LIST_CHANGED
  */
 function onMediaDevicesListChanged(devices) {
+    var initialDeviceList = currentlyAvailableMediaDevices.length == 0;
```
tsareg: I don't think this is the best way to do this. The more correct way is in https://github.com/jitsi/lib-jitsi-meet/pull/139.
Let me explain the problem. The code the polls for ```enumerateDevices()``` is a workaround until ```"devicechange"``` event is supported in browsers. And if in future we will switch to that event natively, it won't compare old and new device, it will just fire and that's it. And with your new implementation we will just loose the first event.

modules/RTC/RTCUtils.js
```diff
@@ -1181,6 +1183,31 @@ var RTCUtils = {
...
+        return currentlyAvailableMediaDevices;
+    },
+
+    /**
+     * Returns event data for device to be reported to stats.
+     * @returns {MediaDeviceInfo} device.
+     */
+    getEventDataForActiveDevice: function (device) {
+        var devices = [];
+        var deviceData = {
```
tsareg: There is internal function mediaDeviceInfoToJSON here https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/RTCUtils.js#L307
I think we can expose it and re-use. WDYT?

JitsiConference.js
```diff
@@ -332,6 +332,16 @@ JitsiConference.prototype.addTrack = function (track) {
     this.room.addListener(XMPPEvents.SENDRECV_STREAMS_CHANGED,
         track.ssrcHandler);
 
+    // Report active device to statistics
+    var devices = RTC.getCurrentlyAvailableMediaDevices();
+    device = devices.find(function (d) {
```
tsareg: There is already similar code in JitsiLocalTrack._setRealDeviceIdFromDeviceList method: https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/JitsiLocalTrack.js#L107. We can probably expose it as public method and re-use

tsareg: Or we can use ```JitsiLocalTrack.getDeviceId()``` and select device by ID from the list.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -55,7 +55,12 @@ function JitsiLocalTrack(stream, track, mediaType, videoType, resolution,
     // called.
     this._realDeviceId = this.deviceId === '' ? undefined : this.deviceId;
 
-    this._onDeviceListChanged = function (devices) {
+    this._onDeviceListChanged = function (devices, isInitial) {
+
+        // skip initial event
+        if (isInitial)
```
tsareg: See my comments below regarding initial event

modules/RTC/RTCUtils.js
damencho: Ok I got that, but even when this is available we can implement something similar. 
And what about the output device, logging? This also solves the problem, when to send to callstats the inital output device data, what is your proposal, when we should do that if we merge PR#139?

JitsiConference.js
damencho: _setRealDeviceIdFromDeviceList is already public cause it is used by jitsi-meet, which I don't like and would prefer to remove that. Also I'm not sure when the deviceId is updated, for the above reason.

modules/RTC/RTCUtils.js
tsareg: We can fire one more event, e.g. RTCEvents.DEVICE_LIST_AVAILABLE, and log audioOutput device on that event:
```javascript
function onReady (options, GUM) {
    rtcReady = true;
    eventEmitter.emit(RTCEvents.RTC_READY, true);
    screenObtainer.init(options, GUM);

    if (RTCUtils.isDeviceListAvailable() && rawEnumerateDevicesWithCallback) {
        rawEnumerateDevicesWithCallback(function (devices) {
            currentlyAvailableMediaDevices = devices.splice(0);
            
            eventEmitter.emit(RTCEvents.DEVICE_LIST_AVAILABLE, currentlyAvailableMediaDevices);

            if (isDeviceChangeEventSupported) {
                navigator.mediaDevices.addEventListener('devicechange', function () {
                    RTCUtils.enumerateDevices(onMediaDevicesListChanged);
                });
            } else {
                window.setTimeout(pollForAvailableMediaDevices,
                    AVAILABLE_DEVICES_POLL_INTERVAL_TIME);
            }
        });
    }
}
```

And this way we will get the initial device list and also won't need to re-do anything after switching to native "devicechange" event.

damencho:
I will create new PR in order to remove d09b8dd (and keep history and avoid force pushing) and will merge https://github.com/jitsi/lib-jitsi-meet/pull/139.

modules/RTC/RTCUtils.js
damencho: Nope it is different as it returns a string not a json object.

PR #152, opened 2016-06-28
Jwt login service

paweldomas:
Integrates external login service which provides JWT token once the user authenticates.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -165,7 +165,8 @@ JitsiConference.prototype.getTokenAuthUrl = function (roleUpgrade) {
  * Check if external authentication is enabled for this conference.
  */
 JitsiConference.prototype.isExternalAuthEnabled = function () {
-    return this.room && this.room.moderator.isExternalAuthEnabled();
+    return this.room &&
+        (this.room.moderator.isExternalAuthEnabled() || this.getTokenAuthUrl());
 };
 
 /**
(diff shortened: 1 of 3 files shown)


hristoterezov:
Do we really need support for tokenAuthUrl in lib-jitsi-meet?

We are just replacing the room name, etc in passed TokenAuthUrl. I think we have all the information for this in jitsi-meet project and I think it fits better there. Am I missing something? WDYT?

paweldomas:
This way it fits the same flow we have for for Shibboleth authentication for opening the popup window. Although the other flows are different, so not sure if that makes sense.

paweldomas:
No matter how easy the logic for generating the URL is, it's better to keep it in the library for eventual app integration. Otherwise apps other than meet will have to duplicate the logic.

Also inside the lib we know more about all state and if it comes to adding new parameters it will be easier.

hristoterezov:
I'm not concerned that the logic is easy. I'm worried that we are passing the url to the library and we are not using it there actually( we are not sending it and it doesn't changes the behavior of the library), we are just returning it. 

For the external auth we are getting the url with requests to jicofo I think. 

So I think this is utility method since the library is not using it. Also if we need the state of the library I think it's better to expose what we need. WDYT?

If you think this method is going to be useful for other services I would suggest to add it as a utility not passing it to JitsiConference. Although it is just string replace for now and personally I don't think we need utility method for this. Maybe if it become more complex in future we can add it as utility method. WDYT?

paweldomas:
> For the external auth we are getting the url with requests to jicofo I think.

Yes, but here we can't do this, because we have to get the URL, before XMPP connection is established(token is required to connect to XMPP server).

> So I think this is utility method since the library is not using it. Also if we need the state of the library I
> think it's better to expose what we need. WDYT?

I don't think it's better to expose stuff and ask the user to deal with that instead of hiding the logic behind the library(in general). But right now it's relatively simple, so I've created a new PR which does the thing with utility method:

https://github.com/jitsi/jitsi-meet/pull/728
https://github.com/jitsi/lib-jitsi-meet/pull/168

PR #155, opened 2016-06-29
Fix logging audio output device to callstats

tsareg:
Do not send event to callstats when no audiooutput devices are available.
Fix non-ASCII letter in Statistics.sendActiveDeviceListEvent() method


damencho:
Jenkins, test this please.

PR #156, opened 2016-06-30
Be forgiving towards API clients

lyubomir:
If the API client happened to try to add one and the same local video
track twice, the library would throw an error. I find that draconian
because:

1. The error message states that a second local video stream cannot be
added but that is incorrect because the API client is not trying to add
a second local video stream but add one and the same local video stream
multiple times.

2. The jsdoc says that a throw will happen if another local video stream
is added but that is misleading because the API client is not trying to
add another local video stream but add the same local video stream.

3. Adding one and the same local video stream multiple times can be
handled gracefully by the lib-jitsi-meet library. Thus, the library
appears more friendly.

Hence, gracefully handle the case of adding the same local video track
multiple times without throwing an error.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -308,17 +308,27 @@ JitsiConference.prototype.setSubject = function (subject) {
  * Adds JitsiLocalTrack object to the conference.
  * @param track the JitsiLocalTrack object.
  * @returns {Promise<JitsiLocalTrack>}
- * @throws will throw and error if track is video track
- * and there is already another video track in the conference.
+ * @throws {Error} if the specified track is a video track and there is already
+ * another video track in the conference.
  */
 JitsiConference.prototype.addTrack = function (track) {
-    if(track.disposed)
-    {
+    if (track.disposed) {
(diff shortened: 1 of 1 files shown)


JitsiConference.js
```diff
@@ -308,17 +308,27 @@ JitsiConference.prototype.setSubject = function (subject) {
...
+    if (track.isVideoTrack()) {
+        // Ensure there's exactly 1 local video track in the conference.
+        var localVideoTrack = this.rtc.getLocalVideoTrack();
+        if (localVideoTrack) {
+            // Don't be excessively harsh and severe if the API client happens
+            // to attempt to add the same local video track twice.
+            if (track === localVideoTrack) {
+                return Promise.resolve(track);
+            } else {
+                throw new Error(
```
tsareg: What about throwing JitsiTrackError instead of just Error to be consistent with other throws in this method? We can have a new type for it, e.g. JitsiTrackErrors.MULTIPLE_VIDEO_TRACKS.

hristoterezov: @tsareg at some point maybe we will support multiple video tracks. So I personally don't think it worths to add new JitsiTrackErrors for this one and remove it later.

PR #160, opened 2016-07-07
Fix JitsiLocalTrack dispose() method

tsareg:
Wrap thrown errors into Promise.reject for consistency in JitsiConference's addTrack() and removeTrack() methods.
Handle case when JitsiLocalTrack#dispose() is called after track was removed from conference.
Set track's conference reference to null when removing track from conference.


damencho:
Jenkins: test this please

PR #162, opened 2016-07-08
Append random hex string to MUC nickname with XMPP login

paweldomas:
When XMPP authenticated user joins the conference his login is used as MUC nickname. Now if his connection gets interrupted and he stays as a ghost until it times out, he will be unable to join from the new connection. That's because Prosody allows to join twice under the same nickname from the same bare JID. Jicofo will not distinguish between those two users and will not invite the second one to the conference.


damencho:
Jenkins, test this please.

PR #163, opened 2016-07-08
False callstats failures

paweldomas:
This series of commits prevent Callstats.io from classifying conference as partially failed when that's not really the case.


damencho:
Jenkins, test this please.

PR #166, opened 2016-07-08
Implements reload functionality for JitsiConnection and JitsiConference

hristoterezov:


modules/RTC/JitsiLocalTrack.js
```diff
@@ -332,22 +332,6 @@ JitsiLocalTrack.prototype.isMuted = function () {
...
-JitsiLocalTrack.prototype._setRTC = function (rtc) {
-    this.rtc = rtc;
-    // We want to keep up with postponed events which should have been fired
-    // on "attach" call, but for local track we not always have the conference
-    // before attaching. However this may result in duplicated events if they
-    // have been triggered on "attach" already.
-    for(var i = 0; i < this.containers.length; i++)
-    {
-        this._maybeFireTrackAttached(this.containers[i]);
-    }
```
damencho: _setRTC was called in addLocalTrack, and now in addLocalTrack we don't call _setConference(which now calls _maybeFireTrackAttached), which means we do not call _maybeFireTrackAttached at that moment. Is this ok?

JitsiConferenceEventManager.js
```diff
@@ -0,0 +1,486 @@
...
+            });
+
+        conference.room.addListener(XMPPEvents.CONNECTION_ICE_FAILED,
+            function (pc) {
+                conference.statistics.sendIceConnectionFailedEvent(pc);
+                conference.room.eventEmitter.emit(
+                    XMPPEvents.CONFERENCE_SETUP_FAILED,
+                    new Error("ICE fail"));
+            });
+
```
damencho: If there is no conference statistics, there will be no XMPPEvents.CONFERENCE_SETUP_FAILED fired. Maybe we need to separate the emit and stats logic.

JitsiConference.js
```diff
@@ -72,6 +67,61 @@ function JitsiConference(options) {
...
+    if(!options)
+        options = {};
+    if(!this.options) {
+        this.options = options;
+    } else {
+        // Override config options
+        var config = options.config || {};
+        for(var key in config)
+            this.options.config[key] = config[key] || this.options.config[key];
+    }
```
lyubomir: I find the above source code implements complex logic that's unjustified (at least I don't see a reason or an explanation about its complexity):
- If there's no this.options, then we're remembering the whole options object.
- But if there's this.options, we care only about the config property of the options object.

Wouldn't it be easier to read & maintain if we paid attention to the whole options object at all times.

Another reason for adopting ES6 is that the above source code of 10 lines could easily be replaced with 1 line:
this.options = Object.assign(this.options || {}, options);

JitsiConferenceEventManager.js
```diff
@@ -0,0 +1,495 @@
...
+        this.conference.eventEmitter);
+}
+
+/**
+ * Setups event listeners related to conference.chatRoom
+ */
+JitsiConferenceEventManager.prototype.setupChatRoomListeners = function () {
+    var conference = this.conference;
+
+    conference.room.addListener(XMPPEvents.ICE_RESTARTING, function () {
```
lyubomir: conference.room is going to appear in the following source code a number of times. I think it makes sense to reduce the complexity of the source code by defining a, for example, var chatRoom = conference.room; just like var conference = this.conference;

JitsiConferenceEventManager.js
```diff
@@ -0,0 +1,495 @@
...
+
+            // mute existing local tracks because this is initial mute from
+            // Jicofo
+            conference.getLocalTracks().forEach(function (track) {
+                if (conference.startAudioMuted && track.isAudioTrack()) {
+                    track.mute();
+                }
+                if (conference.startVideoMuted && track.isVideoTrack()) {
+                    track.mute();
+                }
```
lyubomir: The above conditions seem susceptible to improvement to me because isAudioTrack precludes the possibility of satisfying isVideoTrack as well. I guess 1 could reason 2 ways:

1. Put the isAudioTrack/isVideoTrack first:
````javascript
if (track.isAudioTrack()) {
    conference.startAudioMuted && track.mute();
} else if (track.isVideoTrack()) {
    conference.startVideoMuted && track.mute();
}
````

Of course, one could reason that isAudioTrack/isVideoTrack is slower (because it's more complex a function) than the booleans such as startAudioMuted & startVideoMuted. Since we're in lib-jitsi-meet, we could improve that drawback with a switch on track.getType() because there'll be 1 getType() call:
````javascript
switch (track.getType()) {
case MediaType.AUDIO:
    conference.startAudioMuted && track.mute();
    break;
case MediaType.VIDEO:
    conference.startVideoMuted && track.mute();
    break;
}
````

2. As an alternative to 1. & if we don't want to sacrifice too much, we could at least improve by noticing that if the first if is satisfied, then there's no point in checking the second if. That could be solved with a return in the first if or converting the 2 if to if-else-if.

File: JitsiConferenceEventManager.js [hunk 6]
```diff
@@ -0,0 +1,495 @@
...
+/**
+ * Setups event listeners related to conference.statistics
+ */
+JitsiConferenceEventManager.prototype.setupStatisticsListeners = function () {
+    var conference = this.conference;
+    if(!conference.statistics)
+        return;
+
+    conference.statistics.addAudioLevelListener(function (ssrc, level) {
+        var userId = null;
```
lyubomir: What is var userId = null; about? I don't see userId used in the function.

JitsiConferenceEventManager.js
```diff
@@ -0,0 +1,495 @@
...
+
+        conference.eventEmitter.emit(
+            JitsiConferenceEvents.CONNECTION_STATS, stats);
+    });
+
+    conference.on(JitsiConferenceEvents.TRACK_MUTE_CHANGED,
+        function (track) {
+            if(!track.isLocal())
+                return;
+            var type = (track.getType() === "audio")? "audio" : "video";
```
lyubomir: Why are we using "audio" for track.getType() === "audio" & not MediaType.VIDEO like we do in JitsiTrack?

modules/statistics/statistics.js
```diff
@@ -16,10 +21,12 @@ var JitsiTrackError = require("../../JitsiTrackError");
...
-    ScriptUtil.loadScript(
-            'https://api.callstats.io/static/callstats.min.js',
-            /* async */ true,
-            /* prepend */ true);
+    if(!isCallstatsLoaded)
+        ScriptUtil.loadScript(
+                'https://api.callstats.io/static/callstats.min.js',
+                /* async */ true,
+                /* prepend */ true);
+    isCallstatsLoaded = true;
```
lyubomir: I think {} could help here. There's no point in isCallstatsLoad = true; when the condition !isCallstatsLoaded is false.

modules/statistics/statistics.js
```diff
@@ -211,6 +218,23 @@ Statistics.prototype.startCallStats = function (session, settings) {
 };
 
 /**
+ * Removes the callstats.io instances.
+ */
+Statistics.prototype.stopCallStats = function () {
+    if(this.callStatsIntegrationEnabled && this.callstats) {
```
lyubomir: The condition this.callStatsIntegrationEnabled appears superfluous to me, this.callstats looks like it's enough.

modules/statistics/statistics.js
```diff
@@ -211,6 +218,23 @@ Statistics.prototype.startCallStats = function (session, settings) {
...
+        var callstatsList = Statistics.callsStatsInstances;
+        for(var i = 0;i < callstatsList.length; i++) {
+            if(this.callstats === callstatsList[i]) {
+                Statistics.callsStatsInstances.splice(i, 1);
+                break;
+            }
+        }
```
lyubomir: Hm, isn't the for loop above a reimplementation of Array.prototype.indexOf?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1103,7 +1103,11 @@ JingleSessionPC.prototype.newJingleErrorHandler = function(request, failureCb) {
             error.source = request.tree();
         }
 
-        error.session = this;
+        // Commented to fix JSON.stringify(error) exception for
+        // circular dependancies when we print that error.
+        // FIXME: Maybe we can include
+        // part of the session object
```
lyubomir: I don't see a reason why the FIXME should be split on two lines with so very few words per line. Did it hit a character limit?

modules/xmpp/moderator.js
```diff
@@ -30,6 +30,8 @@ function Moderator(roomName, xmpp, emitter, settings, options) {
     this.externalAuthEnabled = false;
     this.settings = settings;
     this.options = options;
+    this.maxRetries = maxRetries || Infinity;
```
lyubomir: It's not the only place that I see it in this PR but I'd like to use the example to point out that default parameters in ES6 give relatively the same functionality.

modules/xmpp/xmpp.js
```diff
@@ -57,6 +59,45 @@ function XMPP(options, token) {
...
+ * Reloads the XMPP module
+ * @param options {object} options to be overriden
+ */
+XMPP.prototype.reload = function (options) {
+    if(!options)
+        options = {};
+    if(!this.options) {
+        this.options = options;
+    } else {
+        // Override config options
```
lyubomir: That's the product of copy+paste! The comment is incorrect because there's no "config options". Anyway, the bigger concern is that 9 or 10 lines of source code are repeated for what may (if the thing with options.config in my first comment is not a real requirement) turn out to be 1 function which could be optimized & reused.

modules/RTC/JitsiLocalTrack.js
hristoterezov: addLocalTrack is called only from JitsiConference.addTrack and _setConference is called inside addTrack. I think it is the same as before and it should be OK.

JitsiConference.js
```diff
@@ -72,6 +67,61 @@ function JitsiConference(options) {
...
+
+/**
+ * Reinitializes JitsiConference instance
+ * @param options {object} options to be overriden
+ */
+JitsiConference.prototype._reinitialize = function (options) {
+    this._init(options || {});
+    this.eventManager.setupChatRoomListeners();
+    this.eventManager.setupStatisticsListeners();
+    this.join();
```
lyubomir: I find the _reinitialize method hard to understand:

1. I saw that it's calling this.join & I was initially like "How is it sure that it needs to join?". Then I saw that it was called from a method which does leave first. The leave & join could've been kept in 1 place to aid readability:
````javascript
this.leave(true);
_reinitialize(options);
this.join();
````

Or the prerequisite of the method could be described in a comment.

2. Afterwards i.e. after this.join is not here, the source code could be reduced & cleaned up by eliminating duplicate logic between the constructor & _reinitialize:

- Why init(options || {}) in _reinitialize given that the constructor does init(options) only and init checks whether options is falsy?

- One could observe that both this.eventManager.setupChatRoomListeners(); & this.eventManager.setupStatisticsListeners(); need to be executed after this.init(options). In the constructor the same is being done anyway, it's just split differently (because of _setupListeners). Obviously, this.eventManager.setupRTCListeners(); & setupXMPPListeners need only be called when the respectively involved objects are changed (currently only at construction time). I can then relatively easily image a single _init method & no _reinitialize method. Anyway, that's a point about reducing source code in order to facilitate maintenance so I'm not sure whether it agrees with lib-jitsi-meet's policy.

File: JitsiConference.js (hunk 3 above)
Comment (hristoterezov) [inline JitsiConference.js]: Agree. But currently we are not overriding anything from the options. I'll remove it for now.

File: modules/xmpp/xmpp.js (hunk 13 above)
Comment (hristoterezov) [inline modules/xmpp/xmpp.js]: Since we are not using this code currently I'll remove it for now.

File: modules/xmpp/JingleSessionPC.js (hunk 11 above)
Comment (hristoterezov) [inline modules/xmpp/JingleSessionPC.js]: It will fit on 1 line. I'll fix it.

File: modules/util/EventEmitterForwarder.js [hunk 15]
```diff
@@ -0,0 +1,37 @@
...
+ */
+EventEmitterForwarder.prototype.forward = function () {
+    // This line is only for fixing jshint errors.
+    var args = arguments;
+    var srcEvent = args[0];
+    //This will be the "this" value for emit function.
+    args[0] = this.dest;
+    //Using bind.apply to pass the arguments as Array-like object ("arguments")
+    this.src.addListener(srcEvent,
+        this.dest.emit.bind.apply(this.dest.emit, args));
```
lyubomir: Wow, that looks like magic even after we've talked about it...

Anyway, let me just ask about the line this.dest.emit.bind.apply(this.dest.emit, args):

I guess the function this.dest.emit that is the first argument (let me call it fun for the sake of clarity) is needed by the bind function as its this. Because that's to be invoked normally as:
````javacript
fun.bind(thisArg, arg1, arg2, ...);
````

But then why do we need this.dest.emit again at the beginning?

I guess part of the confusion for me are all the dots & repetitions. I think you could've gone for some clarity with:
````javascript
var destEmit = this.destEmit;
...destEmit.bind.apply(destEmit, args)...
````

But then have you tried using Function.prototype.bind because that's likely what you want to get to via this.dest.emit.bind at the beginning?
````javascript
Function.prototype.bind.apply(this.dest.emit, args)
````

JitsiConferenceEventManager.js
hristoterezov: I'll remove this check because anyway track.getType() should return MediaType.AUDIO or MediaType.Video.

modules/util/EventEmitterForwarder.js
hristoterezov: > I guess the function this.dest.emit that is the first argument (let me call it fun for the sake of clarity) is needed by the bind function as its this. Because that's to be invoked normally as:
> 
> `fun.bind(thisArg, arg1, arg2, ...);`
> But then why do we need this.dest.emit again at the beginning?

I'm sorry I'm not sure I understood this part of the comment correctly. You are proposing to change it to `Function.prototype.bind.apply(this.dest.emit, args)` here? Or this is related to different thing? 

`var destEmit = this.dest.emit;` is good idea. 

Do you mean that
`Function.prototype.bind.apply(this.dest.emit, args)` should be equivalent to `this.dest.emit.bind.apply(this.dest.emit, args));` ? I haven't thought about it. I'll try it. Looks better :)

lyubomir: I think that either 1 of the following 2 alternatives gives more clarity:

- var destEmit = this.dest.emit; destEmit.bind.apply(destEmit, args)

- Function.prototype.bind.apply(this.dest.emit, args)

lyubomir:
On the source code level alone I'm OK with this PR in the light of the comments that I already made.

I see nothing strikingly wrong with the reload functionality by reading the source code. I haven't tested this PR manually because of other priorities/lack of time. I guess that the automatic PR testing is at least something.

Besides, I'm not the only one reviewing it.

PR #167, opened 2016-07-11
Clean up event listeners attached inside JitsiConference and related classes

tsareg:
Since we are moving to single page web app and native app in scope of jitsi-meet-react project, it's important for us to work better with memory and do not allow possible memory leaks.

This PR is the first step towards better memory management. Here we clean up all event listeners attached to JitsiConference and some other related classes like ChatRoom, RTC, Moderator, JitsiTrack etc.
Also we clean up events attached to window.

Future PRs will address JitsiConference and XMPP classes and also some static classes like JitsiMeetJS, RTC, RTCUtils.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -53,7 +53,7 @@ function JitsiConference(options) {
             this.options.config.disableThirdPartyRequests,
         roomName: this.options.name
     });
-    setupListeners(this);
+    this._setupListeners();
     this.participants = {};
     this.lastDominantSpeaker = null;
     this.dtmfManager = null;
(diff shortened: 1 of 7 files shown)


tsareg:
This should be revised after https://github.com/jitsi/lib-jitsi-meet/pull/166 is merged

JitsiConference.js
```diff
@@ -114,17 +113,24 @@ JitsiConference.prototype.leave = function () {
...
+        .catch(function (error) {
+            logger.error(error);
+            GlobalOnErrorHandler.callUnhandledRejectionHandler(
+                {promise: this, reason: error});
+            // We are proceeding with leaving the conference because room.leave
+            // may succeed.
+            this._leaveRoomAndRemoveParticipants();
+            return Promise.resolve();
+        }.bind(this))
+        .then(function() {
```
hristoterezov: Why do you need another then clause here?

tsareg: I thought I had some reasons for that, but now I don't see them. Code from this then() can be moved inside _leaveRoomAndRemoveParticipants() method. Also this PR should be fixed after "reload" feature was merged, so I'll take care of it a bit later.

tsareg:
New PR will follow instead

PR #174, opened 2016-07-13
Fix JitsiConferenceEvents.TRACK_REMOVED is fired inconsistently

tsareg:
Currently ```JitsiConferenceEvents.TRACK_REMOVED``` is fired for remote tracks only on ```XMPPEvents.REMOTE_TRACK_REMOVED``` event (https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L1116-L1132). 

However in the case when participant leaves conference at all and we get into ```JitsiConference#onMemberLeft()``` method, we just silently remove remote tracks without firing any events except for ```JitsiConferenceEvents.USER_LEFT``` (https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L623-L634).

This fact causes inconsistencies for library users. So in this PR I propose to always fire  ```JitsiConferenceEvents.TRACK_REMOVED``` whenever track is removed. Also with this PR we are going to fix warnings "Overwriting remote track!" every time user changes his audio/video track (https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/RTC.js#L282).

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -628,7 +628,11 @@ JitsiConference.prototype.onMemberLeft = function (jid) {
     var participant = this.participants[id];
     delete this.participants[id];
 
-    this.rtc.removeRemoteTracks(id);
+    var removedTracks = this.rtc.removeRemoteTracks(id);
+
+    removedTracks.forEach(function (track) {
+        this.eventEmitter.emit(JitsiConferenceEvents.TRACK_REMOVED, track);
+    }.bind(this));
 
     this.eventEmitter.emit(JitsiConferenceEvents.USER_LEFT, id, participant);
 };
@@ -1122,8 +1126,15 @@ function setupListeners(conference) {
                         && tracks[i].getStreamId() == streamId
                         && tracks[i].getTrackId() == trackId) {
                         var track = participant._tracks.splice(i, 1)[0];
-                        conference.eventEmitter.emit(
-                            JitsiConferenceEvents.TRACK_REMOVED, track);
+
+                        track = conference.rtc.removeRemoteTrack(
+                            participant.getId(), track);
+
+                        if (track) {
+                            conference.eventEmitter.emit(
+                                JitsiConferenceEvents.TRACK_REMOVED, track);
+                        }
+
                         return;
                     }
                 }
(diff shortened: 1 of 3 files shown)


hristoterezov:
@tsareg Could you please rebase the PR and fix the conflicts?

modules/RTC/RTC.js
```diff
@@ -287,18 +287,57 @@ RTC.prototype.createRemoteTrack = function (event) {
...
+ *
+ * @param {string} resource - The resource part of the MUC JID.
+ * @returns {JitsiRemoteTrack[]}
  */
 RTC.prototype.removeRemoteTracks = function (resource) {
-    var remoteTracks = this.remoteTracks[resource];
+    var remoteTracksForResource = this.remoteTracks[resource];
+    var removedTracks = [];
+
+    if (remoteTracksForResource) {
```
hristoterezov: Why do we need these changes? The old code seems simpler for me and I think both versions are doing the same thing.

modules/RTC/RTC.js
```diff
@@ -287,18 +287,57 @@ RTC.prototype.createRemoteTrack = function (event) {
...
+
+/**
+ * Removes specified JitsiRemoteTrack associated with given MUC nickname
+ * (resource part of the JID). Returns removed track if any.
+ *
+ * @param {string} resource - The resource part of the MUC JID.
+ * @param {JitsiRemoteTrack} track - Track to remove.
+ * @returns {JitsiRemoteTrack|undefined}
+ */
+RTC.prototype.removeRemoteTrack = function (resource, track) {
```
hristoterezov: The code here seems very complex for me. Couldn't we just assume that we have one track per type as we are assuming when the remoteTracks map is created? This way you can pass the type to the method instead of the track instance and use it to delete the entry from remoteTracks.

JitsiConference.js
```diff
@@ -1122,8 +1126,15 @@ function setupListeners(conference) {
                         && tracks[i].getStreamId() == streamId
                         && tracks[i].getTrackId() == trackId) {
                         var track = participant._tracks.splice(i, 1)[0];
-                        conference.eventEmitter.emit(
-                            JitsiConferenceEvents.TRACK_REMOVED, track);
+
+                        track = conference.rtc.removeRemoteTrack(
```
hristoterezov: Why do you overriding the track here? Do you expect the track variable to be changed after removeRemoteTrack?

PR #181, opened 2016-07-20
Implements interfaces for data channel communication

hristoterezov:


modules/RTC/DataChannels.js
```diff
@@ -247,4 +239,37 @@ DataChannels.prototype._some = function (callback, thisArg) {
     }
 };
 
+/**
+ * Sends passed object via the first found open datachannel
+ * @param jsonObject {object} the object that will be sent
+ */
+DataChannels.prototype.send = function (jsonObject) {
```
bgrozev: Can we log a message if sending fails because the channel is not open? This could mask problems and it is a known issue that data channels sometimes open with a 3 second delay (because of a bug in the bridge).

bgrozev:
Guys, I am reverting this because I see multiple problems, some of which are serious, and will require major changes. I will explain the problems and proposed solutions in an e-mail.

modules/RTC/DataChannels.js
hristoterezov: @bgrozev But we are checking if the channel is open before sending. What is the use case you are referring to?

bgrozev: I am referring to the situation in which the data channel has not opened
(for example because of the jvb bug), in which case calls to
DataChannels.send silently fail (that is, do not result in a message being
sent).

Boris

PR #189, opened 2016-07-25
Changes methods that use RTCDatachanel#send to throw exception on fail

hristoterezov:


bgrozev:
Awesome, this addresses my concern with the silent failures. Pushed a small clean-up commit in jitsi-meet#751, will wait or tests and merge them both.

hristoterezov:
@bgrozev Thanks!

paweldomas:
Since we now have decided to throw an error that there are no data channels open it would be good to have a method for doing sanity check and make use of that in meet. Now on beta we can see many misleading errors(thrown globally) about data channels not configured etc...

hristoterezov:
We can do that but I don't know if there is a way to implement a method that will also check whether RTCDatachannel.send will throw error. And my opinion is that it is going to be strange to have a method that guarantees successful sending only in half of the use cases.

paweldomas:
My point is to avoid calling this method when we know that the conference is not ready yet or the channels are not open yet. So we need a method to check if the data channel is open.

PR #191, opened 2016-07-28
Format callstats application logs as JSON

bgrozev:


modules/version/ComponentsVersions.js
```diff
@@ -63,23 +63,19 @@ function(node, mucResource, mucJid) {
 
         var version = item.value;
         if (this.versions[componentName] !== version) {
-            if(this.versions[componentName] &&
```
hristoterezov: @bgrozev Why did you delete this code?

paweldomas: Will XMPP version(Prosody in our case) change always be incompatible ? Btw. why we did not compare if componentName === ComponentsVersions.XMPP_SERVER_COMPONENT in the first place ?

bgrozev: > @bgrozev Why did you delete this code?

The code fires an "incompatible versions event" if the version of the XMPP server changes during a conference, regardless of the version. We can't know whether a version is compatible or not without looking at said version.

Since I very much doubt that we actually encounter changes to the XMPP server version during a conference, I assume the code is effectively dead. And since I could not figure out what the actual intent is, I deleted it in order to spare other people the time and confusion.

hristoterezov: I think the condition is wrong I can check that. But other than that if JitsiConnection._reload is triggered the client can end up on different server side deployment. So we need this code, I can add a comment there and change the name. WDYT?

PR #195, opened 2016-08-02
Update to use latest AdapterJS, fix filename to match upstream AdapterJS name.

PetroccoCo:
This fix:
- renames adapter.screenshare.js to match the adapter.js name that it has upstream in AdapterJS
  - See commit [95cd0ab7506a458e0c26067fd5b0dfe3861f5e59](https://github.com/jitsi/lib-jitsi-meet/commit/95cd0ab7506a458e0c26067fd5b0dfe3861f5e59#diff-aeb31caf77eb7fd96afd4bc9a9c97ddd) for where the two files diverged
- includes a yet to be incorporated fix from AdapterJS, allowing for a custom pluginInfo object to be set
  - https://github.com/Temasys/AdapterJS/pull/211
  - particularly useful if you have a paid version of the Temasys plugin


PetroccoCo:
Those links appear to be broken, if it matters I'm part of the same organization that @dwilson6 is part of. Not sure if we have an corporation agreement though.

dwilson6:
We do have a corporate agreement.

bgrozev:
Confirmed, Jenkins doesn't know that you have it :)

Jenkins add to whitelist

saghul:
This no longer applies as we have already started to transition to adapter.js. Sorry!

PR #200, opened 2016-08-05
_setMute refactoring to allow muting/unmuting local tracks when they are not in conference

tsareg:
For some reason it's not possible to mute/unmute a JitsiLocalTrack when it's not inside a conference. But we need this functionality for React project, where we handle local tracks differently.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -225,8 +225,8 @@ JitsiLocalTrack.prototype._setMute = function (mute) {
                     });
 
                     if(!streamInfo) {
-                        // FIXME Introduce a new JitsiTrackError.
-                        throw new Error('track.no_stream_found');
+                        throw new JitsiTrackError(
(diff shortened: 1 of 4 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -131,157 +131,197 @@ JitsiLocalTrack.prototype.unmute = function () {
...
+};
+
+/**
+ * Synchronizes muted stated of track with muted state of conference.
+ *
+ * @param {boolean} mute - If track is muted.
+ * @private
+ * @returns {Promise}
+ */
+JitsiLocalTrack.prototype._setConferenceMute = function(mute) {
```
hristoterezov: I think the description and the name of the method are confusing. This method will actually send presence packet with information about the local tracks mute status to the other participants of the conference. Maybe we can use "sendMuteStatus" or something like that.

tsareg:
@hristoterezov I've changed method name and description

PR #201, opened 2016-08-05
Analytics move

damencho:


JitsiConnection.js
```diff
@@ -25,12 +26,19 @@ function JitsiConnection(appID, token, options) {
...
                 (msg === "item-not-found" || msg === "host-unknown")) {
                     // FIXME: don't report the error if we are going to reload
                     this._reload();
                 }
         }.bind(this));
+
+    this.addEventListener(JitsiConnectionEvents.CONNECTION_DISCONNECTED,
+        function (msg) {
+            if(msg)
+                AnalyticsAdapter.sendEvent('connection.disconnected.' + msg);
```
gpolitis: Is there any guarantee that we always have a message when a disconnect happens?

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -84,4 +88,37 @@ JitsiRemoteTrack.prototype._setVideoType = function (type) {
...
+        var type = (this.isVideoTrack() ? 'video' : 'audio');
+
+        var now = window.performance.now();
+        console.log("(TIME) Render " + type + ":\t", now);
+
+        var ttfm = now
+            - (this.conference.getConnectionTimes()["session.initiate"]
+            - this.conference.getConnectionTimes()["muc.joined"]);
+        this.conference.getConnectionTimes()[type + ".ttfm"] = ttfm;
+        console.log("(TIME) TTFM " + type + ":\t", ttfm);
```
gpolitis: Is there any reason why we wouldn't want to log this into the analytics as well? Is it because we have this in callstats?

JitsiConnection.js
damencho: Nope, the case is we eventually can have an error message. And there are cases when you just close the tab and strophe had enough time to close and if there is no check for message you see in the stats event: 'connection.disconnected.undefined'

modules/RTC/JitsiRemoteTrack.js
damencho: Its logged in next commit, let me check, strange.

damencho: Ups, it is coming :)

JitsiConnection.js
gpolitis: My concern is that with the current implementation we don't log when msg is falsy (undefined, null, '', etc.). If we don't care about these cases, then we're good :)

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -84,4 +89,39 @@ JitsiRemoteTrack.prototype._setVideoType = function (type) {
...
+    if (this.isVideoTrack())
+        ttfmTrackerVideoAttached = true;
+
+    // FIXME: this is not working for temasys
+    container.addEventListener("canplay", function () {
+        var type = (this.isVideoTrack() ? 'video' : 'audio');
+
+        var now = window.performance.now();
+        console.log("(TIME) Render " + type + ":\t", now);
+        this.conference.getConnectionTimes()[type + ".render"] = now;
```
gpolitis: Do we care to report render time? (I think we did that before the move.)

damencho: Yeah we don't care for going into analytics. The important one is ttfm, and we also added muc.idle so you can add muc.idle and ttfm and you have it.

damencho: there is no point of having it in the analytics cause it depends how long the first is idle in the muc alone and subsequent participants will have it more close to ttfm.

PR #208, opened 2016-08-11
Analytics update

damencho:


modules/RTC/JitsiRemoteTrack.js
```diff
@@ -107,8 +115,7 @@ JitsiRemoteTrack.prototype._attachTTFMTracker = function (container) {
     if (this.isVideoTrack())
         ttfmTrackerVideoAttached = true;
 
-    // FIXME: this is not working for temasys
-    container.addEventListener("canplay", function () {
+    var playCallback = function () {
```
hristoterezov: Isn't it going to be better if we declare that function outside of _attachTTFMTracker method? WDYT?

PR #210, opened 2016-08-12
Safe ScriptUtil#loadScript() method

tsareg:
Some environments (e.g. React-Native) do not have ```<script>``` tags and do not load JS files in this way, so we need add additional safe checks in "loadScript" method.


lyubomir:
I'm afraid I disagree with this PR. It merely prevents the ScriptUtil#loadScript() from throwing an error. But I find that prevention to be worse than throwing an error because the function still fails to carry out its task and now it fails silently.

PR #212, opened 2016-08-15
Skips selecting participant on stage when alone in a room.

damencho:


tsareg:
Works for me. In React project this will allow us to remove some convoluted try-catch blocks.

JitsiConference.js
```diff
@@ -595,6 +595,12 @@ JitsiConference.prototype.unlock = function () {
  * @throws NetworkError or InvalidStateError or Error if the operation fails.
  */
 JitsiConference.prototype.selectParticipant = function(participantId) {
+    // if there is no participants there is no point in sending message
+    // who is on stage, also in this moment datachannel
+    // may be not yet established
+    if (!this.participants || Object.keys(this.participants).length === 0)
+        return;
```
paweldomas: The docs say "@throws NetworkError or InvalidStateError or Error if the operation fails".
This code just silently does nothing.

I guess that when this method is called the participantId is either "local" or invalid one and that's what we should verify. Or the docs have to be updated that the method will do nothing in that case.

damencho: But this is not correct, cause while in a call with multiple participants we need to handle switching to the local user?
And when there are multiple participants we can still have the errors in the docs.

damencho:
Will create new PR, as discussed offline with Pawel. We need to cache the selected endpoint and send it on opened channel. And in the case when there is no call in progress do not throw an error.

PR #215, opened 2016-08-18
Transcriptions

nikvaessen:
Adds a transcription module 

The transcription module currently used a Sphinx server to send the audio to:
https://github.com/jitsi/Sphinx4-HTTP-server

This code requires the URL of the server to be set in config.js of Jitsi Meet as 
`sphinxURL: "<URL here>"`

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -371,7 +371,8 @@ JitsiConference.prototype.setSubject = function (subject) {
 };
 
 /**
- *
+ * Get a transcriber object for all current participants in this conference
+ * @return {Transcriber} the transcriber object
  */
 JitsiConference.prototype.getTranscriber = function(){
     if(this.transcriber === undefined){
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -18,7 +18,6 @@ var GlobalOnErrorHandler = require("./modules/util/GlobalOnErrorHandler");
 var JitsiConferenceEventManager = require("./JitsiConferenceEventManager");
 var Transcriber = require("./modules/transcription/transcriber");
 var AudioRecorder = require("./modules/transcription/audioRecorder");
-var transcriberHolder = require("./modules/transcription/transcriberHolder");
 
 /**
  * Creates a JitsiConference object with the given name and properties.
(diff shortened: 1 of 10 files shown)


nikvaessen:
Jenkins add to whitelist

damencho:
Adding to whitelist needs to be done by an admin, otherwise everybody can add themselves and run their code.
Jenkins: add to whitelist

bgrozev:
Jenkins ok to test

nikvaessen:
I reset the branch to "88235aa" and added the other changes again, so the new commit should only contain relevant code changes

bgrozev:
A few comments with regard to the glue code in JitsiConference and JitsiMeetJS:

1. Why do we need the transcriberHolder if we can access the transcriber object from the JitsiConference object.
2. Why do we store the audioRecorder in JitsiConference?
3. The sphinx service will throw an error unless there's a global 'config' object with the needed property. I suspect that this is why tests are failing.
4. I don't see where the new tracks created by the library are handled (while the transcriber/recorder is running)

I would suggest that we remove the transcriberHolder and the additional code in JitsiMeetJS, and remove the creation of a recorder/transcriber instance with JitsiConference. Instead, add a getTranscriber() method to JitsiConference, which will create an instance if necessary.

nikvaessen:
> 1. Why do we need the transcriberHolder if we can access the transcriber object from the JitsiConference object.


I added this because I didn't know how to acces it in the console otherwise.

> 2. Why do we store the audioRecorder in JitsiConference?

If we only create the transcriber object when it's requested, how would the transcriber get access to all the tracks, if it's not being stored in the audioRecorder from the start? Also, when a "removed track event" happens in JitsiConferenceEventManger, it needs to tell the audioRecorder that.  
 

> 3.The sphinx service will throw an error unless there's a global 'config' object with the needed 
> property. I suspect that this is why tests are failing.
> 

I changed it so it now just logs the error instead of throwing it, let's see if the test passes now. 

>  4. I don't see where the new tracks created by the library are handled (while the transcriber/recorder is running)


I forgot to write these methods. They should now be commited.

bgrozev:
Thanks for addressing these, Nik.

> If we only create the transcriber object when it's requested, how would the transcriber get access to all the tracks, if it's not being stored in the audioRecorder from the start?

I was thinking that the code which initializes the transcribed will loop through the existing tracks and add them the the transcriber. The recorder can be private to the transcriber.

> Also, when a "removed track event" happens in JitsiConferenceEventManger, it needs to tell the audioRecorder that.

If we only initialize a recorder/transcriber if one is requested, this doesn't need to be handled specially -- if there is an instance, call removeTrack, otherwise don't.

PR #217, opened 2016-08-19
Surrounds code in successcallback with try/catch and call errorcallba…

damencho:
…ck on exception.

If there is a problem with ssrc in successcallback of peerconnection create answer, no callback is called as a result of execution TraceablePeerConnection.createAnswer.

### DIFF
--- a/modules/xmpp/TraceablePeerConnection.js
+++ b/modules/xmpp/TraceablePeerConnection.js
@@ -584,32 +584,45 @@ TraceablePeerConnection.prototype.createAnswer
     this.trace('createAnswer', JSON.stringify(constraints, null, ' '));
     this.peerconnection.createAnswer(
         function (answer) {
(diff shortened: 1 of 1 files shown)


modules/xmpp/TraceablePeerConnection.js
```diff
@@ -584,32 +584,45 @@ TraceablePeerConnection.prototype.createAnswer
...
+
+                successCallback(answer);
+            } catch (e) {
+                // there can be error modifying the answer, for example
+                // for ssrcReplacement there was a track with ssrc that is null
+                // and if we do not catch the error no callback is called
+                // at all
+                self.trace('createAnswerOnError', e);
+                self.trace('createAnswerOnError', dumpSDP(answer));
+                console.error('createAnswerOnError', e, dumpSDP(answer));
```
hristoterezov: Could you please use the logger object?

PR #219, opened 2016-08-22
Adds property to enable/disable reloads.

damencho:
Adds property to enable/disable reloads, by default reloads are disabled.


damencho:
Closing as we will remove reloads code.

PR #221, opened 2016-08-23
Fixes the issue with not able to send feedback after JitsiConference.leave

hristoterezov:


modules/statistics/statistics.js
```diff
@@ -270,8 +270,11 @@ Statistics.prototype.startCallStats = function (session, settings) {
 Statistics.prototype.stopCallStats = function () {
     if(this.callstats) {
         var index = Statistics.callsStatsInstances.indexOf(this.callstats);
-        Statistics.callsStatsInstances.splice(index, 1);
-        this.callstats = null;
+        if(index > -1)
+            Statistics.callsStatsInstances.splice(index, 1);
+        // The next line is commented because we need to be able to send
```
paweldomas: I would rather see new method JitsiConference.dispose() added to JitsiConference and make  .leave() keep some parts of the conference alive(including Call stats) than doing this kind of hacks. But it's only my personal opinion.

hristoterezov: That's good idea! But let's add JitsiConference.dispose() with another PR. Right now if we include only callstats = null, the users of lib-jitsi-meet have to call one more method but they won't get any actual benefits. We have to check what else we are not cleaning and put it there.

PR #225, opened 2016-08-29
Improve device errors

damencho:
Retrieves real device error in the case where we request audio and video and one of them fails, but we receive successful callback from getUserMedia with one of the medias missing.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -1036,11 +1036,35 @@ var RTCUtils = {
                                     devices.push("video");
                                 }
 
-                                reject(new JitsiTrackError(
-                                    { name: "UnknownError" },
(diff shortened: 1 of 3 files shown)


modules/RTC/RTCUtils.js
```diff
@@ -905,11 +914,11 @@ var RTCUtils = {
             this.getUserMedia(constraints,
                 function (stream) {
                     logger.log('onUserMediaSuccess');
-                    setAvailableDevices(um, true);
+                    setAvailableDevices(um, stream);
                     success_callback(stream);
                 },
                 function (error) {
-                    setAvailableDevices(um, false);
+                    setAvailableDevices(um, stream);
```
hristoterezov: We don't have `stream` here. Maybe we'll want to pass `[]` or something else?

PR #228, opened 2016-08-31
Implement audio problem detection

hristoterezov:


JitsiConferenceEventManager.js
```diff
@@ -553,6 +553,17 @@ JitsiConferenceEventManager.prototype.setupStatisticsListeners = function () {
     conference.statistics.addAudioProblemListener(function (ssrc) {
         conference._reportAudioProblem(ssrc);
     });
+
+    conference.statistics.addByteSentStatsListener(function (stats) {
+        var tracks = conference.getLocalTracks();
+        conference.getLocalTracks().forEach(function (track) {
+            var ssrc = track.getSSRC();
```
damencho: tracks variable is not used

modules/RTC/JitsiLocalTrack.js
```diff
@@ -55,6 +55,10 @@ function JitsiLocalTrack(stream, track, mediaType, videoType, resolution,
     // enumerateDevices() list.
     this._trackEnded = false;
 
+    this._bytesSent = null;
+
+    this._testByteSent = true;
+
```
damencho: Maybe add comments what are these used for.

PR #230, opened 2016-09-02
Fixes onended handler issues

hristoterezov:


modules/RTC/JitsiTrack.js
```diff
@@ -79,13 +80,38 @@ function JitsiTrack(conference, stream, track, streamInactiveHandler, trackMedia
...
         }
-        addMediaStreamInactiveHandler(stream, streamInactiveHandler);
     }
+    this.handlers[type] = handler;
+}
+
+/**
+ * Sets the stream property of JitsiTrack object and sets all stored handlers
+ * to it.
+ * @param stream {MediaStream} the new stream.
```
paweldomas: The correct docs syntax will be:
``
@param {MediaStream} stream the new stream
``

PR #232, opened 2016-09-02
Implement interface for adding handlers to MediaStreamTrack object

hristoterezov:


modules/RTC/JitsiLocalTrack.js
```diff
@@ -101,6 +107,19 @@ function JitsiLocalTrack(stream, track, mediaType, videoType, resolution,
 
     RTCUtils.addListener(RTCEvents.DEVICE_LIST_CHANGED,
         this._onDeviceListChanged);
+
+    if(this.isVideoTrack() && this.videoType === VideoType.CAMERA) {
+        this._setHandler("track_mute", function () {
+            if(!this._checkForCameraIssues())
```
paweldomas: Adding return here on reversed condition looks like unnecessary complication and is less readable. I'd rather see it as:
```
if (this._checkForCameraIssues() {
   this.eventEmitter.emit(JitsiTrackEvents.NO_DATA_FROM_SOURCE);
}
```

modules/RTC/JitsiLocalTrack.js
```diff
@@ -101,6 +107,19 @@ function JitsiLocalTrack(stream, track, mediaType, videoType, resolution,
...
+                return;
+            this.eventEmitter.emit(JitsiTrackEvents.NO_DATA_FROM_SOURCE);
+        }.bind(this));
+        this._setHandler("track_ended", function () {
+            if(!this._checkForCameraIssues())
```
paweldomas: The same as above

modules/RTC/JitsiLocalTrack.js
```diff
@@ -515,5 +534,57 @@ JitsiLocalTrack.prototype.getCameraFacingMode = function () {
...
+        // In older version of the spec there is no muted property and
+        // readyState can have value muted. In the latest versions
+        // readyState can have values "live" and "ended" and there is
+        // muted boolean property. If the stream is muted that means that
+        // we aren't receiving any data from the source. We want to notify
+        // the users for error if the stream is muted or ended on it's
+        // creation.
+        if((!("readyState" in track) || track.readyState === "live")
+            && (!("muted" in track) || track.muted === false)) {
+            isReceivingData = true;
```
paweldomas: You can assign the return value of .some() instead. Or am I missing something ?

modules/RTC/RTC.js
```diff
@@ -79,7 +81,17 @@ function RTC(conference, options) {
 RTC.obtainAudioAndVideoPermissions = function (options) {
     return RTCUtils.obtainAudioAndVideoPermissions(options).then(
         function (tracksInfo) {
-            return createLocalTracks(tracksInfo, options);
+            var tracks = createLocalTracks(tracksInfo, options);
+            var allTracksAreReceivingData = true;
+            tracks.some(function (track) {
+                if(!track._isReceivingData()) {
+                    allTracksAreReceivingData = false;
```
paweldomas: The result of !tracks.some can be assigned directly instead.

PR #234, opened 2016-09-05
add polyfills for es6 methods

blackneck:
usage of Array.prototype.find() or String.prototype.startsWith() cause IE to fail.
By @NxTec.

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -1,3 +1,4 @@
+require("babel-polyfill");
 var logger = require("jitsi-meet-logger").getLogger(__filename);
 var AuthUtil = require("./modules/util/AuthUtil");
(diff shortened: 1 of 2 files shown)


blackneck:
Hi!
I have already sign your CLA

JitsiMeetJS.js
```diff
@@ -1,3 +1,4 @@
+require("babel-polyfill");
```
nbermudezs: Maybe add a check if `global._babelPolyfill` is already defined? Requiring `babel-polyfill` multiple times throws an error and it is possible that some other dependency (outside of lib-jitsi-meet) is also doing this.

saghul:
Closing. We're using webpack with the babel loader, which should take care of these now.

PR #235, opened 2016-09-06
Enable lipsync by default + fixes

paweldomas:
This PR enables lipsync by default and removes the option from the conference options.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -203,6 +203,7 @@ This objects represents the server connection. You can create new ```JitsiConnec
             - muc
             - anonymousdomain
         3. useStunTurn -
+        4. enableLipSync - (optional) boolean property which enables the lipsync feature. Currently works only in Chrome and is enabled by default.
 
 2. connect(options) - establish server connection
     - options - JS Object with ```id``` and ```password``` properties.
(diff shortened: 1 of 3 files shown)


doc/API.md
```diff
@@ -51,7 +51,8 @@ The ```options``` parameter is JS object with the following properties:
     10. disableSimulcast - boolean property. Enables/disables simulcast.
     11. enableWindowOnErrorHandler - boolean property (default false). Enables/disables attaching global onerror handler (window.onerror).
     12. disableThirdPartyRequests - if true - callstats will be disabled and the callstats API won't be included.
-    13. analyticsScriptUrl - (optional) custom url to search for the analytics lib, if missing js file will be expected to be next to the library file (the location it is sourced from) 
+    13. analyticsScriptUrl - (optional) custom url to search for the analytics lib, if missing js file will be expected to be next to the library file (the location it is sourced from)
+    14. enableLipSync - (optional) boolean property which enables the lipsync feature. Currently works only in Chrome and is enabled by default.
```
hristoterezov: Isn't this option related to the JitsiConnection object not JitsiMeetJS.init?

PR #238, opened 2016-09-08
Adds a option to set custom callstats script.

damencho:


doc/API.md
```diff
@@ -51,7 +51,8 @@ The ```options``` parameter is JS object with the following properties:
     10. disableSimulcast - boolean property. Enables/disables simulcast.
     11. enableWindowOnErrorHandler - boolean property (default false). Enables/disables attaching global onerror handler (window.onerror).
     12. disableThirdPartyRequests - if true - callstats will be disabled and the callstats API won't be included.
-    13. analyticsScriptUrl - (optional) custom url to search for the analytics lib, if missing js file will be expected to be next to the library file (the location it is sourced from) 
+    13. analyticsScriptUrl - (optional) custom url to search for the analytics lib, if missing js file will be expected to be next to the library file (the location it is sourced from)
+    14. callStatsCustomScriptUrl - (optional) custom url to access callstats client script
```
paweldomas: Is it possible to override this property with #config.analyticsScriptUrl on Meet ?

PR #242, opened 2016-09-09
Talk while muted

damencho:


JitsiConference.js
```diff
@@ -18,6 +18,7 @@ var GlobalOnErrorHandler = require("./modules/util/GlobalOnErrorHandler");
 var JitsiConferenceEventManager = require("./JitsiConferenceEventManager");
 var VideoType = require('./service/RTC/VideoType');
 var Transcriber = require("./modules/transcription/transcriber");
+var TalkMutedDetection = require("./modules/talkmuted/TalkMutedDetection");
```
lyubomir: I realize the following is not your fault because you created your PR before the introduction of ES6 in lib-jitsi-meet. However, I'd really like us to start taking advantage of ES6 features. Since rewriting the whole library in one go is unrealistic, I believe we could try to use ES6 in new PRs, especially ones adding new functionality. I've rewritten this PR using ES6 features and I'll push my modifications after publishing this note. The goal of this comment is to get my thoughts to you for future PRs.

JitsiConferenceEvents.js
```diff
@@ -156,3 +156,7 @@ export const USER_ROLE_CHANGED = "conference.roleChanged";
  * User status changed.
  */
 export const USER_STATUS_CHANGED = "conference.statusChanged";
+/**
+ * Event indicates that local user is talking while he muted himself
+ */
+export const TALK_WHILE_MUTED = "conference.talk_while_muted";
```
lyubomir: I prefer alphabetical sorting because it allows easy searching in the fashion of searching through a phone book. Since (1) I've sorted the existing constants in alphabetical order and (2) I see no advantage to adding `TALK_WHILE_MUTED` at the bottom, I'll move it.

modules/statistics/RTPStatsCollector.js
```diff
@@ -749,9 +749,12 @@ StatsCollector.prototype.processAudioLevelReport = function () {
...
         var now = this.currentAudioLevelsReport[idx];
 
-        //if we don't have "packetsReceived" this is local stream
-        if (now.type != 'ssrc' || !getStatValue(now, 'packetsReceived')) {
+        if (now.type != 'ssrc')
             continue;
+
+        var isLocal = false;
+        if (!getStatValue(now, 'packetsReceived')) {
+            isLocal = true;
```
lyubomir: That's a very long way of saying `var isLocal = !getStatValue(now, 'packetsReceived');`.

JitsiConference.js
```diff
@@ -458,6 +472,9 @@ JitsiConference.prototype.addTrack = function (track) {
                 this.room.setVideoMute(track.isMuted());
             }
 
+            if (this.talkMutedDetection && track.isAudioTrack())
+                this.talkMutedDetection.addTrack(track);
+
```
lyubomir: I don't like that because it gets us further from the goal of having `TalkMutedDetection` as a black box. Since `this.talkMutedDetection` is already listening to `TRACK_MUTE_CHANGED`, cannot it listen to `TRACK_ADDED` instead of the above?

service/statistics/Events.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
     /**
-     * FIXME: needs documentation.
+     * Notifies for audio level in rtp statistics by ssrc.
```
lyubomir: Thank you for adding documentation!

When documenting events, I'd like to have documentation on the parameters carried by the event as well. Otherwise, one has to search through the source code itself in order to determine what arguments and in what order the listener function is to expect.

JitsiConference.js
```diff
@@ -116,6 +117,19 @@ JitsiConference.prototype._init = function (options) {
...
+        this.talkMutedDetection
+            = new TalkMutedDetection(() => {
+                this.eventEmitter.emit(JitsiConferenceEvents.TALK_WHILE_MUTED);
+            });
+        this.statistics.addAudioLevelListener(
+            this.talkMutedDetection.audioLevelListener
+                .bind(this.talkMutedDetection));
+        this.eventEmitter.on(
+            JitsiConferenceEvents.TRACK_MUTE_CHANGED,
+            this.talkMutedDetection.muteChanged.bind(this.talkMutedDetection));
```
lyubomir: I think that `TalkMutedDetection` should receive this `JItsiConference` in its constructor, add the listeners it's interested in and, thus, move closer to a black box that `JitsiConference` merely activates by `new TalkMutedDetection`.

JitsiConference.js
```diff
@@ -116,6 +117,19 @@ JitsiConference.prototype._init = function (options) {
     // Always add listeners because on reload we are executing leave and the
     // listeners are removed from statistics module.
     this.eventManager.setupStatisticsListeners();
+
+    if (this.options.config.enableTalkWhileMuted) {
+        this.talkMutedDetection
+            = new TalkMutedDetection(() => {
+                this.eventEmitter.emit(JitsiConferenceEvents.TALK_WHILE_MUTED);
```
lyubomir: Why is `TALK_WHILE_MUTED` an event of `JitsiConference`? Reading `TalkMutedDetection`, it looks to me like it's mostly about the local audio track so I'm wondering why `TALK_WHILE_MUTED` is not an event of the local audio track.

modules/talkmuted/TalkMutedDetection.js
```diff
@@ -0,0 +1,57 @@
...
+    /**
+     * Mute changed for a track.
+     * @param track the track which mute state has changed.
+     */
+    muteChanged(track) {
+        if (!track.isLocal() || !track.isAudioTrack())
+            return;
+
+        if (track.isMuted())
+            this.eventFired = false;
```
lyubomir: 1. Obviously, the total number of lines above can be reduced:

```` javascript
if (track.isLocal() && track.isAudioTrack() && track.isMuted())
    this.eventFired = false;
````

Since you haven't done that, I'm thinking that you have a reason. The only reason that I see is that you want to make it clear, explicit that regardless of what `muteChanged` does, it will not do it for non-local, non-audio tracks. But then isn't that a condition for the whole `TalkMutedDetection`? If it is, why isn't `addTrack` using the same condition and it's using just a part of it instead? I find such differences to be adding complexity because one has to go and look at the method call sites to determine that indeed the condition will be met.

File: JitsiConference.js [hunk 9]
```diff
@@ -116,6 +117,19 @@ JitsiConference.prototype._init = function (options) {
...
+            });
+        this.statistics.addAudioLevelListener(
+            this.talkMutedDetection.audioLevelListener
+                .bind(this.talkMutedDetection));
```
lyubomir: 1. `JitsiConference`'s `statistics` doesn't appear to be (totally & absolutely) private. It's at least internal, package private (in the terms of Java) because `JitsiConferenceEventManager` uses it. I wouldn't stop you from accessing it inside `TalkMutedDetection` in order to add a listener. I realize though that it would add a dependency on top of accessing a property that is possibly intended to be private but... enlarging `JitsiConference` with each new feature...

2. However, why do you have to listen to `statistics` at all. Shouldn't you be able to get notified about audio level changes from the track itself?

modules/talkmuted/TalkMutedDetection.js
```diff
@@ -0,0 +1,72 @@
...
+        }
+    }
+
+    /**
+     * Adds local tracks. We are interested only in the audio one.
+     * @param track
+     * @private
+     */
+    _trackAdded(track) {
+        if (!track.isAudioTrack())
```
lyubomir: `JitsiConference.TRACK_ADDED` is fired for both local and remote tracks. Given that `TalkMutedDetection` works for the local audio track only, I don't understand why you're only checking whether the track is an audio track. Why are you not checking whether it's a local track?

damencho: When we were adding the track to talk while muted feature from conference it was doing it only for local tracks. I forgot to update it once I moved the adding of the track inside the feature. It is now fixed.

PR #254, opened 2016-09-20
Sends non cached values only if datachannel is open.

damencho:


modules/RTC/RTC.js
```diff
@@ -127,6 +130,7 @@ RTC.prototype.onCallEnded = function() {
         // The reference is cleared to disable any logic related to the data
         // channels.
         this.dataChannels = null;
+        this.dataChannelsOpen = false;
```
paweldomas: It can happen that the data channel is closed during the call. Then we'll probably fail with en error when trying to send and maybe we would be fine with that, but this flag will not be accurate.

PR #256, opened 2016-09-21
Improve camera/mic issue detection

hristoterezov:
- Adds logging to callstats and analytics

- Waiting for 3s after muted event before reporting camera problem. If unmute event is received in this period of time the problem won't be reported.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -10,6 +10,7 @@ var RTCUtils = require("./RTCUtils");
 var MediaType = require('../../service/RTC/MediaType');
 var VideoType = require('../../service/RTC/VideoType');
 var CameraFacingMode = require('../../service/RTC/CameraFacingMode');
+var Statistics = require("../statistics/statistics");
 
 /**
  * Represents a single media track(either audio or video).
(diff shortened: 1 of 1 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -79,6 +80,13 @@ function JitsiLocalTrack(stream, track, mediaType, videoType, resolution,
      */
     this.stopStreamInProgress = false;
 
+    /**
+     * On mute event we are waiting for 3s to check if the stream is going to
+     * be still muted before firing the event for camera issue detected
+     * (NO_DATA_FROM_SOURCE).
+     */
+    this._noDataFromSourceTimout = null;
```
lyubomir: The word 'Timeout' is misspelled as 'Timout'.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -134,6 +131,48 @@ JitsiLocalTrack.prototype.isEnded = function () {
...
+                let now = window.performance.now();
+                this._noDataFromSourceTimout = setTimeout(() => {
+                    this._noDataFromSourceTimout = null;
+                    //unset the handler
+                    this._setHandler("track_unmute", undefined);
+                    if(this._checkForCameraIssues())
+                        this._fireNoDataFromSourceEvent();
+                }, 3000);
+                this._setHandler("track_unmute", () => {
+                    if(this._noDataFromSourceTimout) {
```
lyubomir: Since the `track_unmute` handler is unset after `this._noDataFromSourceTimout = null`, the check `this._noDataFromSourceTimout` is superfluous and forces another level of indentation. I think the source code will be simplified by removing that check.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -134,6 +131,48 @@ JitsiLocalTrack.prototype.isEnded = function () {
...
+                        Statistics.sendEventToAll(
+                            this.getType() + ".track_unmute",
+                            window.performance.now() - now);
+                    }
+                });
+            }
+        });
+        this._setHandler("track_ended", () => {
+            if(this._checkForCameraIssues())
+                this._fireNoDataFromSourceEvent();
```
lyubomir: If `track_ended` arrives within three seconds of `track_mute`, it looks to me like the above piece of source code will be executed twice: on `track_ended` and when `_noDataFromSourceTimout` fires. I think this can be avoided and the source code can be simplified by having `track_ended` and `setTimeout` execute one and the same function.

PR #259, opened 2016-09-26
Adds participant connection status notifications

paweldomas:
This PR adds PARTICIPANT_CONN_STATUS_CHANGED conference event that will be fired when JVB sends notification over the data channel the the user is having problems with media connection.

On Chrome 'onmute' and 'onunmute' events are fired whenever the video tracks stops receiving data and the video freezes. We take those into account and trigger participant connection status updates when those events get out of sync with the video muted signalling.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -340,7 +340,7 @@ JitsiLocalTrack.prototype._setMute = function (mute) {
             return self._sendMuteStatus(mute);
         })
         .then(function() {
-            self.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED);
+            self.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED, this);
         });
 };
 
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -85,7 +85,7 @@ JitsiRemoteTrack.prototype.setMute = function (value) {
         this.stream.muted = value;
 
     this.muted = value;
-    this.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED);
+    this.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED, this);
 };
 
 /**
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -1,6 +1,10 @@
+/* global __filename, module */
+var logger = require("jitsi-meet-logger").getLogger(__filename);
+
 var JitsiTrack = require("./JitsiTrack");
 import * as JitsiTrackEvents from "../../JitsiTrackEvents";
 var RTCBrowserType = require("./RTCBrowserType");
+var RTCEvents = require("../../service/RTC/RTCEvents");
 var Statistics = require("../statistics/statistics");
 var AdapterJS = require("./adapter.screenshare");
 
@@ -9,7 +13,7 @@ var ttfmTrackerVideoAttached = false;
 
 /**
  * Represents a single media track (either audio or video).
- * @param RTC the rtc instance.
+ * @param rtc {RTC} the RTC service instance.
  * @param ownerJid the MUC JID of the track owner
  * @param stream WebRTC MediaStream, parent of the track
  * @param track underlying WebRTC MediaStreamTrack for new JitsiRemoteTrack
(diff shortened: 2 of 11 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -340,7 +340,7 @@ JitsiLocalTrack.prototype._setMute = function (mute) {
             return self._sendMuteStatus(mute);
         })
         .then(function() {
-            self.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED);
+            self.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED, this);
```
hristoterezov: Why do you need this parameter? You have to attach the listener to the track and this means that the listener knows which is the track.

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -19,22 +23,52 @@ var ttfmTrackerVideoAttached = false;
...
 JitsiRemoteTrack.prototype.constructor = JitsiRemoteTrack;
 
+JitsiRemoteTrack.prototype._bindMuteHandlers = function() {
+    // Bind 'onmute'
+    // FIXME it would be better to use recently added '_setHandler' method, but
+    // 1. It does not allow to set more than one handler to the event
+    // 2. It does mix MediaStream('inactive') with MediaStreamTrack events
+    // 3. Allowing to bind more than one event handler requires too much
+    //    refactoring around camera issues detection.
+    this.track.addEventListener('mute', function () {
```
hristoterezov: I think here we should use the _setHandler method. We don't need currently multiple handlers because we are adding only one here.

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -19,22 +23,52 @@ var ttfmTrackerVideoAttached = false;
...
+
+        logger.debug(
+            '"onmute" event(' + Date.now() + '): ',
+            this.getParticipantId(), this.getType(), this.getSSRC());
+
+        this.rtc.eventEmitter.emit(RTCEvents.REMOTE_TRACK_MUTE, this);
```
hristoterezov: I don't see why we are using the event emitter instance from rtc here. I think we should use the event emitter from JitsiRemoteTrack. That way we won't need the rtc instance and we can remove it from the constructor.

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -19,22 +23,52 @@ var ttfmTrackerVideoAttached = false;
...
+    }.bind(this));
+
+    // Bind 'onunmute'
+    this.track.addEventListener('unmute', function () {
+
+        logger.debug(
+            '"onunmute" event(' + Date.now() + '): ',
+            this.getParticipantId(), this.getType(), this.getSSRC());
+
+        this.rtc.eventEmitter.emit(RTCEvents.REMOTE_TRACK_UNMUTE, this);
```
hristoterezov: I don't see why we are using the event emitter instance from rtc here. I think we should use the event emitter from JitsiRemoteTrack. That way we won't need the rtc instance and we can remove it from the constructor.

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -51,7 +85,7 @@ JitsiRemoteTrack.prototype.setMute = function (value) {
         this.stream.muted = value;
 
     this.muted = value;
-    this.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED);
+    this.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED, this);
```
hristoterezov: Why do you need this parameter? You have to attach the listener to the track and this means that the listener knows which is the track.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -0,0 +1,306 @@
...
+
+/**
+ * Class is responsible for emitting
+ * JitsiConferenceEvents.PARTICIPANT_CONN_STATUS_CHANGED events.
+ *
+ * @constructor
+ * @param rtc {RTC} the RTC service instance
+ * @param conference {JitsiConference} parent conference instance
+ */
+function ParticipantConnectionStatus(rtc, conference) {
```
hristoterezov: It's going to be better if we implement this class with the ES6 syntax since we support it now.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -0,0 +1,306 @@
...
+ * @param {JitsiParticipant} participant to be checked for muted video tracks
+ *
+ * @return {boolean} <tt>true</tt> if given <tt>participant</tt> contains any
+ * video <tt>MediaStreamTrack</tt>s muted according to their 'muted' field.
+ */
+var hasRtcMutedVideoTrack = function (participant) {
+    return participant.getTracks().some(function(jitsiTrack) {
+        var rtcTrack = jitsiTrack.getTrack();
+        return jitsiTrack.getType() === MediaType.VIDEO
+            && rtcTrack && rtcTrack.muted === true;
```
hristoterezov: I think this check should better fit as a method of JitsiTrack or JitsiRemoteTrack object instead of accessing rtcTrack.muted property which should be private.

modules/RTC/JitsiRemoteTrack.js
paweldomas: Because it makes implementation of such listeners harder and single listener instance can not be reused. I basically have to create separate listener instance for every track and store it somewhere in order to unregister it later. Like here for example I have one instance, but would have to create some map or whatever to store all of these:
https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/connectivity/ParticipantConnectionStatus.js#L236

Without passing the source object event listener interface feels incomplete.

modules/RTC/JitsiLocalTrack.js
paweldomas: Because it makes implementation of such listeners harder and single listener instance can not be reused. I basically have to create separate listener instance for every track and store it somewhere in order to unregister it later. Like here for example I have one instance, but would have to create some map or whatever to store all of these:
https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/connectivity/ParticipantConnectionStatus.js#L236

Without passing the source object event listener interface feels incomplete.

modules/RTC/JitsiRemoteTrack.js
paweldomas: If I would use _setHandler, then the person who will want to add some new unrelated logic which needs 'mute' event would have to refactor this code instead of adding new listener. Let's say you want to add some logic in base class JitsiTrack. If I have used _setHandler here, then you will have to refactor JitsiRemoteTrack.

modules/RTC/JitsiRemoteTrack.js
paweldomas: This is RTC event and all RTC events are currently triggered by the RTC service. Just tried to follow the convention.

modules/RTC/JitsiRemoteTrack.js
paweldomas: This is RTC event and all RTC events are currently triggered by the RTC service. Just tried to follow the convention.

paweldomas:
> Isn't it going to be better to fire the event JitsiConferenceEvents.PARTICIPANT_CONN_STATUS_CHANGED from the participant instead of the conference? Also if we create the instance of the class in JitsiParticipant and pass to the constructor the participant instead of conference I think the code is going to be simpler and we are going to reduce a lot of dependencies and code. Also this class seems to me logically more related to JitsiParticipant than JitsiConference.

Currently we don't have any "participant events". It's easy to find bunch of other events which are clearly related to participant, yet all of them are conference events: USER_ROLE_CHANGED, USER_STATUS_CHANGED, TRACK_ADDED, TRACK_REMOVED. So I just tried to follow the convention. If you want I can move all of them, but this modifies public library API and would rather not do that now.

modules/RTC/JitsiLocalTrack.js
hristoterezov: You can also use bind or function to pass the remote track there. But you are right that this way is simpler.

modules/RTC/JitsiRemoteTrack.js
hristoterezov: But JitsiRemoteTrack is already interface class already. For example`JitsiTrackEvents.TRACK_MUTE_CHANGED` is fired from there and it is public. So I think here you can fire `RTCEvents.REMOTE_TRACK_UNMUTE `. Maybe you'll have to rename the event somehow and move it from RTCEvents.

paweldomas: RTCEvents are library private events. RTCEvents.REMOTE_TRACK_MUTE is designed to be library private event as well as it's triggered when the muted status of the underlying WebRTC MediaStreamTrack changes. JitsiTrackEvents.TRACK_MUTE_CHANGED is library public event which takes into account XMPP muted signalling and is meant to be consumed by the app which uses the lib-jitsi-meet.

I don't see why we would want to expose RTCEvents.REMOTE_TRACK_MUTE in the public API. It would be confusing.

hristoterezov: I meant it can be again private event, I'm just proposing to fire it from JitsiRemoteTrack's event emitter. And to do this probably you should move it from RTCEvents file and rename it.

PR #264, opened 2016-09-28
Fires CONFERENCE_LEFT event once we receive our presence unavailable.

damencho:


modules/xmpp/ChatRoom.js
```diff
@@ -510,10 +513,14 @@ ChatRoom.prototype.onPresenceUnavailable = function (pres, from) {
             delete this.members[i];
             this.onParticipantLeft(i, member.isFocus);
         }
+        this.connection.emuc.doLeave(this.roomjid);
+        this.eventEmitter.emit(XMPPEvents.MUC_LEFT);
```
paweldomas: Is it possible that when we're being kicked out both MUC_LEFT and KICKED events will be triggered ?

damencho: Yes you are right, the presence when kicked out is having both <status code='307'/> and <status code='110'/> so we will fire both. 
Will reorder those if statements.

damencho: Changed that, so now we process the 110 as we did before and fire muc_left only if it is not kick and then process the kick part.

PR #273, opened 2016-10-04
Set the FPS constraints only when video has been requested

dmurtari:
Since FPS only applies when video is requested, move this block to where the rest of the video constraints are applied.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -172,6 +172,18 @@ function getConstraints(um, options) {
             });
         }
 
+        if (options.minFps || options.maxFps || options.fps) {
+            // for some cameras it might be necessary to request 30fps
+            // so they choose 30fps mjpg over 10fps yuy2
+            if (options.minFps || options.fps) {
+                options.minFps = options.minFps || options.fps; //Fall back to options.fps for backwards compatibility
+                constraints.video.mandatory.minFrameRate = options.minFps;
+            }
+            if (options.maxFps) {
+                constraints.video.mandatory.maxFrameRate = options.maxFps;
+            }
+        }
+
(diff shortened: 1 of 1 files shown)


dmurtari:
I believe our organization has already signed the CLA (@dwilson6 will be able to verify).

dwilson6:
Yes, we have signed the CLA

dwilson6:
@bgrozev any problems with this one?  I'm trying to get some of our older PRs merged to make it easier to pull in new versions of lib-jitsi-meet.

bgrozev:
Jenkins ok to test

bgrozev:
I don't understand this well enough to merge. Does it solve a problem for you? Does this not affect screen sharing, too?

With the code in the new position, the "if (!constraints.video)" block is unnecessary, so please remove it.

dmurtari:
This solves an issue we were seeing with redundant permission requests in Firefox, when requesting video/audio together with FPS constraints on the video. Since the behavior in `obtainAudioAndVideoPermissions()` is to split a video/audio GUM request into two separate requests for video and audio, we were seeing the FPS constraints passed in to the audio requests. This caused Firefox to prompt for both video/audio instead of just audio since the current behavior is to add back the video constraint if it does not exist when the FPS constraint is set, even if only audio is being requested (I hope that scenario makes sense). 

Screen sharing FPS isn't impacted, so should still be 3. 

I think I'm missing the `if (!constraints.video)` that you're referring to? There was one in the FPS block originally but I did get rid of that one.

bgrozev:
> I think I'm missing the if (!constraints.video) that you're referring to?

Yeah, I must have misread, sorry.

bgrozev:
Thanks for the fix!

dmurtari:
No problem, thank you!

PR #275, opened 2016-10-06
feat(analytics): Add support for permanent properties and new parameters format

hristoterezov:


modules/statistics/AnalyticsAdapter.js
```diff
@@ -15,63 +16,40 @@ AnalyticsAdapter.eventsQueue = [];
  * Sends analytics event.
  * @param action
  * @param data
```
paweldomas: So now "data" became some structure. I would love to see some documentation on what it is.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -195,7 +195,8 @@ ParticipantConnectionStatus.prototype._changeConnectionStatus
             }));
 
         // and analytics
-        Statistics.analytics.sendEvent('peer.conn.status', null, newStatus);
+        Statistics.analytics.sendEvent('peer.conn.status',
+            {value: null, label: newStatus});
```
lyubomir: I guess that with the old API we had to specify a `value` of `null` because we didn't care about it but wanted to specify a `label` that we cared about. Why do we specify the `value` with the new API though?

modules/statistics/AnalyticsAdapter.js
```diff
@@ -5,6 +5,7 @@ NoopAnalytics.prototype.sendEvent = function () {};
 
 function AnalyticsAdapter() {
     this.browserName = RTCBrowserType.getBrowserName();
+    this.permanentProperties = {};
```
lyubomir: I don't understand what permanent properties are. Technically, `this.permanentProperties` declares an equivalent of a Java class field. Then we're kind of supposed to document it with jsdocs anyway.

modules/statistics/AnalyticsAdapter.js
hristoterezov: Fixed!

modules/connectivity/ParticipantConnectionStatus.js
hristoterezov: Fixed!

modules/statistics/AnalyticsAdapter.js
hristoterezov: Fixed!

modules/statistics/AnalyticsAdapter.js
paweldomas: Thanks !

PR #279, opened 2016-10-07
Peer conn status comments

paweldomas:
Addresses comments from https://github.com/jitsi/lib-jitsi-meet/pull/259


paweldomas:
Changes included in https://github.com/jitsi/lib-jitsi-meet/pull/292

PR #284, opened 2016-10-12
Fix JitsiConference.leave

hristoterezov:
Fixes when the promise is resoved and rejected.
Removes unnecessary removal of local streams and participants.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -958,10 +958,29 @@ ChatRoom.prototype.onMute = function (iq) {
 
 /**
  * Leaves the room. Closes the jingle session.
+ * @returns {Promise} which is resolved if XMPPEvents.MUC_LEFT is received less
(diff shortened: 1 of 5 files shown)


modules/xmpp/ChatRoom.js
```diff
@@ -958,10 +958,29 @@ ChatRoom.prototype.onMute = function (iq) {
...
+            clearTimeout(timeout);
+            if(doReject) {
+                // the timeout expired
+                reject(new Error("The timeout for the confirmation about " +
+                    "leaving the room expired."));
+            } else {
+                resolve();
+            }
+        }
+        eventEmitter.on(XMPPEvents.MUC_LEFT, onMucLeft);
```
hristoterezov: I think assigning to a local variable wouldn't make it more readable and when you are debugging or logging errors from it the function is going to be displayed as anonymous function.

I agree that defining the function before using it looks confusing but the function is using the timeout variable and event emitter which are defined just before the function. So I don't think that moving the function will make it less confusing.

PR #292, opened 2016-10-20
Fix participant connection status stuck after device switch

paweldomas:
First part of the PR applies some of the comments for the original PR:
https://github.com/jitsi/lib-jitsi-meet/pull/259
The second part addresses an issue where participant's connection status would end up stuck in "disconnected" on device switch/screen sharing(sometimes).

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -18,8 +18,8 @@ var GlobalOnErrorHandler = require("./modules/util/GlobalOnErrorHandler");
 var JitsiConferenceEventManager = require("./JitsiConferenceEventManager");
 var VideoType = require('./service/RTC/VideoType');
 var Transcriber = require("./modules/transcription/transcriber");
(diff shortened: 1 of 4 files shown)


hristoterezov:
I think currently the code is very complicated and this PR is making it even more complicated. My proposal is to change it if it is possible to something like this:

- Implement a method that is collecting the data needed (mute statuses, the video bridge data, etc) and than the method should decide what is the connection status for the participant and fire the event if needed.

- All the events that the module is subscribing for should result in calling the method described above.

I'm not sure if it is possible to implement such a method that follows the current logic but if it is - that will simplify the code and the logic a lot.

For example onTrackRtcUnmuted and onTrackRtcMuted look very similar and we can combine them easily into one method.

paweldomas:
@hristoterezov yes, that sounds like a good idea to make this complex thing simpler

PR #302, opened 2016-11-02
Cleans the statistics code a bit

bgrozev:


bgrozev:
jenkins: ok to test

modules/statistics/RTPStatsCollector.js
```diff
@@ -101,7 +101,7 @@ function acceptReport(id, type) {
 }
 
 /**
- * Peer statistics data holder.
+ * Holds "statistics" for a single SSRC.
```
paweldomas: If it holds the statistics for a single SSRC then the class name should be changed from the PeerStats to something that reflect that fact(I'm pasting this comment without seeing next commits, so it's kind of uniformed comment)

PR #308, opened 2016-11-08
Adds a getParticipantCount function, cleanup.

bgrozev:


JitsiConference.js
```diff
@@ -843,13 +843,16 @@ function (jingleSession, jingleOffer, now) {
...
     // add info whether call is cross-region
     var crossRegion = null;
-    if (window.jitsiRegionInfo)
+    if (window.jitsiRegionInfo) {
         crossRegion = window.jitsiRegionInfo["CrossRegion"];
-    Statistics.analytics.sendEvent("session.initiate",{
+    }
+    Statistics.analytics.sendEvent(
+        "session.initiate",
+        {
```
paweldomas: Nope, we use different braces alignment in JS, the opening one should be on the previous line

paweldomas:
Other than this one minor comment it looks good to me

JitsiConference.js
bgrozev: For some reason I thought it was different for object literals. Will fix (although I find it confusing).

PR #309, opened 2016-11-09
Fixes instanceof JitsiTrackError

hristoterezov:
JitsiTrackError extends Error. Due to an issue with babel, extending built-in classes seems not very good idea(https://github.com/babel/babel/issues/3083). 

This PR fixes `instanceof JitsiTrackError` calls.

### DIFF
--- a/JitsiTrackError.js
+++ b/JitsiTrackError.js
@@ -29,108 +29,110 @@ TRACK_ERROR_TO_MESSAGE_MAP[JitsiTrackErrors.TRACK_MUTE_UNMUTE_IN_PROGRESS]
 TRACK_ERROR_TO_MESSAGE_MAP[JitsiTrackErrors.NO_DATA_FROM_SOURCE]
     = "The track has stopped receiving data from it's source";
 
+
+// FIXME: Using prototype inheritance because otherwise instanceof is not
+// working properly (see https://github.com/babel/babel/issues/3083)
+
 /**
+ *
  * Represents an error that occurred to a JitsiTrack. Can represent various
  * types of errors. For error descriptions (@see JitsiTrackErrors).
  *
  * @extends Error
+ *
+ *
+ * @constructor
+ * @param {Object|string} error - error object or error name
+ * @param {Object|string} (options) - getUserMedia constraints object or
+ * error message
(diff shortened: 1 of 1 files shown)


JitsiTrackError.js
```diff
@@ -29,109 +29,111 @@ TRACK_ERROR_TO_MESSAGE_MAP[JitsiTrackErrors.TRACK_MUTE_UNMUTE_IN_PROGRESS]
 TRACK_ERROR_TO_MESSAGE_MAP[JitsiTrackErrors.NO_DATA_FROM_SOURCE]
     = "The track has stopped receiving data from it's source";
 
+
+// FIXME: Using prototype inheritance because otherwise instanceof is not
```
paweldomas: The comment is a bit misleading about the bug. From the issues description one can figure out that it's not recommended to extend a built-in type (and so does the PR description), but the comment looks like we're waiting for the bug to be fixed.

JitsiTrackError.js
```diff
@@ -29,109 +29,111 @@ TRACK_ERROR_TO_MESSAGE_MAP[JitsiTrackErrors.TRACK_MUTE_UNMUTE_IN_PROGRESS]
...
+// working properly (see https://github.com/babel/babel/issues/3083)
+
 /**
+ *
  * Represents an error that occurred to a JitsiTrack. Can represent various
  * types of errors. For error descriptions (@see JitsiTrackErrors).
  *
  * @extends Error
```
paweldomas: It does not extend the Error anymore, is it ?

Do we have any usages of JitsiTrackError that rely on that fact ?

hristoterezov: It still extends Error. See the lines after the constructor:
```
JitsiTrackError.prototype = Object.create(Error.prototype);
JitsiTrackError.prototype.constructor = JitsiTrackError;
```

I'm not sure where and why we are using this but even if we are not using it I think it's good to extend `Error` since we are reporting error.

paweldomas: ok, makes sense, thanks for the explanation !

JitsiTrackError.js
paweldomas: No, I got the code wrong and thought that we will not longer extend Error - it's ok as it is.

PR #312, opened 2016-11-11
Suspend video

bgrozev:


paweldomas:
Hmm how does it work ? If we're not doing signalling updates then will not it trigger video track muted event on the receiver's side ?

modules/RTC/RTCUtils.js
```diff
@@ -821,16 +821,13 @@ var RTCUtils = {
                             : SDPUtil.filter_special_chars(id));
                 };
                 // DTLS should now be enabled by default but..
```
hristoterezov: Maybe the comment should be removed too?

bgrozev:
Video is stopped if the bandwidth estimation for the sender is below the minimum (something very low, I think 30kbps). It will trigger no events on the receiver.

modules/RTC/RTCUtils.js
```diff
@@ -821,16 +821,13 @@ var RTCUtils = {
...
-                this.pc_constraints = {'optional': [
-                    {'DtlsSrtpKeyAgreement': 'true'}
```
hristoterezov: Why do we need this change?

modules/RTC/RTCUtils.js
```diff
@@ -821,16 +821,13 @@ var RTCUtils = {
...
-                ]};
+                this.pc_constraints = {'optional': [] };
+
                 if (options.useIPv6) {
                     // https://code.google.com/p/webrtc/issues/detail?id=2828
                     this.pc_constraints.optional.push({googIPv6: true});
                 }
-                if (RTCBrowserType.isAndroid()) {
```
hristoterezov: Why do we need this change?

modules/RTC/RTCUtils.js
hristoterezov: Sorry I didn't noticed!

modules/RTC/RTCUtils.js
hristoterezov: I don't see problem but I haven't tested it. I wasn't sure what's the reason behind this removal. But why we are removing "DtlsSrtpKeyAgreement"? Something changed on the latest versions?

bgrozev: Not in recent versions. I don't know when it was last required to use this constraint, but it wasn't recently.

modules/RTC/RTCUtils.js
bgrozev: See below.

PR #313, opened 2016-11-15
Connection problems indication stalled when video muted

paweldomas:
This PR fixes issue when video muted signalling arrive after the timeout which adjusts user's status to disconnected. It also refactors the logic from event based like advised during https://github.com/jitsi/lib-jitsi-meet/pull/292 review.

### DIFF
--- a/modules/connectivity/ParticipantConnectionStatus.js
+++ b/modules/connectivity/ParticipantConnectionStatus.js
@@ -58,6 +58,28 @@ export default class ParticipantConnectionStatus {
         this.rtcMuteTimeout
             = typeof rtcMuteTimeout === 'number'
                 ? rtcMuteTimeout : DEFAULT_RTC_MUTE_TIMEOUT;
+        /**
+         * This map holds a timestamp indicating  when participant's video track
+         * was RTC muted (it is assumed that each participant can have only 1
+         * video track at a time). The purpose of storing the timestamp is to
+         * avoid the transition to disconnected status in case of legitimate
+         * video mute operation where the signalling video muted event can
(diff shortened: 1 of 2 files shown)


modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -271,70 +279,87 @@ export default class ParticipantConnectionStatus {
...
-        if (!isConnectionActive && isConnActiveByJvb && !hasAnyVideoRTCMuted) {
-            // FIXME adjust the log level or remove the message completely once
-            // the feature gets mature enough.
-            logger.info(
-                "Remote track removed for disconnected" +
-                " participant, when the status according to" +
-                " the JVB is connected. Adjusting to the JVB value.");
-            this._changeConnectionStatus(endpointId, isConnActiveByJvb);
+        var isConnectionActive = true;
+        if (!isVideoMuted && (isVideoTrackFrozen || !isConnActiveByJvb)) {
```
paweldomas: Probably it is, but I can't easily confirm that this one line statement is the same thing, so I find it more readable in separate "ifs". That's because there are 2 major cases when connection is considered "not active" and writing two separately allowed me to comment on each one of them. Maybe the comments could be improved though...

hristoterezov: Ok. I see now. 

I think the comments are OK.

So don't change it. Maybe this way is really more clear than the other.

paweldomas: After having more input "off the list" from other team members I will be changing it to the one line solution as more of the people find it easier to read.

PR #321, opened 2016-11-24
Configure babel the same as jitsi-meet

lyubomir:


lyubomir:
Jenkins, test this please!

PR #326, opened 2016-12-10
Add karma and jasmine unit testing to lib-jitsi-meet.  also add a ver…

bbaldino:
…y basic example test for SDPUtil

### DIFF
--- a/.jshintignore
+++ b/.jshintignore
@@ -3,6 +3,7 @@ libs
 lib-jitsi-meet.js
 lib-jitsi-meet.min.js
 lib-jitsi-meet.js.map
+karma.conf.js
 doc
 
 
@@ -30,3 +31,4 @@ JitsiParticipant.js
 JitsiMeetJS.js
 JitsiConnection.js
 JitsiConference.js
+modules/xmpp/SampleSdpStrings.js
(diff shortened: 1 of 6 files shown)


bgrozev:
jenkins add to whitelist

.jshintignore
```diff
@@ -30,3 +31,4 @@ JitsiParticipant.js
 JitsiMeetJS.js
 JitsiConnection.js
 JitsiConference.js
+modules/xmpp/SampleSdpStrings.js
```
hristoterezov: Is this added by mistake?

bbaldino: yes it is...good catch.  this is from the testing i have on my other branch

.jshintignore
```diff
@@ -3,6 +3,7 @@ libs
 lib-jitsi-meet.js
 lib-jitsi-meet.min.js
 lib-jitsi-meet.js.map
+karma.conf.js
```
hristoterezov: Why do we need to exclude karma.conf.js?

webpack.config.js is not excluded for example, what is the difference?

bbaldino: oops.  originally there were some symbol complaints there but now looks like it's just line lengths.  i'll fix that.

PR #327, opened 2016-12-12
feat(electron): basic Electron SS support

hristoterezov:


modules/RTC/ScreenObtainer.js
```diff
@@ -514,6 +502,39 @@ function waitForExtensionAfterInstall(options, waitInterval, retries) {
...
 /**
+ * Handles response from external application / extension and calls GUM to
+ * receive the desktop streams or reports error.
+ * @param {object} response
+ * @param {string} response.streamId - the streamId for the desktop stream
+ * @param {string} response.error - error to be reported.
+ * @param {Function} onSuccess - callback for success.
+ * @param {Function} onFailure - callback for failure.
+ */
+function onGetStreamResponse(response, onSuccess, onFailure) {
```
paweldomas: Just a note here, that this change about extracting "onGetStreamResponse" method would be good candidate for a separate refactoring commit. But of course it does not block this PR from being merged as it's relatively small.

PR #329, opened 2016-12-14
Add events to notify ICE state changes and ICE candidates info

nbermudezs:
I've been working on a 'troubleshooting' tool and I added some new events related to ICE:
- notify state changes
- notify the list of candidates or its absence.

In my case I use the list of candidates to double check that what is being advertised matches our expectations and also to check if one of the REST endpoints in the JVBs is reachable for the TCP 443 candidates.

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -241,6 +241,15 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function () {
             this.conference.connectionIsInterrupted = false;
         });
 
+    this.chatRoomForwarder.forward(XMPPEvents.NO_TCP_ICE_CANDIDATES,
+        JitsiConferenceEvents.NO_TCP_ICE_CANDIDATES);
+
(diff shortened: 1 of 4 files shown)


nbermudezs:
Already signed the ICLA

JitsiConferenceEventManager.js
```diff
@@ -167,6 +167,26 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function () {
...
                 new Error("ICE fail"));
+            conference.eventEmitter.emit(
+                JitsiConferenceEvents.ICE_STATE_CHANGED,
+                conference.myUserId(),
+                "failed");
+        });
+
+    chatRoom.addListener(XMPPEvents.CONNECTION_RESTORED,
+        () => {
+            conference.eventEmitter.emit(
```
gpolitis: Hi @nbermudezs, thanks for the PR! A restored XMPP connection does not imply that ICE has been restored. Maybe the wording needs to be different, but, at first glance, this doesn't look right.

PR #331, opened 2016-12-15
ref(libs): moves the libs directory to doc/example

hristoterezov:
The files from libs directory are used only for doc/example.


bbaldino:
seems like an odd location for these.  why move them to doc/example? these are lib files that are used, yeah?

gpolitis:
Are we sure these files are not used in the client code and are only used in the example?

hristoterezov:
@gpolitis @bbaldino These are dependancies of the library but they are not actually used. The have to be included by the application that uses the library (jitsi-meet project). But they are needed by the example page.

PR #333, opened 2016-12-19
Caps

hristoterezov:
Implements XEP-0115 support. Depends on #332 .


paweldomas:
Was this implementation tested against anything else that itself (lib-jitsi-meet with lib-jitsi-meet) ? It appears to be broken. Jicofo started logging this error:
impl.protocol.jabber.ScServiceDiscoveryManager.discoverInfo().520 Invalid DiscoverInfo for http://jitsi.org/jitsimeet#z1aBBLuF0kcC4cvtqaI3m4YQdkI=: org.jivesoftware.smackx.packet.DiscoverInfo@60c1b4bd

PR #337, opened 2017-01-03
Support pinning multiple users

dwilson6:
This adds support for sending the PinnedEndpointsChangedEvent over the data channel to pin multiple users when last-n is enabled.


jallamsetty1:
This is now possible by calling [`setReceiverConstraints`](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3389).

PR #338, opened 2017-01-04
Sdp overhaul

bbaldino:
can ignore this for now.  want to run the smoke tests...some are failing on my instance but i'm wondering if it's a config thing.


bbaldino:
cool...guess there must be a config difference on my jitsi meet dev server that causes smoke tests to fail.  @hristoterezov do you want to take a look at this?  there's still more cleanup to do, so no need to merge yet.  but the idea is this:

- rework the modify sources queue / processing loop: rather than having logic in the processing loop, move the logic to the tasks themselves and keep the loop simple. 
- add a shim to anything still using the modifySources queue to keep things working as expected
- change the 'video camera device change' flow to use the new loop directly.  the other scenarios (video mute, toggle screenshare, etc.) will eventually all be moved over to this style.
- because of the shim, both old and new scenarios are processed through the same queue, so shouldn't be any out-of-order or concurrency problems

this is a small piece of a larger project to get some things cleaned up and more efficient (no intermediary states when doing compound operations like toggle screenshare or change camera), so expect even the code in this diff to evolve in future PRs as i get to more of the scenarios.

bbaldino:
@hristoterezov this is officially ready for you to take a look at :) we can go through it in person tomorrow and i can give you a summary.  tests are passing and i've bounced some ideas off of george, so should be in pretty decent shape already.

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
+var transform = require('sdp-transform');
```
hristoterezov: Why not using `import`? 

Also maybe transform should be declared with `const` or `let` if we are using ES6.

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Begin helper functions
+ */
+function getPrimarySsrc (videoMLine) {
```
hristoterezov: JS docs missing

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
...
+            }
+            let fidGroup = findGroup(videoMLine, "FID");
+            if (fidGroup) {
+                return parseInt(fidGroup.ssrcs.split(" ")[0]);
+            }
+        }
+    }
+}
+
+function getRtxSsrc (videoMLine, primarySsrc) {
```
hristoterezov: JS docs missing

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
...
+                        }
+                    });
+                }
+            }
+        }
+        return transform.write(parsedSdp);
+    }
+}
+
+module.exports = SdpConsistency;
```
hristoterezov: Why not `export default`?

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
...
+
+/**
+ * Handles the work of keeping video ssrcs consistent across multiple
+ * o/a cycles, making it such that all stream operations can be
+ * kept local and do not need to be signaled.
+ * NOTE: This only keeps the 'primary' video ssrcs consistent: meaning
+ * the primary video stream and an associated RTX stream, if it exists
+ */
+class SdpConsistency {
+    constructor () {
```
hristoterezov: JS docs missing

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
...
+        this.clearSsrcCache();
+    }
+
+    clearSsrcCache () {
```
hristoterezov: JS docs missing

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
...
+        this.cachedPrimarySsrc = null;
+        this.cachedPrimaryRtxSsrc = null;
+    }
+
+    setPrimarySsrc (primarySsrc) {
```
hristoterezov: JS docs missing

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,139 @@
...
+        this.cachedPrimarySsrc = primarySsrc;
+    }
+
+    makeVideoPrimarySsrcsConsistent (sdpStr) {
```
hristoterezov: JS docs missing

JitsiConference.js
```diff
@@ -552,24 +500,98 @@ JitsiConference.prototype.removeTrack = function (track) {
...
+        .then(() => {
+            if (newStream) {
+                // Now handle the addition of the newStream at the JitsiConference level
+                this._setupNewTrack(newStream);
+            }
+            return Promise.resolve();
+        });
+};
+
+JitsiConference.prototype._setupNewTrack = function (newTrack) {
```
hristoterezov: JS docs missing

modules/xmpp/JingleSessionPC.js
```diff
@@ -600,34 +552,9 @@ JingleSessionPC.prototype.onTerminated = function (reasonCondition,
...
-    logger.log('addssrc', new Date().getTime());
-    logger.log('ice', this.peerconnection.iceConnectionState);
-
-    this.readSsrcInfo(elem);
-
-    var sdp = new SDP(this.peerconnection.remoteDescription.sdp);
-    var mySdp = new SDP(this.peerconnection.localDescription.sdp);
-
-    $(elem).each(function (idx, content) {
+JingleSessionPC.prototype._parseSsrcInfoFromSourceAdd = function (sourceAddElem, currentRemoteSdp) {
```
hristoterezov: JS docs missing

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+ * });
+ */
+JingleSessionPC.prototype._processQueueTasks = function (task, finishedCallback) {
+    task(finishedCallback);
+};
+
+/**
+ * Takes in a jingle offer iq, returns the new sdp offer
+ */
+JingleSessionPC.prototype._processNewJingleOfferIq = function(offerIq) {
```
hristoterezov: `@param` and `@returns` docs missing

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+    if (this.failICE) {
+        remoteSdp.failICE = true;
+    }
+
+    remoteSdp.fromJingle(offerIq);
+    this.readSsrcInfo($(offerIq).find(">content"));
+    return remoteSdp;
+};
+
+JingleSessionPC.prototype._processRemoteRemoveSource = function (removeSsrcInfo) {
```
hristoterezov: JS docs missing

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+            remoteSdp.media[idx] = remoteSdp.media[idx].replace(line + '\r\n', '');
+        });
+    });
+    remoteSdp.raw = remoteSdp.session + remoteSdp.media.join('');
+
+    return remoteSdp;
+};
 
-    $(elem).each(function (idx, content) {
+JingleSessionPC.prototype._processRemoteAddSource = function (addSsrcInfo) {
```
hristoterezov: JS docs missing

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+        remoteSdp.media[idx] += lines;
+    });
+    remoteSdp.raw = remoteSdp.session + remoteSdp.media.join('');
+
+    return remoteSdp;
+};
+
+/**
+ * Do a new o/a flow using the existing remote description
+ * @param sdp: optional remote sdp to use.  if not provided, the remote
```
hristoterezov: It's good to have the exact name of the parameter and type:
`@param {Type} exampleVar <description>`

Also `@returns` is missing.

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+ *  by the modification queue.
+ */
+JingleSessionPC.prototype.addStreamToPeerConnection = function (stream, ssrcInfo) {
+    let actualStream = stream && stream.getOriginalStream ? stream.getOriginalStream() : stream;
+    if (this.peerconnection) {
+        this.peerconnection.addStream(actualStream, ssrcInfo);
+    }
+};
+
+JingleSessionPC.prototype._parseSsrcInfoFromSourceRemove = function (sourceRemoveElem, currentRemoteSdp) {
```
hristoterezov: JS docs missing

modules/xmpp/SdpConsistency.js
bbaldino: done (changed to use import)

modules/xmpp/SdpConsistency.js
bbaldino: done

modules/xmpp/SdpConsistency.js
bbaldino: done

modules/xmpp/SdpConsistency.js
bbaldino: i already have a js doc above the class, is there something specific in there that i'm missing?

modules/xmpp/SdpConsistency.js
bbaldino: done

modules/xmpp/SdpConsistency.js
bbaldino: done

modules/xmpp/SdpConsistency.js
bbaldino: done

modules/xmpp/SdpConsistency.js
bbaldino: done

JitsiConference.js
bbaldino: done

modules/xmpp/JingleSessionPC.js
bbaldino: done

modules/xmpp/JingleSessionPC.js
bbaldino: done

modules/xmpp/JingleSessionPC.js
bbaldino: done

modules/xmpp/JingleSessionPC.js
bbaldino: done

modules/xmpp/JingleSessionPC.js
bbaldino: done

modules/xmpp/JingleSessionPC.js
bbaldino: done

JitsiConference.js
```diff
@@ -552,24 +500,98 @@ JitsiConference.prototype.removeTrack = function (track) {
...
-    if(!this.room){
-        if(this.rtc) {
-            this.onTrackRemoved(track);
+/**
+ * Replaces oldStream with newStream and performs a single offer/answer
+ *  cycle after both operations are done.  Either oldStream or newStream
+ *  can be null; replacing a valid 'oldStream' with a null 'newStream'
+ *  effectively just removes 'oldStream'
+ * @param oldStream the current stream in use to be replaced
+ * @param newStream the new stream to use
```
bbaldino: done

modules/xmpp/JingleSessionPC.js
```diff
@@ -322,73 +327,20 @@ JingleSessionPC.prototype.acceptOffer = function(jingleOffer,
...
+                success();
+                finishedCallback();
+            }, (error) => {
+                logger.info("Error renegotiating after setting new remote offer: " + error);
+                failure(error);
+                JingleSessionPC.onJingleFatalError(this, error);
+                finishedCallback(error);
+            });
+    };
+    this.modificationQueue.push(workFunction);
```
bbaldino: yeah i just thought this way looked cleaner and the name makes it more obvious what's being done i thought.  i can change it if you don't like it this way though.

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+ *  cycle after both operations are done.  Either oldStream or newStream
+ *  can be null; replacing a valid 'oldStream' with a null 'newStream'
+ *  effectively just removes 'oldStream'
+ * @param oldStream the current stream in use to be replaced
+ * @param newStream the new stream to use
+ * @returns {Promise}
+ */
+JingleSessionPC.prototype.replaceStream = function (oldStream, newStream) {
+    return new Promise((resolve, reject) => {
+        let workFunction = (finishedCallback) => {
```
bbaldino: same as my comment above...i just think it's easier to read and is more self-documenting with the 'workFunction' name.  i can change it though.

JitsiConference.js
```diff
@@ -425,16 +426,6 @@ JitsiConference.prototype.addTrack = function (track) {
...
-        if(ssrcMap[this.getMSID()]){
-            this._setSSRC(ssrcMap[this.getMSID()]);
-            conference.room.removeListener(XMPPEvents.SENDRECV_STREAMS_CHANGED,
-                this.ssrcHandler);
-        }
-    }.bind(track, this);
-    this.room.addListener(XMPPEvents.SENDRECV_STREAMS_CHANGED,
-        track.ssrcHandler);
-
     if(track.isAudioTrack() || (track.isVideoTrack() &&
```
bbaldino: yeah this is a bit awkward i guess...we don't call this method anywhere anymore but it was kept to not break the API.  Basically it gets called here as well because you don't call both flow (addTrack and replaceStream) for the same track...it's one or the other.

bbaldino: oops...forgot we called replaceStream from there.  ignore that last comment :)

bbaldino: done

JitsiConference.js
```diff
@@ -552,24 +500,98 @@ JitsiConference.prototype.removeTrack = function (track) {
...
+/**
+ * Replaces oldStream with newStream and performs a single offer/answer
+ *  cycle after both operations are done.  Either oldStream or newStream
+ *  can be null; replacing a valid 'oldStream' with a null 'newStream'
+ *  effectively just removes 'oldStream'
+ * @param oldStream the current stream in use to be replaced
+ * @param newStream the new stream to use
+ * @returns {Promise}
+ */
+JitsiConference.prototype.replaceStream = function (oldStream, newStream) {
```
bbaldino: there is quite a bit of inconsistency in the usage of track vs. stream.  i started with 'replaceTrack' but all the methods beneath it that are called are named with 'stream'.

fixed the disposed checks

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+                        let localUfrag = getUfrag(answer.sdp);
+                        if (localUfrag != this.localUfrag) {
+                            this.localUfrag = localUfrag;
+                            this.room.eventEmitter.emit(
+                                    XMPPEvents.LOCAL_UFRAG_CHANGED, localUfrag);
+                        }
+                        this.peerconnection.setLocalDescription(
+                            answer,
+                            () => { resolve(); },
+                            (error) => { reject(error, "setLocalDescription failed"); }
```
bbaldino: done

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+                        );
+                    },
+                    (error) => { reject(error, "createAnswer failed"); },
```
bbaldino: done

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +581,256 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+                    media_constraints
+                );
+            },
+            (error) => { reject(error, "setRemoteDescription failed"); }
```
bbaldino: done

JitsiConference.js
```diff
@@ -552,24 +500,98 @@ JitsiConference.prototype.removeTrack = function (track) {
...
+    // First do the removal of the oldStream at the JitsiConference level
+    if (oldStream) {
+        this.onTrackRemoved(oldStream);
```
bbaldino: done

modules/xmpp/SdpConsistency.js
hristoterezov: We are usually adding JS doc for the contructor as well.

JitsiConference.js
hristoterezov: At list for all public APIs we have only the concept of tracks I think. And the objects that you are passing are of type JitsiLocalTrack so ...

modules/xmpp/JingleSessionPC.js
hristoterezov: I prefer the other way personally.

modules/xmpp/JingleSessionPC.js
hristoterezov: I prefer the other way personally.

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,176 @@
...
+ *  (if there is one) for that video ssrc
+ * @param {object} videoMLine the video MLine from which to extract the
+ *  rtx video ssrc
+ * @param {number} primarySsrc the video ssrc for which to find the
+ *  corresponding rtx ssrc
+ * @returns {number} the rtx ssrc (or undefined if there isn't one)
+ */
+function getRtxSsrc (videoMLine, primarySsrc) {
+    if (videoMLine.ssrcGroups) {
+        return videoMLine
```
hristoterezov: Personally I would implement this in the following way:
```
let res;
videoMLine.ssrcGroups.find(ssrcGroup => {
    if(ssrcGroup.semantics === "FID"
        &&  parseInt(ssrcGroup.ssrcs.split(" ")[0]) === primarySsrc) {
        res = parseInt(ssrcGroup.ssrcs.split(" ")[1]);
        return true; 
    }
    return false;
})
```
or with simple loop or for Each.
This way it should do less operations (in this case maybe it won't matter) and also personally I find it easier to understand.

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,176 @@
...
+ * @returns {number} the primary video ssrc
+ */
+function getPrimarySsrc (videoMLine) {
+    if (!videoMLine.ssrcs) {
+        return;
+    }
+    let numSsrcs = videoMLine.ssrcs
+        .filter(ssrcInfo => ssrcInfo.attribute === "msid")
+        .map(ssrcInfo => ssrcInfo.id)
+        .filter((ssrc, index, array) => array.indexOf(ssrc) === index)
```
hristoterezov: Why do we need the last filter?

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,176 @@
...
+        .length;
+    if (numSsrcs === 1) {
+        return videoMLine.ssrcs
```
hristoterezov: Isn't it:
```
return videoMLine.ssrcs
           .filter(ssrcInfo => ssrcInfo.attribute === "msid")[0].id
```
simpler.

Also If we have more than one places in the function where we use `videoMLine.ssrcs.filter(ssrcInfo => ssrcInfo.attribute === "msid")` it's better to execute it once and assign it to a variable.

modules/xmpp/SdpConsistency.js
hristoterezov: or:
```
let group = videoMLine.ssrcGroups.find(ssrcGroup => {
    if(ssrcGroup.semantics === "FID"
        &&  parseInt(ssrcGroup.ssrcs.split(" ")[0]) === primarySsrc) {
        return true; 
    }
    return false;
});
return  group? parseInt(group.ssrcs.split(" ")[1]) : undefined;
```

bbaldino: i don't find the above to be very easy to read (primarily because of the long line with the call to parseInt inside the if), and i think assigning to an outside variable inside the 'find' doesn't feel great, but i did change this to use find to find the specific group and then move the rtxSsrc parsing (once the group has been found) outside so the .map/.filter chain is gone.

modules/xmpp/SdpConsistency.js
bbaldino: this list contains multiple entries for each ssrc (because it's each ssrc line in the sdp, so has one for msid, one for cname, etc.) so the last line ensures only unique entries are added.

bbaldino: ah....oops...yes i see i am double filtering here...i think i forgot to remove the msid filtering line before (i think it's cleaner not to assume there is an msid line necessarily).  i'll clean this up.

modules/xmpp/SdpConsistency.js
bbaldino: i think we can make this even simpler actually, if there is only 1 ssrc we can just do:

  return videoMLine.ssrcs[0].id;

paweldomas:
@bbaldino I just went through the whole PR and I find it very interesting (and nice) how you have separated the changes into commits by providing intermediate shims for the code that is not complete yet. Also I like the modify sources cleanup that it submits actual task functions rather than executing all the stuff in modify sources method. The replaceStreams method is also something I've been thinking about for a while never getting to it. All these changes are big improvement to the current code. I have few question about some changes, but will post them on per commit basis.

modules/xmpp/SdpConsistency.js
```diff
@@ -0,0 +1,175 @@
...
+    }
+
+    /**
+     * Explicitly set the primary ssrc to be used in
+     *  makeVideoPrimarySsrcsConsistent
+     * @param {number} primarySsrc the primarySsrc to be used
+     *  in future calls to makeVideoPrimarySsrcsConsistent
+     */
+    setPrimarySsrc (primarySsrc) {
+        this.cachedPrimarySsrc = primarySsrc;
```
paweldomas: I haven't looked at the simulcast lib if there are any changes related to SSRC caching, but I wonder if the non-primary SSRCs will be consistent as well ? Does the simulcast lib cache non-primary SSRCs as long as the primary one remains the same ?

paweldomas: Oh actually I see that now the SSRC cache is being set on the simulcast lib when doing "unmute"

modules/xmpp/JingleSessionPC.js
paweldomas: I like the approach with var named worked function. As Brian wrote it self documents the fact that this is a worker function that is about to be submitted to the modification queue.

modules/xmpp/SdpConsistency.js
bbaldino: yup...the division of responsibilities is:
SdpConsistency keep the primary video ssrc consistent
SdpSimulcast adds simulcast ssrcs and takes the responsibility for keeping the sim ssrcs consistent
(And, in a future PR: RtxModifier adds rtx ssrcs and keeps them consistent)

modules/xmpp/JingleSessionPC.js
```diff
@@ -663,51 +591,285 @@ JingleSessionPC.prototype.addSource = function (elem) {
...
+                    finishedCallback();
+                });
+        };
+        this.modificationQueue.push(workFunction);
+    });
+};
+
+/**
+ * Just add the stream to the peerconnection
+ * @param stream either the low-level webrtc MediaStream or
```
paweldomas: a correct way for documenting this would be:
`@param {JitsiLocalTrack|MediaStream} stream either the low-level webrtc MediaStream or...`
this way some IDEs will recognise the type and provide code completion.

modules/xmpp/SdpConsistency.js
```diff
@@ -55,13 +52,15 @@ function getPrimarySsrc (videoMLine) {
...
 function getRtxSsrc (videoMLine, primarySsrc) {
     if (videoMLine.ssrcGroups) {
-        return videoMLine
-            .ssrcGroups
-            .filter(ssrcGroup => ssrcGroup.semantics === "FID")
-            .filter(ssrcGroup => 
-                parseInt(ssrcGroup.ssrcs.split(" ")[0]) === primarySsrc)
-            .map(ssrcGroup => ssrcGroup.ssrcs.split(" ")[1])
-            .map(ssrcStr => parseInt(ssrcStr))[0];
+                let fidGroup = videoMLine.ssrcGroups.find(group => {
```
paweldomas: Not a big deal, but the indentation seems to be messed up.

modules/xmpp/JingleSessionPC.js
bbaldino: fixed

modules/xmpp/SdpConsistency.js
bbaldino: oops...thanks.  fixed.

modules/xmpp/JingleSessionPC.js
```diff
@@ -962,98 +963,106 @@ JingleSessionPC.prototype.generateNewStreamSSRCInfo = function () {
...
+ * NOTE: must be called within a work function being executed
+ *  by the modification queue.
+ */
+JingleSessionPC.prototype.removeStreamFromPeerConnection = function (stream, stopStream) {
+    let actualStream = stream && stream.getOriginalStream ? stream.getOriginalStream() : stream;
+    if (!this.peerconnection) {
         return;
     }
-
     if (RTCBrowserType.getBrowserType() ===
```
paweldomas: I know that it was like that before the changes and I'm not asking you to change that now, but in general it seems to make more sense to distinguish such cases on per feature basis rather than on per browser type basis. In this case the RTP sender API is used to avoid sLD/sRD cycle. Hence it would make sense to add a method to BrowserType.isRTPSenderAPISupported() and include check for Firefox browser type there. This way it will be easier to change than in future and keep the logic in one place. I did something like that for the onmute event support when a video track is frozen: https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/RTCBrowserType.js#L125

modules/xmpp/JingleSessionPC.js
```diff
@@ -724,173 +886,24 @@ JingleSessionPC.prototype.removeSource = function (elem) {
...
-
-    /**
-     * Implements a failure callback which reports an error message and an
-     * optional error through (1) logger, (2) GlobalOnErrorHandler, and (3)
-     * queueCallback.
-     *
-     * @param {string} errmsg the error message to report
-     * @param {*} error an optional error to report in addition to errmsg
-     */
-    function reportError(errmsg, err) {
```
hristoterezov: Why not using this code on some of the errors from SLD, SRD, etc?

modules/xmpp/JingleSessionPC.js
bbaldino: yeah i think that sounds like a better take on it, if you don't think it's too bad for now though i agree it'd be nice to have that in a separate PR

JitsiConference.js
bbaldino: fixed this (change to replaceTrack)

PR #341, opened 2017-01-06
fix(caps): add event to notify when caps version for a user is changed

hristoterezov:


JitsiConferenceEventManager.js
```diff
@@ -506,6 +506,16 @@ JitsiConferenceEventManager.prototype.setupRTCListeners = function () {
  */
 JitsiConferenceEventManager.prototype.setupXMPPListeners = function () {
     var conference = this.conference;
+    conference.xmpp.caps.addListener(XMPPEvents.PARTCIPANT_FEATURES_CHANGED,
+        from => {
+            const participant = conference.getParticipantId(
```
paweldomas: I just noticed that there is no such method: JitsiConference.getParticipantId

PR #343, opened 2017-01-19
cleanup: Make desktop sharing options consistent and drop deprecated …

saghul:
…ones

Make sure there are options to set the extension ID and an enabled flag for
both Chrome and Firefox.

Removed checking for deprecated configuration options: desktopSharing and
chromeExtensionId.


saghul:
This came up on my 1-1 with @damencho the other day, so here it is :-)  Anyone is welcome to review!

bgrozev:
Yay :)

PR #344, opened 2017-01-21
Fix crash on screensharing when video muted

paweldomas:
The app will crash if the user tries to start screen sharing from "video muted" state.

### DIFF
--- a/modules/xmpp/TraceablePeerConnection.js
+++ b/modules/xmpp/TraceablePeerConnection.js
@@ -2,7 +2,6 @@
 
 import { getLogger } from "jitsi-meet-logger";
 const logger = getLogger(__filename);
-import RTC from '../RTC/RTC';
 import SdpConsistency from "./SdpConsistency.js";
 var RTCBrowserType = require("../RTC/RTCBrowserType.js");
 var XMPPEvents = require("../../service/xmpp/XMPPEvents");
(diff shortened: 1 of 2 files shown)


modules/xmpp/TraceablePeerConnection.js
```diff
@@ -322,11 +321,8 @@ TraceablePeerConnection.prototype.addStream = function (stream, ssrcInfo) {
     }
 };
 
-TraceablePeerConnection.prototype.removeStream = function (stream, stopStreams) {
+TraceablePeerConnection.prototype.removeStream = function (stream) {
     this.trace('removeStream', stream.id);
-    if (stopStreams) {
-        RTC.stopMediaStream(stream);
```
bbaldino: so do we not still need this functionality?

paweldomas: I searched for usages and the method is never called with true as the second argument. I don't know if it was used before. Maybe I better double check on that.

bbaldino: did you find anything?  let me know if you want me to take a look too

paweldomas: Yeah, maybe it would be good if you could check if this is used currently. For me it looks like it's not and it could be removed.

bbaldino: the only place (in the code before my changes) that i see passing a boolean it at all is: https://github.com/jitsi/lib-jitsi-meet/blob/f7efba4aecda8fe6b3d6366a0d3bf1a054f85b8c/modules/xmpp/JingleSessionPC.js#L1031, so yeah i think we're ok

PR #346, opened 2017-01-26
Enable rtx sources for streams

bbaldino:


modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,242 @@
...
+ * Begin helper functions
+ */
+/**
+ * Given a videoMLine, returns a list of the video
+ *  ssrcs (those used to actually send video, not
+ *  any associated secondary streams)
+ * @param videoMLine media line object from transform.parse
+ * @returns list of ssrcs (as strings)
+ */
+function getPrimaryVideoSsrcs (videoMLine) {
```
paweldomas: It feels like those helper functions could be extracted to some utility class, so that they would be more visible for someone unfamiliar with RtcModifier. The same applies to the SdpConsistency. Something like SdpTransformUtil that would gather helpers for operating with the 'sdp-transform' lib.

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,242 @@
...
+        .filter((ssrc, index, array) => array.indexOf(ssrc) === index);
+
+    if (videoMLine.ssrcGroups) {
+        videoMLine.ssrcGroups.forEach((ssrcGroupInfo) => {
+            // Right now, FID groups are the only ones we parse to 
+            //  disqualify streams.  If/when others arise we'll
+            //  need to add support for them here
+            if (ssrcGroupInfo.semantics === "FID") {
+                // secondary FID streams should be filtered out
+                let secondarySsrc = ssrcGroupInfo.ssrcs.split(" ")[1];
```
paweldomas: I see that split(" ") then parsing to number is duplicated in many places. Can we wrap that into some util/class ?

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,242 @@
...
+            if (group.semantics === "FID") {
+                let primarySsrc = group.ssrcs.split(" ")[0];
+                let rtxSsrc = group.ssrcs.split(" ")[1];
+                if (!ssrcExists(primarySsrc)) {
+                    ssrcsToRemove.push(rtxSsrc);
+                }
+            }
+        });
+        videoMLine.ssrcs = videoMLine.ssrcs
+            .filter(ssrc => ssrcsToRemove.indexOf(ssrc.id + "") === -1);
```
paweldomas: I see that there is this problem of converting between strings and numbers when dealing with sdp-transform. Maybe we should review our strategy to store SSRCs as strings instead ?

bbaldino: yeah this does get annoying.  for SdpConsistency i changed everything that returned them or took them in as args to use numbers and always convert away from string.  here there's a bit of a mix because always forcing them felt awkward.  i think part of it comes from the fact that the 'id' field in the sdp ssrc line uses a number, but the ssrc group field has the line as a string of space-separated ssrc strings

modules/xmpp/RtxModifier.js
bbaldino: yeah i think it'd be a good idea to write a 'parse ssrcs from group' helper method or something.  i'll do that.

bbaldino: fixed this

paweldomas:
I find SDPUtil slightly confusing, because they are not related with "sdp-transform" lib and mix operation on objects coming from it with manual SDP parsing, but it's fine as long as the arguments are documented, so I'm going to merge it anyway.

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
...
+function getRtxSsrc (videoMLine, primarySsrc) {
+    if (videoMLine.ssrcGroups) {
+        let fidGroup = videoMLine.ssrcGroups.find(group => {
+            if (group.semantics === "FID") {
+                let groupPrimarySsrc = SDPUtil.parseGroupSsrcs(group)[0];
+                return groupPrimarySsrc === primarySsrc;
+            }
+        });
+        if (fidGroup) {
+          return parseInt(fidGroup.ssrcs.split(" ")[1]);
```
paweldomas: Would it be suitable to call SDPUtil.parseGroupSsrcs here as well ? (sorry for being so picky :P)

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
...
+export default class RtxModifier {
+    /**
+     * Constructor
+     */
+    constructor () {
+        /**
+         * Map of video ssrc to corresponding RTX
+         *  ssrc
+         */
+        this.correspondingRtxSsrcs = {};
```
saghul: nitpick: maybe use a Map? (ES6 ALL THE THINGS)

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,263 @@
...
+
+        it ("should re-use the same rtx ssrc for a primary ssrc it's seen before", function() {
+          // Have rtxModifier generate an rtx ssrc via modifyRtxSsrcs.  Then call it again
+          //  with the same primary ssrc in the sdp (but no rtx ssrc).  It should use
+          //  the same rtx ssrc as before.
+          let newSdpStr = this.rtxModifier.modifyRtxSsrcs(this.transform.write(this.singleVideoSdp));
+          let newSdp = transform.parse(newSdpStr);
+
+          let fidGroup = getVideoGroups(newSdp, "FID")[0];
+          let fidGroupRtxSsrc = parseInt(fidGroup.ssrcs.split(" ")[1]);
```
paweldomas: "SDPUtil parse groups" should be used here as well ? (please fix also other occurrences in this file)

modules/xmpp/TraceablePeerConnection.js
```diff
@@ -583,18 +616,35 @@ TraceablePeerConnection.prototype.getStats = function(callback, errback) {
...
             group: {ssrcs: ssrcInfo.ssrcs.join(" "), semantics: "SIM"}});
-        return ssrcInfo;
+        ssrcInfo;
     } else {
-        return {ssrcs: [RandomUtil.randomInt(1, 0xffffffff)], groups: []};
+        ssrcInfo = {ssrcs: [SDPUtil.generateSsrc()], groups: []};
+    }
+    if (!this.session.room.options.disableRtx) {
+        let currNumSsrcs = ssrcInfo.ssrcs.length;
+        for (let i = 0; i < currNumSsrcs; ++i) {
```
saghul: for..of

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
+import { getLogger } from "jitsi-meet-logger";
+const logger = getLogger(__filename);
+import * as transform from 'sdp-transform';
+var SDPUtil = require("./SDPUtil");
```
saghul: const?

saghul:
@bbaldino left a couple of drive-by comments.

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,263 @@
+/*eslint-disable max-len*/
+/*jshint maxlen:false*/
+var RtxModifier = require("./RtxModifier.js");
```
paweldomas: Can we stick with ES6 "import" with the new code ? (please fix all occurrences)
In general we should try to use ES6 with the new code. I'm not sure if that works with the tests, does it ?

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,263 @@
...
+};
+
+var getVideoGroups = function (parsedSdp, groupSemantics) {
+  let videoMLine = parsedSdp.media.find(m => m.type === "video");
+  videoMLine.ssrcGroups = videoMLine.ssrcGroups || [];
+  return videoMLine.ssrcGroups
+    .filter(g => g.semantics === groupSemantics);
+};
+
+describe ("RtxModifier", function() {
```
paweldomas: can we replace all function() with () => if possible (not sure if that works with the tests)

modules/xmpp/SDPUtil.js
```diff
@@ -1,7 +1,7 @@
 import {getLogger} from "jitsi-meet-logger";
 const logger = getLogger(__filename);
 var RTCBrowserType = require("../RTC/RTCBrowserType");
-
+var RandomUtil = require('../util/RandomUtil');
```
paweldomas: please use "import" instead

modules/xmpp/RtxModifier.js
bbaldino: guess i missed this one...fixed.

modules/xmpp/RtxModifier.spec.js
bbaldino: fixed

modules/xmpp/TraceablePeerConnection.js
bbaldino: fixed

bbaldino: i find the syntax of for..of very unfortunate btw:
`for (let primarySsrc of ssrcInfo.ssrcs) {`
doesn't read very nicely to me.

modules/xmpp/RtxModifier.js
bbaldino: fixed

modules/xmpp/RtxModifier.spec.js
bbaldino: fixed

modules/xmpp/TraceablePeerConnection.js
paweldomas: in this particular case it does read better when a ssrc, becomes a primary SSRC, but I don't have strong opinion

modules/xmpp/RtxModifier.spec.js
bbaldino: unfortunately (or fortunately, maybe?) jasmine passes some context in a passed 'this', so some of the functions (`beforeEach` and `it`) are best using the old-style functions.  so our option is to use arrow syntax in some places but not others, or just be consistent with old-style functions everywhere.  what do you think?

modules/xmpp/SDPUtil.js
bbaldino: fixed

modules/xmpp/RtxModifier.spec.js
paweldomas: If that's causing issues leave it as is.

modules/xmpp/RtxModifier.js
bbaldino: fixed.  also not crazy about the `set` syntax of `Map`.  to me
`myMap[key] = value;`
is clearer than
`myMap.set(key, value);`
but oh well.

modules/xmpp/RtxModifier.spec.js
bbaldino: i agree...i think being consistent (always use old-style functions in tests) will be more straightforward and less error prone.

bbaldino:
@paweldomas let me know if this looks good, but please don't merge it quite yet.

modules/xmpp/TraceablePeerConnection.js
bbaldino: actually i had to change it back...we add to the list being iterated over here, so i explicitly grabbed the original length so we only iterated over the original items.  i added a comment here too to make that more clear.

bbaldino:
ps @paweldomas i agree with your sentiment about SDPUtil.  it's a bit of a catch-all at the moment, and honestly i think there is a lot of sdp functionality that would be useful as a lib of its own.  maybe that would build on top of sdp-transform or not, i dunno.

paweldomas:
> ps @paweldomas i agree with your sentiment about SDPUtil. it's a bit of a catch-all at the moment, and honestly i think there is a lot of sdp functionality that would be useful as a lib of its own. maybe that would build on top of sdp-transform or not, i dunno.

Yes, I agree. Have been thinking about the same (possibly using sdp-transform on top). Like I wrote it is fine as long as it's documented (the methods are still SDP utilities).

paweldomas:
Looks good, I don't want to hold it off anymore. Is it ready to be merged now ?

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
+import { getLogger } from "jitsi-meet-logger";
+const logger = getLogger(__filename);
+import * as transform from 'sdp-transform';
+const SDPUtil = require("./SDPUtil");
```
hristoterezov: Why not `import * as SDPUtil from "./SDPUtil"`?

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,268 @@
+/*eslint-disable max-len*/
```
hristoterezov: Why do we have mixed requires with imports and vars with const/let in this file?

modules/xmpp/SampleSdpStrings.js
```diff
@@ -0,0 +1,162 @@
+/*eslint-disable max-len*/
```
hristoterezov: Can't we use let/const in this file?

modules/xmpp/SampleSdpStrings.js
```diff
@@ -0,0 +1,162 @@
...
+"a=ssrc-group:FID 1089111804 2963867077\r\n" +
+"a=ssrc-group:SIM 1757014965 1479742055 1089111804\r\n" +
+"a=rtcp-mux\r\n";
+
+var simulcastSdp = baseSessionSdp + baseAudioMLineSdp + simulcastVideoMLineSdp + baseDataMLineSdp;
+var simulcastRtxSdp = baseSessionSdp + baseAudioMLineSdp + simulcastRtxVideoMLineSdp + baseDataMLineSdp;
+var plainVideoSdp = baseSessionSdp + baseAudioMLineSdp + plainVideoMLineSdp + baseDataMLineSdp;
+var rtxVideoSdp = baseSessionSdp + baseAudioMLineSdp + rtxVideoMLineSdp + baseDataMLineSdp;
+
+exports.simulcastSdp = transform.parse(simulcastSdp);
```
hristoterezov: Why not export `export const simulcastSdp = transform.parse(simulcastSdp);`?

PS: The question is valid for the next lines also.

modules/xmpp/SdpConsistency.js
```diff
@@ -1,4 +1,5 @@
 import * as transform from 'sdp-transform';
+var SDPUtil = require("./SDPUtil");
```
hristoterezov: Why not import * as SDPUtil from "./SDPUtil"?

modules/xmpp/SampleSdpStrings.js
```diff
@@ -0,0 +1,162 @@
+/*eslint-disable max-len*/
+/*jshint maxlen:false*/
+var transform = require('sdp-transform');
```
hristoterezov: Why not using import?

modules/xmpp/SampleSdpStrings.js
```diff
@@ -0,0 +1,162 @@
...
+
+var baseSessionSdp = "" +
```
hristoterezov: Could you please add comments for every variable?

modules/xmpp/SdpConsistency.js
```diff
@@ -141,36 +118,28 @@ export default class SdpConsistency {
...
             }
-            let newPrimaryRtxSsrc =
-                getRtxSsrc(videoMLine, newPrimarySsrc);
             if (!this.cachedPrimarySsrc) {
                 this.cachedPrimarySsrc = newPrimarySsrc;
-                this.cachedPrimaryRtxSsrc = newPrimaryRtxSsrc;
-                console.log("Sdp-consistency caching primary ssrc " +
-                    this.cachedPrimarySsrc + " and rtx " +
-                    this.cachedPrimaryRtxSsrc);
+                console.log("Sdp-consistency caching primary ssrc " +
```
hristoterezov: I see that this is old code but I was wondering why we are using console.log instead of logger.log? Do you have any idea?

If there isn't reason for this maybe it's good moment to change them to logger since you are changing them. WDYT?

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
...
+        logger.info("Setting ssrc cache to ", ssrcMapping);
+        this.correspondingRtxSsrcs = ssrcMapping;
+    }
+
+    /**
+     * Adds RTX ssrcs for any video ssrcs that don't
+     *  already have them.  If the video ssrc has been
+     *  seen before, and already had an RTX ssrc generated,
+     *  the same RTX ssrc will be used again.
+     * @param {string} sdpStr sdp in raw string format
```
hristoterezov: I believe there should be `@returns`.

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
...
+        });
+        return transform.write(parsedSdp);
+    }
+
+    /**
+     * Remove all reference to any rtx ssrcs that 
+     *  don't correspond to the primary stream.
+     * Must be called *after* any simulcast streams
+     *  have been imploded
+     * @param {string} sdpStr sdp in raw string format
```
hristoterezov: I believe there should be `@returns`.

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,268 @@
...
+/*jshint maxlen:false*/
+import RtxModifier from "./RtxModifier.js";
+var SampleSdpStrings = require("./SampleSdpStrings.js");
+var transform = require('sdp-transform');
+var SDPUtil = require("./SDPUtil.js");
+
+var numVideoSsrcs = (parsedSdp) => {
```
hristoterezov: 1) Missing JSDoc
2) Why not `function numVideoSsrcs()...`?

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,268 @@
...
+  let videoMLine = parsedSdp.media.find(m => m.type === "video");
+  return videoMLine.ssrcs
+    .map(ssrcInfo => ssrcInfo.id)
+    .filter((ssrc, index, array) => array.indexOf(ssrc) === index)
+    .length;
+};
+
+var getPrimaryVideoSsrc = function (parsedSdp) {
```
hristoterezov: 1) Missing JSDoc
2) Why not `function getPrimaryVideoSsrc()...`?

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,268 @@
...
+  let videoMLine = parsedSdp.media.find(m => m.type === "video");
+  return parseInt(SDPUtil.parsePrimaryVideoSsrc(videoMLine));
+};
+
+// Only handles parsing 2 scenarios right now:
+// 1) Single video ssrc
+// 2) Multiple video ssrcs in a single simulcast group
+var getPrimaryVideoSsrcs = function (parsedSdp) {
```
hristoterezov: 1) Missing JSDoc
2) Why not `function getPrimaryVideoSsrcs()...`?

modules/xmpp/RtxModifier.spec.js
```diff
@@ -0,0 +1,268 @@
...
+    let simGroups = getVideoGroups(parsedSdp, "SIM");
+    if (simGroups.length > 1) {
+      return;
+    }
+    let simGroup = simGroups[0];
+    return SDPUtil.parseGroupSsrcs(simGroup);
+  }
+};
+
+var getVideoGroups = function (parsedSdp, groupSemantics) {
```
hristoterezov: 1) Missing JSDoc
2) Why not `function getVideoGroups()...`?

modules/xmpp/SDPUtil.js
```diff
@@ -362,7 +362,69 @@ var SDPUtil = {
         line += ' ';
         line += cand.getAttribute('generation') || '0';
         return line + '\r\n';
-    }
+    },
+
+    parsePrimaryVideoSsrc: function(videoMLine) {
```
hristoterezov: Missing JSDoc.

modules/xmpp/SDPUtil.js
```diff
@@ -362,7 +362,69 @@ var SDPUtil = {
...
+                    .find(group => group.semantics === "SIM");
+                if (simGroup) {
+                    primarySsrc = simGroup.ssrcs.split(" ")[0];
+                }
+            }
+        }
+        return primarySsrc;
+    },
+
+    generateSsrc: function() {
```
hristoterezov: Missing JSDoc.

modules/xmpp/SDPUtil.js
```diff
@@ -362,7 +362,69 @@ var SDPUtil = {
...
+        return RandomUtil.randomInt(1, 0xffffffff);
+    },
+
+    getSsrcAttribute: function (mLine, ssrc, attributeName) {
```
hristoterezov: Missing JSDoc.

hristoterezov:
There are a lot of places where you have used `let` but actually the variable is never changed, so could you please change those variables to `const`?

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
...
+ * Begin helper functions
+ */
+/**
+ * Given a videoMLine, returns a list of the video
+ *  ssrcs (those used to actually send video, not
+ *  any associated secondary streams)
+ * @param {object} videoMLine media line object from transform.parse
+ * @returns {list<string>} list of primary video ssrcs
+ */
+function getPrimaryVideoSsrcs (videoMLine) {
```
hristoterezov: If we have simulcast enabled isn't it going to return also all simulcast ssrcs?

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
...
+ *  corresponding rtx ssrc
+ * @returns {number} the rtx ssrc (or undefined if there isn't one)
+ */
+function getRtxSsrc (videoMLine, primarySsrc) {
+    if (videoMLine.ssrcGroups) {
+        let fidGroup = videoMLine.ssrcGroups.find(group => {
+            if (group.semantics === "FID") {
+                let groupPrimarySsrc = SDPUtil.parseGroupSsrcs(group)[0];
+                return groupPrimarySsrc === primarySsrc;
+            }
```
hristoterezov: Maybe it's good to add `return false;` by default. WDYT?

hristoterezov: Also you can save one `SDPUtil.parseGroupSsrcs` call if you store the result from it.

modules/xmpp/TraceablePeerConnection.js
```diff
@@ -317,7 +318,29 @@ TraceablePeerConnection.prototype.addStream = function (stream, ssrcInfo) {
         this.peerconnection.addStream(stream);
     if (ssrcInfo && ssrcInfo.type === "addMuted") {
         this.sdpConsistency.setPrimarySsrc(ssrcInfo.ssrc.ssrcs[0]);
-        this.simulcast.setSsrcCache(ssrcInfo.ssrc.ssrcs);
+        let simGroup =
```
hristoterezov: Isn't it going to be better to first check for simulcast and rtx enabled/disabled before doing all those operations?

modules/xmpp/RtxModifier.js
bbaldino: fixed

modules/xmpp/RtxModifier.js
bbaldino: fixed

modules/xmpp/RtxModifier.js
bbaldino: fixed

modules/xmpp/RtxModifier.spec.js
bbaldino: fixed

modules/xmpp/RtxModifier.spec.js
bbaldino: fixed (both parts)

modules/xmpp/RtxModifier.spec.js
bbaldino: fixed (both parts)

modules/xmpp/RtxModifier.spec.js
bbaldino: fixed (both parts)

modules/xmpp/RtxModifier.spec.js
bbaldino: fixed (both parts)

modules/xmpp/SampleSdpStrings.js
bbaldino: fixed

modules/xmpp/SampleSdpStrings.js
bbaldino: fixed

modules/xmpp/SdpConsistency.js
bbaldino: yeah i dunno...fixed it.

modules/xmpp/RtxModifier.js
```diff
@@ -0,0 +1,276 @@
...
+
+    /**
+     * Adds RTX ssrcs for any video ssrcs that don't
+     *  already have them.  If the video ssrc has been
+     *  seen before, and already had an RTX ssrc generated,
+     *  the same RTX ssrc will be used again.
+     * @param {string} sdpStr sdp in raw string format
+     */
+    modifyRtxSsrcs (sdpStr) {
+        let parsedSdp = transform.parse(sdpStr);
```
hristoterezov: Maybe it's good to limit the number of transform.parse calls. Maybe we should have one transformed copy in higher level which we can pass here and whenever it is needed. 

I think the same is valid for transform.write. WDYT?

I'm not sure if the refactoring related to this is not too big to be included into this PR. Maybe we should do this in separate PR.

modules/xmpp/SDPUtil.js
bbaldino: fixed

modules/xmpp/SDPUtil.js
bbaldino: fixed

modules/xmpp/SDPUtil.js
bbaldino: fixed

modules/xmpp/RtxModifier.js
bbaldino: yeah that's ok, that's what we want (since we'll add rtx streams for each one)

modules/xmpp/RtxModifier.js
bbaldino: the result will be `undefined` by default already, won't it?  and that's what i wanted in the case of it not finding one anyway

modules/xmpp/TraceablePeerConnection.js
bbaldino: yup good idea

bbaldino: oh, well actually this is basically doing that by checking for the presence of the sim and fid groups

modules/xmpp/RtxModifier.js
bbaldino: hm, but this file only calls it twice?  once for each of the flows (modifyRtxSsrcs on the local side, and implodeRemoteRtxSsrcs on the remote side)

bbaldino:
> There are a lot of places where you have used let but actually the variable is never changed, so could you please change those variables to const?

i've been using let but not const as i haven't seen it much.  i'll make an effort to try and get RtxModifier.js cleaned up, but probably won't go further than that for now...will get in the habit of using it by default and only falling back to `let` when necessary (i assume that's what we want to do?)

modules/xmpp/SdpConsistency.js
bbaldino: fixed

modules/xmpp/SampleSdpStrings.js
bbaldino: fixed

modules/xmpp/SampleSdpStrings.js
bbaldino: fixed

modules/xmpp/TraceablePeerConnection.js
hristoterezov: I meant checking `this.session.room.options.disableRtx`

PR #349, opened 2017-01-30
Fix conference being broken after someone without any video device joins

paweldomas:
Video SSRC has to be generated for the recvonly stream if there are no video tracks in the PeerConnection.

### DIFF
--- a/modules/xmpp/TraceablePeerConnection.js
+++ b/modules/xmpp/TraceablePeerConnection.js
@@ -489,6 +489,19 @@ TraceablePeerConnection.prototype.createAnswer
                  *  about unmapped ssrcs)
                  */
                 if (!RTCBrowserType.isFirefox()) {
+                    // If there were no video streams added to
+                    // the PeerConnection, there would be no primary SSRC cached
+                    if (!this.sdpConsistency.cachedPrimarySsrc) {
+                        const newPrimarySSRC
+                            = RandomUtil.randomInt(1, 0xffffffff);
+                        this.sdpConsistency.setPrimarySsrc(newPrimarySSRC);
+                        // FIXME an error could be printed here if there *are*
+                        // video tracks added, but TPC doesn't have enough info
+                        logger.warn(
+                            "Generated 'recvonly' SSRC: "
+                            + newPrimarySSRC + " (it is fine only if there are"
+                            + " no local video streams)");
+                    }
                     answer.sdp = this.sdpConsistency.makeVideoPrimarySsrcsConsistent(answer.sdp);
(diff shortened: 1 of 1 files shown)


bbaldino:
it seems like it would be nice if this would follow the same code path as starting-as-video-muted.  the generation for the ssrcs for that happens here: https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L911.  would it work to have this user that path as well?

paweldomas:
Well.. not really, because the case we're talking about is when there is no video track added to the conference

bbaldino:
so would it not work to add an empty/muted video track in this scenario like we do when we start-as-muted?

paweldomas:
When we start muted, there was a video track previously created, but it was muted, before the conference started. Muted track will have no underlying WebRTC stream at this point, but there is local video track in the conference (which can potentially be unmuted). When there is no device or no permissions were granted there will be no local video track. It does not seem right to me to add fake track in this case.
This fix does not comply with the whole "ssrc info" architecture, but I'm working on getting rid of it and it will no longer be the case. I find this issue critical, so it would be good to have it merged, but I'm not willing to do more refactoring now just to comply with it. Eventually we can delay it until I have my refactoring done.

bbaldino:
@paweldomas sounds good.  what about adding a check for the lack of tracks around https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L883 and generating the ssrc/setting the cache there?  then at least we're covering both of the different scenarios (starting with no track at all and starting with muted tracks) in the same general area?  this would basically be adding:

```
if (!this.rtc.localTracks) {
  // No local track, make sure we generate a recvonly video ssrc
  // ... your logic goes here
}
```

modules/xmpp/TraceablePeerConnection.js
```diff
@@ -489,6 +489,19 @@ TraceablePeerConnection.prototype.createAnswer
                  *  about unmapped ssrcs)
                  */
                 if (!RTCBrowserType.isFirefox()) {
+                    // If there were no video streams added to
+                    // the PeerConnection, there would be no primary SSRC cached
+                    if (!this.sdpConsistency.cachedPrimarySsrc) {
+                        const newPrimarySSRC
+                            = RandomUtil.randomInt(1, 0xffffffff);
```
bbaldino: I added a helper method for this:
`SDPUtil.generateSsrc`

paweldomas:
Yes, that sounds good. I will give it a try.

paweldomas:
Here's another version: https://github.com/jitsi/lib-jitsi-meet/pull/350
Please let me know what do you think.

PR #350, opened 2017-01-30
Fix conference being broken after someone without any video device joins (ver 2)

paweldomas:
Video SSRC has to be generated for the recvonly stream if there are no video tracks in the PeerConnection.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2,6 +2,7 @@
 
 var logger = require("jitsi-meet-logger").getLogger(__filename);
 import RTC from "./modules/RTC/RTC";
+import * as MediaType from "./service/RTC/MediaType";
 var XMPPEvents = require("./service/xmpp/XMPPEvents");
 var EventEmitter = require("events");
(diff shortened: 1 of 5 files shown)


bbaldino:
i think this looks better, though hopefully we'll be able to clean it up more in the future.  i wish we had a better trigger point than when we first get the local tracks though (like doing it at the point of failure acquiring the video track instead of when we first try to retrieve them).  or maybe there isn't an obvious point at failure-time to do this?

paweldomas:
For me it should be up to the TraceablePeerConnection to make sure everything is right with the SDP. It should be aware of which local tracks are in the PC and it should be able to take appropriate actions based on the current state. JitsiConference is too much involved into the details which annoys me very much and makes all modifications a real pain in the ass.

I'm not sure if there is a good point of failure currently to handle that. Such failure could occur either initially when local video track is being created or when existing video track is being unmuted. Such track may not be added to the conference yet, so things are getting complex.

bbaldino:
ok, let's go with this then.  hopefully the modifySSRCs cleanup can make this flow a bit nicer too.

PR #352, opened 2017-02-06
fix: remove unneeded test when muting a local stream

saghul:
Some runtimes such as React Native don't have a window.location.protocol
property, so don't check for it, assume removeStream is available.

I tried to do some "git archeology" in order to find why the check was introduced, but I couldn't find anything, it was introduced in https://github.com/jitsi/lib-jitsi-meet/commit/2328443ecb5d4aff8d284a2f92b097d5936c2f7e#diff-922e8edf325ba2a951da3fe2c72eaad4 as part of a larger refactor (the check wasn't present prior to the refactor).


saghul:
@hristoterezov Can you please review this? Thanks!

PR #353, opened 2017-02-07
Temasys plugin detection

BeatC:
In this PR I updated logic of conference initialization.

Now it's being rejected if user opens app in old browser, MS Edge (Temasys plugin doesn't support MS Edge http://support.temasys.com.sg/support/solutions/articles/5000654345-can-the-temasys-webrtc-plugin-be-used-with-microsoft-edge-) or if Temasys plugin is not installed. It helps us to show appropriate page informing user that her browser is not supported by the app or plugin installation is required.

### DIFF
--- a/modules/RTC/adapter.screenshare.js
+++ b/modules/RTC/adapter.screenshare.js
@@ -4013,24 +4013,21 @@ if ( (navigator.mozGetUserMedia ||
     }
 
     var downloadLink = AdapterJS.WebRTCPlugin.pluginInfo.downloadLink;
-
-    // If should show download link.
-    if (downloadLink) {
+    if(downloadLink) { // if download link
       var popupString;
-
-      // Check if portal link for the plugin is provided.
-      if (AdapterJS.WebRTCPlugin.pluginInfo.portalLink) {
+      if (AdapterJS.WebRTCPlugin.pluginInfo.portalLink) { // is portal link
        popupString = 'This website requires you to install the ' +
         ' <a href="' + AdapterJS.WebRTCPlugin.pluginInfo.portalLink +
         '" target="_blank">' + AdapterJS.WebRTCPlugin.pluginInfo.companyName +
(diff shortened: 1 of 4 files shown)


modules/RTC/adapter.screenshare.js
```diff
@@ -4013,21 +4013,24 @@ if ( (navigator.mozGetUserMedia ||
     }
 
     var downloadLink = AdapterJS.WebRTCPlugin.pluginInfo.downloadLink;
-    if(downloadLink) { // if download link
+
```
paweldomas: We should try to not modify this file. On update we used to copy it directly from the Temasys repo and add this line:
https://github.com/BeatC/lib-jitsi-meet/blob/35636308c361485b0d6b003fef810ae9e6f6bedc/modules/RTC/adapter.screenshare.js#L2

If we make changes it may be hard to figure out.

paweldomas: I think there was some option that would disable the notification bar...

paweldomas: Can you try this ? https://github.com/BeatC/lib-jitsi-meet/blob/35636308c361485b0d6b003fef810ae9e6f6bedc/modules/RTC/adapter.screenshare.js#L17

BeatC: Ok, I can restore changes in this file.

paweldomas:
I gave this code a try and one things appears to be broken. The conference will not start automatically, after the Temasys plugin is installed. It worked, before the PR changes. I tested it on Safari/Mac, but will probably behave the same on IE/Win. To test that scenario you either have to uninstall the plugin on Windows or remove it from /Library/Internet Plug-Ins on Mac.

lyubomir:
Superseded by https://github.com/jitsi/lib-jitsi-meet/pull/369.

PR #355, opened 2017-02-07
if enabled, prefer to use h264 (if it exists in the offer)

bbaldino:
the idea here is to let clients configure a preference for h264 if it's in the offer.  the plan for now is that just ios would exercise this.

### DIFF
--- a/modules/xmpp/SDPUtil.js
+++ b/modules/xmpp/SDPUtil.js
@@ -466,11 +466,13 @@ var SDPUtil = {
      */
     preferVideoCodec: function(videoMLine, codecName) {
         let payloadType = null;
-        videoMLine.rtp.forEach(rtp => {
-            if (rtp.codec === codecName) {
-                payloadType = rtp.payload;
-            }
-        });
+        for (let i = 0; i < videoMLine.rtp.length; ++i) {
+          const rtp = videoMLine.rtp[i];
+          if (rtp.codec === codecName) {
+              payloadType = rtp.payload;
+              break;
+          }
+        }
         if (payloadType) {
             const payloadTypes = videoMLine.payloads.split(" ").map(p => parseInt(p));
             const payloadIndex = payloadTypes.indexOf(payloadType);
--- a/modules/xmpp/SDPUtil.js
+++ b/modules/xmpp/SDPUtil.js
@@ -456,7 +456,10 @@ var SDPUtil = {
     /**
      * Sets the given codecName as the preferred codec by
      *  moving it to the beginning of the payload types
-     *  list (modifies the given mline in place)
+     *  list (modifies the given mline in place).  If there
+     *  are multiple options within the same codec (multiple h264
+     *  profiles, for instance), this will prefer the first one
+     *  that is found.
      * @param {object} videoMLine the video mline object from
      *  an sdp as parsed by transform.parse
      * @param {string} the name of the preferred codec
@@ -469,7 +472,7 @@ var SDPUtil = {
             }
         });
         if (payloadType) {
-            let payloadTypes = videoMLine.payloads.split(" ").map(p => parseInt(p));
+            const payloadTypes = videoMLine.payloads.split(" ").map(p => parseInt(p));
             const payloadIndex = payloadTypes.indexOf(payloadType);
             payloadTypes.splice(payloadIndex, 1);
             payloadTypes.unshift(payloadType);
--- a/modules/xmpp/SDPUtil.js
+++ b/modules/xmpp/SDPUtil.js
@@ -453,6 +453,29 @@ var SDPUtil = {
     getMedia: function (sdp, type) {
         return sdp.media.find(m => m.type === type);
     },
+    /**
+     * Sets the given codecName as the preferred codec by
+     *  moving it to the beginning of the payload types
+     *  list (modifies the given mline in place)
+     * @param {object} videoMLine the video mline object from
+     *  an sdp as parsed by transform.parse
+     * @param {string} the name of the preferred codec
+     */
+    preferVideoCodec: function(videoMLine, codecName) {
+        let payloadType = null;
+        videoMLine.rtp.forEach(rtp => {
+            if (rtp.codec === codecName) {
+                payloadType = rtp.payload;
+            }
+        });
+        if (payloadType) {
+            let payloadTypes = videoMLine.payloads.split(" ").map(p => parseInt(p));
+            const payloadIndex = payloadTypes.indexOf(payloadType);
+            payloadTypes.splice(payloadIndex, 1);
+            payloadTypes.unshift(payloadType);
+            videoMLine.payloads = payloadTypes.join(" ");
+        }
+    },
 };
 
 module.exports = SDPUtil;
--- a/modules/xmpp/TraceablePeerConnection.js
+++ b/modules/xmpp/TraceablePeerConnection.js
@@ -389,6 +389,13 @@ TraceablePeerConnection.prototype.setRemoteDescription
     description = this.simulcast.mungeRemoteDescription(description);
     this.trace('setRemoteDescription::postTransform (simulcast)', dumpSDP(description));
 
+    if (this.session.room.options.preferH264) {
+        const parsedSdp = transform.parse(description.sdp);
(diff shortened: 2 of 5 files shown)


modules/xmpp/SDPUtil.js
```diff
@@ -443,6 +443,30 @@ var SDPUtil = {
...
+    /**
+     * Sets the given codecName as the preferred codec by
+     *  moving it to the beginning of the payload types
+     *  list (modifies the given mline in place)
+     * @param {object} videoMLine the video mline object from
+     *  an sdp as parsed by transform.parse
+     * @param {string} the name of the preferred codec
+     */
+    preferVideoCodec: function(videoMLine, codecName) {
+        let payloadType = null;
```
saghul: Endpoints may offer H264 several times, with different profiles, we should accommodate for the case of having the same codec more than once.

modules/xmpp/TraceablePeerConnection.js
```diff
@@ -392,6 +392,13 @@ TraceablePeerConnection.prototype.setRemoteDescription
     description.sdp = this.rtxModifier.implodeRemoteRtxSsrcs(description.sdp);
     this.trace('setRemoteDescription::postTransform (implodeRemoteRtxSsrcs)', dumpSDP(description));
 
+    if (this.session.room.options.preferH264) {
```
saghul: Where should this option be documented?

modules/xmpp/SDPUtil.js
bbaldino: this is a good point...but what do you think about keeping it simple for now and modifying it once we hit scenarios where there are more?  i don't want to make it too complicated now

modules/xmpp/TraceablePeerConnection.js
bbaldino: this would go in config.js

modules/xmpp/SDPUtil.js
paweldomas: I would be ok with that as long as we detect that and print an error or something. It would be good t know prior to debugging eventual issues caused by that.

modules/xmpp/SDPUtil.js
```diff
@@ -443,6 +443,30 @@ var SDPUtil = {
...
+
+    /**
+     * Sets the given codecName as the preferred codec by
+     *  moving it to the beginning of the payload types
+     *  list (modifies the given mline in place)
+     * @param {object} videoMLine the video mline object from
+     *  an sdp as parsed by transform.parse
+     * @param {string} the name of the preferred codec
+     */
+    preferVideoCodec: function(videoMLine, codecName) {
```
lyubomir: Why do you continue to mix ES5 and ES6? Could we please drop `: function` for the ES6 object initializer shorthand?

modules/xmpp/SDPUtil.js
```diff
@@ -443,6 +443,30 @@ var SDPUtil = {
...
+        let payloadType = null;
+        videoMLine.rtp.forEach(rtp => {
+            if (rtp.codec === codecName) {
+                payloadType = rtp.payload;
+            }
+        });
+        if (payloadType) {
+            let payloadTypes = videoMLine.payloads.split(" ").map(p => parseInt(p));
```
lyubomir: payloadTypes is not a `let`, it's a `const`, right?

modules/xmpp/TraceablePeerConnection.js
hristoterezov: Probably it's good to add it to API.md in lib-jitsi-meet project too.

modules/xmpp/SDPUtil.js
saghul: Chrome already has an open issue about adding H264 high profile together with baseline, which seems mostly ready for landing: https://bugs.chromium.org/p/webrtc/issues/detail?id=6337

modules/xmpp/SDPUtil.js
bbaldino: i was keeping consistent with the rest of the file.  would you rather new functions use the new syntax but be inconsistent with the rest of the file?

modules/xmpp/SDPUtil.js
bbaldino: i guess so in the sense that the variable is never redefined (though it is changed).  confusing definition of const in javascript...will change it.

modules/xmpp/SDPUtil.js
bbaldino: also, please remember i am a relatively new javascript developer :)

lyubomir: I'd rather we stop writing ES5 source code right now because someone will have to translate it to ES6 eventually.

I understand the consistency argument but I think it's weakened by the consistency situation within lib-jitsi-meet.

I'm perfectly OK with you being new to JavaScript. I consider myself new because I sat down studying JavaScript in May 2016.

modules/xmpp/SDPUtil.js
lyubomir: The meaning of `const` equals the protection the language provides for it. If the JavaScript `const` is confusing to you, I'd guess you're used to a `const` protection that includes banning mutating methods i.e. a language with a notion of `const` methods which JavaScript is not.

modules/xmpp/SDPUtil.js
bbaldino: @saghul do you know if there's an example sdp with multiple profiles anywhere?  without that i wouldn't be able to test it reliably.

saghul: @bbaldino the LifeSize devices do that, so maybe @aaronkvanmeerten can get one for you? (Aaron, if you start pjsua in the foreground it will print SIP traces to stdout).

modules/xmpp/TraceablePeerConnection.js
bbaldino: @hristoterezov will do, but i'm having a hard time determining which section of the doc it should go in.  do you know?

modules/xmpp/SDPUtil.js
bbaldino: mixing let/const in a file is one thing but, in my opinion, mixing function definitions doesn't seem like a good idea.  even in projects that aren't consistent across an entire codebase, usually it is kept consistent at least within a file.  if the team disagrees i'll change but to me it seems like consistency for function definitions would trump the incremental amount of work we'd save whenever we go through and change this file over to es6.

modules/xmpp/SDPUtil.js
bbaldino: yes i do understand that, lyubo (which is why i said it in my note).  just a little off-hand commentary.  changed to const.

modules/xmpp/SDPUtil.js
bbaldino: i don't think the existing behavior is that bad: basically if there are multiple h264's, we will prefer the first one we find, so i believe not even an error is really that necessary as the current behavior won't fail in that scenario (if it helps, i can add some more comments describing that behavior).  also, consider the two scenarios:
1) user is using jicofo: here there is only 1 h264 profile available
2) user is using their own signaling: here the user is in control of what the offer looks like, so they have some flexibility in what they offer already anyway.

not disagreeing that it's a good idea, i'm just not sure it feels urgent to worry about it now.  what do you think @paweldomas @saghul ?

bbaldino: ok, i documented the behavior to prefer the first seen and added a test case to verify it.  being able to prefer something based on sub-parameters of the codec is actually quite a bit harder, as the parameters are in a single line and vary from one codec to another, so i think it'd be best to wait and see how we want to best differentiate between different sub-types of a single codec.

modules/xmpp/TraceablePeerConnection.js
saghul: AFAIS options are kinda global for the library, so you can add it to the init section here: https://github.com/jitsi/lib-jitsi-meet/blob/master/doc/API.md#jitsimeetjs

I noticed `disableRtx` is missing there. Can you add it too?

bbaldino: thanks saul...added them both

PR #357, opened 2017-02-08
don't add rtx streams on firefox

bbaldino:


bbaldino:
things appear to work fine with just this fix, but the sdp still doesn't technically look right.  don't merge this yet, let me see if i can do something to clean the rest up

gpolitis:
Merging as per our discussion.

bbaldino:
doesn't seem to be updating here, but the checks did pass. @gpolitis

PR #358, opened 2017-02-09
Strip rtx

bbaldino:
turns out sdp-simulcast already stripped rtx streams/groups that were related to the sim streams it removed, so we don't need `implodeRemoteRtxSsrcs`.  however, after discussion, we decided to strip all rtx from the offer for firefox so we don't run into any weirdness when doing sdp-interop.

actually, @gpolitis, maybe the check should be based on whether or not we're doing unified plan (instead of checking fore firefox)? what do you think?

### DIFF
--- a/modules/xmpp/RtxModifier.js
+++ b/modules/xmpp/RtxModifier.js
@@ -222,15 +222,13 @@ export default class RtxModifier {
     }
 
     /**
-     * Remove all reference to any rtx ssrcs that 
-     *  don't correspond to the primary stream.
-     * Must be called *after* any simulcast streams
-     *  have been imploded
+     * Strip all rtx streams from the given sdp
      * @param {string} sdpStr sdp in raw string format
+     * @returns {string} sdp string with all rtx streams stripped
      */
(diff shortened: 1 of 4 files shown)


modules/xmpp/RtxModifier.js
```diff
@@ -222,55 +222,43 @@ export default class RtxModifier {
...
+        const videoMLine = 
             parsedSdp.media.find(mLine => mLine.type === "video");
         if (videoMLine.direction === "inactive" ||
                 videoMLine.direction === "recvonly") {
             logger.info("RtxModifier doing nothing, video " +
                 "m line is inactive or recvonly");
             return sdpStr;
         }
+        if (!videoMLine.ssrcs) {
+          logger.info("RtxModifier doing nothing, no video ssrcs present");
```
saghul: should this be info? IMHO debug level matches better.

bbaldino: i think you're right...i moved all the logs in this file to debug

gpolitis:
tests passed http://ci.jitsi.org/view/pr-testing/job/lib-jitsi-meet-PR-test/727/

PR #362, opened 2017-02-17
Peer to peer

paweldomas:
Work in progress...

Corresponding PR in jitsi-meet: 
https://github.com/jitsi/jitsi-meet/pull/1329
and jitsi-meet-torture:
https://github.com/jitsi/jitsi-meet-torture/pull/68

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -263,15 +263,25 @@ JitsiConference.prototype.getExternalAuthUrl = function (urlForPopup) {
 JitsiConference.prototype.getLocalTracks = function (mediaType) {
     let tracks = [];
     if (this.rtc) {
-        tracks = this.rtc.localTracks.slice();
-    }
-    if (mediaType !== undefined) {
-        tracks = tracks.filter(
-            (track) => {
-                return track && track.getType && track.getType() === mediaType;
-            });
+        tracks = this.rtc.getLocalTracks(mediaType);
     }
     return tracks;
+};
+
+/**
+ * Obtains local audio track.
+ * @return {JitsiLocalTrack|undefined}
+ */
+JitsiConference.prototype.getLocalAudioTrack = function () {
+    return this.rtc ? this.rtc.getLocalAudioTrack() : undefined;
+};
+
+/**
+ * Obtains local video track.
+ * @return {JitsiLocalTrack|undefined}
+ */
+JitsiConference.prototype.getLocalVideoTrack = function () {
+    return this.rtc ? this.rtc.getLocalVideoTrack() : undefined;
 };
 
 /**
@@ -395,18 +405,15 @@ JitsiConference.prototype.getTranscriber = function(){
     if (this.transcriber === undefined){
         this.transcriber = new Transcriber();
         //add all existing local audio tracks to the transcriber
-        // FIXME accessing localTracks field directly
-        this.rtc.localTracks.forEach(function (localTrack) {
-            if (localTrack.isAudioTrack()){
-                this.transcriber.addTrack(localTrack);
-            }
-        }.bind(this));
+        this.getLocalTracks(MediaType.AUDIO).forEach(
+            function (localAudio) {
+                this.transcriber.addTrack(localAudio);
+            }, this);
         //and all remote audio tracks
         this.rtc.getRemoteTracks(MediaType.AUDIO).forEach(
             function (remoteTrack){
                 this.transcriber.addTrack(remoteTrack);
-            }.bind(this)
-        );
+            }, this);
     }
     return this.transcriber;
 };
@@ -1030,7 +1037,7 @@ function (jingleSession, jingleOffer, now) {
 
     this.rtc.initializeDataChannels(jingleSession.peerconnection);
     // Add local Tracks to the ChatRoom
-    this.rtc.localTracks.forEach(function(localTrack) {
+    this.getLocalTracks().forEach(function(localTrack) {
         var ssrcInfo = null;
         /**
          * We don't do this for Firefox because, on Firefox, we keep the
@@ -1126,7 +1133,7 @@ JitsiConference.prototype.onCallEnded
     // will learn what their SSRC from the new PeerConnection which will be
     // created on incoming call event.
     var self = this;
-    this.rtc.localTracks.forEach(function(localTrack) {
+    this.getLocalTracks().forEach(function(localTrack) {
         // Reset SSRC as it will no longer be valid
         localTrack._setSSRC(null);
         // Bind the handler to fetch new SSRC, it will un register itself once
@@ -1179,21 +1186,25 @@ JitsiConference.prototype.myUserId = function () {
 };
 
 JitsiConference.prototype.sendTones = function (tones, duration, pause) {
+    // FIXME P2P 'dtmfManager' must be cleared, after switching jingleSessions
     if (!this.dtmfManager) {
-        var connection = this.xmpp.connection.jingle.activecall.peerconnection;
-        if (!connection) {
-            logger.warn("cannot sendTones: no conneciton");
+        if (!this.jingleSession) {
+            logger.warn("cannot sendTones: no jingle session");
             return;
         }
 
-        var tracks = this.getLocalTracks().filter(function (track) {
-            return track.isAudioTrack();
-        });
-        if (!tracks.length) {
+        const peerConnection = this.jingleSession.peerconnection;
+        if (!peerConnection) {
+            logger.warn("cannot sendTones: no peer connection");
+            return;
+        }
+
+        const localAudio = this.getLocalAudioTrack();
+        if (!localAudio) {
             logger.warn("cannot sendTones: no local audio stream");
             return;
         }
-        this.dtmfManager = new JitsiDTMFManager(tracks[0], connection);
+        this.dtmfManager = new JitsiDTMFManager(localAudio, peerConnection);
     }
 
     this.dtmfManager.sendTones(tones, duration, pause);
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -550,13 +550,15 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
  * the process or <tt>null</t> if the method should act as "add track"
  * @param {JitsiLocalTrack|null} newTrack the new track to be added or
  * <tt>null</tt> if the method should act as "remove track"
- * @return {Promise}
+ * @return {Promise} resolved when the process is done or rejected with a string
+ * which describes the error.
  * @private
  */
 JitsiConference.prototype._doReplaceTrack = function (oldTrack, newTrack) {
     if (this.jingleSession) {
         return this.jingleSession.replaceTrack(oldTrack, newTrack);
     } else {
+        logger.info("_doReplaceTrack - no JVB JingleSession");
         return Promise.resolve();
     }
 };
@@ -623,38 +625,33 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
  * Add track stream.
  * @param {JitsiLocalTrack} track the local track that will be added as part of
  * the unmute operation.
- * @param {function} callback callback executed, after successful stream
- * addition.
- * @param {function} errorCallback callback executed if stream addition fails.
+ * @return {Promise} resolved when the process is done or rejected with a string
+ * which describes the error.
  */
-JitsiConference.prototype._addLocalTrackAsUnmute
-    = function (track, callback, errorCallback) {
+JitsiConference.prototype._addLocalTrackAsUnmute = function (track) {
     if (this.jingleSession) {
-        this.jingleSession.addTrackAsUnmute(track)
-            .then(callback).catch(errorCallback);
+        return this.jingleSession.addTrackAsUnmute(track);
     } else {
         // We are done immediately
         logger.warn(
             "Add local MediaStream as unmute - no JingleSession started yet");
-        callback();
+        return Promise.resolve();
     }
 };
 
 /**
  * Removes given local track, as part of the mute operation.
  * @param {JitsiLocalTrack} track the local track that will be removed.
- * @param {function} callback callback executed after successful track removal.
- * @param {function} errorCallback callback executed if track removal fails.
+ * @return {Promise}
  */
-JitsiConference.prototype._removeTrackAsMute
-    = function (track, callback, errorCallback) {
+JitsiConference.prototype._removeTrackAsMute = function (track) {
     if (this.jingleSession) {
-        this.jingleSession.removeTrackAsMute(track)
-            .then(callback).catch(errorCallback);
+        return this.jingleSession.removeTrackAsMute(track);
     } else {
         // We are done immediately
-        logger.warn("Remove local MediaStream - no JingleSession started yet");
-        callback();
+        logger.warn(
+            "Remove local MediaStream - no JVB JingleSession started yet");
+        return Promise.resolve();
     }
 };
 
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -553,13 +553,6 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
             }
             return Promise.resolve();
         }, (error) => {
-            if (newTrack) {
-                // NOTE It's is unclear from this method perspective on where is
-                // the local track added, but it is added in _setupNewTrack
-
-                // Remove the local stream if the process fails
-                this.rtc.removeLocalTrack(newTrack);
-            }
             return Promise.reject(new Error(error));
         });
 };
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1122,8 +1122,10 @@ JitsiConference.prototype._rejectIncomingCall
  * @private
  */
 JitsiConference.prototype._startRemoteStats = function () {
-    if (this.jvbJingleSession)
-        this.statistics.startRemoteStats(this.jvbJingleSession.peerconnection);
+    const activePeerConnection = this.getActivePeerConnection();
+    if (activePeerConnection) {
+        this.statistics.startRemoteStats(activePeerConnection);
+    }
 };
 
 /**
@@ -1212,12 +1214,7 @@ JitsiConference.prototype.myUserId = function () {
 JitsiConference.prototype.sendTones = function (tones, duration, pause) {
     // FIXME P2P 'dtmfManager' must be cleared, after switching jingleSessions
     if (!this.dtmfManager) {
-        if (!this.jvbJingleSession) {
-            logger.warn("cannot sendTones: no jingle session");
-            return;
-        }
-
-        const peerConnection = this.jvbJingleSession.peerconnection;
+        const peerConnection = this.getActivePeerConnection();
         if (!peerConnection) {
             logger.warn("cannot sendTones: no peer connection");
             return;
@@ -1318,6 +1315,16 @@ JitsiConference.prototype.getPhonePin = function () {
     if (this.room)
         return this.room.getPhonePin();
     return null;
+};
+
+/**
+ * Obtains currently active <tt>TraceablePeerConnection</tt> (if any).
+ * @return {TraceablePeerConnection|null}
+ * @protected
+ */
+JitsiConference.prototype.getActivePeerConnection = function () {
+    return this.jvbJingleSession
+        ? this.jvbJingleSession.peerconnection : null;
 };
 
 /**
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -3,7 +3,6 @@
 var logger = require("jitsi-meet-logger").getLogger(__filename);
 import RTC from "./modules/RTC/RTC";
 import * as MediaType from "./service/RTC/MediaType";
-var RTCEvents = require("./service/RTC/RTCEvents");
 var EventEmitter = require("events");
 import * as JitsiConferenceErrors from "./JitsiConferenceErrors";
 import * as JitsiConferenceEvents from "./JitsiConferenceEvents";
@@ -17,7 +16,6 @@ var ComponentsVersions = require("./modules/version/ComponentsVersions");
 var GlobalOnErrorHandler = require("./modules/util/GlobalOnErrorHandler");
 var JitsiConferenceEventManager = require("./JitsiConferenceEventManager");
 var VideoType = require('./service/RTC/VideoType');
-var RTCBrowserType = require("./modules/RTC/RTCBrowserType.js");
 var Transcriber = require("./modules/transcription/transcriber");
 import ParticipantConnectionStatus
     from "./modules/connectivity/ParticipantConnectionStatus";
@@ -472,15 +470,12 @@ JitsiConference.prototype._fireMuteChangeEvent = function (track) {
  * @param track the JitsiLocalTrack object.
  */
 JitsiConference.prototype.onLocalTrackRemoved = function (track) {
-    track._setSSRC(null);
     track._setConference(null);
     this.rtc.removeLocalTrack(track);
     track.removeEventListener(JitsiTrackEvents.TRACK_MUTE_CHANGED,
         track.muteHandler);
     track.removeEventListener(JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED,
         track.audioLevelHandler);
-    this.rtc.removeListener(RTCEvents.SENDRECV_STREAMS_CHANGED,
-        track.ssrcHandler);
 
     // send event for stopping screen sharing
     // FIXME: we assume we have only one screen sharing track
@@ -523,17 +518,6 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
             return Promise.reject(
                 new JitsiTrackError(JitsiTrackErrors.TRACK_IS_DISPOSED));
         }
-        // Set up the ssrcHandler for the new track before we add it at the lower levels
-        newTrack.ssrcHandler = function (conference, ssrcMap) {
-            if (ssrcMap[this.getMSID()]) {
-                this._setSSRC(ssrcMap[this.getMSID()]);
-                conference.rtc.removeListener(
-                    RTCEvents.SENDRECV_STREAMS_CHANGED,
-                    this.ssrcHandler);
-            }
-        }.bind(newTrack, this);
-        this.rtc.addListener(RTCEvents.SENDRECV_STREAMS_CHANGED,
-            newTrack.ssrcHandler);
     }
     // Now replace the stream at the lower levels
     return this._doReplaceTrack(oldTrack, newTrack)
@@ -630,7 +614,7 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
 
 /**
  * Adds loca WebRTC stream to the conference.
- * @param {MediaStream} stream new stream that will be added.
+ * @param {JitsiLocalTrack} track new stream that will be added.
  * @param {function} callback callback executed after successful stream addition.
  * @param {function(error)} errorCallback callback executed if stream addition fail.
  * @param {object} ssrcInfo object with information about the SSRCs associated with the
@@ -640,13 +624,25 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
  * started. That is before the 'session-accept' is sent.
  */
 JitsiConference.prototype._addLocalStream
-    = function (stream, callback, errorCallback, ssrcInfo, dontModifySources) {
+    = function (track, callback, errorCallback, dontModifySources) {
     if (this.jingleSession) {
         this.jingleSession.addStream(
-            stream, callback, errorCallback, ssrcInfo, dontModifySources);
+            track, callback, errorCallback, dontModifySources);
     } else {
         // We are done immediately
         logger.warn("Add local MediaStream - no JingleSession started yet");
+        callback();
+    }
+};
+
+JitsiConference.prototype.addLocalWebRTCStreamAsUnmute
+    = function (track, callback, errorCallback, dontModifySources) {
+    if(this.jingleSession) {
+        this.jingleSession.addStreamAsUnmute(
+            track, callback, errorCallback, dontModifySources);
+    } else {
+        // We are done immediately
+        logger.warn("Add local MediaStream as unmute - no JingleSession started yet");
         callback();
     }
 };
@@ -656,33 +652,17 @@ JitsiConference.prototype._addLocalStream
  * @param {MediaStream} stream the stream that will be removed.
  * @param {function} callback callback executed after successful stream removal.
  * @param {function} errorCallback callback executed if stream removal fail.
- * @param {object} ssrcInfo object with information about the SSRCs associated
- * with the stream.
  */
-JitsiConference.prototype.removeLocalStream
-    = function (stream, callback, errorCallback, ssrcInfo) {
-    if (this.jingleSession) {
-        this.jingleSession.removeStream(
-            stream, callback, errorCallback, ssrcInfo);
+JitsiConference.prototype.removeTrackAsMute
+    = function (stream, callback, errorCallback) {
+    if(this.jingleSession) {
+        this.jingleSession.removeTrackAsMute(
+            stream, callback, errorCallback);
     } else {
         // We are done immediately
         logger.warn("Remove local MediaStream - no JingleSession started yet");
         callback();
     }
-};
-
-/**
- * Generate ssrc info object for a stream with the following properties:
- * - ssrcs - Array of the ssrcs associated with the stream.
- * - groups - Array of the groups associated with the stream.
- */
-JitsiConference.prototype._generateNewStreamSSRCInfo = function () {
-    if (!this.jingleSession) {
-        logger.warn("The call haven't been started. " +
-            "Cannot generate ssrc info at the moment!");
-        return null;
-    }
-    return this.jingleSession.generateNewStreamSSRCInfo();
 };
 
 /**
@@ -1037,46 +1017,10 @@ function (jingleSession, jingleOffer, now) {
     this.rtc.initializeDataChannels(jingleSession.peerconnection);
     // Add local Tracks to the ChatRoom
     this.getLocalTracks().forEach(function(localTrack) {
-        let ssrcInfo = null;
-        /**
-         * We don't do this for Firefox because, on Firefox, we keep the
-         *  stream in the peer connection and just set 'enabled' on the
-         *  track to false (see JitsiLocalTrack::_setMute).  This means
-         *  that if we generated an ssrc here and set it in the cache, it
-         *  would clash with the one firefox generates (since, unlike chrome,
-         *  the stream is still attached to the peer connection) and causes
-         *  problems between sdp-interop and trying to keep the ssrcs
-         *  consistent
-         */
-        if (localTrack.isVideoTrack() && localTrack.isMuted() && !RTCBrowserType.isFirefox()) {
-            /**
-             * Handles issues when the stream is added before the peerconnection
-             * is created. The peerconnection is created when second participant
-             * enters the call. In that use case the track doesn't have
-             * information about it's ssrcs and no jingle packets are sent. That
-             * can cause inconsistent behavior later.
-             *
-             * For example:
-             * If we mute the stream and than second participant enter it's
-             * remote SDP won't include that track. On unmute we are not sending
-             * any jingle packets which will brake the unmute.
-             *
-             * In order to solve issues like the above one here we have to
-             * generate the ssrc information for the track .
-             */
-            localTrack._setSSRC(this._generateNewStreamSSRCInfo());
-            ssrcInfo = {
-                mtype: localTrack.getType(),
-                type: "addMuted",
-                ssrcs: localTrack.ssrc.ssrcs,
-                groups: localTrack.ssrc.groups,
-                msid: localTrack.storedMSID
-            };
-        }
         try {
             this._addLocalStream(
-                localTrack.getOriginalStream(), function () {}, function () {},
-                ssrcInfo, true /* don't modify SSRCs */);
+                localTrack, function () {}, function () {},
+                true /* don't modify SSRCs */);
         } catch(e) {
             GlobalOnErrorHandler.callErrorHandler(e);
             logger.error(e);
@@ -1126,21 +1070,6 @@ JitsiConference.prototype.onCallEnded
     this.jingleSession = null;
     // Let the RTC service do any cleanups
     this.rtc.onCallEnded();
-    // PeerConnection has been closed which means that SSRCs stored in
-    // JitsiLocalTrack will not match those assigned by the old PeerConnection
-    // and SSRC replacement logic will not work as expected.
-    // We want to re-register 'ssrcHandler' of our local tracks, so that they
-    // will learn what their SSRC from the new PeerConnection which will be
-    // created on incoming call event.
-    var self = this;
-    this.getLocalTracks().forEach(function(localTrack) {
-        // Reset SSRC as it will no longer be valid
-        localTrack._setSSRC(null);
-        // Bind the handler to fetch new SSRC, it will un register itself once
-        // it reads the values
-        self.rtc.addListener(
-            RTCEvents.SENDRECV_STREAMS_CHANGED, localTrack.ssrcHandler);
-    });
 };
 
 /**
@@ -1414,16 +1343,27 @@ JitsiConference.prototype.isCallstatsEnabled = function () {
 /**
  * Handles track attached to container (Calls associateStreamWithVideoTag method
  * from statistics module)
- * @param track the track
+ * @param {JitsiLocalTrack|JitsiRemoteTrack} track the track
  * @param container the container
  */
 JitsiConference.prototype._onTrackAttach = function(track, container) {
-    var ssrc = track.getSSRC();
+    const isLocal = track.isLocal();
+    let ssrc = null;
+    if (isLocal) {
+        // Local tracks have SSRC stored on per peer connection basis
+        const peerConnection
+            = this.jingleSession && this.jingleSession.peerconnection;
+        if (peerConnection) {
+            ssrc = peerConnection.getLocalSSRC(track);
+        }
+    } else {
+        ssrc = track.getSSRC();
+    }
     if (!container.id || !ssrc) {
         return;
     }
     this.statistics.associateStreamWithVideoTag(
-        ssrc, track.isLocal(), track.getUsageLabel(), container.id);
+        ssrc, isLocal, track.getUsageLabel(), container.id);
 };
 
 /**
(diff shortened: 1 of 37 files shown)


JitsiConference.js
```diff
@@ -874,78 +1030,94 @@ function (jingleSession, jingleOffer, now) {
...
-    // happen in case if user doesn't have or denied permission to
-    // both camera and microphone.
-    this.statistics.startCallStats(jingleSession);
-    this.statistics.startRemoteStats(jingleSession.peerconnection);
+/**
+ * Method called to start remote stats for the current peer connection (if
+ * available).
+ * @private
+ */
+JitsiConference.prototype._ensureRemoteStatsRunning = function () {
```
bbaldino: nit: seems like just `_startRemoteStats` might be a more appropriate name?

JitsiConference.js
```diff
@@ -968,24 +1147,9 @@ JitsiConference.prototype.onCallEnded
         this.statistics.stopCallStats();
     }
     // Current JingleSession is invalid so set it to null on the room
```
bbaldino: looks like this comment needs to be updated (we no longer go through the room)

JitsiConference.js
```diff
@@ -968,24 +1147,9 @@ JitsiConference.prototype.onCallEnded
...
-    this.room.setJingleSession(null);
+    this.jingleSession = null;
```
bbaldino: i'm just curious, what's the thinking behind moving jinglesession out of room and directly to jitsiconference?

JitsiConference.js
```diff
@@ -1031,21 +1195,25 @@ JitsiConference.prototype.myUserId = function () {
 };
 
 JitsiConference.prototype.sendTones = function (tones, duration, pause) {
+    // FIXME P2P 'dtmfManager' must be cleared, after switching jingleSessions
```
bbaldino: is this fixed now?

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+export default class P2PEnabledConference extends JitsiConference {
+    /**
+     * Creates new <tt>P2PEnabledConference</tt>.
+     * @param options see description in {@link JitsiConference} constructor.
+     * @param {number} [options.config.backToP2PDelay=5] a delay given in
+     * seconds, before the conference switches back to P2P after the 3rd
+     * participant has left.
+     */
+    constructor(options) {
+        // Call super
```
bbaldino: seems like an unnecessary comment :)

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+        /**
+         * Fake <tt>ChatRoom</tt> passed to {@link peerToPeerSession}.
+         * @type {FakeChatRoomLayer}
+         */
+        this.peerToPeerFakeRoom = null;
+        /**
+         * A JingleSession for the direct peer to peer connection.
+         * @type {JingleSessionPC}
+         */
+        this.peerToPeerSession = null;
```
bbaldino: some places uses `p2p`, others use `peerToPeer`, would be nice to keep them consistent

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+     * element.
+     * @private
+     */
+    _acceptP2PIncomingCall (jingleSession, jingleOffer) {
+
+        jingleSession.setSSRCOwnerJid(this.room.myroomjid);
+
+        // Accept the offer
+        this.peerToPeerSession = jingleSession;
+        // FIXME .P2P should be set initially in strophe.jingle.js
```
bbaldino: is this fixed?

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+        const localTracks = this.getLocalTracks();
+
+        logger.debug("Adding " + localTracks + " to P2P...");
+        this.peerToPeerSession.addLocalTracks(localTracks).then(
+            () => {
+                logger.debug("Add " + localTracks + " to P2P done!");
+                this.peerToPeerSession.acceptOffer(
+                    jingleOffer,
+                    () => {
+                        logger.debug("Got RESULT for P2P 'session-accept'");
```
bbaldino: what does "RESULT" refer to here?

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+                    error);
+            });
+    }
+
+    /**
+     * @inheritDoc
+     * @override
+     */
+    _addLocalTrackAsUnmute (track) {
+        const all = [super._addLocalTrackAsUnmute(track)];
```
bbaldino: could we use a more descriptive variable name here?  at first i thought this was referring to 'all local tracks', but i see below it's actually a set of promises

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+    _createFakeRoom(isInitiator) {
+        return new FakeChatRoomLayer(this, isInitiator);
+    }
+
+    /**
+     * @inheritDoc
+     * @override
+     */
+    _doReplaceTrack (oldTrack, newTrack) {
+        const all = [super._doReplaceTrack(oldTrack, newTrack)];
```
bbaldino: same comment as above

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+        switch (eventType) {
+            // Log events which may be of interest for the P2P implementation
+            case JitsiConferenceEvents.CONNECTION_INTERRUPTED:
+            case JitsiConferenceEvents.CONNECTION_RESTORED:
+            case JitsiConferenceEvents.P2P_STATUS:
+                logger.debug(
+                    "_emitIntercept: block? " + shouldBlock, arguments);
+                break;
+        }
+        if (!shouldBlock)
```
bbaldino: i don't know what our style convention is, but personally if's without braces give me anxiety :)

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     * @override
+     * @private
+     */
+    _ensureRemoteStatsRunning () {
+        if (this.p2pEstablished) {
+            // Use remote starts from the P2P connection
```
bbaldino: "starts" -> "stats"

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+        // Start remote stats
+        logger.info("Starting remote stats with p2p connection");
+        this._ensureRemoteStatsRunning();
+    }
+
+    /**
+     * Detaches local tracks from the JVB connection.
+     * @private
+     */
+    _removeLocalTracksFromJVB() {
```
bbaldino: i think it'd be a bit clearer if the method name here matched the description more closely.  e.g. `_removeLocalTracksFromJvbPeerConnection`

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+            });
+    }
+
+    /**
+     * Removes from the conference remote tracks associated with the P2P
+     * connection.
+     * @private
+     */
+    _removeP2PRemoteTracks () {
+        this._removeRemoteTracks("P2P", this.peerToPeerSession);
```
bbaldino: wondering if `this.peerToPeerSession` should be something like `this.peerToPeerJingleSession` to be more consistent.  it's still a `JingleSession`, right?  i know it's much longer, but i think some way to make the distinctions between `this.JingleSession` and `this.peerToPeerSession` might be good.

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+
+    /**
+     * Sets new P2P status and updates some events/states hijacked from
+     * the <tt>JitsiConference</tt>.
+     * @param {boolean} newStatus the new P2P status value, <tt>true</tt> means
+     * that P2P is now in use, <tt>false</tt> means that the JVB connection is
+     * now in use.
+     * @private
+     */
+    _setNewP2PStatus (newStatus) {
```
bbaldino: `_setP2PStatus` seems ok?  the 'New' seems a bit redundant

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+    }
+
+    /**
+     * Starts new P2P session.
+     * @param {string} peerJid the JID of the remote participant
+     * @private
+     */
+    _startPeer2PeerSession(peerJid) {
+        if (this.deferredStartP2P) {
+            // Make not that the task has been executed
```
bbaldino: "not" -> "note"

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+            this.deferredStartP2P = null;
+        }
+        if (this.peerToPeerSession) {
+            logger.error("P2P session already started!");
+            return;
+        }
+
+        /*if (!peerJid) {
```
bbaldino: what's up with this code? is it needed?

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+    /**
+     * Method when called will decide whether it's the time to start or stop the
+     * P2P session.
+     * @param {boolean} userLeftEvent if <tt>true</tt> it means that the call
+     * originates from the user left event.
+     * @private
+     */
+    _startStopP2PSession (userLeftEvent) {
+        if (this.options.config.disableAutoP2P) {
+            logger.info("Auto P2P disabled");
```
bbaldino: what does "auto p2p" refer to?

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+                const myId = this.myUserId();
+                const peersId = peer.getId();
+                if (myId > peersId) {
+                    logger.debug(
+                        "Everyone's a moderator - "
+                            + "the other peer should start P2P", myId, peersId);
+                    // Abort
+                    return;
+                } else if (myId == peersId) {
+                    logger.error("The same IDs ? ", myId, peersId);
```
bbaldino: should we throw here?

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+            logger.info("Doing setAnswer");
+            this.peerToPeerSession.setAnswer(answer);
+        }
+    }
+
+    /**
+     * @inheritDoc
+     * @override
+     */
+    onCallEnded (JingleSession, reasonCondition, reasonText) {
```
bbaldino: did you mean to start this first arg with a capital letter? (`JingleSession`)

P2PEnabledConference.js
```diff
@@ -0,0 +1,899 @@
...
+     * Manually stops the current P2P session (should be used only in the tests)
+     */
+    stopPeer2PeerSession() {
+        this._stopPeer2PeerSession();
+    }
+}
+
+/**
+ * This is a fake {@link ChatRoom} passed to the P2P {@link JingleSessionPC}
+ * in order to capture events emitted on it's event emitter (Jingle session uses
```
bbaldino: "it's" -> "its", and maybe should do "Jingle session" -> "JingleSession"

modules/RTC/JitsiLocalTrack.js
```diff
@@ -47,8 +52,9 @@ function JitsiLocalTrack(stream, track, mediaType, videoType, resolution,
 
     this.deviceId = deviceId;
     this.startMuted = false;
-    this.initialMSID = this.getMSID();
+    this.storedMSID = this.getMSID();
     this.inMuteOrUnmuteProgress = false;
+    this.peerConnections = [];
```
bbaldino: can a track belong to more than one PeerConnection at a time?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -553,9 +582,11 @@ JitsiLocalTrack.prototype._setByteSent = function (bytesSent) {
     // the conference(and through the XMPP chat room ???) instead
     let iceConnectionState
         = this.conference ? this.conference.getConnectionState() : null;
+    // FIXME take "completed" into account for P2P
```
bbaldino: is this fixed?

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -12,23 +12,30 @@ var ttfmTrackerVideoAttached = false;
...
- * @param stream WebRTC MediaStream, parent of the track
- * @param track underlying WebRTC MediaStreamTrack for new JitsiRemoteTrack
- * @param mediaType the MediaType of the JitsiRemoteTrack
- * @param videoType the VideoType of the JitsiRemoteTrack
- * @param ssrc the SSRC number of the Media Stream
- * @param muted initial muted state of the JitsiRemoteTrack
+ * @param {RTC} rtc the RTC service instance.
+ * @param {JitsiConference} conference the conference to which this track
+ *        belongs to
+ * @param {string} owner the endpoint ID of the track owner
```
bbaldino: is it worth renaming this to `ownerEndpointId` to be more self-documenting?

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -12,23 +12,30 @@ var ttfmTrackerVideoAttached = false;
...
+ * @param {MediaStream} stream WebRTC MediaStream, parent of the track
+ * @param {MediaStreamTrack} track underlying WebRTC MediaStreamTrack for
+ *        the new JitsiRemoteTrack
+ * @param {MediaType} mediaType the type of the media
+ * @param {VideoType} videoType the type of the video if applicable
+ * @param {string} ssrc the SSRC number of the Media Stream
+ * @param {boolean} muted the initial muted state
+ * @param {boolean} isP2P indicates whether or not this track belongs to a P2P
```
bbaldino: so will the same "physical" track show up as a different JitsiRemoteTrack if it's forwarded directly over p2p vs via the bridge? i.e. will we get an entirely new JitsiRemoteTrack instance for the same track when it stops sending p2p and is sent via the bridge?

modules/RTC/JitsiTrack.js
```diff
@@ -60,10 +60,9 @@ function addMediaStreamInactiveHandler(mediaStream, handler) {
  *        onended/oninactive events of the stream.
  * @param trackMediaType the media type of the JitsiTrack
  * @param videoType the VideoType for this track if any
- * @param ssrc the SSRC of this track if known
  */
-function JitsiTrack(conference, stream, track, streamInactiveHandler, trackMediaType,
-                    videoType, ssrc)
+function JitsiTrack(conference, stream, track, streamInactiveHandler,
```
bbaldino: probably missed it elsewhere, but why remove ssrc from here?

modules/RTC/MungeLocalSdp.js
```diff
@@ -0,0 +1,295 @@
...
+        if (!transformer.selectMedia("audio")) {
+            logger.error(
+                "Unable to hack local audio track SDP - no 'audio' media");
+            return false;
+        }
+
+        let modified = false;
+
+        // FIXME not sure about this directions
+        if (["sendrecv", "recvonly", "sendonly"]
```
bbaldino: maybe easier to check if `transformer.mediaDirection !== 'inactive'` ?

modules/RTC/MungeLocalSdp.js
```diff
@@ -0,0 +1,295 @@
...
+                            // Inject the group
+                            logger.debug(
+                                "Injecting SIM group for " + videoTrack, group);
+                            transformer.addSSRCGroup(group);
+                        }
+                    }
+                    // Insert RTX
+                    // FIXME in P2P RTX is used by Chrome regardless of this
+                    // option status
+                    if (!this.pc.options.disableRtx) {
```
bbaldino: i think we should try and find a way for this to be able to leverage the existing RtxModifier logic.  i don't entirely follow the purpose of this class yet though, maybe we can talk more about that.

JitsiConference.js
```diff
@@ -47,6 +46,11 @@ function JitsiConference(options) {
     this._init(options);
     this.componentsVersions = new ComponentsVersions(this);
     this.participants = {};
+    /**
+     * Jingle Session instance
+     * @type {JingleSessionPC}
+     */
+    this.jingleSession = null;
```
paweldomas: Yes, previously JingleSession was stored in ChatRoom. I moved it in this commit:
https://github.com/jitsi/lib-jitsi-meet/pull/362/commits/bee0b28b933807272bd693e5f35429feb25f6688

JitsiConference.js
```diff
@@ -505,34 +536,52 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
...
+                this.onLocalTrackRemoved(oldTrack);
             }
             if (newTrack) {
                 // Now handle the addition of the newTrack at the JitsiConference level
                 this._setupNewTrack(newTrack);
             }
             return Promise.resolve();
         }, (error) => {
+            if (newTrack) {
+                // NOTE It's is unclear from this method perspective on where is
```
paweldomas: hmm yes, you're right. I remember it was fixing some problem, but will double check that. Maybe I messed up something when rebasing.

JitsiConference.js
```diff
@@ -505,34 +536,52 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
...
+ * <tt>null</tt> if the method should act as "remove track"
+ * @return {Promise} resolved when the process is done or rejected with a string
+ * which describes the error.
+ * @private
+ */
+JitsiConference.prototype._doReplaceTrack = function (oldTrack, newTrack) {
+    if (this.jingleSession) {
+        return this.jingleSession.replaceTrack(oldTrack, newTrack);
+    } else {
+        logger.info("_doReplaceTrack - no JVB JingleSession");
```
paweldomas: No, it can be only Jvb JingleSession and maybe it should be renamed this way. P2P session is stored in P2PEnabledConference

JitsiConference.js
paweldomas: For me it fells weird that ChatRoom was involved in dealing with the JingleSession. I guess it was because we were missing some information for creating remote tracks like which user owns which track, video type etc. I created a SignallingLayer to be able to split that out. If you look at code in the master branch you will see that it's not doing anything useful with the session, but being just another "hop" in the execution chain.

P2PEnabledConference.js
paweldomas: Which one do you prefer :) ?

P2PEnabledConference.js
paweldomas: Still working on that

P2PEnabledConference.js
paweldomas: It refers to a confirmation/ack packet as defined by the XMPP:
https://xmpp.org/rfcs/rfc3920.html#stanzas-semantics-iq

P2PEnabledConference.js
paweldomas: No, I forgot to remove it. It will go away. Thanks !

P2PEnabledConference.js
paweldomas: It disables automatic switching between JVB and P2P and then it can be done manually using this methods: https://github.com/jitsi/lib-jitsi-meet/blob/peer-to-peer/P2PEnabledConference.js#L763. I use this property in torture tests to enforce switch from JVB to P2P at specific state - when user is video muted. I will add better description to the method and will describe this config option.

P2PEnabledConference.js
paweldomas: Yes, it should "abort" at least, but I would not "throw", because if it's executed from Strophe packet processor for the "user joined room" presence it will remove this handler and break the whole app. It can continue without P2P though.

P2PEnabledConference.js
paweldomas: No, will fix that, thanks !

modules/RTC/JitsiLocalTrack.js
paweldomas: According to the current switching logic between P2P and JVB it can not, but technically, as defined by the whole logic about add/remove, attach/detach and video mute it can.

paweldomas: I should also describe this field.

modules/RTC/JitsiLocalTrack.js
paweldomas: No, still need to fix that, thanks !

modules/RTC/JitsiRemoteTrack.js
paweldomas: yes, the can have the same "msid", but will be "spawned" by different PeerConnection and will be a different JitsiRemoteTrack instance.

modules/RTC/JitsiTrack.js
paweldomas: Ssrc is no longer part of the JitsiLocalTrack, but only of JitsiRemoteTrack. For local tracks SSRCs have to be obtained through the TraceablePeerConnection. I guess it could still be stored in JitsiLocalTrack, but would have to be mapped on per peer connection basis.

modules/RTC/MungeLocalSdp.js
paweldomas: From what I could understand about RtxModifier it's responsibility is to make sure that RTX and SSRCs are present in the SDP and that they remain the same, after unmuting video track. The purpose of this class is to fake localDescription to prevent from sending source-add/source-remove and get rid of "modified SSRC info" state. It is also faking the SDP for detached tracks. I'm not sure what you mean exactly by leveraging the logic. Is it about exposing some methods and using them here ?

JitsiConference.js
paweldomas: I think it was fixing some intermittent problem which no longer exists and I messed something during rebase, so I will remove this commit completely, before merge happens. For now I will push a revert commit.

JitsiConference.js
paweldomas: I will push a fix soon, thanks !

P2PEnabledConference.js
paweldomas: I will change to p2p, because it's shorter

P2PEnabledConference.js
paweldomas: I'll push a fix proposal soon

P2PEnabledConference.js
paweldomas: Removed that code while refactoring for DTMF fix.

P2PEnabledConference.js
paweldomas: Yes, thanks ! Maybe not exactly throw, but at least "abort". Throwing could break Strophe's presence handler attached to member left/joined.

P2PEnabledConference.js
paweldomas: yes, thanks !

P2PEnabledConference.js
paweldomas: Jingle session was intentional just like chat room was not ChatRoom, so I will keep it this way if it's not a big deal for you. Or I will have to change both.

P2PEnabledConference.js
bbaldino: gotcha.  so is this basically just disabling the ability to go to p2p? if so we could just call it `disableP2P` maybe.

P2PEnabledConference.js
bbaldino: ok

modules/RTC/JitsiRemoteTrack.js
bbaldino: ok thanks

modules/RTC/MungeLocalSdp.js
bbaldino: this has a lot of overlap with the stuff in RtxModifier (and other things in the setLocalDescription transform chain in TPC), i think we should find a way to re-use the logic so if we find bugs or make changes we don't have to do it in two places

paweldomas: I guess we need to discuss it tomorrow, as I'm not sure how do you imagine it exactly

JitsiConference.js
```diff
@@ -899,6 +900,32 @@ JitsiConference.prototype.onRemoteTrackAdded = function (track) {
...
 /**
+ * Callback called by the Jingle plugin when 'session-answer' is received.
+ * @param {JingleSessionPC} session the Jingle session for which an answer was
+ * received.
+ * @param {jQuery} answer a jQuery selector pointing to 'jingle' IQ element
+ */
+// eslint-disable-next-line no-unused-vars
+JitsiConference.prototype.onCallAccepted = function (session, answer) {
+    // JVB conference does not care
+    logger.error("onAcceptedCall - not implemented in JVB conference");
```
saghul: error seems a bit harsh here, debug?

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+        }
+    }
+
+    /**
+     * @inheritDoc
+     * @override
+     */
+    onCallAccepted (jingleSession, answer) {
+        if (this.peerToPeerSession === jingleSession) {
+            logger.info("Doing setAnswer");
```
saghul: debug? also mention that this is p2p related?

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+            // FIXME isFocus seems to be unreliable ? See fix note in ChatRoom
+            jingleSession.isP2P = !this.room.isFocus(jingleSession.peerjid);
+            // It is important to print that, as long as isFocus is unreliable.
+            logger.info(
+                "Marking session from " + jingleSession.peerjid
+                + (jingleSession.isP2P ? " as P2P" : " as *not* P2P"));
+        }
+        if (jingleSession.isP2P) {
+            const role = this.room.getMemberRole(jingleSession.peerjid);
+            if ('moderator' === role) {
```
saghul: doesn't this read better the other way around?

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+    }
+
+    /**
+     * @inheritDoc
+     * @override
+     */
+    onMemberLeft (jid) {
+        super.onMemberLeft(jid);
+
+        this._startStopP2PSession();
```
saghul: how about renaming this to `maybeStartStopP2PSession` ?

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+        }
+    }
+
+    /**
+     * Manually starts new P2P session (should be used only in the tests).
+     */
+    startPeer2PeerSession() {
+        const peers = this.getParticipants();
+        // Start peer to peer session
+        if (peers.length > 0) {
```
saghul: shouldn't this be exactly 1?

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+        this._stopPeer2PeerSession();
+    }
+}
+
+/**
+ * This is a fake {@link ChatRoom} passed to the P2P {@link JingleSessionPC}
+ * in order to capture events emitted on it's event emitter (Jingle session uses
+ * chat room's emitter to send events).
+ */
+class FakeChatRoomLayer {
```
saghul: Since this a fake ChatRoom, why not call it FakeChatRoom? That "layer" doesn't add much, does it?

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+         * @type {Array}
+         */
+        this.connectionTimes = [];
+
+        /**
+         * Maps options of the original <tt>ChatRoom</tt>
+         */
+        this.options = p2pConference.room.options;
+        if (!this.options) {
+            logger.error("ChatRoom.options are undefined");
```
saghul: when can this happen? why is it a problem?

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+            room.addPresenceListener(name, handler);
+        }));
+    }
+
+    /**
+     * @see SignallingLayer.getMediaPresenceInfo
+     */
+    getMediaPresenceInfo (endpointId, mediaType) {
+        let result = null;
+        this._forwardToChatRoom((room) =>{
```
saghul: style: `room => `

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -191,7 +194,8 @@ JitsiRemoteTrack.prototype._attachTTFMTracker = function (container) {
  * @return {string}
  */
 JitsiRemoteTrack.prototype.toString = function () {
-    return "RemoteTrack[" + this.owner + ", " + this.getType() + "]";
+    return "RemoteTrack[" + this.owner + ", " + this.getType()
```
saghul: template string?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1465,55 +1499,64 @@ TraceablePeerConnection.prototype.createAnswer
...
-                answer.sdp = localDescription.raw;
+                if (!isOffer) {
+                    let remoteDescription = new SDP(this.remoteDescription.sdp);
+                    let localDescription = new SDP(resultSdp.sdp);
+                    _fixAnswerRFC4145Setup(remoteDescription, localDescription);
+                    resultSdp.sdp = localDescription.raw;
+                }
 
-                let ssrcMap = extractSSRCMap(answer);
+                let ssrcMap = extractSSRCMap(resultSdp);
```
saghul: const

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -313,6 +320,18 @@ export default class ParticipantConnectionStatus {
     }
 
     /**
+     * Goes over every participant and updates connectivity status.
+     * Should be called when parameter which affects all of the participants
```
saghul: a parameter

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -313,6 +320,18 @@ export default class ParticipantConnectionStatus {
...
+     * is changed (P2P for example).
+     */
+    refreshStatusForAll() {
+        this.conference.getParticipants().forEach(
```
saghul: for.. of

JitsiConference.js
```diff
@@ -80,6 +84,7 @@ function JitsiConference(options) {
     this.connectionIsInterrupted = false;
 
 }
+JitsiConference.prototype.constructor = JitsiConference;
```
saghul: I see why this is needed, but given the magnitude of this change already, how about making JitsiConference an ES6 class as part of the process?

JitsiConference.js
```diff
@@ -47,6 +46,11 @@ function JitsiConference(options) {
     this._init(options);
     this.componentsVersions = new ComponentsVersions(this);
     this.participants = {};
+    /**
+     * Jingle session instance for the JVB connection.
+     * @type {JingleSessionPC}
+     */
+    this.jvbJingleSession = null;
```
saghul: where was this up until now?

JitsiConference.js
```diff
@@ -591,6 +633,40 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
...
+ * the unmute operation.
+ * @return {Promise} resolved when the process is done or rejected with a string
+ * which describes the error.
+ */
+JitsiConference.prototype._addLocalTrackAsUnmute = function (track) {
+    if (this.jvbJingleSession) {
+        return this.jvbJingleSession.addTrackAsUnmute(track);
+    } else {
+        // We are done immediately
+        logger.warn(
```
saghul: in the above function this is an info, we should pick one...

JitsiConference.js
```diff
@@ -591,6 +633,40 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
...
+ * @param {JitsiLocalTrack} track the local track that will be removed.
+ * @return {Promise}
+ */
+JitsiConference.prototype._removeTrackAsMute = function (track) {
+    if (this.jvbJingleSession) {
+        return this.jvbJingleSession.removeTrackAsMute(track);
+    } else {
+        // We are done immediately
+        logger.warn(
+            "Remove local MediaStream - no JVB JingleSession started yet");
```

P2PEnabledConference.js
```diff
@@ -0,0 +1,886 @@
...
+         * and running.
+         * @type {boolean}
+         */
+        this.isP2PConnectionInterrupted = false;
+        /**
+         * Flag set to <tt>true</tt> when P2P session has been established
+         * (ICE has been connected).
+         * @type {boolean}
+         */
+        this.p2pEstablished = false;
```
saghul: arent these 2 flags expressing a too similar thing?

P2PEnabledConference.js
```diff
@@ -0,0 +1,886 @@
...
+
+    /**
+     * Accept incoming P2P Jingle call.
+     * @param {JingleSessionPC} jingleSession the session instance
+     * @param {jQuery} jingleOffer a jQuery selector pointing to 'jingle' IQ
+     * element.
+     * @private
+     */
+    _acceptP2PIncomingCall (jingleSession, jingleOffer) {
+
```
saghul: extra newline?

P2PEnabledConference.js
```diff
@@ -0,0 +1,886 @@
...
+            // Log events which may be of interest for the P2P implementation
+            case JitsiConferenceEvents.CONNECTION_INTERRUPTED:
+            case JitsiConferenceEvents.CONNECTION_RESTORED:
+            case JitsiConferenceEvents.P2P_STATUS:
+                logger.debug(
+                    "_emitIntercept: block? " + shouldBlock, arguments);
+                break;
+        }
+        if (!shouldBlock) {
+            this._originalEmit.apply(this.eventEmitter, arguments);
```
saghul: you could use ES6 rest parameters here, for extra coolness!

P2PEnabledConference.js
```diff
@@ -0,0 +1,886 @@
...
+     * @return {EventEmitter}
+     * @private
+     */
+    _createEventEmitter () {
+        const self = this;
+        return {
+            emit: function (type) {
+                logger.debug("Fake emit: ", type, arguments);
+                switch (type) {
+                    case XMPPEvents.CONNECTION_ESTABLISHED:
```
saghul: who is triggering these ?

saghul:
I went through most of the PR and left some comments.  Here are some more general comments that apply to it all:

* I think I understand where the difference between the JitsiConference and the P2PEnabled one comes from, but the lines seems quite blurry, since both classes are somewhat aware of each other. I know it's a lot of code, but did you consider adding to the current JitsiConference instead of adding a new entity? If so, why did you discard the idea?

* The need for a fake chat room is not clear to me: any reason not to use DMs across the MUC we already have for the JVB connection?

* As we discussed on the phone: I wonder if starting a P2P peer connection just with a datachannel, then meassuring things and then deciding if moving streams is appropriate or not would simplify the current approach.

Please take these comments with a grain of salt, since I'm not neck deep into the implementation, so those are some things I could think of at 10k ft.

P2PEnabledConference.js
paweldomas: It does not disable P2P capabilities, but only automatic switching between P2P and JVB. The jitsi-meet-torture tests are doing it explicitly through public methods start and stop.

paweldomas: This line is important to understand why we need that option:
https://github.com/jitsi/lib-jitsi-meet/blob/peer-to-peer/JitsiConnection.js#L94

JitsiConference.js
paweldomas: Not really, if it would be debug we would never see it. The thing is that this method should never be called unless this a P2PEnabledConference which overrides this method. We will never see this message logged unless the code is broken.

JitsiConference.js
paweldomas: It was in the ChatRoom. I moved it in this commit: bee0b28
Brian asked the same question and you should be able to find our discussion above (sorry I don't know how to paste link to it)

JitsiConference.js
paweldomas: Please have mercy :) This will bring merge conflicts with any change on the master, because the indentation has to be changed. I did that for the JingleSessionPC already and had a lot of trouble afterwards.

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
+/* global __filename */
+
+import {getLogger} from "jitsi-meet-logger";
+const logger = getLogger(__filename);
```
paweldomas: Yes, I agree. I did it only, because I saw it done like this in few other places already, but will change that.

P2PEnabledConference.js
```diff
@@ -0,0 +1,726 @@
...
+
+    /**
+     * Creates fake {@link ChatRoom} which is to be used by the P2P Jingle
+     * Session.
+     * @param {boolean} isInitiator indicates whether the room is to be create
+     * for 'initiator' or 'responder'
+     * @return {FakeChatRoomLayer}
+     * @private
+     */
+    _createFakeRoom(isInitiator) {
```
paweldomas: Yes... there was more code initially, but now it's a one liner. Will remove it.

P2PEnabledConference.js
paweldomas: It can happen when someone breaks the code. Just trying to detect this early and indicate what is missing. I could wait for some inner piece of code to break, but given that compiled/minimised code will provide "a, b, c" names it may be hard to guess anything out of it.

JitsiConference.js
saghul: I see. Can you add a comment so someone can tackle that sooner rather than later? :-)

JitsiConference.js
saghul: Ack

P2PEnabledConference.js
paweldomas: Now I started thinking if it's the right place to do this check though...

JitsiConference.js
saghul: Ah, ok.

JitsiConference.js
```diff
@@ -1138,13 +1295,26 @@ JitsiConference.prototype.getPhonePin = function () {
 };
 
 /**
+ * Obtains currently active <tt>TraceablePeerConnection</tt> (if any).
+ * @return {TraceablePeerConnection|undefined}
+ * @protected
+ */
+JitsiConference.prototype.getActivePeerConnection = function () {
+    return this.jvbJingleSession
+        ? this.jvbJingleSession.peerconnection : undefined;
```
paweldomas: We don't, but it would be nice to agree on one. I will change it to null.

P2PEnabledConference.js
paweldomas: No, "p2pEstablished" means that the ICE has connected and that we're now actively using P2P connection. The interrupted flag is a transient state indicating there are problems with P2P, but we're still using it. I will update the docs.

P2PEnabledConference.js
paweldomas: What do you mean ? Should this line be removed or extra line added (where ?) ?

P2PEnabledConference.js
paweldomas: JingleSessionPC

P2PEnabledConference.js
paweldomas: Not sure what you mean exactly, but sounds interesting :) Will try to learn something about that

paweldomas:
> I think I understand where the difference between the JitsiConference and the P2PEnabled one comes from, but the lines seems quite blurry, since both classes are somewhat aware of each other. I know it's a lot of code, but did you consider adding to the current JitsiConference instead of adding a new entity? If so, why did you discard the idea?

I wanted to avoid making JitsiConference even bigger than it is already. I disagree that both classes are aware of each other. Only P2PEnabledConference is aware on what's going on in JitsiConference. There is one place in JitsiConference that logs error message which is to be removed/fixed. Right now P2PEnabledConference contains all the logic and tricks required for the P2P to happen. Maybe it would be worth asking others in the team what do they think about this solution. If they find it strange and unreadable then it needs to be changed.

paweldomas:
> The need for a fake chat room is not clear to me: any reason not to use DMs across the MUC we already have for the JVB connection?

I'm sorry, but I have no idea what a DM is ?

paweldomas:
> As we discussed on the phone: I wonder if starting a P2P peer connection just with a datachannel, then meassuring things and then deciding if moving streams is appropriate or not would simplify the current approach.

That is interesting idea, but I see it more as an improvement and a next step rather than simplification. We will still need attach/detach tricks to keep the JVB connection alive.

saghul:
> Also in the long run I imagine we don't want to maintain "non P2P" conference version.

This was the main reason for my suggestion.

> I'm sorry, but I have no idea what a DM is ? Direct message ?

Yep, direct messages between users in the room.

> That is interesting idea, but I see it more as an improvement and a next step rather than simplification. We will still need attach/detach tricks to keep the JVB connection alive.

Sure, just putting it on the record, maybe someone else has some thoughts on a similar direction.

P2PEnabledConference.js
saghul: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/rest_parameters

P2PEnabledConference.js
saghul: there seems to be an empty line after the function declaration

P2PEnabledConference.js
saghul: Thanks!

paweldomas:
> The need for a fake chat room is not clear to me: any reason not to use DMs across the MUC we already have for the JVB connection?

Fake chat room was added to be able to isolate P2P JingleSessionPC instance and be able to control it. It's about intercepting the interaction, between JingleSessionPC and ChatRoom instance. I don't see how DM are related to that, but maybe I can think of trying to get rid of that layer. I added it in very early stage of development and many things were separated. It may turn out it's possible to handle the events in a more elegant way.

saghul:
> Fake chat room was added to be able to isolate P2P JingleSessionPC instance and be able to control it. It's about intercepting the interaction, between JingleSessionPC and ChatRoom instance. I don't see how DM are related to that, but maybe I can think of trying to get rid of that layer. I added it in very early stage of development and many things were separated. It may turn out it's possible to handle the events in a more elegant way. I will take a look into that.

My train of thought goes as follows: when joining a conference, we always join a room. Then Jicofo sends each participant the session-initiate and we are connected through the JVB. (of course you know that, I'm trying to lay the whole picture here) So, given that, it sounded natural to me that in P2P mode users would send the session-initiate stanzas through the same medium.

paweldomas:
> My train of thought goes as follows: when joining a conference, we always join a room. Then Jicofo sends each participant the session-initiate and we are connected through the JVB. (of course you know that, I'm trying to lay the whole picture here) So, given that, it sounded natural to me that in P2P mode users would send the session-initiate stanzas through the same medium.

Yes, they still do use ChatRoom for that. It's doesn't change anything around that. The thing about passing ChatRoom to JingleSessionPC is confusing, but the only thing that happens there is that it uses it's eventEmitter to sends it's XMPPEvents and for obtaining some missing information about remote tracks (but that part was isolated in SignallingLayer). I will look into removing this dependency completely from JingleSessionPC.

P2PEnabledConference.js
paweldomas: For me in this case it's easier to grab whole 'arguments' array and pass it to 'apply'. Otherwise I would have to concat 'eventType' for 'apply', no ? How would you write it ? I must be missing something

paweldomas:
@bbaldino @saghul I am going to push some commits that fix these typos and renaming stuff, but after that I will be creating PRs gradually by pushing refactoring changes step by step. I'm going to keep this PR open for a while just to be able to run integration tests, but you can forget about it for now.

PR #365, opened 2017-02-21
rtx: set disableRtx default to true

saghul:
It used to be true by default, enforced by the library, without any
configuration option. This is the case for tag 1631, for example.

Later on, the default was changed to false and the configuration option is now
honored.

This creates a problem in the current situation:

* user A (using a new lib-jitsi-meet) creates the room and passes disableRtx as
  false (the new default)
* user B (using the old lib-jitsi-meet) joins the room
* RTX will be used because user B advertised RTX support and the conference has
  it enabled
* user B will fail to join because its lib-jitsi-meet is not RTX ready

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -894,7 +894,7 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function () {
     } else {
         ssrcInfo = {ssrcs: [SDPUtil.generateSsrc()], groups: []};
     }
-    if (!this.options.disableRtx) {
+    if (!this.options.disableRtx && !RTCBrowserType.isFirefox()) {
         // Specifically use a for loop here because we'll
         //  be adding to the list we're iterating over, so we
         //  only want to iterate through the items originally
--- a/modules/xmpp/moderator.js
+++ b/modules/xmpp/moderator.js
@@ -147,13 +147,11 @@ Moderator.prototype.createConferenceIq =  function () {
                 value: this.options.conference.channelLastN
             }).up();
     }
-    if (this.options.conference.disableRtx !== undefined) {
-        elem.c(
-            'property', {
-                name: 'disableRtx',
-                value: this.options.conference.disableRtx
-            }).up();
-    }
+    elem.c(
+        'property', {
+            name: 'disableRtx',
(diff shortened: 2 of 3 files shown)


saghul:
The avid reader might have noticed that having a "disableXXX" with a default value of "true" is more convoluted that it needs to be, as opposed to a "enableXXX" with a default value of "false".

With this in mind, I propose we change this to "useRtx" with a default value of "false".

modules/xmpp/TraceablePeerConnection.js
```diff
@@ -547,7 +547,9 @@ TraceablePeerConnection.prototype.createAnswer
                         dumpSDP(answer));
                 }
 
-                if (!this.session.room.options.disableRtx && !RTCBrowserType.isFirefox()) {
+                if (typeof this.session.room.options.disableRtx !== 'undefined'
```
bbaldino: so, if i'm reading this right, this is changing "disableRtx not defined" to be treated the same as "disableRtx: true", which i think is inconsistent with the way other values are treated in the config.  if we're gonna do this, i think it would be good to go all the way through with your other idea (change 'disableRtx' to 'useRtx'), which i think is a good idea anyway.

bbaldino: and actually, isn't the 'undefined' case eliminated here by the logic you added in moderator.js?

gpolitis:
@saghul  I think we need to discuss the scenario that you're describing above, because we want disableRtx to be the false by default, or, in other words, we want to enable RTX by default. The reason is because RTX allows for probing for bandwidth while protecting for losses.

saghul:
Update the PR, PTAL. As discussed:

- the default value is false
- always pass the value to jicofo
- fixed a few checks for RTX support (check the option + !firefox)

saghul:
Rebased.

saghul:
Rebased again. @bbaldino @gpolitis can you PTAL?

gpolitis:
LGTM

saghul:
Rebased one more time. If nobody objects to the change, can someone please merge it? I'd rather not be rebasing every couple of days...

gpolitis:
I'm sorry @saghul, this PR fell out of my radar.

saghul:
Thanks George! My day is glorious now! ✨

PR #371, opened 2017-02-23
Make sure the minimum resolution is no less than 320 x 180

saghul:
Otherwise Chrome has been seen using a simulcast layer with a resolution of 240
x 135, which can break implementations not supporting odd resolutions, such as
WebRTC on Android.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -100,26 +100,20 @@ var isDeviceChangeEventSupported = false;
 var rtcReady = false;
 
 function setResolutionConstraints(constraints, resolution) {
-    var isAndroid = RTCBrowserType.isAndroid();
+    // XXX Google Chrome has been observed to fall down to 240x135 which is an
+    // odd resolution and odd resolutions may cause crashes in WebRTC on Android
+    // at least:
+    // - https://bugs.chromium.org/p/webrtc/issues/detail?id=6651
+    // - https://bugs.chromium.org/p/webrtc/issues/detail?id=7206
+    // As a mitigation, do not fall bellow the following minimal resolution.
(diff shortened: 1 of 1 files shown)


saghul:
Relevant Chrome issues:

    * https://bugs.chromium.org/p/webrtc/issues/detail?id=6651
    * https://bugs.chromium.org/p/webrtc/issues/detail?id=7206

They are both fixed in WebRTC master, and apparently odd resolutions could still happen when desktop sharing... So I'm not sure what to do here. On one hand this patch seems to avoid the issue in the usual cases (a multi-party conference) but OTOH it doesn't cover everything... We do need a solution short term.

saghul:
Calling the cavalry! @jitsi/jitsi-meet

modules/RTC/RTCUtils.js
```diff
@@ -100,26 +100,20 @@ var isDeviceChangeEventSupported = false;
...
-        constraints.video.mandatory.minHeight = Resolutions[resolution].height;
+        constraints.video.mandatory.maxHeight = Resolutions[resolution].height;
+        constraints.video.mandatory.maxWidth = Resolutions[resolution].width;
     }
-    else if (isAndroid) {
-        // FIXME can't remember if the purpose of this was to always request
-        //       low resolution on Android ? if yes it should be moved up front
-        constraints.video.mandatory.minWidth = 320;
-        constraints.video.mandatory.minHeight = 180;
-        constraints.video.mandatory.maxFrameRate = 15;
```
hristoterezov: Why we don't need this anymore?

saghul: In all honesty: I don't know. I did some git archeology and found no clear reason for it, and given today we un RN on Android I thought it was a good idea to take it out and reevaluate if necessary.

PR #372, opened 2017-02-24
Move JingleSessionPC from ChatRoom to JitsiConference

paweldomas:
This PR detaches components one from another to end up with JitsiConference managing the jingle session instance instead of the ChatRoom.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -47,6 +47,11 @@ function JitsiConference(options) {
     this._init(options);
     this.componentsVersions = new ComponentsVersions(this);
     this.participants = {};
+    /**
+     * Jingle Session instance
+     * @type {JingleSessionPC}
+     */
+    this.jingleSession = null;
     this.lastDominantSpeaker = null;
     this.dtmfManager = null;
     this.somebodySupportsDTMF = false;
@@ -181,6 +186,11 @@ JitsiConference.prototype.leave = function () {
             // ChatRoom instance.
             this.getParticipants().forEach(
                 participant => this.onMemberLeft(participant.getJid()));
+            // Close the JingleSession
+            if (this.jingleSession) {
+                this.jingleSession.close();
+                this.jingleSession = null;
+            }
         });
     }
 
@@ -517,7 +527,7 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
             newTrack.ssrcHandler);
     }
     // Now replace the stream at the lower levels
-    return this.room.replaceStream (oldTrack, newTrack)
+    return this._doReplaceTrack(oldTrack, newTrack)
         .then(() => {
             if (oldTrack) {
                 this.onTrackRemoved(oldTrack);
@@ -530,6 +540,25 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
         }, (error) => {
             return Promise.reject(new Error(error));
         });
+};
+
+/**
+ * Replaces the tracks at the lower level by going through the Jingle session
+ * and WebRTC peer connection. The method will resolve immediately if there is
+ * currently no JingleSession started.
+ * @param {JitsiLocalTrack|null} oldTrack the track to be removed during
+ * the process or <tt>null</t> if the method should act as "add track"
+ * @param {JitsiLocalTrackn|null} newTrack the new track to be added or
+ * <tt>null</tt> if the method should act as "remove track"
+ * @return {Promise}
+ * @private
+ */
+JitsiConference.prototype._doReplaceTrack = function (oldTrack, newTrack) {
+    if (this.jingleSession) {
+        return this.jingleSession.replaceStream(oldTrack, newTrack);
+    } else {
+        return Promise.resolve();
+    }
 };
 
 /**
@@ -588,6 +617,62 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
         this.statistics.sendScreenSharingEvent(true);
 
     this.eventEmitter.emit(JitsiConferenceEvents.TRACK_ADDED, newTrack);
+};
+
+/**
+ * Adds loca WebRTC stream to the conference.
+ * @param {MediaStream} stream new stream that will be added.
+ * @param {function} callback callback executed after successful stream addition.
+ * @param {function(error)} errorCallback callback executed if stream addition fail.
+ * @param {object} ssrcInfo object with information about the SSRCs associated with the
+ * stream.
+ * @param {boolean} [dontModifySources] if true _modifySources won't be called.
+ * Used for streams added before the call start.
+ */
+JitsiConference.prototype.addLocalWebRTCStream
+    = function (stream, callback, errorCallback, ssrcInfo, dontModifySources) {
+    if(this.jingleSession) {
+        this.jingleSession.addStream(
+            stream, callback, errorCallback, ssrcInfo, dontModifySources);
+    } else {
+        // We are done immediately
+        logger.warn("Add local MediaStream - no JingleSession started yet");
+        callback();
+    }
+};
+
+/**
+ * Remove local WebRTC media stream.
+ * @param {MediaStream} stream the stream that will be removed.
+ * @param {function} callback callback executed after successful stream removal.
+ * @param {function} errorCallback callback executed if stream removal fail.
+ * @param {object} ssrcInfo object with information about the SSRCs associated
+ * with the stream.
+ */
+JitsiConference.prototype.removeLocalWebRTCStream
+    = function (stream, callback, errorCallback, ssrcInfo) {
+    if(this.jingleSession) {
+        this.jingleSession.removeStream(
+            stream, callback, errorCallback, ssrcInfo);
+    } else {
+        // We are done immediately
+        logger.warn("Remove local MediaStream - no JingleSession started yet");
+        callback();
+    }
+};
+
+/**
+ * Generate ssrc info object for a stream with the following properties:
+ * - ssrcs - Array of the ssrcs associated with the stream.
+ * - groups - Array of the groups associated with the stream.
+ */
+JitsiConference.prototype._generateNewStreamSSRCInfo = function () {
+    if(!this.jingleSession) {
+        logger.warn("The call haven't been started. " +
+            "Cannot generate ssrc info at the moment!");
+        return null;
+    }
+    return this.jingleSession.generateNewStreamSSRCInfo();
 };
 
 /**
@@ -875,7 +960,7 @@ function (jingleSession, jingleOffer, now) {
     }
 
     // Accept incoming call
-    this.room.setJingleSession(jingleSession);
+    this.jingleSession = jingleSession;
     this.room.connectionTimes["session.initiate"] = now;
     // Log "session.restart"
     if (this.wasStopped) {
@@ -927,8 +1012,7 @@ function (jingleSession, jingleOffer, now) {
              * In order to solve issues like the above one here we have to
              * generate the ssrc information for the track .
              */
-            localTrack._setSSRC(
-                this.room.generateNewStreamSSRCInfo());
+            localTrack._setSSRC(this._generateNewStreamSSRCInfo());
             ssrcInfo = {
                 mtype: localTrack.getType(),
                 type: "addMuted",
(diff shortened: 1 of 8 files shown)


paweldomas:
Requires https://github.com/jitsi/jitsi-meet-torture/pull/69

JitsiConference.js
```diff
@@ -591,6 +620,62 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
...
+ * Adds loca WebRTC stream to the conference.
+ * @param {MediaStream} stream new stream that will be added.
+ * @param {function} callback callback executed after successful stream addition.
+ * @param {function(error)} errorCallback callback executed if stream addition fail.
+ * @param {object} ssrcInfo object with information about the SSRCs associated with the
+ * stream.
+ * @param {boolean} [dontModifySources] if true _modifySources won't be called.
+ * Used for streams added before the call start.
+ */
+JitsiConference.prototype.addLocalWebRTCStream
```
bbaldino: do we need `WebRTC` in the name here? i don't think we use that elsewhere

JitsiConference.js
```diff
@@ -591,6 +620,62 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
...
+
+/**
+ * Remove local WebRTC media stream.
+ * @param {MediaStream} stream the stream that will be removed.
+ * @param {function} callback callback executed after successful stream removal.
+ * @param {function} errorCallback callback executed if stream removal fail.
+ * @param {object} ssrcInfo object with information about the SSRCs associated
+ * with the stream.
+ */
+JitsiConference.prototype.removeLocalWebRTCStream
```
bbaldino: same as above

JitsiConference.js
```diff
@@ -1160,8 +1244,8 @@ JitsiConference.prototype.getPhonePin = function () {
  * for its session.
  */
 JitsiConference.prototype.getConnectionState = function () {
-    if(this.room)
-        return this.room.getConnectionState();
+    if(this.jingleSession)
```
bbaldino: would you mind adding braces here?

modules/xmpp/JingleSessionPC.js
```diff
@@ -23,11 +23,30 @@ import * as JingleSessionState from "./JingleSessionState";
...
+ * @param ice_config the ICE servers config object as defined by the WebRTC
+ * standard.
+ * @param {object} options a set of config options
+ * @param {boolean} options.webrtcIceUdpDisable <tt>true</tt> to block UDP
+ * candidates.
+ * @param {boolean} options.webrtcIceTcpDisable <tt>true</tt> to block TCP
+ * candidates.
+ * @param {boolean} options.failICE it's an option used in the tests. Set to
+ * <tt>true</tt> to block any real candidates and make the ICE fail.
+ */
```
bbaldino: you're a hero

modules/xmpp/TraceablePeerConnection.js
```diff
@@ -12,9 +12,26 @@ var SDPUtil = require("./SDPUtil");
...
+ * @param {boolean} options.disableRtx if set to 'true' will disable the RTX
+ * @param {boolean} options.preferH264 if set to 'true' H264 will be preferred
+ * over other video codecs.
+ * @param {EventEmitter} eventEmitter the emitter which wil be used by the new
+ * instance to emit events.
+ *
+ * @constructor
+ */
+function TraceablePeerConnection(ice_config,
+                                 constraints, options, eventEmitter) {
```
bbaldino: wow, so we can remove all reference to JingleSession from TPC completely?  that's awesome...huge decoupling right there.

JitsiConference.js
paweldomas: we don't will remove it

JitsiConference.js
paweldomas: will do, no problem

modules/xmpp/TraceablePeerConnection.js
paweldomas: yes, those were tiny little references without any functional meaning. Meant only to make things more complex

JitsiConference.js
```diff
@@ -533,6 +543,25 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
...
+ * @param {JitsiLocalTrackn|null} newTrack the new track to be added or
+ * <tt>null</tt> if the method should act as "remove track"
+ * @return {Promise}
+ * @private
+ */
+JitsiConference.prototype._doReplaceTrack = function (oldTrack, newTrack) {
+    if (this.jingleSession) {
+        return this.jingleSession.replaceStream(oldTrack, newTrack);
+    } else {
+        return Promise.resolve();
```
saghul: shouldn't this fail instead?

JitsiConference.js
```diff
@@ -533,6 +543,25 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
...
+ * @param {JitsiLocalTrack|null} oldTrack the track to be removed during
+ * the process or <tt>null</t> if the method should act as "add track"
+ * @param {JitsiLocalTrackn|null} newTrack the new track to be added or
+ * <tt>null</tt> if the method should act as "remove track"
+ * @return {Promise}
+ * @private
+ */
+JitsiConference.prototype._doReplaceTrack = function (oldTrack, newTrack) {
+    if (this.jingleSession) {
+        return this.jingleSession.replaceStream(oldTrack, newTrack);
```
saghul: this function deals with tracks, while the underlying one is called `replaceStream`. Can me we make this consistent?

JitsiConference.js
```diff
@@ -591,6 +620,62 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
 };
 
 /**
+ * Adds loca WebRTC stream to the conference.
+ * @param {MediaStream} stream new stream that will be added.
+ * @param {function} callback callback executed after successful stream addition.
+ * @param {function(error)} errorCallback callback executed if stream addition fail.
+ * @param {object} ssrcInfo object with information about the SSRCs associated with the
+ * stream.
+ * @param {boolean} [dontModifySources] if true _modifySources won't be called.
```
saghul: can you clarify a bit when this is needed? is the public API of this function supposed to care? Does it make sense to have some internal `_addLocalStream` which allows for that option to be set but not the public one?

JitsiConference.js
```diff
@@ -591,6 +620,62 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
...
+ */
+JitsiConference.prototype.addLocalStream
+    = function (stream, callback, errorCallback, ssrcInfo, dontModifySources) {
+    if(this.jingleSession) {
+        this.jingleSession.addStream(
+            stream, callback, errorCallback, ssrcInfo, dontModifySources);
+    } else {
+        // We are done immediately
+        logger.warn("Add local MediaStream - no JingleSession started yet");
+        callback();
```
saghul: shouldn't this fail?

JitsiConference.js
```diff
@@ -591,6 +620,62 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
...
+ */
+JitsiConference.prototype.removeLocalStream
+    = function (stream, callback, errorCallback, ssrcInfo) {
+    if(this.jingleSession) {
+        this.jingleSession.removeStream(
+            stream, callback, errorCallback, ssrcInfo);
+    } else {
+        // We are done immediately
+        logger.warn("Remove local MediaStream - no JingleSession started yet");
+        callback();
```

JitsiConference.js
```diff
@@ -591,6 +620,62 @@ JitsiConference.prototype._setupNewTrack = function (newTrack) {
...
+};
+
+/**
+ * Generate ssrc info object for a stream with the following properties:
+ * - ssrcs - Array of the ssrcs associated with the stream.
+ * - groups - Array of the groups associated with the stream.
+ */
+JitsiConference.prototype._generateNewStreamSSRCInfo = function () {
+    if(!this.jingleSession) {
+        logger.warn("The call haven't been started. " +
```
saghul: hasn't

JitsiConference.js
```diff
@@ -1160,9 +1244,11 @@ JitsiConference.prototype.getPhonePin = function () {
  * for its session.
  */
 JitsiConference.prototype.getConnectionState = function () {
-    if(this.room)
-        return this.room.getConnectionState();
-    return null;
+    if(this.jingleSession) {
+        return this.jingleSession.getIceConnectionState();
+    } else {
```
saghul: no need for the else clause

modules/xmpp/ChatRoom.js
```diff
@@ -982,7 +900,6 @@ export default class ChatRoom extends Listenable {
      * rejected.
      */
     leave () {
-        this._dispose();
```
saghul: how is closing the session triggered now?

modules/xmpp/JingleSessionPC.js
saghul: 😍 😍

modules/xmpp/TraceablePeerConnection.js
saghul: 🎉

modules/xmpp/JingleSessionPC.js
```diff
@@ -87,7 +106,8 @@ JingleSessionPC.prototype.doInitialize = function () {
     this.peerconnection = new TraceablePeerConnection(
             this.connection.jingle.ice_config,
             RTC.getPCConstraints(),
-            this);
+            this.room.options,
+            this.room.eventEmitter);
```
saghul: maybe add a comment on why we pass the room's emitter?

modules/xmpp/JingleSessionPC.js
```diff
@@ -87,7 +106,8 @@ JingleSessionPC.prototype.doInitialize = function () {
     this.peerconnection = new TraceablePeerConnection(
             this.connection.jingle.ice_config,
             RTC.getPCConstraints(),
-            this);
+            this.room.options,
```
saghul: how about we copy the options and pass them along? it looks a bit weird that we pass some ChatRoom options to a PC.

JitsiConference.js
paweldomas: Not really, because the underlying pipe starting at JingleSessionPC accepts either JitsiLocalTrack or MediaStream. I have plans on fixing that in next PRs, but as long as it accepts either one it will always be confusing. Unless I call it reaplceStreamOrTrack

JitsiConference.js
paweldomas: No, it only means that we're done with replacing immediately, because the session does not exist yet. That's the case when we're sitting alone in the room and switching between "camera" and "screen" for example.

JitsiConference.js
paweldomas: Please see my answer above to similar question.

JitsiConference.js
saghul: If it accepts both I'd go for `reaplceStreamOrTrack`. 👍

JitsiConference.js
saghul: 👍

JitsiConference.js
paweldomas: In fact this whole method is kind of "lib-jitsi-meet" internal. I will rename it this way. I will describe when it can be true.

JitsiConference.js
paweldomas: Same as above

JitsiConference.js
saghul: 👍

JitsiConference.js
saghul: 👍

JitsiConference.js
saghul: 👍

JitsiConference.js
paweldomas: no need, but is it so big deal ? Brian asked my for braces, so I added them everywhere. I will leave it as is if you don't mind.

modules/xmpp/ChatRoom.js
paweldomas: It is done by JitsiConference which holds the instance:
https://github.com/jitsi/lib-jitsi-meet/pull/372/files#diff-40e35c919311b508dc5b45a7bf028f54R189

modules/xmpp/JingleSessionPC.js
paweldomas: but you mean to explicitly copy each option not the whole object, right ?

JitsiConference.js
saghul: No big deal! Even with braces you don't need the else :-)  We want to eventually use the same style as we do in jitsi-meet here. We are not there yet, but ideally we should try to align all new code with the new rules. For example, this will complain for having no space after `if`.

modules/xmpp/ChatRoom.js
saghul: Gotcha! 👍

modules/xmpp/JingleSessionPC.js
saghul: Ideally, yes.

PR #374, opened 2017-02-28
Convert to es6

paweldomas:
Converts MediaType, JingleSession and JingleSessionPC to ES6 syntax.

### DIFF
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -99,7 +99,7 @@ export default class JingleSession {
      * Note: currently only used on transport-info
      */
     // eslint-disable-next-line no-unused-vars
-    addIceCandidates (contents) {
+    addIceCandidates(contents) {
 
     }
 
@@ -107,7 +107,7 @@ export default class JingleSession {
      * Returns current state of this <tt>JingleSession</tt> instance.
      * @returns {JingleSessionState} the current state of this session instance.
      */
-    getState () {
+    getState() {
         return this.state;
     }
 
@@ -117,7 +117,7 @@ export default class JingleSession {
      * @param contents an array of Jingle 'content' elements.
      */
     // eslint-disable-next-line no-unused-vars
-    addSources (contents) {
+    addSources(contents) {
 
     }
 
@@ -127,7 +127,7 @@ export default class JingleSession {
      * @param contents an array of Jingle 'content' elements.
      */
     // eslint-disable-next-line no-unused-vars
-    removeSources (contents) {
+    removeSources(contents) {
 
     }
 
@@ -141,7 +141,7 @@ export default class JingleSession {
      * response is received.
      */
     // eslint-disable-next-line no-unused-vars
-    terminate (reason, text, success, failure) {
+    terminate(reason, text, success, failure) {
 
     }
 
@@ -156,7 +156,7 @@ export default class JingleSession {
      *        unrecoverable anyway)
      */
     // eslint-disable-next-line no-unused-vars
-    acceptOffer (jingle, success, failure) {
+    acceptOffer(jingle, success, failure) {
 
     }
 }
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -151,9 +151,8 @@ export default class JingleSession {
      * @param success callback called when we the incoming session has been
      * accepted
      * @param failure callback called when we fail for any reason, will supply
-     *        error object with details(which is meant more to be printed to
-     *        the logger than analysed in the code, as the error is
-     *        unrecoverable anyway)
+     * error object with details(which is meant more to be printed to the logger
+     * than analysed in the code, as the error is unrecoverable anyway)
      */
     // eslint-disable-next-line no-unused-vars
     acceptOffer(jingle, success, failure) {
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -13,8 +13,8 @@ export default class JingleSession {
 
     /**
      * Creates new <tt>JingleSession</tt>.
-     * @param {string} me our JID
      * @param {string} sid the Jingle session identifier
+     * @param {string} me our JID
      * @param {string} peerjid the JID of the remote peer
      * @param {Strophe.Connection} connection the XMPP connection
      * @param {Object} media_constraints the media constraints object passed to
@@ -22,9 +22,9 @@ export default class JingleSession {
      * @param {Object} ice_config the ICE servers config object as defined by
      * the WebRTC. Passed to the PeerConnection's constructor.
      */
-    constructor(me, sid, peerjid, connection, media_constraints, ice_config) {
-        this.me = me;
+    constructor(sid, me, peerjid, connection, media_constraints, ice_config) {
         this.sid = sid;
+        this.me = me;
         this.peerjid = peerjid;
         this.connection = connection;
         this.media_constraints = media_constraints;
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -1,8 +1,8 @@
-
-import {getLogger} from "jitsi-meet-logger";
-const logger = getLogger(__filename);
-
+/* global __filename */
+import { getLogger } from "jitsi-meet-logger";
 import * as JingleSessionState from "./JingleSessionState";
+
+const logger = getLogger(__filename);
 
 /*
  * JingleSession provides an API to manage a single Jingle session. We will
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -71,8 +71,8 @@ export default class JingleSession {
     initialize(isInitiator, room, rtc) {
         if (this.state !== null) {
             const errmsg
-                = 'attempt to initiate on session ' + this.sid + 'in state '
-                    + this.state;
+                = `attempt to initiate on session ${this.sid}
+                   in state ${this.state}`;
             logger.error(errmsg);
             throw new Error(errmsg);
         }
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -14,7 +14,7 @@ export default class JingleSession {
     /**
      * Creates new <tt>JingleSession</tt>.
      * @param {string} sid the Jingle session identifier
-     * @param {string} me our JID
+     * @param {string} localJid our JID
      * @param {string} peerjid the JID of the remote peer
      * @param {Strophe.Connection} connection the XMPP connection
      * @param {Object} media_constraints the media constraints object passed to
@@ -22,9 +22,10 @@ export default class JingleSession {
      * @param {Object} ice_config the ICE servers config object as defined by
      * the WebRTC. Passed to the PeerConnection's constructor.
      */
-    constructor(sid, me, peerjid, connection, media_constraints, ice_config) {
+    constructor(sid,
+                localJid, peerjid, connection, media_constraints, ice_config) {
         this.sid = sid;
-        this.me = me;
+        this.localJid = localJid;
         this.peerjid = peerjid;
         this.connection = connection;
         this.media_constraints = media_constraints;
@@ -79,8 +80,8 @@ export default class JingleSession {
         this.room = room;
         this.rtc = rtc;
         this.state = JingleSessionState.PENDING;
-        this.initiator = isInitiator ? this.me : this.peerjid;
-        this.responder = !isInitiator ? this.me : this.peerjid;
+        this.initiator = isInitiator ? this.localJid : this.peerjid;
+        this.responder = !isInitiator ? this.localJid : this.peerjid;
         this.doInitialize();
     }
 
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -34,7 +34,6 @@ export default class JingleSession {
         /**
          * Whether to use dripping or not. Dripping is sending trickle
          * candidates not one-by-one.
-         * Note: currently we do not support 'false'.
          */
         this.usedrip = true;
 
--- a/modules/xmpp/JingleSession.js
+++ b/modules/xmpp/JingleSession.js
@@ -1,148 +1,161 @@
-/*
- * JingleSession provides an API to manage a single Jingle session. We will
- * have different implementations depending on the underlying interface used
- * (i.e. WebRTC and ORTC) and here we hold the code common to all of them.
- */
+
 import {getLogger} from "jitsi-meet-logger";
 const logger = getLogger(__filename);
 
 import * as JingleSessionState from "./JingleSessionState";
 
-function JingleSession(me, sid, peerjid, connection,
-                       media_constraints, ice_config) {
-    /**
-     * Our JID.
-     */
-    this.me = me;
-
-    /**
-     * The Jingle session identifier.
-     */
-    this.sid = sid;
-
-    /**
-     * the JID of the remote peer.
-     */
-    this.peerjid = peerjid;
-
-    /**
-     * The XMPP connection.
-     */
-    this.connection = connection;
-
-    /**
-     * Whether to use dripping or not. Dripping is sending trickle candidates
-     * not one-by-one.
-     * Note: currently we do not support 'false'.
-     */
-    this.usedrip = true;
-
-    /**
-     *  When dripping is used, stores ICE candidates which are to be sent.
-     */
-    this.drip_container = [];
-
-    // Media constraints. Is this WebRTC only?
-    this.media_constraints = media_constraints;
-
-    // ICE servers config (RTCConfiguration?).
-    this.ice_config = ice_config;
-
-    // The chat room instance associated with the session.
-    this.room = null;
-
-    /**
-     * Jingle session state - uninitialized until {@link initialize} is called
-     * @type {JingleSessionState}
-     */
-    this.state = null;
-
-    /**
-     * The RTC service instance
-     * @type {RTC}
-     */
-    this.rtc = null;
-}
-
-/**
- * Prepares this object to initiate a session.
- * @param isInitiator whether we will be the Jingle initiator.
- * @param room <tt>ChatRoom<tt> for the conference associated with this session
- * @param {RTC} rtc the RTC service instance
+/*
+ * JingleSession provides an API to manage a single Jingle session. We will
+ * have different implementations depending on the underlying interface used
+ * (i.e. WebRTC and ORTC) and here we hold the code common to all of them.
  */
-JingleSession.prototype.initialize = function(isInitiator, room, rtc) {
-    if (this.state !== null) {
-        var errmsg
-            = 'attempt to initiate on session ' + this.sid + 'in state '
-                + this.state;
-        logger.error(errmsg);
-        throw new Error(errmsg);
+export default class JingleSession {
+
+    /**
+     * Creates new <tt>JingleSession</tt>.
+     * @param {string} me our JID
+     * @param {string} sid the Jingle session identifier
+     * @param {string} peerjid the JID of the remote peer
+     * @param {Strophe.Connection} connection the XMPP connection
+     * @param {Object} media_constraints the media constraints object passed to
+     * the PeerConnection onCreateAnswer/Offer as defined by the WebRTC.
+     * @param {Object} ice_config the ICE servers config object as defined by
+     * the WebRTC. Passed to the PeerConnection's constructor.
+     */
+    constructor(me, sid, peerjid, connection, media_constraints, ice_config) {
+        this.me = me;
+        this.sid = sid;
+        this.peerjid = peerjid;
+        this.connection = connection;
+        this.media_constraints = media_constraints;
+        this.ice_config = ice_config;
+
+        /**
+         * Whether to use dripping or not. Dripping is sending trickle
+         * candidates not one-by-one.
+         * Note: currently we do not support 'false'.
+         */
+        this.usedrip = true;
+
+        /**
+         *  When dripping is used, stores ICE candidates which are to be sent.
+         */
+        this.drip_container = [];
+
+        /**
+         * The chat room instance associated with the session.
+         * @type {ChatRoom}
+         */
+        this.room = null;
+
+        /**
+         * Jingle session state - uninitialized until {@link initialize} is
+         * called @type {JingleSessionState}
+         */
+        this.state = null;
+
+        /**
+         * The RTC service instance
+         * @type {RTC}
+         */
+        this.rtc = null;
     }
-    this.room = room;
-    this.rtc = rtc;
-    this.state = JingleSessionState.PENDING;
-    this.initiator = isInitiator ? this.me : this.peerjid;
-    this.responder = !isInitiator ? this.me : this.peerjid;
-    this.doInitialize();
-};
 
-/**
- * Finishes initialization.
- */
-JingleSession.prototype.doInitialize = function() {};
+    /**
+     * Prepares this object to initiate a session.
+     * @param {boolean} isInitiator whether we will be the Jingle initiator.
+     * @param {ChatRoom} room the chat room for the conference associated with
+     * this session
+     * @param {RTC} rtc the RTC service instance
+     */
+    initialize(isInitiator, room, rtc) {
+        if (this.state !== null) {
+            const errmsg
+                = 'attempt to initiate on session ' + this.sid + 'in state '
+                    + this.state;
+            logger.error(errmsg);
+            throw new Error(errmsg);
+        }
+        this.room = room;
+        this.rtc = rtc;
+        this.state = JingleSessionState.PENDING;
+        this.initiator = isInitiator ? this.me : this.peerjid;
+        this.responder = !isInitiator ? this.me : this.peerjid;
+        this.doInitialize();
+    }
 
-/**
- * Adds the ICE candidates found in the 'contents' array as remote candidates?
- * Note: currently only used on transport-info
- */
-// eslint-disable-next-line no-unused-vars
-JingleSession.prototype.addIceCandidates = function(contents) {};
+    /**
+     * The implementing class finishes initialization here. Called at the end of
+     * {@link initialize}.
+     */
+    doInitialize() {
 
-/**
- * Returns current state of this <tt>JingleSession</tt> instance.
- * @returns {JingleSessionState} the current state of this session instance.
- */
-JingleSession.prototype.getState = function () {
-    return this.state;
-};
+    }
 
-/**
- * Handles an 'add-source' event.
- *
- * @param contents an array of Jingle 'content' elements.
- */
-// eslint-disable-next-line no-unused-vars
-JingleSession.prototype.addSources = function(contents) {};
+    /**
+     * Adds the ICE candidates found in the 'contents' array as remote
+     * candidates?
+     * Note: currently only used on transport-info
+     */
+    // eslint-disable-next-line no-unused-vars
+    addIceCandidates (contents) {
 
-/**
- * Handles a 'remove-source' event.
- *
- * @param contents an array of Jingle 'content' elements.
- */
-// eslint-disable-next-line no-unused-vars
-JingleSession.prototype.removeSources = function(contents) {};
+    }
 
-/**
- * Terminates this Jingle session by sending session-terminate
- * @param reason XMPP Jingle error condition
- * @param text some meaningful error message
- * @param success a callback called once the 'session-terminate' packet has been
- * acknowledged with RESULT.
- * @param failure a callback called when either timeout occurs or ERROR response
- * is received.
- */
-// eslint-disable-next-line no-unused-vars
-JingleSession.prototype.terminate = function(reason, text, success, failure) {};
+    /**
+     * Returns current state of this <tt>JingleSession</tt> instance.
+     * @returns {JingleSessionState} the current state of this session instance.
+     */
+    getState () {
+        return this.state;
+    }
 
-/**
- * Handles an offer from the remote peer (prepares to accept a session).
- * @param jingle the 'jingle' XML element.
- * @param success callback called when we the incoming session has been accepted
- * @param failure callback called when we fail for any reason, will supply error
- *        object with details(which is meant more to be printed to the logger
- *        than analysed in the code, as the error is unrecoverable anyway)
- */
-// eslint-disable-next-line no-unused-vars
-JingleSession.prototype.acceptOffer = function(jingle, success, failure) {};
+    /**
+     * Handles an 'add-source' event.
+     *
+     * @param contents an array of Jingle 'content' elements.
+     */
+    // eslint-disable-next-line no-unused-vars
+    addSources (contents) {
 
-module.exports = JingleSession;
+    }
+
+    /**
+     * Handles a 'remove-source' event.
+     *
+     * @param contents an array of Jingle 'content' elements.
+     */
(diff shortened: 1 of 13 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1475 +19,1549 @@ import * as JingleSessionState from "./JingleSessionState";
...
+        this.lasticecandidate = false;
+        this.closed = false;
+
+        this.modifyingLocalStreams = false;
+        /**
+         * Used to keep state about muted/unmuted video streams
+         *  so we can prevent errant source-add/source-removes
+         *  from happening
+         */
+        this.modifiedSSRCs = {};
```
bbaldino: map?

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1475 +19,1549 @@ import * as JingleSessionState from "./JingleSessionState";
...
+
+        /**
+         * A map that stores SSRCs of remote streams. And is used only locally
+         * We store the mapping when jingle is received, and later is used
+         * onaddstream webrtc event where we have only the ssrc
+         * FIXME: This map got filled and never cleaned and can grow durring long
+         * conference
+         * @type {{}} maps SSRC number to jid
+         */
+        this.ssrcOwners = {};
```
bbaldino: map?

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1475 +19,1549 @@ import * as JingleSessionState from "./JingleSessionState";
...
+     * 'transport-replace'.
+     * @param success callback called when we receive 'RESULT' packet for
+     *        'transport-replace'
+     * @param failure function(error) called when we receive an error response
+     *        or when the request has timed out.
+     */
+    sendTransportReject(success, failure) {
+        // Send 'transport-reject', so that the focus will
+        // know that we've failed
+        let tReject = $iq({to: this.peerjid, type: 'set'})
```
bbaldino: my vote would be to just name this the longer 'transportReject' so it's clearer

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1475 +19,1549 @@ import * as JingleSessionState from "./JingleSessionState";
...
-    // Calling tree() to print something useful
-    term = term.tree();
-    logger.info("Sending session-terminate", term);
+    /**
+     * @inheritDoc
+     */
+    terminate (reason, text, success, failure) {
+        this.state = JingleSessionState.ENDED;
+
+        let term = $iq({ to: this.peerjid, type: 'set' })
```
bbaldino: same here "terminate" (or even "terminateIqMsg")

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1475 +19,1549 @@ import * as JingleSessionState from "./JingleSessionState";
...
-        let addSsrcInfo = this._parseSsrcInfoFromSourceAdd(elem, sdp);
+        this.readSsrcInfo(elem);
 
-        let newRemoteSdp = this._processRemoteAddSource(addSsrcInfo);
+        let workFunction = (finishedCallback) => {
+            const sdp = new SDP(this.peerconnection.remoteDescription.sdp);
+            const mySdp = new SDP(this.peerconnection.localDescription.sdp);
+            const addSsrcInfo = this._parseSsrcInfoFromSourceAdd(elem, sdp);
+
+        const newRemoteSdp = this._processRemoteAddSource(addSsrcInfo);
```
bbaldino: it looks like the indentation got messed up here?

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1475 +19,1549 @@ import * as JingleSessionState from "./JingleSessionState";
...
- *  should be added to the remote SDP
- * @returns type {SDP Object} the new remote SDP (after removing the lines
- *  in removeSsrcInfo
- */
-JingleSessionPC.prototype._processRemoteAddSource = function (addSsrcInfo) {
-    let remoteSdp = new SDP(this.peerconnection.remoteDescription.sdp);
-    addSsrcInfo.forEach(function(lines, idx) {
-        remoteSdp.media[idx] += lines;
-    });
-    remoteSdp.raw = remoteSdp.
```
bbaldino: i think this is because the diff is messing up...this wasn't changed, right?

bbaldino: was this like this before?  why not log the string directly?

service/RTC/MediaType.js
```diff
@@ -1,15 +1,9 @@
...
-     */
-    AUDIO: "audio",
-    /**
-     * The video type.
-     */
-    VIDEO: "video"
-};
-module.exports = MediaType;
\ No newline at end of file
+export const AUDIO = "audio";
```
bbaldino: is it better to do it this way then have an actual type that encompasses the two values?  maybe we want to at least name this like "MEDIA_TYPE_AUDIO" or something?  I actually think MediaType.AUDIO is nicer, but i dunno what the idiomatic way is.

modules/xmpp/JingleSessionPC.js
paweldomas: thanks !

service/RTC/MediaType.js
paweldomas: Hmm I know what you mean, but I noticed that we use this convention in few other places which have been converted to the ES6 already, so I tried to comply:
https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConferenceErrors.js
https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConferenceEvents.js

modules/xmpp/JingleSessionPC.js
paweldomas: hmm actually it looks good here. It must be Github

paweldomas: nope it's wrong. I'm blind :P

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
+     * @param {string} sid the Jingle session identifier
+     * @param {string} peerjid the JID of the remote peer
+     * @param {Strophe.Connection} connection the XMPP connection
+     * @param {Object} media_constraints the media constraints object passed to
+     * the PeerConnection onCreateAnswer/Offer as defined by the WebRTC.
+     * @param {Object} ice_config the ICE servers config object as defined by
+     * the WebRTC. Passed to the PeerConnection's constructor.
      */
-    this.me = me;
+    constructor(me, sid, peerjid, connection, media_constraints, ice_config) {
```
saghul: IMHO the sid should be the first parameter.

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
+        this.connection = connection;
+        this.media_constraints = media_constraints;
+        this.ice_config = ice_config;
+
+        /**
+         * Whether to use dripping or not. Dripping is sending trickle
+         * candidates not one-by-one.
+         * Note: currently we do not support 'false'.
+         */
+        this.usedrip = true;
```
saghul: Is support for it planned? Ever? If not maybe we can just remove the flag?

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
+     * Prepares this object to initiate a session.
+     * @param {boolean} isInitiator whether we will be the Jingle initiator.
+     * @param {ChatRoom} room the chat room for the conference associated with
+     * this session
+     * @param {RTC} rtc the RTC service instance
      */
-    this.sid = sid;
+    initialize(isInitiator, room, rtc) {
+        if (this.state !== null) {
+            const errmsg
```
saghul: use a template string

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
+        this.doInitialize();
+    }
 
     /**
-     * the JID of the remote peer.
+     * The implementing class finishes initialization here. Called at the end of
+     * {@link initialize}.
      */
-    this.peerjid = peerjid;
+    doInitialize() {
```
saghul: Is this private? then how about `_doInitialize` ?

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
+                = 'attempt to initiate on session ' + this.sid + 'in state '
+                    + this.state;
+            logger.error(errmsg);
+            throw new Error(errmsg);
+        }
+        this.room = room;
+        this.rtc = rtc;
+        this.state = JingleSessionState.PENDING;
+        this.initiator = isInitiator ? this.me : this.peerjid;
+        this.responder = !isInitiator ? this.me : this.peerjid;
```
saghul: Do we need these 2? Can't we simplify it by using a boolean?

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
 
     /**
-     * The XMPP connection.
+     * Adds the ICE candidates found in the 'contents' array as remote
+     * candidates?
+     * Note: currently only used on transport-info
      */
-    this.connection = connection;
+    // eslint-disable-next-line no-unused-vars
+    addIceCandidates (contents) {
```
saghul: private?

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
 
     /**
-     * Whether to use dripping or not. Dripping is sending trickle candidates
-     * not one-by-one.
-     * Note: currently we do not support 'false'.
+     * Returns current state of this <tt>JingleSession</tt> instance.
+     * @returns {JingleSessionState} the current state of this session instance.
      */
-    this.usedrip = true;
+    getState () {
```
saghul: I assume we have this because `this.state` is supposed to be private? If so, a more idiomatic way to do it isn using ES6 getters:

```
get state() {
    return this._state;
}
```

And we make state a private attribute of the instance. Consumers can access it as `session.state` just fine, but they cannot set it.

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
     /**
-     * Jingle session state - uninitialized until {@link initialize} is called
-     * @type {JingleSessionState}
+     * Handles a 'remove-source' event.
+     *
+     * @param contents an array of Jingle 'content' elements.
      */
-    this.state = null;
+    // eslint-disable-next-line no-unused-vars
+    removeSources (contents) {
```
saghul: style: no space before (

saghul: I see this one a number of times.

modules/xmpp/JingleSession.js
```diff
@@ -1,148 +1,161 @@
...
- */
-// eslint-disable-next-line no-unused-vars
-JingleSession.prototype.acceptOffer = function(jingle, success, failure) {};
+    /**
+     * Handles an offer from the remote peer (prepares to accept a session).
+     * @param jingle the 'jingle' XML element.
+     * @param success callback called when we the incoming session has been
+     * accepted
+     * @param failure callback called when we fail for any reason, will supply
+     *        error object with details(which is meant more to be printed to
```
saghul: the indentation is not consistent with the above.

modules/xmpp/JingleSessionPC.js
```diff
@@ -2,15 +2,15 @@
...
-var SDPDiffer = require("./SDPDiffer");
-var SDPUtil = require("./SDPUtil");
-var SDP = require("./SDP");
-var async = require("async");
-var XMPPEvents = require("../../service/xmpp/XMPPEvents");
-var RTCBrowserType = require("../RTC/RTCBrowserType");
-var GlobalOnErrorHandler = require("../util/GlobalOnErrorHandler");
-var Statistics = require("../statistics/statistics");
+import JingleSession from "./JingleSession";
+const SDPDiffer = require("./SDPDiffer");
```
saghul: import!

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
 
     /**
-     * A map that stores SSRCs of remote streams. And is used only locally
-     * We store the mapping when jingle is received, and later is used
-     * onaddstream webrtc event where we have only the ssrc
-     * FIXME: This map got filled and never cleaned and can grow durring long
-     * conference
-     * @type {{}} maps SSRC number to jid
+     * Creates new <tt>JingleSessionPC</tt>
+     * @param {string} me our JID
```
saghul: same as earlier: localjid?

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+     * @param media_constraints the media constraints object passed to
+     * createOffer/Answer, as defined by the WebRTC standard
+     * @param ice_config the ICE servers config object as defined by the WebRTC
+     * standard.
+     * @param {object} options a set of config options
+     * @param {boolean} options.webrtcIceUdpDisable <tt>true</tt> to block UDP
+     * candidates.
+     * @param {boolean} options.webrtcIceTcpDisable <tt>true</tt> to block TCP
+     * candidates.
+     * @param {boolean} options.failICE it's an option used in the tests. Set to
```
saghul: can we make this _failICE since it's just used for testing stuff?

modules/xmpp/JingleSessionPC.js
saghul: yes!

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+
+        /**
+         * A map that stores SSRCs of remote streams. And is used only locally
+         * We store the mapping when jingle is received, and later is used
+         * onaddstream webrtc event where we have only the ssrc
+         * FIXME: This map got filled and never cleaned and can grow during
+         * long conference
+         * @type {{}} maps SSRC number to jid
+         */
+        this.ssrcOwners = {};
```
saghul: Map

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-        this,
-        this.connection.jingle.ice_config,
-        /* Options */
-        {
-            disableSimulcast: this.room.options.disableSimulcast,
-            disableRtx: this.room.options.disableRtx,
-            preferH264: this.room.options.preferH264
-        });
+    doInitialize () {
+        const self = this;
```
saghul: 🔥 use arrow functions

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-            (self.peerconnection.signalingState === 'closed'
+            self.sendIceCandidate(candidate);
+        };
+        // Note there is a change in the spec about closed:
+        // This value moved into the RTCPeerConnectionState enum in
+        // the May 13, 2016 draft of the specification, as it reflects the state
+        // of the RTCPeerConnection, not the signaling connection. You now
+        // detect a closed connection by checking for connectionState to be
+        // "closed" instead.
+        // I suppose at some point this will be moved to onconnectionstatechange
```
saghul: does this  comment still stand?

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-            XMPPEvents.ICE_CONNECTION_STATE_CHANGED,
-            self.peerconnection.iceConnectionState);
-        switch (self.peerconnection.iceConnectionState) {
-            case 'connected':
-
-                // Informs interested parties that the connection has been restored.
-                if (self.peerconnection.signalingState === 'stable' && self.isreconnect)
-                    self.room.eventEmitter.emit(XMPPEvents.CONNECTION_RESTORED);
-                self.isreconnect = false;
+            logger.log("(TIME) ICE " + self.peerconnection.iceConnectionState +
```
saghul: use a template string

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-            sid: this.sid});
-
-    const localSDP = new SDP(this.peerconnection.localDescription.sdp);
-    for (var mid = 0; mid < localSDP.media.length; mid++) {
-        var cands = candidates.filter(function (el) { return el.sdpMLineIndex == mid; });
-        var mline = SDPUtil.parse_mline(localSDP.media[mid].split('\r\n')[0]);
-        if (cands.length > 0) {
-            var ice = SDPUtil.iceparams(localSDP.media[mid], localSDP.session);
+
+    sendIceCandidate (candidate) {
```
saghul: style: extra space before (

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+        const self = this;
```
saghul: 🔥

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+    sendIceCandidates (candidates) {
+        logger.log('sendIceCandidates', candidates);
+        const cand = $iq({to: this.peerjid, type: 'set'})
+            .c('jingle', {xmlns: 'urn:xmpp:jingle:1',
+                action: 'transport-info',
+                initiator: this.initiator,
+                sid: this.sid});
+
+        const localSDP = new SDP(this.peerconnection.localDescription.sdp);
+        for (let mid = 0; mid < localSDP.media.length; mid++) {
```
saghul: for..of ?

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+            if (cands.length > 0) {
+                const ice
+                    = SDPUtil.iceparams(localSDP.media[mid], localSDP.session);
+                ice.xmlns = 'urn:xmpp:jingle:transports:ice-udp:1';
+                cand.c('content', {
+                    creator: this.initiator == this.me
+                                    ? 'initiator' : 'responder',
+                    name: (cands[0].sdpMid ? cands[0].sdpMid : mline.media)
+                }).c('transport', ice);
+                for (let i = 0; i < cands.length; i++) {
```
saghul: for..of

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-                        = new SDP(this.peerconnection.localDescription.sdp);
-                    this.sendTransportAccept(localSDP, success, failure);
-                },
-                failure);
-        },
-        failure
-    );
-};
+    /**
+     * Does accept incoming Jingle 'session-initiate' and should send
```
saghul: s/Does accept/Accepts/

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+        localSDP.toJingle(
+            accept,
+            this.initiator == this.me ? 'initiator' : 'responder',
+            null);
+        this.fixJingle(accept);
+
+        // Calling tree() to print something useful
+        accept = accept.tree();
+        logger.info("Sending session-accept", accept);
+        const self = this;
```
saghul: 🔥

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+     * 'transport-replace'.
+     * @param localSDP the 'SDP' object with local session description
+     * @param success callback called when we receive 'RESULT' packet for
+     *        'transport-replace'
+     * @param failure function(error) called when we receive an error response
+     *        or when the request has timed out.
+     */
+    sendTransportAccept(localSDP, success, failure) {
+        const self = this;
+        let tAccept = $iq({to: this.peerjid, type: 'set'})
```
saghul: transportAccept ?

modules/xmpp/JingleSessionPC.js
saghul: agreed!

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+     *  into sdp lines
+     * @param {jquery xml element} sourceAddElem the source-add
+     *  element from jingle
+     * @param {SDP object} currentRemoteSdp the current remote
+     *  sdp (as of this new source-add)
+     * @returns {list} a list of SDP line strings that should
+     *  be added to the remote SDP
+     */
+    _parseSsrcInfoFromSourceAdd (sourceAddElem, currentRemoteSdp) {
+        let addSsrcInfo = [];
```
saghul: const, since you're not re-assigning.

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-    if (!this.peerconnection.localDescription) {
-        logger.warn("removeSource - localDescription not ready yet");
-        setTimeout(() => this.removeRemoteStream(elem), 200);
-        return;
+    /**
+     * Handles a Jingle source-add message for this Jingle session.
+     * @param elem An array of Jingle "content" elements.
+     */
+    addRemoteStream (elem) {
+        // FIXME: dirty waiting
```
saghul: fixme how? if there is no realistic way to fix it I vote we do away with the comment.

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+        if (!this.peerconnection.localDescription) {
+            logger.warn("addSource - localDescription not ready yet");
+            setTimeout(() => this.addRemoteStream(elem), 200);
```
saghul: use setTimeout consistently, sometimes I've seen `window.setTimeout`. I'd use just `setTimeout` since it's more portable, I think.

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+            return;
+        }
+        logger.log('Processing add remote stream');
+        logger.log(
+            'ICE connection state: ', this.peerconnection.iceConnectionState);
+
+        this.readSsrcInfo(elem);
+
+        let workFunction = (finishedCallback) => {
```
saghul: const

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+            this._renegotiate(newRemoteSdp)
+                .then(() => {
+                    logger.info("Remote source-add processed");
+                    const newSdp
+                        = new SDP(this.peerconnection.localDescription.sdp);
+                    logger.log("SDPs", mySdp, newSdp);
+                    this.notifyMySSRCUpdate(mySdp, newSdp);
+                    finishedCallback();
+                }, (error) => {
+                    logger.error("Error renegotiating after processing"
```
saghul: use a template string

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-            });
-    };
-    this.modificationQueue.push(workFunction);
-};
+    /**
+     * Handles a Jingle source-remove message for this Jingle session.
+     * @param elem An array of Jingle "content" elements.
+     */
+    removeRemoteStream (elem) {
+        // FIXME: dirty waiting
```
saghul: same as above

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
-};
+        //TODO(brian): in the code below there are 2 chunks of code that relate
+        //  to observing changes in local and remove ufrags.  since they
+        //  just need to read and observe the SDPs, we should create the
+        //  notion of an SDP observer in TraceablePeerConnection that
+        //  gets notified of all SDP changes.  Code like the ufrag
+        //  logic below could listen to that and be separated from
+        //  core flows like this.
+        return new Promise((resolve, reject) => {
+            let remoteUfrag = JingleSessionPC.getUfrag(remoteDescription.sdp);
```
saghul: const

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+            if (remoteUfrag != this.remoteUfrag) {
```
saghul: loose equality? is this intentional?

modules/xmpp/JingleSessionPC.js
```diff
@@ -19,1476 +19,1561 @@ import * as JingleSessionState from "./JingleSessionState";
...
+                this.remoteUfrag = remoteUfrag;
+
```
saghul: same as above

saghul: ditto

saghul: use a template string

saghul: ditto

saghul: style: put the body of the function on the next line
style: no parens needed for single parameter arrow functions

saghul: ditto

saghul: s/Stream/Track/

saghul: use template strings

saghul: const

saghul: spurious new line

saghul: is this fixme really necessary?

saghul: then this must be private, right? then it needs a leading underscore.

saghul: Remove a stream.

saghul: in the JS multiverse callbacks usually come last in the list of parameters

saghul: Don't we have the isFirefox helper for this?

saghul: sends

saghul: use a template string

saghul: ditto

saghul: ditto

saghul: Can you reword this? Doesn't parse well for me. How about "Build a error handler to be used with Stophe's connection.sendIQ..."

saghul: spurious newline

saghul: reverse the check: typeof X === "foo"

saghul: why static if we then want to set some this parameter?

saghul: great place to use a template string!

service/RTC/MediaType.js
saghul: Do we put these values into anything? Otherwise consider using a Symbol.

modules/xmpp/JingleSessionPC.js
paweldomas: I think it's possible to end up in an endless loop here, so the FIXME should stand, but with different comment.

modules/xmpp/JingleSessionPC.js
paweldomas: I don't know, but I don't want to make any functional changes in this PR and removing the comment would require to change the code. I don't have enough time to investigate that now.

modules/xmpp/JingleSession.js
paweldomas: Actually the only not supported thing is changing this var from true to false. I'll remove this comment.

modules/xmpp/JingleSession.js
paweldomas: I don't want to refactor this now.

modules/xmpp/JingleSession.js
paweldomas: it's protected, will add JS doc

modules/xmpp/JingleSession.js
paweldomas: old code, not touching that now

modules/xmpp/JingleSession.js
paweldomas: sounds good, but I can't fix all of the old code now

modules/xmpp/JingleSessionPC.js
paweldomas: not changing old code in this PR. this will require changes in torture as well

modules/xmpp/JingleSessionPC.js
paweldomas: not changing old code

modules/xmpp/JingleSessionPC.js
paweldomas: not changing old code

modules/xmpp/JingleSessionPC.js
paweldomas: out of scope of this PR

modules/xmpp/JingleSessionPC.js
paweldomas: out of scope

modules/xmpp/JingleSessionPC.js
paweldomas: out of scope

modules/xmpp/JingleSessionPC.js
paweldomas: out of scope

modules/xmpp/JingleSessionPC.js
paweldomas: out of scope

modules/xmpp/JingleSessionPC.js
paweldomas: out of scope

modules/xmpp/JingleSessionPC.js
paweldomas: I don't know, but don't want to change it know in case it could break

service/RTC/MediaType.js
paweldomas: Sounds interesting, but that's too much for this PR. I could break something for sure

bbaldino: ok, i think it's a bit of a bummer to lose some 'scoping' for the term and having everything essentially in the top-level namespace means we have to build the scoping into the name itself.  i think this is fine since you were consistent, i'll have to read more about the idiomatic way to do this in es6.

PR #375, opened 2017-03-01
Fix broken 'transport-replace'

paweldomas:
This PR fixes a crash when conference is being moved to new JVB.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -412,8 +412,9 @@ JingleSessionPC.prototype.replaceTransport = function (jingleOfferElem,
             this.setOfferCycle(
                 originalOffer,
                 () => {
-                    // Set local description OK, now localSDP up to date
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -412,8 +412,9 @@ JingleSessionPC.prototype.replaceTransport = function (jingleOfferElem,
             this.setOfferCycle(
                 originalOffer,
                 () => {
-                    // Set local description OK, now localSDP up to date
-                    this.sendTransportAccept(this.localSDP, success, failure);
+                    const localSDP
+                        = new SDP(this.peerconnection.localDescription.sdp);
+                    this.sendTransportAccept(localSDP, success, failure);
```
bbaldino: so this was broken because `this.localSDP` was removed, right?  i think i may have done that so maybe i caused this bug.  if so then lgtm.

paweldomas: yes, this.localSDP was removed. I tested this PR and it works fine

PR #377, opened 2017-03-02
Revert "Make sure the minimum resolution is no less than 320 x 180"

saghul:
This reverts commit 886bc1519de7830314819f9fded692956b1bbedf.

Turns out Chrome decides to not choose the maximum possible resolution when
given a range. Oh well!


saghul:
Sorry everyone, I was so fixated on the Android crash I didn't notice the lack of glorious HD.

PR #378, opened 2017-03-03
Fix screenshare without video

dwilson6:
This pull request fixes two issues:
1. the recvonly ssrc that is generated when the jingle session is initiated with no video track was broken by recent refactoring
2. screen sharing without a camera was broken because the jingle signaling wasn't happening to switch the video mline from recvonly to sendrecv

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -542,6 +542,10 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
         .then(() => {
             if (oldTrack) {
                 this.onLocalTrackRemoved(oldTrack);
+                if(!newTrack && oldTrack.isVideoTrack() && oldTrack.videoType === VideoType.DESKTOP) {
+                    this.removeCommand("videoType");
+                    this.room.sendPresence();
+                }
             }
             if (newTrack) {
                 // Now handle the addition of the newTrack at the JitsiConference level
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1085,7 +1085,7 @@ function (jingleSession, jingleOffer, now) {
     }.bind(this));
     // Generate the 'recvonly' SSRC in case there are no video tracks
     if (!this.getLocalTracks(MediaType.VIDEO).length) {
-        this.room.generateRecvonlySsrc();
+        jingleSession.generateRecvonlySsrc();
     }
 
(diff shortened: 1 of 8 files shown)


dwilson6:
@paweldomas here's a PR for the issues we were discussing on the mailing list

paweldomas:
Hi there !

Thanks for looking into that ! The first commit is fine, but the second one - I don't think it's the right way for fixing it. We should probably clear the SSRC cache, so that a new one will be generated and the source-remove/source-add will be sent (I haven't tested if that will work yet). I will try to take a look.

dwilson6:
Yeah, I wasn't sure what the right approach should be.  I can look at clearing the SSRC cache.  That does sound like a better solution.

modules/RTC/RTC.js
```diff
@@ -486,7 +486,7 @@ export default class RTC extends Listenable {
             delete this.remoteTracks[
                 toBeRemoved.getParticipantId()][toBeRemoved.getType()];
 
-            this.rtc.eventEmitter.emit(
```
paweldomas: so was this broken during refactoring as well ?

dwilson6: it looks like it was changed a couple months ago. I was seeing errors that the object was undefined (since this.rtc is naturally not defined in the RTC module).

paweldomas:
@dwilson6 thanks, that's what I meant !
I actually have worked on the fix as well and there is one more case to be handled. Can you give a try to this PR and let me know if it fixes your problems ? https://github.com/jitsi/lib-jitsi-meet/pull/380

JitsiConference.js
```diff
@@ -542,6 +542,10 @@ JitsiConference.prototype.replaceTrack = function (oldTrack, newTrack) {
         .then(() => {
             if (oldTrack) {
                 this.onLocalTrackRemoved(oldTrack);
+                if(!newTrack && oldTrack.isVideoTrack() && oldTrack.videoType === VideoType.DESKTOP) {
```
paweldomas: Are you testing these problems with jitsi-meet or some custom app ?
I imagine this line here fixes display issue which appears also when someone who is sharing the screen video mutes. I'm not comfortable with this change, because the root problem lies elsewhere.

dwilson6:
Closing this in favor of https://github.com/jitsi/lib-jitsi-meet/pull/380

PR #379, opened 2017-03-03
Add SdpTransformUtil.js

paweldomas:
SdpTransformUtil is a utility which encapsulates common operations on 'sdp-transform' lib objects. Currently used by SdpConsistency and RtxModifier, but there are plans for adding new classes in future refactoring towards the peer to peer.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -380,8 +380,8 @@ JitsiLocalTrack.prototype._addStreamToConferenceAsUnmute = function () {
             {
                 mtype: self.type,
                 type: "unmute",
-                ssrcs: self.ssrc ? self.ssrc.ssrcs : undefined,
-                groups: self.ssrc ? self.ssrc.groups: undefined,
+                ssrcs: self.ssrc && self.ssrc.ssrcs,
+                groups: self.ssrc && self.ssrc.groups,
                 msid: self.getMSID()
             });
     });
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -380,7 +380,8 @@ JitsiLocalTrack.prototype._addStreamToConferenceAsUnmute = function () {
             {
                 mtype: self.type,
                 type: "unmute",
-                ssrc: self.ssrc,
+                ssrcs: self.ssrc ? self.ssrc.ssrcs : undefined,
+                groups: self.ssrc ? self.ssrc.groups: undefined,
                 msid: self.getMSID()
             });
     });
@@ -406,7 +407,8 @@ function (successCallback, errorCallback) {
         {
             mtype: this.type,
             type: "mute",
-            ssrc: this.ssrc
+            ssrcs: this.ssrc ? this.ssrc.ssrcs : undefined,
+            groups: this.ssrc ? this.ssrc.groups : undefined
         });
 };
 
@@ -521,7 +523,7 @@ JitsiLocalTrack.prototype._setConference = function(conference) {
  */
 JitsiLocalTrack.prototype.getSSRC = function () {
     if(this.ssrc && this.ssrc.groups && this.ssrc.groups.length)
-        return this.ssrc.groups[0].primarySSRC;
+        return this.ssrc.groups[0].ssrcs[0];
     else if(this.ssrc && this.ssrc.ssrcs && this.ssrc.ssrcs.length)
         return this.ssrc.ssrcs[0];
     else
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -370,19 +370,17 @@ JitsiLocalTrack.prototype._addStreamToConferenceAsUnmute = function () {
         return Promise.resolve();
     }
 
-    var self = this;
-
-    return new Promise(function(resolve, reject) {
-        self.conference._addLocalStream(
-            self.stream,
+    return new Promise((resolve, reject) => {
+        this.conference._addLocalStream(
+            this.stream,
             resolve,
             (error) => reject(new Error(error)),
             {
-                mtype: self.type,
+                mtype: this.type,
                 type: "unmute",
-                ssrcs: self.ssrc && self.ssrc.ssrcs,
-                groups: self.ssrc && self.ssrc.groups,
-                msid: self.getMSID()
+                ssrcs: this.ssrc && this.ssrc.ssrcs,
+                groups: this.ssrc && this.ssrc.groups,
+                msid: this.getMSID()
             });
     });
 };
@@ -407,8 +405,8 @@ function (successCallback, errorCallback) {
         {
             mtype: this.type,
             type: "mute",
-            ssrcs: this.ssrc ? this.ssrc.ssrcs : undefined,
-            groups: this.ssrc ? this.ssrc.groups : undefined
+            ssrcs: this.ssrc && this.ssrc.ssrcs,
+            groups: this.ssrc && this.ssrc.groups
         });
 };
 
@@ -424,11 +422,9 @@ JitsiLocalTrack.prototype._sendMuteStatus = function(mute) {
         return Promise.resolve();
     }
 
-    var self = this;
-
-    return new Promise(function(resolve) {
-        self.conference.room[
-            self.isAudioTrack()
+    return new Promise((resolve) => {
+        this.conference.room[
+            this.isAudioTrack()
                 ? 'setAudioMute'
                 : 'setVideoMute'](mute, resolve);
     });
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -417,15 +417,15 @@ function extractSSRCMap(desc) {
         return;
     }
 
-    session.media.forEach(function (bLine) {
-        if (!Array.isArray(bLine.ssrcs))
+    session.media.forEach(function (mLine) {
+        if (!Array.isArray(mLine.ssrcs))
         {
             return;
         }
 
-        if (typeof bLine.ssrcGroups !== 'undefined' &&
-            Array.isArray(bLine.ssrcGroups)) {
-            bLine.ssrcGroups.forEach(function (group) {
+        if (typeof mLine.ssrcGroups !== 'undefined' &&
+            Array.isArray(mLine.ssrcGroups)) {
+            mLine.ssrcGroups.forEach(function (group) {
                 if (typeof group.semantics !== 'undefined' &&
                     typeof group.ssrcs !== 'undefined') {
                     // Parse SSRCs and store as numbers
@@ -439,7 +439,7 @@ function extractSSRCMap(desc) {
                 }
             });
         }
-        bLine.ssrcs.forEach(function (ssrc) {
+        mLine.ssrcs.forEach(function (ssrc) {
             if(ssrc.attribute !== 'msid')
                 return;
             ssrcList[ssrc.value]
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -428,7 +428,12 @@ function extractSSRCMap(desc) {
             bLine.ssrcGroups.forEach(function (group) {
                 if (typeof group.semantics !== 'undefined' &&
                     typeof group.ssrcs !== 'undefined') {
-                    var primarySSRC = Number(group.ssrcs.split(' ')[0]);
+                    // Parse SSRCs and store as numbers
+                    const groupSSRCs
+                        = group.ssrcs.split(' ').map(
+                            (ssrcStr) => { return Number(ssrcStr); });
+                    const primarySSRC = groupSSRCs[0];
+                    group.ssrcs = groupSSRCs;
                     ssrcGroups[primarySSRC] = ssrcGroups[primarySSRC] || [];
                     ssrcGroups[primarySSRC].push(group);
                 }
@@ -437,13 +442,12 @@ function extractSSRCMap(desc) {
         bLine.ssrcs.forEach(function (ssrc) {
             if(ssrc.attribute !== 'msid')
                 return;
-            ssrcList[ssrc.value] = ssrcList[ssrc.value] ||
-                {groups: [], ssrcs: []};
+            ssrcList[ssrc.value]
+                = ssrcList[ssrc.value] || { groups: [], ssrcs: [] };
             ssrcList[ssrc.value].ssrcs.push(ssrc.id);
             if(ssrcGroups[ssrc.id]){
                 ssrcGroups[ssrc.id].forEach(function (group) {
-                    ssrcList[ssrc.value].groups.push(
-                        {primarySSRC: ssrc.id, group: group});
+                    ssrcList[ssrc.value].groups.push(group);
                 });
             }
         });
@@ -568,26 +572,23 @@ TraceablePeerConnection.prototype.addStream = function (stream, ssrcInfo) {
     if (stream)
         this.peerconnection.addStream(stream);
     if (ssrcInfo && ssrcInfo.type === "addMuted") {
-        this.sdpConsistency.setPrimarySsrc(ssrcInfo.ssrc.ssrcs[0]);
+        this.sdpConsistency.setPrimarySsrc(ssrcInfo.ssrcs[0]);
         const simGroup =
-            ssrcInfo.ssrc.groups.find(groupInfo => {
-                return groupInfo.group.semantics === "SIM";
+            ssrcInfo.groups.find(groupInfo => {
+                return groupInfo.semantics === "SIM";
             });
         if (simGroup) {
-            const simSsrcs = SDPUtil.parseGroupSsrcs(simGroup.group);
-            this.simulcast.setSsrcCache(simSsrcs);
+            this.simulcast.setSsrcCache(simGroup.ssrcs);
         }
         const fidGroups =
-            ssrcInfo.ssrc.groups.filter(groupInfo => {
-                return groupInfo.group.semantics === "FID";
+            ssrcInfo.groups.filter(groupInfo => {
+                return groupInfo.semantics === "FID";
             });
         if (fidGroups) {
             const rtxSsrcMapping = new Map();
             fidGroups.forEach(fidGroup => {
-                const fidGroupSsrcs =
-                    SDPUtil.parseGroupSsrcs(fidGroup.group);
-                const primarySsrc = fidGroupSsrcs[0];
-                const rtxSsrc = fidGroupSsrcs[1];
+                const primarySsrc = fidGroup.ssrcs[0];
+                const rtxSsrc = fidGroup.ssrcs[1];
                 rtxSsrcMapping.set(primarySsrc, rtxSsrc);
             });
             this.rtxModifier.setSsrcCache(rtxSsrcMapping);
@@ -887,10 +888,8 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function () {
         for (let i = 0; i < SIMULCAST_LAYERS; i++) {
             ssrcInfo.ssrcs.push(SDPUtil.generateSsrc());
         }
-        ssrcInfo.groups.push({
-            primarySSRC: ssrcInfo.ssrcs[0],
-            group: {ssrcs: ssrcInfo.ssrcs.join(" "), semantics: "SIM"}});
-        ssrcInfo;
+        ssrcInfo.groups.push(
+            { ssrcs: ssrcInfo.ssrcs.slice(), semantics: "SIM" });
     } else {
         ssrcInfo = {ssrcs: [SDPUtil.generateSsrc()], groups: []};
     }
@@ -905,11 +904,8 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function () {
             const rtxSsrc = SDPUtil.generateSsrc();
             ssrcInfo.ssrcs.push(rtxSsrc);
             ssrcInfo.groups.push({
-                primarySSRC: primarySsrc,
-                group: {
-                    ssrcs: primarySsrc + " " + rtxSsrc,
-                    semantics: "FID"
-                }
+                ssrcs: [primarySsrc, rtxSsrc],
+                semantics: "FID"
             });
         }
     }
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -421,9 +421,9 @@ function extractSSRCMap(desc) {
     const ssrcMap = new Map();
     /**
      * Groups mapped by primary SSRC number
-     * @type {Map<number,SSRCGroupInfo>}
+     * @type {Map<number,Array<SSRCGroupInfo>>}
      */
-    const ssrcGroups = new Map();
+    const groupsMap = new Map();
 
     if (typeof desc !== 'object' || desc === null ||
         typeof desc.sdp !== 'string') {
@@ -453,10 +453,10 @@ function extractSSRCMap(desc) {
                     const primarySSRC = groupSSRCs[0];
                     // Note that group.semantics is already present
                     group.ssrcs = groupSSRCs;
-                    if (!ssrcGroups.has(primarySSRC)) {
-                        ssrcGroups.set(primarySSRC, []);
+                    if (!groupsMap.has(primarySSRC)) {
+                        groupsMap.set(primarySSRC, []);
                     }
-                    ssrcGroups.get(primarySSRC).push(group);
+                    groupsMap.get(primarySSRC).push(group);
                 }
             }
         }
@@ -480,8 +480,10 @@ function extractSSRCMap(desc) {
 
             ssrcInfo.ssrcs.push(ssrcNumber);
 
-            if (ssrcGroups.has(ssrcNumber)) {
-                for (const group of ssrcGroups[ssrcNumber]) {
+            if (groupsMap.has(ssrcNumber)) {
+                const ssrcGroups = groupsMap.get(ssrcNumber);
+
+                for (const group of ssrcGroups) {
                     ssrcInfo.groups.push(group);
                 }
             }
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -882,10 +882,15 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function () {
         for (let i = 0; i < SIMULCAST_LAYERS; i++) {
             ssrcInfo.ssrcs.push(SDPUtil.generateSsrc());
         }
-        ssrcInfo.groups.push(
-            { ssrcs: ssrcInfo.ssrcs.slice(), semantics: "SIM" });
+        ssrcInfo.groups.push({
+            ssrcs: ssrcInfo.ssrcs.slice(),
+            semantics: "SIM"
+        });
     } else {
-        ssrcInfo = {ssrcs: [SDPUtil.generateSsrc()], groups: []};
+        ssrcInfo = {
+            ssrcs: [SDPUtil.generateSsrc()],
+            groups: []
+        };
     }
     if (!this.options.disableRtx) {
         // Specifically use a for loop here because we'll
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -406,52 +406,49 @@ function extractSSRCMap(desc) {
     if (typeof desc !== 'object' || desc === null ||
         typeof desc.sdp !== 'string') {
         logger.warn('An empty description was passed as an argument.');
-        return desc;
-    }
-
-    var ssrcList = {};
-    var ssrcGroups = {};
-    var session = transform.parse(desc.sdp);
-    if (!Array.isArray(session.media))
-    {
         return;
     }
 
-    session.media.forEach(function (mLine) {
-        if (!Array.isArray(mLine.ssrcs))
-        {
-            return;
+    const ssrcList = { };
+    const ssrcGroups = { };
+    const session = transform.parse(desc.sdp);
+    if (!Array.isArray(session.media)) {
+        return;
+    }
+
+    for (const mLine of session.media) {
+        if (!Array.isArray(mLine.ssrcs)) {
+            continue;
         }
 
-        if (typeof mLine.ssrcGroups !== 'undefined' &&
-            Array.isArray(mLine.ssrcGroups)) {
-            mLine.ssrcGroups.forEach(function (group) {
+        if (Array.isArray(mLine.ssrcGroups)) {
+            for (const group of mLine.ssrcGroups) {
                 if (typeof group.semantics !== 'undefined' &&
                     typeof group.ssrcs !== 'undefined') {
                     // Parse SSRCs and store as numbers
                     const groupSSRCs
-                        = group.ssrcs.split(' ').map(
-                            (ssrcStr) => { return Number(ssrcStr); });
+                        = group.ssrcs.split(' ')
+                                     .map(ssrcStr => parseInt(ssrcStr));
                     const primarySSRC = groupSSRCs[0];
                     group.ssrcs = groupSSRCs;
                     ssrcGroups[primarySSRC] = ssrcGroups[primarySSRC] || [];
                     ssrcGroups[primarySSRC].push(group);
                 }
-            });
+            }
         }
-        mLine.ssrcs.forEach(function (ssrc) {
-            if(ssrc.attribute !== 'msid')
-                return;
+        for (const ssrc of mLine.ssrcs) {
+            if (ssrc.attribute !== 'msid')
+                continue;
             ssrcList[ssrc.value]
                 = ssrcList[ssrc.value] || { groups: [], ssrcs: [] };
             ssrcList[ssrc.value].ssrcs.push(ssrc.id);
-            if(ssrcGroups[ssrc.id]){
-                ssrcGroups[ssrc.id].forEach(function (group) {
+            if (ssrcGroups[ssrc.id]) {
+                for (const group of ssrcGroups[ssrc.id]) {
                     ssrcList[ssrc.value].groups.push(group);
-                });
+                }
             }
-        });
-    });
+        }
+    }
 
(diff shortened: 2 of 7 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -380,7 +380,8 @@ JitsiLocalTrack.prototype._addStreamToConferenceAsUnmute = function () {
             {
                 mtype: self.type,
                 type: "unmute",
-                ssrc: self.ssrc,
+                ssrcs: self.ssrc ? self.ssrc.ssrcs : undefined,
```
saghul: `self.ssrc && self.ssrc.ssrcs` ?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -417,33 +417,37 @@ function extractSSRCMap(desc) {
...
+    session.media.forEach(function (mLine) {
+        if (!Array.isArray(mLine.ssrcs))
         {
             return;
         }
 
-        if (typeof bLine.ssrcGroups !== 'undefined' &&
-            Array.isArray(bLine.ssrcGroups)) {
-            bLine.ssrcGroups.forEach(function (group) {
+        if (typeof mLine.ssrcGroups !== 'undefined' &&
```
saghul: simplify test; no need to check for undefined

modules/RTC/TraceablePeerConnection.js
```diff
@@ -417,33 +417,37 @@ function extractSSRCMap(desc) {
         return;
     }
 
-    session.media.forEach(function (bLine) {
-        if (!Array.isArray(bLine.ssrcs))
+    session.media.forEach(function (mLine) {
+        if (!Array.isArray(mLine.ssrcs))
```
saghul: not in your patch, but the { seems misplaced

modules/RTC/TraceablePeerConnection.js
```diff
@@ -568,26 +572,23 @@ TraceablePeerConnection.prototype.addStream = function (stream, ssrcInfo) {
     if (stream)
         this.peerconnection.addStream(stream);
     if (ssrcInfo && ssrcInfo.type === "addMuted") {
-        this.sdpConsistency.setPrimarySsrc(ssrcInfo.ssrc.ssrcs[0]);
+        this.sdpConsistency.setPrimarySsrc(ssrcInfo.ssrcs[0]);
         const simGroup =
-            ssrcInfo.ssrc.groups.find(groupInfo => {
-                return groupInfo.group.semantics === "SIM";
+            ssrcInfo.groups.find(groupInfo => {
+                return groupInfo.semantics === "SIM";
```
saghul: do you think a `SIMULCAST_GROUP` constant would help?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1412,16 +1412,15 @@ export default class JingleSessionPC extends JingleSession {
...
                         ssrc + "\"]");
                     sourceNode.remove();
                 });
-                ssrcObj.ssrc.groups.forEach(function (group) {
+                ssrcObj.groups.forEach(function (group) {
                     const groupNode = desc.find(">ssrc-group[semantics=\"" +
-                        group.group.semantics + "\"]:has(source[ssrc=\"" +
-                        group.primarySSRC +
-                        "\"])");
+                        group.semantics + "\"]:has(source[ssrc=\"" +
```
saghul: style: use single quotes for strings. Also avoids the need for escaping double quotes

modules/xmpp/RtxModifier.js
```diff
@@ -1,116 +1,63 @@
...
+ * be used to modify video media description
+ * @param {object} primarySsrcInfo the info (ssrc, msid & cname) for the
  *  primary ssrc
  * @param {number} rtxSsrc the rtx ssrc to associate with the primary ssrc
  */
-function updateAssociatedRtxStream (videoMLine, primarySsrcInfo, rtxSsrc) {
-    logger.debug("Updating mline to associate " + rtxSsrc + 
+function updateAssociatedRtxStream (sdpTransformer, primarySsrcInfo, rtxSsrc) {
+    logger.debug(
+        "Updating mline to associate " + rtxSsrc +
```
saghul: use template strings

modules/xmpp/RtxModifier.js
```diff
@@ -1,116 +1,63 @@
...
     let primarySsrcCname = primarySsrcInfo.cname;
 
-    let previousAssociatedRtxStream = 
-        getRtxSsrc (videoMLine, primarySsrc);
-    if (previousAssociatedRtxStream === rtxSsrc) {
-        logger.debug(rtxSsrc + " was already associated with " +
-            primarySsrc);
+    let previousRtxSSRC = sdpTransformer.getRtxSSRC(primarySsrc);
+    if (previousRtxSSRC === rtxSsrc) {
+        logger.debug(rtxSsrc + " was already associated with " + primarySsrc);
```
saghul: ditto

modules/xmpp/RtxModifier.js
```diff
@@ -1,116 +1,63 @@
...
         return;
     }
-    if (previousAssociatedRtxStream) {
-        logger.debug(primarySsrc + " was previously assocaited with rtx " +
-            previousAssociatedRtxStream + ", removing all references to it");
+    if (previousRtxSSRC) {
+        logger.debug(
```
saghul: ditto

modules/xmpp/RtxModifier.js
```diff
@@ -1,116 +1,63 @@
...
-        logger.debug(JSON.stringify(videoMLine.ssrcGroups));
-        videoMLine.ssrcGroups = videoMLine.ssrcGroups
-            .filter(groupInfo => {
-                return groupInfo
-                    .ssrcs
-                    .indexOf(previousAssociatedRtxStream + "") === -1;
-            });
+        sdpTransformer.removeSSRC(previousRtxSSRC);
+
+        logger.debug("groups before filtering for " + previousRtxSSRC);
```
saghul: ditto

modules/xmpp/RtxModifier.js
```diff
@@ -163,62 +110,64 @@ export default class RtxModifier {
...
+            return sdpStr;
+        }
+        const direction = sdpTransformer.mediaDirection;
+        if (direction === "inactive" || direction === "recvonly") {
             logger.debug("RtxModifier doing nothing, video " +
                 "m line is inactive or recvonly");
             return sdpStr;
         }
-        if (!videoMLine.ssrcs) {
+        if (sdpTransformer.getSSRCCount() < 1) {
```
saghul: test for 0 explicitly? it may make things more readable

modules/xmpp/RtxModifier.js
```diff
@@ -163,62 +110,64 @@ export default class RtxModifier {
...
         }
         logger.debug("Current ssrc mapping: ", this.correspondingRtxSsrcs);
-        let primaryVideoSsrcs = getPrimaryVideoSsrcs(videoMLine);
+        let primaryVideoSsrcs = sdpTransformer.getPrimaryVideoSSRCs();
         logger.debug("Parsed primary video ssrcs ", primaryVideoSsrcs, " " +
             "making sure all have rtx streams");
-        primaryVideoSsrcs.forEach(ssrc => {
-            let msid = SDPUtil.getSsrcAttribute(videoMLine, ssrc, "msid");
-            let cname = SDPUtil.getSsrcAttribute(videoMLine, ssrc, "cname");
+        for (const ssrc of primaryVideoSsrcs) {
```
saghul: ❤️

modules/xmpp/RtxModifier.js
```diff
@@ -163,62 +110,64 @@ export default class RtxModifier {
...
+            let msid = sdpTransformer.getSSRCAttrValue(ssrc, "msid");
+            let cname = sdpTransformer.getSSRCAttrValue(ssrc, "cname");
             let correspondingRtxSsrc = this.correspondingRtxSsrcs.get(ssrc);
             if (correspondingRtxSsrc) {
                 logger.debug("Already have an associated rtx ssrc for " +
-                    " video ssrc " + ssrc + ": " + 
+                    " video ssrc " + ssrc + ": " +
```
saghul: template strings!

modules/xmpp/SdpConsistency.js
```diff
@@ -78,6 +38,9 @@ export default class SdpConsistency {
      *  in future calls to makeVideoPrimarySsrcsConsistent
      */
     setPrimarySsrc (primarySsrc) {
+        if (typeof primarySsrc !== 'number') {
```
saghul: maybe simplify by assigning the return value of `parseInt` ? That would take care of raising the appropriate exception.

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+/**
+ * Parses the secondary SSRC of given SSRC group.
+ * @param {object} group the SSRC group object as defined by the 'sdp-transform'
+ * @return {Number} the secondary SSRC number
+ */
+export function parseSecondarySSRC(group) {
+    return parseInt(group.ssrcs.split(" ")[1]);
+}
+
+function findGroup(mLine, groupSemantics, ssrcs) {
```
saghul: docs?

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+    constructor(rawSDP) {
+        this.parsedSDP = transform.parse(rawSDP);
+        this.selectedMLine = null;
+    }
+
+    /**
+     * Returns the direction of currently selected media.
+     * @return {string} the media direction name as defined in the SDP.
+     */
+    get mediaDirection() {
```
saghul: ❤️

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+     * @return {boolean} <tt>true</tt> if the media of given type has been found
+     * and selected or <tt>false</tt> otherwise.
+     */
+    selectMedia(mediaType) {
+        this.selectedMLine
+            = this.parsedSDP.media.find(mLine => mLine.type === mediaType);
+        return !!(this.selectedMLine);
+    }
+
+    get _ssrcs () {
```
saghul: a private getter?

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+        this._assertMLineSelected();
+
+        const mediaType = this.selectedMLine.type;
+
+        if (mediaType !== 'video') {
+            throw new Error(
+                "getPrimarySsrc doesn't work with '" + mediaType +"'");
+        }
+
+        let numSsrcs = getSSRCCount(this.selectedMLine);
```
saghul: const

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+        if (numSsrcs === 1) {
+            // Not using _ssrcs on purpose here
+            return this.selectedMLine.ssrcs[0].id;
+        } else {
+            // Look for a SIM or FID group
+            if (this.selectedMLine.ssrcGroups) {
+                let simGroup = this.findGroup("SIM");
```
saghul: const

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+                if (simGroup) {
+                    return parsePrimarySSRC(simGroup);
+                }
+                let fidGroup = this.findGroup("FID");
```
saghul: const

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+     * secondary SSRC of the first "FID" group found)
+     * @param {number} primarySsrc the video ssrc for which to find the
+     * corresponding rtx ssrc
+     * @returns {number|undefined} the rtx ssrc (or undefined if there isn't
+     * one)
+     */
+    getRtxSSRC (primarySsrc) {
+        this._assertMLineSelected();
+
+        let fidGroup = this.findGroupByPrimarySSRC("FID", primarySsrc);
```
saghul: const

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+        return fidGroup ? parseSecondarySSRC(fidGroup) : undefined;
```
saghul: `return fidGroup && parseSecondarySSRC(fidGroup);`

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+     * @return {Array.<number>} an array of all primary video SSRCs as numbers.
+     */
+    getPrimaryVideoSSRCs () {
+        this._assertMLineSelected();
+
+        const mediaType = this.selectedMLine.type;
+
+        if (mediaType !== 'video') {
+            throw new Error(
+                "getPrimaryVideoSSRCs doesn't work with '" + mediaType +"'");
```
saghul: template strings!

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+        }
+
+        let videoSSRCs = this.getSSRCs();
+
+        this.forEachSSRCGroup((ssrcGroupInfo) => {
```
saghul: no need for parens

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+        return videoSSRCs;
+    }
+
+    /**
+     * Checks if there is currently any media selected or throws
+     * an <tt>Error</tt> otherwise.
+     * @private
+     */
+    _assertMLineSelected() {
+        // FIXME if possible wrap all methods instead of adding this call to
```
saghul: Not sure about the current state, but this should be doable with decorators: https://babeljs.io/docs/plugins/transform-decorators/

modules/RTC/TraceablePeerConnection.js
bbaldino: also in many places i think we (or at least i) have just done `if (mLine.ssrcGroups)`, since undefined resolves to falsey.  checking if it's actually an array is going the extra mile, but, technically then we should be checking all types are the ones we expect :)

modules/RTC/TraceablePeerConnection.js
```diff
@@ -417,33 +417,37 @@ function extractSSRCMap(desc) {
...
+        if (typeof mLine.ssrcGroups !== 'undefined' &&
+            Array.isArray(mLine.ssrcGroups)) {
+            mLine.ssrcGroups.forEach(function (group) {
                 if (typeof group.semantics !== 'undefined' &&
                     typeof group.ssrcs !== 'undefined') {
-                    var primarySSRC = Number(group.ssrcs.split(' ')[0]);
+                    // Parse SSRCs and store as numbers
+                    const groupSSRCs
+                        = group.ssrcs.split(' ').map(
+                            (ssrcStr) => { return Number(ssrcStr); });
```
bbaldino: oh cool, is `Number(xxx)` better than `parseInt(xxx)`?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -417,33 +417,37 @@ function extractSSRCMap(desc) {
...
                 }
             });
         }
-        bLine.ssrcs.forEach(function (ssrc) {
+        mLine.ssrcs.forEach(function (ssrc) {
             if(ssrc.attribute !== 'msid')
                 return;
-            ssrcList[ssrc.value] = ssrcList[ssrc.value] ||
-                {groups: [], ssrcs: []};
+            ssrcList[ssrc.value]
```
bbaldino: i think it reads better with the `=` on the same line as `ssrcList[ssrc.value]`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -568,26 +572,23 @@ TraceablePeerConnection.prototype.addStream = function (stream, ssrcInfo) {
...
             });
         if (simGroup) {
-            const simSsrcs = SDPUtil.parseGroupSsrcs(simGroup.group);
-            this.simulcast.setSsrcCache(simSsrcs);
+            this.simulcast.setSsrcCache(simGroup.ssrcs);
         }
         const fidGroups =
-            ssrcInfo.ssrc.groups.filter(groupInfo => {
-                return groupInfo.group.semantics === "FID";
+            ssrcInfo.groups.filter(groupInfo => {
```
bbaldino: i have a feeling that i had some debug info here because this could be shorter, feel free to change to this (if it won't cause conflicts, if so no big deal for now):

`ssrcInfo.groups.filter(groupInfo => groupInfo.semantics === "FID"`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -887,10 +888,8 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function () {
         for (let i = 0; i < SIMULCAST_LAYERS; i++) {
             ssrcInfo.ssrcs.push(SDPUtil.generateSsrc());
         }
-        ssrcInfo.groups.push({
-            primarySSRC: ssrcInfo.ssrcs[0],
-            group: {ssrcs: ssrcInfo.ssrcs.join(" "), semantics: "SIM"}});
-        ssrcInfo;
+        ssrcInfo.groups.push(
+            { ssrcs: ssrcInfo.ssrcs.slice(), semantics: "SIM" });
```
bbaldino: personally i think it's easier to read if the fields of an object are split across new lines (similar to the way it was before), but maybe just a personal preference

modules/xmpp/RtxModifier.js
bbaldino: this is all (my) existing code, so not pawel's fault :) guessing this will cause too many conflicts for him to fix now

modules/RTC/TraceablePeerConnection.js
paweldomas: I'm renaming a var here :P can't fix all the code. This one is going away anyway.

paweldomas: but it makes sense to change it actually

modules/RTC/TraceablePeerConnection.js
paweldomas: I think we used to place = on the next line most of the time.

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+ * if (transformer.selectMedia('video)) {
+ *     transformer.addSSRCAttribute({
+ *         id: 2342343,
+ *         attribute: "cname",
+ *         value: "someCname"
+ *     });
+ *     rawSdp = transformer.toRawSdp();
+ * }
+ */
+export class SdpTransformWrap {
```
bbaldino: i'm not a fan of this as an instance class that takes in the raw sdp, i think it would be better to have this be a set of static utility functions that take in an sdp and perform an operation (more of a functional-type approach).  i think it also becomes clearer in the methods that need to do something for them to take the sdp directly and call, for example, `SdpUtil.addRtxSsrc(mLine, rtx ssrc info...)` then `sdpTransformWrap.addRtxSsrc(rtx ssrc info)` because it's a bit confusing which sdp is being modified (since it's hidden).  i think mainly i find the obfuscation of the underlying sdp by the wrapper a bit weird, but maybe i'm making a big deal out of nothing, so open to discussing it.

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,439 @@
...
+    }
+
+    /**
+     * Selects the first media SDP of given name.
+     * @param {string} mediaType the name of the media e.g. 'audio', 'video',
+     * 'data'.
+     * @return {boolean} <tt>true</tt> if the media of given type has been found
+     * and selected or <tt>false</tt> otherwise.
+     */
+    selectMedia(mediaType) {
```
bbaldino: state like this also feels a bit weird, i'd rather have a util function to get an mLine and then the mLine passed to a static function to be modified

modules/RTC/TraceablePeerConnection.js
paweldomas: yes it will work as well

modules/RTC/TraceablePeerConnection.js
paweldomas: yes, I agree

modules/xmpp/JingleSessionPC.js
paweldomas: Do we have a rule for that already ? I feel that we're using double quotes most of the time. But I find the single quotes approach better in that case, because there's no need for escaping.

modules/xmpp/SdpTransformUtil.js
paweldomas: Yes, I was not sure if others will like this approach, but I think we need to decide, before I proceed any further. I'm going to try converting it back to static, but then this whole PR goes to trash.

paweldomas: Ok, so I've pushed another commit which extracts separate instance called MLineWrap to get rid of this weird "select m line" state. I think it's better compared to static methods for which I've opened a separate PR#382, because:
1. There's no real benefit of functional approach/stateless  unless you're doing parallel processing
2. You have to pass the m line instance to almost every single utility call, so in my opinion it's more convenient to have a wrapper object around that m line.

modules/xmpp/SdpTransformUtil.js
paweldomas: I agree it feels weird and I've pushed a separate commit to address that problem (see my other comment above)

paweldomas:
jenkins test this please

bbaldino:
ok, lgtm.

modules/xmpp/JingleSessionPC.js
paweldomas: I plan to remove this code, so I'd rather not touch it as it could be broken by refactoring and then we'll waste time on fixing it

modules/RTC/TraceablePeerConnection.js
paweldomas: yes, but I don't know where to put and also I plan to remove this code.

modules/xmpp/RtxModifier.js
paweldomas: I'm just trying to handle all cases even those that are "impossible". What if for any reason getSSRCCount would return -1 ? The code would try to execute with some possibly unexpected behaviour or crash.

modules/xmpp/SdpConsistency.js
paweldomas: My intention here is to crash the code instead of handling two different types (string or number)

modules/xmpp/SdpTransformUtil.js
paweldomas: convenience getter, is that wrong ?

modules/xmpp/SdpTransformUtil.js
paweldomas: sorry about that, but without some highlighting I find it hard to get used to

modules/xmpp/SdpTransformUtil.js
paweldomas: I've removed this completely, but thanks for the suggestion !

modules/RTC/TraceablePeerConnection.js
paweldomas: I don't know I just used what it was here before

modules/RTC/JitsiLocalTrack.js
```diff
@@ -406,7 +407,8 @@ function (successCallback, errorCallback) {
         {
             mtype: this.type,
             type: "mute",
-            ssrc: this.ssrc
+            ssrcs: this.ssrc ? this.ssrc.ssrcs : undefined,
```
saghul: `this.ssrc && this.ssrc.ssrcs`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -406,48 +406,49 @@ function extractSSRCMap(desc) {
     if (typeof desc !== 'object' || desc === null ||
         typeof desc.sdp !== 'string') {
         logger.warn('An empty description was passed as an argument.');
-        return desc;
+        return;
```
saghul: should we return null to signal that there is no mapping?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -406,48 +406,49 @@ function extractSSRCMap(desc) {
...
     }
 
-    var ssrcList = {};
-    var ssrcGroups = {};
-    var session = transform.parse(desc.sdp);
-    if (!Array.isArray(session.media))
-    {
+    const ssrcList = { };
```
saghul: use a Map? Also ssrcMap sounds like a more appropriate name, since it's not a list

saghul: Also, I don't think we use a space for object literals

modules/xmpp/JingleSessionPC.js
```diff
@@ -1412,16 +1412,15 @@ export default class JingleSessionPC extends JingleSession {
                     ssrcObj.mtype + "\"]>description");
                 if (!desc || !desc.length)
                     return;
-                ssrcObj.ssrc.ssrcs.forEach(function (ssrc) {
+                ssrcObj.ssrcs.forEach(function (ssrc) {
```
saghul: for..of and template strings, since you're here :-)

modules/xmpp/JingleSessionPC.js
saghul: ok

modules/xmpp/RtxModifier.js
```diff
@@ -1,116 +1,62 @@
...
     let primarySsrc = primarySsrcInfo.id;
     let primarySsrcMsid = primarySsrcInfo.msid;
     let primarySsrcCname = primarySsrcInfo.cname;
 
-    let previousAssociatedRtxStream = 
-        getRtxSsrc (videoMLine, primarySsrc);
-    if (previousAssociatedRtxStream === rtxSsrc) {
-        logger.debug(rtxSsrc + " was already associated with " +
-            primarySsrc);
+    let previousRtxSSRC = mLine.getRtxSSRC(primarySsrc);
```
saghul: const?

modules/xmpp/SdpConsistency.js
saghul: Postel's Principle? "Be gentle in what you accept..."

saghul:
> On Mar 7, 2017, at 20:27, Paweł Domas <notifications@github.com> wrote:
> 
> @paweldomas commented on this pull request.
> 
> In modules/xmpp/RtxModifier.js <https://github.com/jitsi/lib-jitsi-meet/pull/379#discussion_r104756278>:
> 
> >              logger.debug("RtxModifier doing nothing, video " +
>                  "m line is inactive or recvonly");
>              return sdpStr;
>          }
> -        if (!videoMLine.ssrcs) {
> +        if (sdpTransformer.getSSRCCount() < 1) {
> I'm just trying to handle all cases even those that are "impossible". What if for any reason getSSRCCount would return -1 ? The code would try to execute with some possibly unexpected behaviour or crash.
> 
> 

If it returns a value < 0 the code is already broken, and this test may hide the problem. I believe it’s always better to have loud failures than to sweep things under a rug.

--
Saúl

modules/xmpp/SdpConsistency.js
```diff
@@ -93,9 +56,12 @@ export default class SdpConsistency {
      *  with ssrcs consistent with this class' cache
      */
     makeVideoPrimarySsrcsConsistent (sdpStr) {
-        let parsedSdp = transform.parse(sdpStr);
-        let videoMLine =
-            parsedSdp.media.find(mLine => mLine.type === "video");
+        const sdpTransformer = new SdpTransformWrap(sdpStr);
+        const videoMLine = sdpTransformer.selectMedia("video");
+        if (!videoMLine) {
+            logger.error("No 'video' media found in the sdp: " + sdpStr);
```
saghul: template strings

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,419 @@
...
+
+/**
+ * Tells how many distinct SSRCs are contained in given media line.
+ * @param {Object} mLine the media line object as defined by 'sdp-transform' lib
+ * @return {number}
+ */
+function getSSRCCount(mLine) {
+    if (!mLine.ssrcs) {
+        return 0;
+    } else {
```
saghul: you can omit the else and save a level of indentation

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,419 @@
...
+     * found
+     * @param {string} attrName the name of the SSRC attribute to be found.
+     * @return {string|undefined} the value of SSRC attribute or
+     * <tt>undefined</tt> if no such attribute exists.
+     */
+    getSSRCAttrValue(ssrcNumber, attrName) {
+        const attribute = this._ssrcs.find(
+            ssrcObj => ssrcObj.id == ssrcNumber
+            && ssrcObj.attribute === attrName);
+        return attribute ? attribute.value : undefined;
```
saghul: &&

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,419 @@
...
+     * Removes all attributes for given SSRC number.
+     * @param {number} ssrcNum the SSRC number for which all attributes will be
+     * removed.
+     */
+    removeSSRC(ssrcNum) {
+        if (!this.mLine.ssrcs || !this.mLine.ssrcs.length) {
+            return;
+        }
+
+        // FIXME it should be possible to remove those values more efficiently
```
saghul: this one should be easy to fix: measure, pick one and be done with it :-)

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,419 @@
...
+        return findGroupByPrimarySSRC(
+            this.mLine, semantics, primarySSRC);
+    }
+
+    /**
+     * Gets the SSRC count for the underlying media description.
+     * @return {number}
+     */
+    getSSRCCount() {
+        return getSSRCCount(this.mLine);
```
saghul: this looks a bit confusing, maybe make this one private? or another name?

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,419 @@
...
+     * Finds the primary video SSRC.
+     * @returns {number|undefined} the primary video ssrc
+     * @throws Error if the underlying media description is not a video
+     */
+    getPrimaryVideoSsrc () {
+        const mediaType = this.mLine.type;
+
+        if (mediaType !== 'video') {
+            throw new Error(
+                "getPrimarySsrc doesn't work with '" + mediaType +"'");
```
saghul: template strings

modules/xmpp/SdpTransformUtil.js
```diff
@@ -0,0 +1,419 @@
...
+                }
+            });
+        }
+    }
+
+    /**
+     * Executes given call back for each SSRC group
+     * @param {function(object)} callback
+     */
+    forEachSSRCGroup(callback) {
```
saghul: why not just use for..of on the object itself?

modules/RTC/JitsiLocalTrack.js
paweldomas: sorry, missed that one...

modules/RTC/TraceablePeerConnection.js
paweldomas: This method was inconsistent already, I merely made it return the same thing in case of failure. There's no JSDoc so who knows what should it return.

saghul: Good time to find out! :-)

modules/xmpp/SdpConsistency.js
paweldomas: In other comment you said that the code should fail loudly and that was my intention here. I added this Error after seeing that number and string types are mixed up. It helped me to fix the problem, before committing any broken code.

paweldomas: Also the JSDoc says it wants a number so it enforces number. I will add throws description as well

modules/xmpp/SdpTransformUtil.js
paweldomas: but this way two returns are aligned with each other :)

modules/xmpp/SdpTransformUtil.js
paweldomas: this comment is outdated.... removing it

modules/xmpp/SdpTransformUtil.js
paweldomas: for..of crashes on undefined

paweldomas:
> If it returns a value < 0 the code is already broken, and this test may hide the problem. I believe it’s always better to have loud failures than to sweep things under a rug.

Well if it would be == 0 then it could sweep things under a rug, because the code could just run and eventually only do weird things without crashing. With < 0 at least we would see that it claims there are no SSRCs and we would know where to investigate. Of course the best would be to detect < 0 and == 0 separately, but I find it too much to handle.

I can also add a method "containsAnySSRCs" if that would make things easier, but honestly I don't know why it's such a big deal.

modules/xmpp/SdpTransformUtil.js
saghul: `for (const x of container || [])` ?

saghul:
> can also add a method "containsAnySSRCs" if that would make things easier, but honestly I don't know why it's such a big deal.

No big deal, just a suggestion :-)

modules/xmpp/SdpTransformUtil.js
paweldomas: I want to add getter to expose `ssrcGroups` to the wrapper, but is it ok to add getter without adding a setter at the same time ?

paweldomas:
@bbaldino @saghul it should be all now. Please let me know what do you think.

Btw. it would make sense to squash all commits on merge.

modules/xmpp/JingleSessionPC.js
paweldomas: It doesn't make sense to fix it only in one place of JingleSessinPC and fixing them in whole file is too big effort at this point. I've spent a lot of time on old code fixes anyway.

PR #380, opened 2017-03-03
Fixes for the recvonly SSRC logic

paweldomas:


modules/RTC/TraceablePeerConnection.js
```diff
@@ -696,6 +696,16 @@ TraceablePeerConnection.prototype.generateRecvonlySsrc = function() {
...
 };
 
+/**
+ * Makes the underlying TraceablePeerConnection forget the current primary video
+ * SSRC.
+ * @deprecated
+ */
+TraceablePeerConnection.prototype.clearRecvonlySsrc = function () {
+    logger.info("Clearing primary video SSRC!");
+    this.sdpConsistency.setPrimarySsrc(undefined);
```
bbaldino: there is a `clearSsrcCache` method for this

modules/xmpp/JingleSessionPC.js
```diff
@@ -922,6 +922,26 @@ export default class JingleSessionPC extends JingleSession {
...
+                if (!oldTrack && newTrack && newTrack.isVideoTrack()) {
+                    // Clearing current primary SSRC will make
+                    // the SdpConsistency generate a new one which will result
+                    // with:
+                    // 1. source-remove for the recvonly
+                    // 2. source-add for the new video stream
+                    this.peerconnection.clearRecvonlySsrc();
+                }
+                // Transition from video to no video
+                if (oldTrack && oldTrack.isVideoTrack() && !newTrack) {
```
bbaldino: wondering if it's clearer if this is an `else if` ?

paweldomas: not sure what kind of clarity this would add... how I personally see it that it would only make me process also the previous statement in combination with this one. Do we really have to be so picky ?

bbaldino: well, using 2 if's back-to-back suggests both could be entered, but that can never happen here, right?

dwilson6:
@paweldomas this PR does fix the problems I was running into.

paweldomas:
@bbaldino both problems should be fixed now

PR #383, opened 2017-03-08
build: add yarn lock file

saghul:
This switched the install script to postinstall, which is functionally
equivalent, but works in projects depending on lib-jitsi-meet instead of
failing.


saghul:
I'm breaking this un into 2-3 PRs. Readme updates, postinstall and we can do Yarn later.

PR #385, opened 2017-03-08
Adds isInLastN to JitsiConference

yanas:


modules/RTC/DataChannels.js
```diff
@@ -13,6 +13,7 @@ function DataChannels(peerConnection, emitter) {
     peerConnection.ondatachannel = this.onDataChannel.bind(this);
     this.eventEmitter = emitter;
 
+    this._lastNEndpoints = null;
```
saghul: I don't think this is the right place for this. The data channel is just a message bus, the last N endpoints are related to the RTC connection and IMHO it feels much better in RTC.js.

webintoit:
Is there any way to `ask` videobridge about lastN endpoints array? There is issue when new user connects to muc, he tries to render all videostreams in conference, but tracks can be `frozen`.

JitsiConference.js
```diff
@@ -775,6 +775,18 @@ JitsiConference.prototype.setLastN = function(lastN) {
 };
 
 /**
+ * Checks if the participant given by participantId is currently included in
+ * the last N.
+ * @param {string} participantId the identifier of the participant we would
+ * like to check
+ * @return {boolean} {true} if the participant with id is in the last N set or
```
paweldomas: I'm not sure if {true}/{false} is the right syntax. I think it's used only to indicate a type (but I may be wrong)

modules/RTC/DataChannels.js
paweldomas: It feels like it's convenient to have this variable here, but like Saul said it would be better to keep DataChannels a  message bus. One idea would be to move `emit(RTCEvents.LASTN_ENDPOINT_CHANGED...` to RTC. I guess DataChannels are using RTC's event emitter anyway.

yanas:
@saghul @paweldomas thank you for your feedback. I have pushed an updated version. Please re-review. Thanks!

yanas:
Just noticed that I maybe need to remove the listener on datachannels closed. I'll add that part.

modules/RTC/RTC.js
```diff
@@ -157,6 +158,12 @@ export default class RTC extends Listenable {
             };
             this.addListener(RTCEvents.DATA_CHANNEL_OPEN,
                 this._dataChannelOpenListener);
+
+            this.addListener(RTCEvents.LASTN_ENDPOINT_CHANGED,
+                lastNEndpoints => {
+                    this._lastNEndpoints = lastNEndpoints;
```
saghul: where is `this._lastNEndpoints` initialized? If we initialize it to [] in the constructor that would simplify the check next...

JitsiConference.js
```diff
@@ -841,6 +841,18 @@ JitsiConference.prototype.setLastN = function(lastN) {
 };
 
 /**
+ * Checks if the participant given by participantId is currently included in
+ * the last N.
+ * @param {string} participantId the identifier of the participant we would
+ * like to check
```
saghul: punctuation

JitsiConference.js
```diff
@@ -841,6 +841,18 @@ JitsiConference.prototype.setLastN = function(lastN) {
...
+ * @return {boolean} true if the participant with id is in the last N set or
+ * if there's no last N set, false otherwise
```

modules/RTC/RTC.js
```diff
@@ -845,4 +852,16 @@ export default class RTC extends Listenable {
...
+
+    /**
+     * Indicates if the endpoint id is currently included in the last N.
+     *
+     * @param {string} id the endpoint id that we check for last N
+     * @returns {boolean} true if the endpoint id is in the last N or if we
+     * don't have data channel support, otherwise we return false
+     */
+    isInLastN(id) {
+        return !this._lastNEndpoints // lastNEndpoints not initialised yet
```
saghul: if you initialize the attribute (we should), this can be greatly simplified as

`return this._lastNEndpoints.indexOf(id) !== -1;`

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -363,7 +363,7 @@ export default class ParticipantConnectionStatus {
             `Figure out conn status, is video muted: ${isVideoMuted
                  } is active(jvb): ${isConnActiveByJvb
                  } video track frozen: ${isVideoTrackFrozen
-                 } is in last N: ${isInLastN}
+                 } is in last N: ${isInLastN
```
paweldomas: does this carrying over the "}" sign get rid of the newline in the printed message ?

modules/RTC/DataChannels.js
```diff
@@ -124,19 +123,17 @@ DataChannels.prototype.onDataChannel = function(event) {
...
                 // last-n at this time i.e. were not in the old list of last-n
                 // endpoint IDs.
                 const endpointsEnteringLastN = obj.endpointsEnteringLastN;
 
-                logger.info(
-                    'Data channel new last-n event: ',
-                        self._lastNEndpoints, endpointsEnteringLastN, obj);
+                logger.info('Data channel new last-n event: ',
+                    lastNEndpoints, endpointsEnteringLastN, obj);
                 self.eventEmitter.emit(RTCEvents.LASTN_ENDPOINT_CHANGED,
```
paweldomas: Another idea would be to just add some method to RTC aka "processLastNEndpointChangedMsg" and emit the RTC event from there. You would not have to deal with the listeners this way, but given that most of the events are emitted directly from here it makes sense to leave it this way.

modules/RTC/RTC.js
paweldomas: yes, it would be good to initialize all members in the constructor, so that their type is documented.

modules/RTC/RTC.js
yanas: This isn't what we want though. If the list isn't initialised we would like to return true. Ok about initialising the variable though, I'm just commenting about the check above.

paweldomas: Sounds good to me

gpolitis:
Hi @webintoit, thank you for your query. To answer your question, currently there is no way to query the bridge but what you're describing is actually a bug (in the bridge) that's fixed with this commit https://github.com/jitsi/jitsi-videobridge/pull/429/commits/7461702a5eaefb81fea9a3896a5b4c73cbcec3a9.

modules/RTC/RTC.js
```diff
@@ -157,6 +165,14 @@ export default class RTC extends Listenable {
             };
             this.addListener(RTCEvents.DATA_CHANNEL_OPEN,
                 this._dataChannelOpenListener);
+
+            // Add Last N change listener.
+            this._lastNChangeListener = lastNEndpoints => {
```
saghul: this can be initialized in the constructor

modules/RTC/RTC.js
```diff
@@ -83,6 +83,13 @@ export default class RTC extends Listenable {
         // this happen
         this.dataChannelsOpen = false;
 
+        // Defines the last N endpoints list. It can be null or an array once
+        // initialised with a datachannel last N event.
+        this._lastNEndpoints = null;
```
paweldomas: It helps if the type is declared like this:
`/*`
  `* @type {Array<string>|null}`
  `*/`
This way IDE and some other tools are able to detect the type.

yanas:
Jenkins test this please.

modules/RTC/RTC.js
paweldomas: can we have this changed, pretty please ?

PR #403, opened 2017-03-10
Improve and simplify some browser feature checks

saghul:


saghul:
Rebased after the eslint commits. @hristoterezov can you please review?

modules/RTC/RTCBrowserType.js
```diff
@@ -1,10 +1,10 @@
-const logger = require('jitsi-meet-logger').getLogger(__filename);
```
hristoterezov: Do you really like more the `import` syntax for this situation or you are changing this only for consistency?

Personally I like the `require` syntax more only for this use case :) But I agree with the consistency argument as well.

modules/RTC/RTCBrowserType.js
```diff
@@ -176,14 +176,6 @@ const RTCBrowserType = {
     },
 
     /**
-     * Whether the browser is running on an android device.
-     * @returns {boolean}
```
hristoterezov: Why we are removing this?

modules/RTC/RTCBrowserType.js
saghul: Consistency++ :-)

modules/RTC/RTCBrowserType.js
saghul: Because it's no longer needed / used. It was there to support a hack which is now gone.

modules/RTC/RTCBrowserType.js
```diff
@@ -163,11 +163,7 @@ const RTCBrowserType = {
     },
 
     usesPlanB() {
-        return (
-            RTCBrowserType.isChrome()
-                || RTCBrowserType.isOpera()
-                || RTCBrowserType.isReactNative()
-                || RTCBrowserType.isTemasysPluginUsed());
+        return !RTCBrowserType.usesUnifiedPlan();
```
hristoterezov: I like this! Actually now I'm think that before this changes maybe we had a bug if you are using Electron and NWJS. They should use Plan B right?

modules/RTC/RTCBrowserType.js
hristoterezov: So now chrome on android works without explicitly setting the resolution to 180p and the frame rate to 15?

saghul: AFAICT, yes.

modules/RTC/RTCBrowserType.js
saghul: Yes, since they use the Chrome engine.

PR #409, opened 2017-03-13
Adds video sip gw handling.

damencho:
Parsing presence and availability of the service. Adding possibility to start/stop a video sip gw session.


JitsiConference.js
```diff
@@ -1626,4 +1626,32 @@ JitsiConference.prototype.isConnectionInterrupted = function() {
...
+ */
+JitsiConference.prototype.isVideoSIPGWAvailable = function() {
+    return this.room.videoSIPGWHandler
+        && this.room.videoSIPGWHandler.isVideoSIPGWAvailable();
+};
+
+/**
+ * Creates a video SIP GW session and returns it if service is enabled.
+ *
+ * @param sipAddress - The sip address to be used.
```
paweldomas: missing type declaration for both 'sipAddress' and 'displayName'

JitsiConference.js
```diff
@@ -1626,4 +1626,32 @@ JitsiConference.prototype.isConnectionInterrupted = function() {
...
+ * @param displayName - The display name to be used for this session.
+ * @returns {JitsiVideoSIPGWSession}
```
paweldomas: From the code I can see that it can be {JitsiVideoSIPGWSession|null}

JitsiConferenceEvents.js
```diff
@@ -143,6 +143,12 @@ export const PHONE_NUMBER_CHANGED = 'conference.phoneNumberChanged';
 export const RECORDER_STATE_CHANGED = 'conference.recorderStateChanged';
 
 /**
+ * Indicates that video SIP GW state changed.
```
paweldomas: Does this event have any arguments ?

JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * Jitsi video SIP GW session. Holding its state and able to start/stop it.
+ */
+export default class JitsiVideoSIPGWSession {
+
+    /**
+     * Creates new session with the desired sip address and display name.
+     *
+     * @param sipAddress - The sip address to use when starting the session.
```
paweldomas: missing type for all arguments

JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,87 @@
...
+    constructor(sipAddress, displayName, chatRoom) {
+        this.sipAddress = sipAddress;
+        this.displayName = displayName;
+        this.chatRoom = chatRoom;
+
+        this.eventEmitter = new EventEmitter();
+    }
+
+    /**
+     * Currently not implemented.
```
paweldomas: Is the JSDoc description up to date ? Is it really not implemented ?

JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,87 @@
...
+    start() {
+        sendJibriIQ(
+            this, 'start', this.connection, this.focusMucJid, this.sipAddress,
+            this.displayName);
+    }
+
+    /**
+     * Changes the state of this session.
+     *
+     * @param newState The new state to set.
```
paweldomas: type not documented

JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,87 @@
...
+     */
+    stop() {
+        sendJibriIQ(this, 'stop', this.chatRoom, null);
+    }
+
+    /**
+     * Starts a new session. Sends an iq to the focus.
+     */
+    start() {
+        sendJibriIQ(
```
paweldomas: Both start and stop method do not seem to care about current state and allow to send IQ in possibly wrong state instead of failing immediately (it doesn't make sense to send stop when the session has been stopped already or has never been started).

JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,87 @@
...
+                newState: this.state
+            }
+        );
+    }
+
+    /**
+     * Subscribes the passed listener to the event for state change of this
+     * session.
+     *
+     * @param listener {Function} the function that will receive the event
```
paweldomas: Type should appear before the argument name:
`@param {Function} listener`

JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,87 @@
...
+     */
+    addStateListener(listener) {
+        this.eventEmitter.addListener(STATE_CHANGED, listener);
+    }
+
+    /**
+     * Unsubscribes the passed handler.
+     *
+     * @param listener {Function} the function to be removed.
```
paweldomas: The same as above

modules/xmpp/ChatRoom.js
```diff
@@ -467,6 +470,12 @@ export default class ChatRoom extends Listenable {
             case 'jibri-recording-status':
                 jibri = node;
                 break;
+            case 'jibri-sip-status':
```
paweldomas: Don't we have a way to register presence node listeners ? It feels like the gateway should be doing that.

modules/xmpp/ChatRoom.js
```diff
@@ -514,6 +532,14 @@ export default class ChatRoom extends Listenable {
                 this.recording.handleJibriPresence(this.lastJibri);
             }
         }
+
+        if (!this.videoSIPGWHandler) {
```
paweldomas: Why is VideoSIPGW inside of the ChatRoom ? It feels like the ChatRoom should be dealing with XMPP MUC stuff and not managing other components.

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
+/* global $, $iq */
+import { getLogger } from 'jitsi-meet-logger';
+const logger = getLogger(__filename);
+
+import XMPPEvents from '../../service/xmpp/XMPPEvents';
+import * as Constants from './VideoSIPGWConstants';
```
paweldomas: import statements are not in alphabetical order

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+
+/**
+ * Main video SIP GW handler. Stores references of all created sessions.
+ */
+export default class VideoSIPGW {
+
+    /**
+     * Creates new handler.
+     *
+     * @param chatRoom - Tha chat room to handle.
```
paweldomas: missing type declaration

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+     */
+    constructor(chatRoom) {
+        this.chatRoom = chatRoom;
+        this.eventEmitter = chatRoom.eventEmitter;
+        logger.info('creating VideoSIPGW');
+        this.sessions = {};
```
paweldomas: Missing type description

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+
+        this.sessionStateChangeListener = this.sessionStateChanged.bind(this);
+    }
+
+    /**
+     * Handles presence nodes with name: jibri-sip-status.
+     *
+     * @param node the presence node to handle.
```
paweldomas: missing type declaration

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+            this.status = newStatus;
+            this.eventEmitter.emit(
+                XMPPEvents.VIDEO_SIP_GW_AVAILABILITY_CHANGED, this.status);
+        }
+    }
+
+    /**
+     * Handles presence nodes with name: jibri-sip-call-state.
+     *
+     * @param node the presence node to handle.
```
paweldomas: missing type declaration

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+            }
+        }
+        }
+    }
+
+
+        /**
+     * Creates new session and stores its reference.
+     *
+     * @param sipAddress - The sip address to use.
```
paweldomas: missing type declaration

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+     */
+    isVideoSIPGWAvailable() {
+        return this.status === Constants.STATUS_AVAILABLE;
+    }
+
+    /**
+     * Listener for session state changed. When a session goes to off or failed
+     * we delete its reference.
+     *
+     * @param address - The SIP address of the session.
```
paweldomas: missing type declaration

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+     * @param event - {name, oldState, newState}
```
paweldomas: event can be described like "options" here:
https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/JingleSessionPC.js#L44

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+     */
+    sessionStateChanged(address, event) {
+        if (event.newState === Constants.STATE_OFF
+            || event.newState === Constants.STATE_FAILED) {
+            const session = this.sessions[address];
+
+            session.removeStateListener(this.sessionStateChangeListener);
+            delete this.sessions[address];
```
paweldomas: It would be worth to mention somewhere in the JSDoc that session instance once enters this OFF or FAILED should not be used again.

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+        }
+    }
+}
+
+/**
+ * Sends a jibri command using an iq.
+ *
+ * @param videoSIPGWSession - The session sending the command.
```
paweldomas: missing type declaration

modules/xmpp/VideoSIPGW.js
```diff
@@ -0,0 +1,184 @@
...
+ * @param action - The action to send.
+ * @param chatRoom - The chat room to send the iq to.
+ */
+export function sendJibriIQ(videoSIPGWSession, action, chatRoom) {
```
paweldomas: what is the reason for exporting this function instead of having it in JitsiVideoSIPGWSession.js ?

paweldomas: it could be a member or am I missing something ?

service/xmpp/XMPPEvents.js
```diff
@@ -184,6 +184,9 @@ const XMPPEvents = {
     // suspending detected
     SUSPEND_DETECTED: 'xmpp.suspend_detected',
 
+    // Indicates that video SIP GW state changed.
```
paweldomas: Event arguments should be described

JitsiConference.js
```diff
@@ -1653,4 +1654,36 @@ JitsiConference.prototype.getSpeakerStats = function() {
...
+    return this.videoSIPGWHandler
+        && this.videoSIPGWHandler.isVideoSIPGWAvailable();
+};
+
+/**
+ * Creates a video SIP GW session and returns it if service is enabled.
+ *
+ * @param {string} sipAddress - The sip address to be used.
+ * @param {string} displayName - The display name to be used for this session.
+ * @returns {JitsiVideoSIPGWSession|null}
```
paweldomas: It would be good to explain when null can be returned. When looking just at the description I would assume that it's returned when the service is available.

When I actually take a look at the code it looks that VideoSIPGW is created regardless whether it's available or not, but that's less important. I think there should be a private getter with lazy initialisation which should be used from both here and isVideoSIPGWAvailable to create "VideoSIPGW" and be able to check service availability. Currently if 'isVideoSIPGWAvailable' is called, before "createVideoSIPGWSession" it will never return true (no "videoSIPGWHandler" instance).

modules/videosipgw/JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * Subscribes the passed listener to the event for state change of this
+     * session.
+     *
+     * @param {Function} listener - The function that will receive the event.
+     */
+    addStateListener(listener) {
+        this.eventEmitter.addListener(STATE_CHANGED, listener);
```
paweldomas: you can use Listenable instead of implementing this stuff manually

https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/util/Listenable.js

modules/videosipgw/JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,152 @@
...
+            type: 'set' })
+            .c('jibri', attributes)
+            .up();
+
+        logger.log('Stop video SIP GW session', iq.nodeTree);
+        this.chatRoom.connection.sendIQ(
+            iq,
+            result => {
+                logger.log('Result', result);
+                const initialState
```
paweldomas: Github doesn't show me the line limit, but do we need to wrap this ? It feels like this would fit into single line (but I'm not sure - have not tried that).

modules/videosipgw/JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,152 @@
...
+        this._sendJibriIQ('stop');
+    }
+
+    /**
+     * Starts a new session. Sends an iq to the focus.
+     */
+    start() {
+        // if state is off, this session was active for some reason
+        // and we should create new one, rather than reusing it
+        if (this.state === Constants.STATE_ON
```
paweldomas: this check feels weird... so the initial value is "undefined" ? Can we initialise 'state' in the constructor and document it's default value ?

modules/videosipgw/JitsiVideoSIPGWSession.js
```diff
@@ -0,0 +1,152 @@
...
+    }
+
+    /**
+     * Sends a jibri command using an iq.
+     *
+     * @private
+     * @param {JitsiVideoSIPGWSession} videoSIPGWSession - The session
+     * sending the command.
+     * @param {string} action - The action to send ('start' or 'stop').
+     * @param {ChatRoom} chatRoom - The chat room to send the iq to.
```
paweldomas: chatRoom is no longer an argument ? the same with videoSIPGWSession

modules/videosipgw/VideoSIPGW.js
```diff
@@ -0,0 +1,144 @@
...
+     */
+    constructor(chatRoom) {
+        this.chatRoom = chatRoom;
+        this.eventEmitter = chatRoom.eventEmitter;
+        logger.info('creating VideoSIPGW');
+        this.sessions = {};
+
+        this.sessionStateChangeListener = this.sessionStateChanged.bind(this);
+
+        chatRoom.addPresenceListener('jibri-sip-status',
```
paweldomas: I see we never remove presence listener. Can you add some comment which explains why ? Like for example "VideoSIPGW, JitsiConference and ChatRoom are not reusable and no more than one VideoSIPGW can be created per JitsiConference, so we don't bother to cleanup".

modules/videosipgw/VideoSIPGW.js
```diff
@@ -0,0 +1,144 @@
...
+            this.handleJibriSIPStatus.bind(this));
+        chatRoom.addPresenceListener('jibri-sip-call-state',
+            this.handleJibriSIPState.bind(this));
+    }
+
+    /**
+     * Handles presence nodes with name: jibri-sip-status.
+     *
+     * @param node the presence node to handle.
```
paweldomas: Missing "node" type. If it's hard to tell then maybe mark as {Object} and explain whether it's a DOM Element or what is it (I don't know) ?

modules/videosipgw/VideoSIPGW.js
```diff
@@ -0,0 +1,144 @@
...
+            this.status = newStatus;
+            this.eventEmitter.emit(
+                XMPPEvents.VIDEO_SIP_GW_AVAILABILITY_CHANGED, this.status);
+        }
+    }
+
+    /**
+     * Handles presence nodes with name: jibri-sip-call-state.
+     *
+     * @param node the presence node to handle.
```
paweldomas: Please see my comment above.

modules/videosipgw/VideoSIPGW.js
```diff
@@ -0,0 +1,144 @@
...
+     * @param {string} sipAddress - The sip address to use.
+     * @param {string} displayName - The display name to use.
+     * @returns {JitsiVideoSIPGWSession}
+     */
+    createVideoSIPGWSession(sipAddress, displayName) {
+        const session = new JitsiVideoSIPGWSession(
+            sipAddress, displayName, this.chatRoom);
+
+        session.addStateListener(this.sessionStateChangeListener);
+        this.sessions[sipAddress] = session;
```
paweldomas: It would be good to check if we don't overwrite any session here and eventually print an error or something.

modules/videosipgw/VideoSIPGW.js
```diff
@@ -0,0 +1,144 @@
...
+     * Listener for session state changed. When a session goes to off or failed
+     * we delete its reference.
+     *
+     * @param {string} address - The SIP address of the session.
+     * @param {options} event - { name, oldState, newState }
+     */
+    sessionStateChanged(address, event) {
+        if (event.newState === Constants.STATE_OFF
+            || event.newState === Constants.STATE_FAILED) {
+            const session = this.sessions[address];
```
paweldomas: The code will crash if for any reason session for 'address' is no longer there. It would be good to detect and log an error instead.

modules/videosipgw/JitsiVideoSIPGWSession.js
damencho: Yep, if it is in one line I got this from lint:
137:62  error  Expected line break before `.attr`  newline-per-chained-call

JitsiConference.js
```diff
@@ -2341,34 +2341,43 @@ JitsiConference.prototype.getSpeakerStats = function() {
...
             return null;
         }
 
-        if (!this.videoSIPGWHandler) {
-            this.videoSIPGWHandler = new VideoSIPGW(this.room);
-            logger.info('Created VideoSIPGW');
-        }
-
-        return this.videoSIPGWHandler
+        return this._getVideoSIPGWHandle()
```
paweldomas: What will happen if 'isVideoSIPGWAvailable' return false, but someone calls this method anyway ?

modules/videosipgw/JitsiVideoSIPGWSession.js
```diff
@@ -28,11 +28,16 @@ export default class JitsiVideoSIPGWSession {
...
+        super();
+
         this.sipAddress = sipAddress;
         this.displayName = displayName;
         this.chatRoom = chatRoom;
 
-        this.eventEmitter = new EventEmitter();
+        // The initial state is undefined. Initial state cannot be STATE_OFF,
+        // the session enters this state when it was in STATE_ON and was stopped
+        // and such session cannot be used anymore
```
paweldomas: Probably the type should be documented with:
`/*`
`* The description...`
` * @type {Constants|undefined}`
`*/`
`this.state = undefined;`
Hmm actually I don't know how to document the type in case 'VideoSIPGWConstants' are imported as 'Constants'...

PR #413, opened 2017-03-15
P2P ver2

paweldomas:
work in progress


modules/RTC/TraceablePeerConnection.js
```diff
@@ -705,89 +1420,106 @@ TraceablePeerConnection.prototype.createDataChannel = function(label, opts) {
...
+    this.trace('setLocalDescription::preTransform', dumpSDP(d));
 
-        // if we're running on FF, transform to Plan A first.
-        if (RTCBrowserType.usesUnifiedPlan()) {
-            d = this.interop.toUnifiedPlan(d);
-            this.trace(
-                'setLocalDescription::postTransform (Plan A)',
-                dumpSDP(d));
-        }
+    // if we're running on FF, transform to Plan A first.
```
saghul: "If we are using unified plan, transform to it first"

modules/RTC/TraceablePeerConnection.js
```diff
@@ -910,68 +1673,148 @@ TraceablePeerConnection.prototype.createAnswer
                  *  about unmapped ssrcs)
                  */
                 if (!RTCBrowserType.isFirefox()) {
-                    answer.sdp
+                    // If there are no local video tracks, then a
```
saghul: the comment seems cut

modules/RTC/TraceablePeerConnection.js
```diff
@@ -910,68 +1673,148 @@ TraceablePeerConnection.prototype.createAnswer
...
         },
         constraints);
-    };
+};
 
-TraceablePeerConnection.prototype.addIceCandidate
+/* eslint-enable max-params */
+
+/**
+ *
```
saghul: some extra docs would help here. Also, what SSRC are we extracting here? Is it the primary one? Then how about extractPrimerySsrc?

paweldomas:
@saghul thanks for the review I will work on the comments ! As for the correctness it's all on me now. We expect to have some unexpected bugs appearing after beta release, but they should go away with time eventually (fixed)

JitsiConference.js
```diff
@@ -678,82 +681,43 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
...
 
-    return this.jingleSession.generateNewStreamSSRCInfo();
+    // We are done immediately
+    logger.warn(
+        'Remove local MediaStream - no JVBJingleSession started yet');
+
+    return Promise.resolve();
 };
 
+/* eslint-enable max-params */
```
paweldomas: this is some mistake during rebase, it needs to go away with the disable above

JitsiConference.js
```diff
@@ -964,6 +928,16 @@ JitsiConference.prototype.onMemberLeft = function(jid) {
     }
 };
 
+/**
+ * Method called on local MUC role change.
+ * @param {string} role the name of new user's role as defined by XMPP MUC.
+ */
+JitsiConference.prototype.onLocalRoleChanged = function(role) {
```
paweldomas: There are plenty of methods like that which makes it a convention :D for this class.
It feels like they should be non-public with "_" in front of the names. I would rather add a FIXME for all of them as this is a bigger cleanup.

JitsiConference.js
```diff
@@ -1132,97 +1125,119 @@ JitsiConference.prototype.onIncomingCall
...
+            );
+
+            // Start callstats as soon as peerconnection is initialized,
+            // do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never
+            // happen in case if user doesn't have or denied permission to
+            // both camera and microphone.
+            this.statistics.startCallStats(jingleSession);
+            this._startRemoteStats();
+        });
+    } catch (e) {
```
paweldomas: I have no idea. I imagine this try catch was introduced to protect from breaking strophe packet handler when something goes wrong, but we would have to ask the original author. Not sure what to do about this one. Remove or add a comment ?

JitsiConference.js
```diff
@@ -1037,6 +1011,32 @@ JitsiConference.prototype.onRemoteTrackAdded = function(track) {
 };
 
 /**
+ * Callback called by the Jingle plugin when 'session-answer' is received.
+ * @param {JingleSessionPC} session the Jingle session for which an answer was
+ * received.
+ * @param {jQuery} answer a jQuery selector pointing to 'jingle' IQ element
+ */
+// eslint-disable-next-line no-unused-vars
+JitsiConference.prototype.onCallAccepted = function(session, answer) {
```
paweldomas: please check on my answer above

JitsiConference.js
```diff
@@ -1439,15 +1429,27 @@ JitsiConference.prototype.getPhonePin = function() {
 };
 
 /**
+ * Obtains currently active <tt>TraceablePeerConnection</tt> (if any).
+ * @return {TraceablePeerConnection|null}
+ * @protected
+ */
+JitsiConference.prototype.getActivePeerConnection = function() {
+    return this.jvbJingleSession
+        ? this.jvbJingleSession.peerconnection : null;
```
paweldomas: will it return null ? do I understand correctly that with && it will return this.jvb.JingleSession ? So if it's null it should return null ?

JitsiConference.js
```diff
@@ -1558,17 +1560,29 @@ JitsiConference.prototype.isCallstatsEnabled = function() {
...
  * Handles track attached to container (Calls associateStreamWithVideoTag method
  * from statistics module)
- * @param track the track
+ * @param {JitsiLocalTrack|JitsiRemoteTrack} track the track
  * @param container the container
  */
 JitsiConference.prototype._onTrackAttach = function(track, container) {
-    const ssrc = track.getSSRC();
+    const isLocal = track.isLocal();
+    let ssrc = null;
```
paweldomas: don't we prefer null over undefined :P ?

JitsiConferenceEventManager.js
```diff
@@ -29,6 +29,12 @@ function JitsiConferenceEventManager(conference) {
             conference.statistics.sendMuteEvent(track.isMuted(),
                 track.getType());
         });
+    conference.on(
```
paweldomas: That is good question. I guess nowhere just like all other event listeners added by this class as there is no way to dispose this class gracefully. I can add FIXME, as it's too big effort to clean this up now.

Anyway it seems like it integral part of JitsiConference instance, so it will be invalid with it's parent. There is no way to reuse JitsiConference instance as far as I know.

JitsiConferenceEventManager.js
```diff
@@ -543,6 +505,27 @@ JitsiConferenceEventManager.prototype.setupRTCListeners = function() {
             }
         });
 
+    rtc.addListener(RTCEvents.LOCAL_UFRAG_CHANGED,
+        (tpc, ufrag) => {
```
paweldomas: like we discussed off the list I will use "tpc"

JitsiConferenceEventManager.js
```diff
@@ -647,15 +630,15 @@ JitsiConferenceEventManager.prototype.setupStatisticsListeners = function() {
             JitsiConferenceEvents.CONNECTION_STATS, stats);
     });
 
-    conference.statistics.addByteSentStatsListener(stats => {
+    conference.statistics.addByteSentStatsListener((tPeerConn, stats) => {
```
paweldomas: will go with "tpc" here

modules/RTC/JitsiLocalTrack.js
```diff
@@ -144,6 +158,35 @@ function JitsiLocalTrack(
 JitsiLocalTrack.prototype = Object.create(JitsiTrack.prototype);
 JitsiLocalTrack.prototype.constructor = JitsiLocalTrack;
 
+JitsiLocalTrack.prototype._addPeerConnection = function(tpc) {
+    if (this._isAttachedToPC(tpc)) {
```
paweldomas: This `_isAttachedToPC(tpc)` one liner is used in TraceablePeerConnection. Should I remove the _ sign and make it public ? I was thinking that it's kind of library internal, but I'm still confused about this sing usage....

modules/RTC/JitsiLocalTrack.js
```diff
@@ -238,6 +281,23 @@ JitsiLocalTrack.prototype._setRealDeviceIdFromDeviceList = function(devices) {
 };
 
 /**
+ * Sets the stream property of JitsiLocalTrack object and sets all stored
+ * handlers to it.
+ * @param {MediaStream} stream the new stream.
+ */
+JitsiLocalTrack.prototype._setStream = function(stream) {
+    JitsiTrack.prototype._setStream.call(this, stream);
```
paweldomas: will it work with this old style prototype class ?

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -45,11 +48,12 @@ function JitsiRemoteTrack(
...
         },
         mediaType,
-        videoType,
-        ssrc);
+        videoType);
     this.rtc = rtc;
+    this.ssrc = ssrc;
     this.ownerEndpointId = ownerEndpointId;
     this.muted = muted;
+    this.isP2P = isP2P;
```
paweldomas: nope, conference deals with both p2p and JVB remote tracks.

modules/RTC/MungeLocalSdp.js
```diff
@@ -0,0 +1,343 @@
...
+     *
+     * NOTE 1 video track is assumed
+     *
+     * @param {SdpTransformWrap} transformer the transformer instance which will
+     * be used to process the SDP.
+     * @return {boolean} <tt>true</tt> if any modification were made.
+     * @private
+     */
+    _describeLocalVideoTracks(transformer) {
+
```
paweldomas: this space before comment lint rule is confusing :P

modules/RTC/MungeLocalSdp.js
```diff
@@ -0,0 +1,343 @@
...
+        for (const videoTrack of localVideos) {
+            const isMuted = videoTrack.isMuted();
+            const muteInProgress = videoTrack.inMuteOrUnmuteProgress;
+            const isAttached = videoTrack._isAttachedToPC(this.pc);
+            const shouldFakeSdp = isMuted || muteInProgress || !isAttached;
+
+            logger.debug(
+                `${videoTrack
+                 } isMuted: ${isMuted
```
paweldomas: I agree, but we used that as a convention for wrapping template string instead of using + sign. I'd rather keep it like this until we decide on strict rule for this case. There are many other places like this not only in this project.

modules/util/JSUtil.js
```diff
@@ -0,0 +1,8 @@
+/**
+ * Gets object's values.
+ * @param {Object} object
+ * @return {Array<Object>}
+ */
+export function getValues(object) {
```
paweldomas: No support in IE, Opera, Safari. Does any of our libs fix that ? https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_objects/Object/values#Browser_compatibility

modules/xmpp/ChatRoom.js
```diff
@@ -534,14 +534,17 @@ export default class ChatRoom extends Listenable {
         // make sure we catch all errors coming from any handler
         // otherwise we can remove the presence handler from strophe
         try {
-            let tagHandler = this.presHandlers[node.tagName];
+            let tagHandlers = this.presHandlers[node.tagName];
 
             if (node.tagName.startsWith('jitsi_participant_')) {
-                tagHandler = this.participantPropertyListener;
+                // NOTE should we allow more than one handler ?
```
paweldomas: The question is whether or not  "jitsi_participant_ events" require more handlers just like other events, but it's accessed via "set" style method, so I'm leaving it as is. Should I remove the comment ?

modules/xmpp/strophe.jingle.js
```diff
@@ -18,16 +20,24 @@ import ConnectionPlugin from './ConnectionPlugin';
...
-    constructor(xmpp, eventEmitter) {
+    constructor(xmpp, eventEmitter, p2pStunServers) {
         super();
         this.xmpp = xmpp;
         this.eventEmitter = eventEmitter;
         this.sessions = {};
-        this.iceConfig = { iceServers: [] };
+        this.jvbIceConfig = { iceServers: [ ] };
+        this.p2pIceConfig = { iceServers: [ ] };
+        if (Array.isArray(p2pStunServers)) {
```
paweldomas: well, but it checks for Array type as well, so I'd rather leave it as is.

service/RTC/RTCEvents.js
```diff
@@ -43,9 +30,7 @@ const RTCEvents = {
 
     /**
      * Indicates that the remote track has been removed from the conference.
-     * 1st event argument is the ID of the parent WebRTC stream to which
-     * the track being removed belongs to.
-     * 2nd event argument is the ID of the removed track.
+     * 1st event argument is the removed {@link JitsiRemoteTrack} instance.
```
paweldomas: but this one is consistent right ? I will go with {@link }

P2PEnabledConference.js
```diff
@@ -0,0 +1,947 @@
...
+import * as JitsiConferenceEvents from './JitsiConferenceEvents';
+import * as RTCEvents from './service/RTC/RTCEvents';
+import * as XMPPEvents from './service/xmpp/XMPPEvents';
+
+const logger = getLogger(__filename);
+
+/**
+ * A peer to peer enabled conference that will try to use direct connection when
+ * available in case there are only 2 participants in the room. The JVB
+ * connection will be kept alive and it will be reused if the 3rd participant
```
paweldomas: is there a rule to add "the" to first, second and so on ?

P2PEnabledConference.js
```diff
@@ -0,0 +1,947 @@
...
+
+        /**
+         * Stores reference to deferred start P2P task. It's created when 3rd
+         * participant leaves the room in order to avoid ping pong effect (it
+         * could be just a page reload).
+         * @type {number|null}
+         */
+        this.deferredStartP2P = null;
+
+        const delay = parseInt(options.config.backToP2PDelay, 10);
```
paweldomas: it's needed to be defined already at this.backToP2PDelay

P2PEnabledConference.js
```diff
@@ -0,0 +1,947 @@
...
+
+        if (this.p2pJingleSession) {
+            allPromises.push(
+                this.p2pJingleSession.replaceTrack(oldTrack, newTrack));
+        }
+
+        return Promise.all(allPromises);
+    }
+
+    /* eslint-disable prefer-rest-params */
```
paweldomas: but they are inconvenient for this particular case, I would have to merge arrays.

P2PEnabledConference.js
```diff
@@ -0,0 +1,947 @@
...
+
+    /**
+     * Intercepts events emitted by parent <tt>JitsiConference</tt>
+     * @private
+     */
+    _emitIntercept(eventType) {
+        const shouldBlock = this._shouldBlockEvent(eventType);
+
+        switch (eventType) {
+
```
paweldomas: nope, it's ESLint rule for comments...

P2PEnabledConference.js
```diff
@@ -0,0 +1,947 @@
...
+                        'Deferred start P2P task\'s been set already!');
+
+                    return;
+                }
+                logger.info(
+                    `Will start P2P with: ${jid
+                         } after ${this.backToP2PDelay} seconds...`);
+                this.deferredStartP2P = window.setTimeout(
+                    this._startPeer2PeerSession.bind(this, jid),
+                    this.backToP2PDelay * 1000);
```
paweldomas: well this way it can be stored in seconds which is more human friendly in my opinion

P2PEnabledConference.js
```diff
@@ -0,0 +1,947 @@
...
+    }
+
+    /**
+     * Creates fake event emitter used to intercept some of the XMPP events
+     * emitted by {@link P2PEnabledConference.p2pJingleSession}.
+     * @return {EventEmitter}
+     * @private
+     */
+    _createEventEmitter() {
+        const self = this;
```
paweldomas: it is needed, there is object defined below, I don't know how to write it otherwise. I guess arrow will not work here

modules/RTC/RTC.js
```diff
@@ -757,20 +627,32 @@ export default class RTC extends Listenable {
      * @param ssrc the ssrc to check.
      */
     getResourceBySSRC(ssrc) {
+        const track = this._getTrackBySSRC(ssrc);
 
-        // FIXME: Convert the SSRCs in whole project to use the same type.
-        // Now we are using number and string.
-        if (this.getLocalTracks().find(
+        return track ? track.getParticipantId() : null;
```
paweldomas: it will make it harder to figure out whether this can return null or undefined

modules/RTC/TraceablePeerConnection.js
```diff
@@ -227,15 +318,133 @@ const dumpSDP = function(description) {
...
+ * @return {Array<JitsiRemoteTrack>}
+ */
+TraceablePeerConnection.prototype.getRemoteTracks
+= function(endpointId, mediaType) {
+    const remoteTracks = [];
+    const endpoints
+        = endpointId ? [ endpointId ] : Object.keys(this.remoteTracks);
+
+    for (const endpoint of endpoints) {
+        const endpointTracks = this.remoteTracks[endpoint] || { };
```
paweldomas: this goes away with Map conversion, but yes I believe there is space for both in the rules

modules/RTC/TraceablePeerConnection.js
```diff
@@ -63,10 +84,71 @@ function TraceablePeerConnection(
...
+     */
+    this.localTracks = {};
+
+    /**
+     * @typedef {Object} TPCGroupInfo
+     * @property {string} semantics the SSRC groups semantics
+     * @property {Array<number>} ssrcs group's SSRCs in order where the first
+     * one is group's primary SSRC, the second one is secondary (RTX) and so
+     * on...
+     */
```
paweldomas: and I don't know if there is any acronym for the 3rd one :P it's neither RTX nor primary, the last layer of the simulcast group ?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -63,10 +84,71 @@ function TraceablePeerConnection(
...
+    /**
+     * @typedef {Object} TPCSSRCInfo
+     * @property {Array<number>} ssrcs an array which holds all track's SSRCs
+     * @property {Array<TPCGroupInfo>} groups an array stores all track's SSRC
+     * groups
+     */
```
paweldomas: what ? it's typedef

modules/RTC/TraceablePeerConnection.js
paweldomas: no, actually there's no space for object

JitsiConference.js
saghul: Ok, sounds good

JitsiConference.js
saghul: Lets's do comment

JitsiConference.js
saghul: it will return null or undefined. falsey nonetheless

JitsiConference.js
saghul: it's a local var which you check for falseyness

JitsiConferenceEventManager.js
saghul: FML. Add a comment please.

modules/RTC/JitsiLocalTrack.js
saghul: there is only one way to know! :-)

modules/RTC/JitsiRemoteTrack.js
saghul: Hum, so generally I think a track shouldn't be instructed where it is. Is there no way of knowing if the conference is in P2P mode so the track could query it?

modules/RTC/MungeLocalSdp.js
saghul: When did we convene about that? Was I not invited to the linting party? Let's stay consistent, whatever was chosen.

modules/util/JSUtil.js
saghul: We can depend on a shim: https://github.com/es-shims/Object.values just like we do for `for..of`.

modules/xmpp/ChatRoom.js
saghul: yeah

modules/xmpp/strophe.jingle.js
saghul: ack

service/RTC/RTCEvents.js
saghul: yes @ link is 👍

P2PEnabledConference.js
saghul: doesn't it actually help? `[...array1, ...array2]`

P2PEnabledConference.js
saghul: ah, ok

P2PEnabledConference.js
saghul: ok

modules/RTC/TraceablePeerConnection.js
saghul: Oh, I didn't know we did that :-)

P2PEnabledConference.js
saghul: ah, I see. Sorry, I had misread! 👍

modules/RTC/JitsiRemoteTrack.js
paweldomas: Conference is aware if P2P is currently active, but I don't see a problem with this field. It's for the other components (including conference) to know where does a track belong to without having to go and fetch it's parent TPC,JingleSessionPC. It doesn't do anything with this field, but it's part of it's identity.

modules/RTC/MungeLocalSdp.js
paweldomas: There was no linting party. I was not involved in other rules introduction as well. It came up during discussion on other PR on the topic of "how to wrap template strings". I'm not sure if it's "official though" :P

modules/util/JSUtil.js
paweldomas: Is this "es-shims" included through some other lib ? I don't see it declared anywhere in package.json ? Should I add it ? If I see correctly it adds few dependencies. I will probably use a Map or something instead and have this one removed.

P2PEnabledConference.js
paweldomas: that is my understanding as well, but I thought that when it comes to order it should come with "the"

P2PEnabledConference.js
paweldomas: How do I write this with rest args ?
`this._originalEmit.apply(this.eventEmitter, arguments);`

paweldomas: there is `eventType` that has to be merged

modules/RTC/JitsiRemoteTrack.js
saghul: Ok then!

modules/RTC/TraceablePeerConnection.js
```diff
@@ -20,13 +24,28 @@ const SIMULCAST_LAYERS = 3;
 /* eslint-disable max-params */
 
 /**
+ * Extracts the ICE username fragment from an SDP string.
+ * @param {string} sdp the SDP in raw text format
+ */
+function getUfrag(sdp) {
```
bbaldino: this looks like a good helper to be put in SDPUtils.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -63,10 +83,71 @@ function TraceablePeerConnection(
...
     /**
+     * Indicates whether or not this instance is used in a peer to peer
+     * connection.
+     * @type {boolean}
+     */
+    this.isP2P = isP2P;
+
+    // FIXME: We should support multiple streams per jid.
+    /**
+     * The map holds remote tracks associated with this peer connection.
```
bbaldino: can you document what's being mapped here? like "maps jid to a map of its tracks, keyed by media type" or something

modules/RTC/TraceablePeerConnection.js
bbaldino: where is the actual typedef itself?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -227,15 +317,140 @@ const dumpSDP = function(description) {
...
+    if (!endpointId) {
+        logger.error('On peerMuteChanged - no endpoint ID');
+
+        return;
+    }
+    const track = this.getRemoteTracks(endpointId, mediaType);
+
+    if (track.length) {
+        // NOTE 1 track per media type is assumed
+        track[0].setMute(isMuted);
```
bbaldino: do you think that, in places like this, we should throw if the assumption isn't true (e.g. here if the number of tracks is >1)?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -227,15 +317,140 @@ const dumpSDP = function(description) {
...
+    if (mediaType !== undefined) {
+        tracks = tracks.filter(track => track.getType() === mediaType);
+    }
+
+    return tracks;
+};
+
+/**
+ * Obtains all remote tracks currently known to this PeerConnection instance.
+ * @param {string} [endpointId] the track owner's identifier (MUC nickname)
```
bbaldino: if these are optional, how about doing them in an options object instead of params?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -689,12 +1088,294 @@ TraceablePeerConnection.prototype.addStream = function(stream, ssrcInfo) {
...
+
+/**
+ * Checks whether given track is attached to this TPC. See
+ * {@link JitsiLocalTrack._isAttachedToPC} and {@link attachTrack} for more
+ * info.
+ * @param {JitsiLocalTrack} localTrack
+ * @return {boolean} <tt>true</tt> if attached or <tt>false</tt> otherwise
+ * @private
+ */
+TraceablePeerConnection.prototype._isTrackAttached = function(localTrack) {
```
bbaldino: why add a helper function for a one-liner?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -689,12 +1088,294 @@ TraceablePeerConnection.prototype.addStream = function(stream, ssrcInfo) {
...
+        if (RTCBrowserType.isFirefox()) {
+            this._handleFirefoxRemoveStream(webRtcStream);
+        } else {
+            this.peerconnection.removeStream(webRtcStream);
+        }
+    }
+};
+
+/**
+ * Removes local track as part of the unmute operation.
```
bbaldino: as part of the "mute" operation?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -689,12 +1088,294 @@ TraceablePeerConnection.prototype.addStream = function(stream, ssrcInfo) {
...
+        logger.warn(
+            `Not removing detached ${localTrack} as unmute from ${this}`);
+
+        return false;
+    }
+
+    if (webRtcStream) {
+        logger.info(
+            `Removing ${localTrack} as mute from ${this}`);
+        this._removeStream(webRtcStream);
```
bbaldino: i thought, if it's just being muted, that would be a 'detach'? but maybe i understood that incorrectly

modules/RTC/TraceablePeerConnection.js
```diff
@@ -705,89 +1386,106 @@ TraceablePeerConnection.prototype.createDataChannel = function(label, opts) {
 };
 
 TraceablePeerConnection.prototype.setLocalDescription
-    = function(description, successCallback, failureCallback) {
-        let d = description;
+= function(description, successCallback, failureCallback) {
+    let d = description;
```
bbaldino: this doesn't seem necessary.

EDIT: oh, was this in the existing code?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -910,68 +1639,149 @@ TraceablePeerConnection.prototype.createAnswer
...
+
+    return null;
+}
+
+/**
+ * Applies SSRC map extracted from the latest local description to local tracks.
+ * @param {Map<string,TrackSSRCInfo>} ssrcMap
+ * @private
+ */
+TraceablePeerConnection.prototype._applyLocalSSRCMap = function(ssrcMap) {
```
bbaldino: we have SDPConsistency for this...did you get rid of that?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -347,9 +369,12 @@ export default class ParticipantConnectionStatus {
         const isVideoTrackFrozen = this.isVideoTrackFrozen(participant);
         let isConnActiveByJvb = this.connStatusFromJvb[id];
 
-        // If no status was received from the JVB it means that it's active
-        // (the bridge does not send notification unless there is a problem).
-        if (typeof isConnActiveByJvb !== 'boolean') {
+        if (this.conference.isP2PEstablished()) {
+            logger.debug('Assuming connection active by JVB - in P2P mode');
+            isConnActiveByJvb = true;
```
bbaldino: i don't get this...we see that p2p is established...what does it mean to set `isConnActiveByJvb`?

modules/xmpp/ChatRoom.js
```diff
@@ -893,15 +896,37 @@ export default class ChatRoom extends Listenable {
...
      * @param name
+     * @param handler
      */
-    removePresenceListener(name) {
-        delete this.presHandlers[name];
+    removePresenceListener(name, handler) {
+        const tagHandlers = this.presHandlers[name];
+        const handlerIdx = tagHandlers ? tagHandlers.indexOf(handler) : -1;
+
+        // eslint-disable-next-line no-negated-condition
```
bbaldino: why not just change it to check that the found index was >= 0?

modules/xmpp/JingleSessionPC.js
```diff
@@ -60,40 +64,42 @@ export default class JingleSessionPC extends JingleSession {
...
-         * Used to keep state about muted/unmuted video streams
-         *  so we can prevent errant source-add/source-removes
-         *  from happening
+         * Stores "delayed" ICE candidates which are added to the PC once
+         * the first sRD/sLD cycle is done.
+         * @type {Array} an array of ICE candidate lines which can be added
+         * directly to the PC
          */
-        this.modifiedSSRCs = {};
+        this.candidiates = [];
```
bbaldino: is this new?  maybe name them to better reflect the documentation so it's more obvious what they are when used in the code.  `delayedIceCandidates`?

bbaldino: (or maybe `queuedIceCandidates`, based on the method they're used in below)

bbaldino: also, if you just want to synchronize adding these with the sRD/sLD cycle, you should just use the modification queue.  adding them as they come will result in more srd/sld cycles, but could also result in a faster connection.

modules/xmpp/JingleSessionPC.js
```diff
@@ -372,6 +453,53 @@ export default class JingleSessionPC extends JingleSession {
...
+                    this.peerconnection.addIceCandidate(
+                        rtcCandidate,
+                        () => logger.debug(`addIceCandidate ok: ${line}`),
+                        error => {
+                            logger.error(
+                                `addIceCandidate failed: ${line}`, error);
+                        });
+                } else {
+                    logger.debug(`Delaying ICE candidate: ${line}`);
+                    this.candidiates.push(rtcCandidate);
```
bbaldino: if we haven't even created the answer yet, we don't need to queue up these candidates separately, right?  when we call createAnswer they will just be there, since the stack is already aware of them.  but maybe i'm misunderstanding the scenario here.

modules/xmpp/JingleSessionPC.js
```diff
@@ -803,31 +1010,22 @@ export default class JingleSessionPC extends JingleSession {
...
-                        'Error renegotiating after processing remote'
-                            + ` source-add: ${error}`);
-                    finishedCallback(error);
-                });
+            return newRemoteSdp;
         };
 
-        this.modificationQueue.push(workFunction);
+        // Queue and execute
+        this._doRenegotiate('source-add', workFunction)
```
bbaldino: i don't get the changes in here.  all renegotiation should be done within the modification queue context, but we don't queue anything here?

modules/xmpp/JingleSessionPC.js
```diff
@@ -843,36 +1041,29 @@ export default class JingleSessionPC extends JingleSession {
...
-                        'Error renegotiating after processing remote'
-                          + ` source-remove: ${error}`);
-                    finishedCallback(error);
-                });
+            return newRemoteSdp;
         };
 
-        this.modificationQueue.push(workFunction);
+        // Queue and execute
+        this._doRenegotiate('source-remove', workFunction)
```
bbaldino: same as above

modules/xmpp/JingleSessionPC.js
```diff
@@ -969,155 +1160,166 @@ export default class JingleSessionPC extends JingleSession {
...
+    /**
+     * Renegotiate cycle implementation for the initiator's case.
+     * @param {object} remoteDescription the SDP object as defined by the WebRTC
+     * which will be used as remote description in the cycle.
+     * @param {function} resolve the success callback
+     * @param {function} reject the failure callback
+     * @private
+     * @private
+     */
+    _initiatorRenegotiate(remoteDescription, resolve, reject) {
```
bbaldino: it seems like we could've just replaced the existing, single `renegotiate` method with a version that added an option for creating a local offer (a scenario which didn't exist before)

modules/xmpp/JingleSessionPC.js
```diff
@@ -1196,203 +1398,248 @@ export default class JingleSessionPC extends JingleSession {
...
+     * remote session descriptions. Will compare the local description, before
+     * and after the renegotiation to update local streams description (sends
+     * "source-add"/"source-remove" notifications).
+     * @param {string} actionName the name of the action which will appear in
+     * the events logged to the logger.
+     * @param {WorkerFunction} workFunction a function that will be executed on
+     * the queue. See type description for moe info.
+     * @private
+     */
+    _doRenegotiate(actionName, workFunction) {
```
bbaldino: this seems to have gotten a lot more complicated (with multiple layers of work functions and wrappers) and it's not clear to my why it has to be that way.  maybe we can talk about that.

modules/xmpp/RtxModifier.js
```diff
@@ -120,17 +120,30 @@ export default class RtxModifier {
...
+    }
+
+    /**
+     * Does the same thing as {@link modifyRtxSsrcs}, but takes the
+     *  {@link MLineWrap} instance wrapping video media as an argument.
+     * @param {MLineWrap} videoMLine
+     * @return {boolean} <tt>true</tt> if the SDP wrapped by
+     *  {@link SdpTransformWrap} has been modified or <tt>false</tt> otherwise.
+     */
+    modifyRtxSsrcs2(videoMLine) {
```
bbaldino: if we have to keep both methods (this and `modifyRtxSsrcs`), maybe we can come up with a better name here?

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -0,0 +1,107 @@
...
+
+import { getLogger } from 'jitsi-meet-logger';
+import * as MediaType from '../../service/RTC/MediaType';
+import * as SignalingEvents from '../../service/RTC/SignalingEvents';
+import SignalingLayer from '../../service/RTC/SignalingLayer';
+
+const logger = getLogger(__filename);
+
+/**
+ * Default XMPP implementation of the {@link SignalingLayer} interface. Obtains
```
bbaldino: this is a great idea

modules/xmpp/JingleSessionPC.js
```diff
@@ -969,155 +1161,166 @@ export default class JingleSessionPC extends JingleSession {
...
+     * @param {object} remoteDescription the SDP object as defined by the WebRTC
+     * which will be used as remote description in the cycle.
+     * @param {function} resolve the success callback
+     * @param {function} reject the failure callback
+     * @private
+     */
+    _responderRenegotiate(remoteDescription, resolve, reject) {
 
-            logger.debug('Renegotiate: setting remote description');
+        // FIXME use WebRTC promise API to simplify things
```
saghul: extra newline?

modules/xmpp/JingleSessionPC.js
```diff
@@ -969,155 +1161,166 @@ export default class JingleSessionPC extends JingleSession {
...
+    }
+
+    /**
+     * Renegotiate cycle implementation for the initiator's case.
+     * @param {object} remoteDescription the SDP object as defined by the WebRTC
+     * which will be used as remote description in the cycle.
+     * @param {function} resolve the success callback
+     * @param {function} reject the failure callback
+     * @private
+     * @private
```
saghul: this one must be super private :-)

modules/xmpp/JingleSessionPC.js
```diff
@@ -969,155 +1161,166 @@ export default class JingleSessionPC extends JingleSession {
...
+     */
+    _initiatorRenegotiate(remoteDescription, resolve, reject) {
+
+        // FIXME use WebRTC promise API to simplify things
+        if (this.peerconnection.signalingState === 'have-local-offer') {
+
```
saghul: extra newline

modules/RTC/JitsiLocalTrack.js
```diff
@@ -62,10 +69,17 @@ function JitsiLocalTrack(
 
     this.deviceId = deviceId;
     this.startMuted = false;
-    this.initialMSID = this.getMSID();
+    this.storedMSID = this.getMSID();
```
paweldomas: yes, MSID would be cleared, but it's needed for LocalSdpMunger

JitsiConference.js
```diff
@@ -91,14 +106,57 @@ function JitsiConference(options) {
     /**
      * Indicates whether the connection is interrupted or not.
      */
-    this.connectionIsInterrupted = false;
+    this.isJvbConnectionIsInterrupted = false;
```
paweldomas: oh yes, thanks !

JitsiConference.js
```diff
@@ -91,14 +106,57 @@ function JitsiConference(options) {
...
+     * @type {boolean}
+     */
+    this.isP2PConnectionInterrupted = false;
 
+    /**
+     * Flag set to <tt>true</tt> when P2P session has been established
+     * (ICE has been connected).
+     * @type {boolean}
+     */
+    this.p2pEstablished = false;
```
paweldomas: "isP2PConnectionEstablished" means "isP2PConnectionInUse", so maybe I should rename it this way

JitsiConference.js
```diff
@@ -91,14 +106,57 @@ function JitsiConference(options) {
...
+
+    /**
+     * A JingleSession for the direct peer to peer connection.
+     * @type {JingleSessionPC}
+     */
+    this.p2pJingleSession = null;
```
paweldomas: we never use something like that in our code for class fields, I would find this confusing

JitsiConference.js
```diff
@@ -937,6 +995,8 @@ JitsiConference.prototype.onMemberJoined = function(jid, nick, role, isHidden) {
             this.updateDTMFSupport();
         },
         error => logger.error(error));
+
+    this._startStopP2PSession();
```
paweldomas: because we are 1 in the room and 2nd person joins

JitsiConference.js
```diff
@@ -1132,139 +1260,151 @@ JitsiConference.prototype.onIncomingCall
...
+        this.jvbJingleSession = null;
 
-    // Stop the stats
-    if (this.statistics) {
-        this.statistics.stopRemoteStats();
-        this.statistics.stopCallStats();
+        // Let the RTC service do any cleanups
+        this.rtc.onCallEnded();
+    } else if (jingleSession === this.p2pJingleSession) {
+        this._stopP2PSession();
```
paweldomas: not yet

JitsiConference.js
```diff
@@ -1069,6 +1185,11 @@ JitsiConference.prototype.onRemoteTrackRemoved = function(removedTrack) {
     }, this);
 
     if (!consumed) {
+        if ((this.p2pEstablished && !removedTrack.isP2P)
+             || (!this.p2pEstablished && removedTrack.isP2P)) {
+            // This track has been removed explicitly by P2PEnabledConference
```
paweldomas: The thing is that it can be either removed by the P2P logic or as a result of 'source-remove'. The P2P removes JVB remote tracks when in P2P and vice versa. But when it's a result of 'source-remove' it is an error which should be logged. Does it make sense ? If this explanation is enough I will add it to the comment.

JitsiConference.js
paweldomas: I am leaving this as is.

JitsiConference.js
```diff
@@ -1132,139 +1260,151 @@ JitsiConference.prototype.onIncomingCall
...
 };
 
 /**
  * Handles the suspend detected event. Leaves the room and fires suspended.
+ * @param {JingleSessionPC} jingleSession
  */
-JitsiConference.prototype.onSuspendDetected = function() {
-    this.leave();
-    this.eventEmitter.emit(JitsiConferenceEvents.SUSPEND_DETECTED);
+JitsiConference.prototype.onSuspendDetected = function(jingleSession) {
```
paweldomas: Yes, each JingleSessionPC instance emits events regardless of it's P2P status (the event is routed in JitsiConferenceEventManager).

JitsiConference.js
```diff
@@ -1558,17 +1709,29 @@ JitsiConference.prototype.isCallstatsEnabled = function() {
...
  * @param container the container
  */
 JitsiConference.prototype._onTrackAttach = function(track, container) {
-    const ssrc = track.getSSRC();
+    const isLocal = track.isLocal();
+    let ssrc = null;
+
+    if (isLocal) {
+        // Local tracks have SSRC stored on per peer connection basis
+        const peerConnection
```
paweldomas: This logic there is about logging association between video tag and local SSRC to CallStats. But because we do CallStats only for the JVB it tries to get it from the JVB connection. I will add a comment.

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+ * @param {JingleSessionPC} session
+ * @private
+ */
+JitsiConference.prototype._onIceConnectionInterrupted = function(session) {
+    if (session.isP2P) {
+        this.isP2PConnectionInterrupted = true;
+    } else {
+        this.isJvbConnectionIsInterrupted = true;
+    }
+    if (session.isP2P === this.p2pEstablished) {
```
paweldomas: No it goes as: "if we're in P2P mode and event comes from P2P then emit the event" and vice versa.

There is no "jvbEstablished", because we don't care. There is confusion about P2P established, but I will rename that to "is P2P in use" like I've mentioned in the comment above.

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+};
+
+/**
+ * Handles {@link XMPPEvents.CONNECTION_ICE_FAILED}
+ * @param {JingleSessionPC} session
+ * @private
+ */
+JitsiConference.prototype._onIceConnectionFailed = function(session) {
+    if (session.isP2P) {
+        this._stopP2PSession('connectivity-error', 'ICE FAILED');
```
paweldomas: No, we don't terminate JVB connection when ICE fails.

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+ * @private
+ */
+JitsiConference.prototype._onIceConnectionRestored = function(session) {
+    if (session.isP2P) {
+        this.isP2PConnectionInterrupted = false;
+    } else {
+        this.isJvbConnectionIsInterrupted = false;
+    }
+
+    if (session.isP2P === this.p2pEstablished) {
```
paweldomas: same response as above ;)

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+            logger.error(`Attach ${localTracks} to JVB failed!`, error);
+        });
+};
+
+/**
+ * Adds remote tracks to the conference associated with the JVB session.
+ * @private
+ */
+JitsiConference.prototype._addRemoteJVBTracks = function() {
+    this._addRemoteTracks('JVB', this.jvbJingleSession);
```
paweldomas: It may seem, so but if additional safety checks would be necessary I would have to duplicate. There is one done on jingleSession instance already.

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+};
+
+/**
+ * Adds remote tracks to the conference associated with the P2P session.
+ * @private
+ */
+JitsiConference.prototype._addRemoteP2PTracks = function() {
+    this._addRemoteTracks('P2P', this.p2pJingleSession);
```
paweldomas: please check on my answer above

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+ * Called when {@link XMPPEvents.CONNECTION_ESTABLISHED} event is
+ * triggered for a {@link JingleSessionPC}. Switches the conference to use
+ * the P2P connection if the event comes from the P2P session.
+ * @param {JingleSessionPC} jingleSession the session instance.
+ * @private
+ */
+JitsiConference.prototype._onP2PConnectionEstablished
+= function(jingleSession) {
+    if (!jingleSession.isP2P) {
+
```
paweldomas: this ESlint rules are so confusing...

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+        return;
+    } else if (this.p2pJingleSession !== jingleSession) {
+        logger.error('CONNECTION_ESTABLISHED - not P2P session ?!');
```
paweldomas: yes, it's about invalid instance, will adjust the msg.

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+    }
+};
+
+/**
+ * Removes from the conference remote tracks associated with the JVB
+ * connection.
+ * @private
+ */
+JitsiConference.prototype._removeRemoteJVBTracks = function() {
+    this._removeRemoteTracks('JVB', this.jvbJingleSession);
```
paweldomas: Same as with the add methods. There is extra check and possible there could be more that would have to be duplicated otherwise.

modules/RTC/JitsiLocalTrack.js
bbaldino: 👍

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+};
+
+/**
+ * Removes from the conference remote tracks associated with the P2P
+ * connection.
+ * @private
+ */
+JitsiConference.prototype._removeRemoteP2PTracks = function() {
+    this._removeRemoteTracks('P2P', this.p2pJingleSession);
```
paweldomas: please check my answer above

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+    this.dtmfManager = null;
+
+    // Update P2P status
+    this.eventEmitter.emit(
+        JitsiConferenceEvents.P2P_STATUS,
+        this,
+        this.p2pEstablished);
+
+    // Refresh connection interrupted/restored
+    this.eventEmitter.emit(
```
paweldomas: No, when we switch back to JVB connection we need to update to reflect JVB's connection status.

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+            : JitsiConferenceEvents.CONNECTION_RESTORED);
+};
+
+/**
+ * Starts new P2P session.
+ * @param {string} peerJid the JID of the remote participant
+ * @private
+ */
+JitsiConference.prototype._startP2PSession = function(peerJid) {
+    this._maybeClearDeferredStartP2P();
```
paweldomas: That's because the receive side could get 'session-initiate' from malicious client.

JitsiConference.js
bbaldino: 👍

JitsiConference.js
```diff
@@ -1623,7 +1786,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+};
+
+/**
+ * Method when called will decide whether it's the time to start or stop
+ * the P2P session.
+ * @param {boolean} userLeftEvent if <tt>true</tt> it means that the call
+ * originates from the user left event.
+ * @private
+ */
+JitsiConference.prototype._startStopP2PSession = function(userLeftEvent) {
```
paweldomas: This is getting long now, how about `_maybeStartOrStopP2P` ?

JitsiConference.js
bbaldino: i get that, it just seems like there are groups of variables that correspond to each of those things that it might be nice to group together.  we could discuss it more and maybe put it on a list for a follow-up task (should we decide it's a good thing)? that way we don't have to block getting this in on it.

JitsiConference.js
bbaldino: gotcha, at this time i thought `_startStopP2PSession()` was starting and stopping (restarting) the session.

modules/RTC/LocalSdpMunger.js
```diff
@@ -0,0 +1,288 @@
...
+
+    /**
+     * Makes sure local SDP for audio media reflects current local tracks status
+     * of the parent TPC.
+     * @param {SdpTransformWrap} transformer the transformer instance which will
+     * be used to process the SDP.
+     * @return {boolean} <tt>true</tt> if any modification were made.
+     * @private
+     */
+    _describeLocalAudioTracks(transformer) {
```
paweldomas: For audio it's handling the case when the local audio track is "being detached"/"detached already". To fake that it's still present in the local description.

JitsiConference.js
bbaldino: ok, i think i get it.  yeah more comments there would be good

modules/RTC/LocalSdpMunger.js
paweldomas: I will try to improve, but any suggestions would be welcome. I'm not sure which details I'm missing here, but that's because I know what it does.

JitsiConference.js
bbaldino: further down in the review i thought i saw places where we start the callstats stuff for p2p (or maybe it's some other stats, not sure)?

JitsiConference.js
bbaldino: 👍

modules/RTC/RTC.js
```diff
@@ -253,34 +254,41 @@ export default class RTC extends Listenable {
         return RTCUtils.getDeviceAvailability();
     }
 
+    /* eslint-disable max-params */
```
paweldomas: the limit is 3 :)

JitsiConference.js
bbaldino: 👍

JitsiConference.js
bbaldino: ok, i'll check it out once you've done the rename and see if it's clearer

JitsiConference.js
bbaldino: mind adding a comment so it's obvious it's purposefully omitted?

modules/RTC/RTC.js
```diff
@@ -356,71 +364,41 @@ export default class RTC extends Listenable {
...
+     * Runs a callback on each PeerConnection currently stored in the RTC
+     * module. If callback return any non-null value during execution
+     * the execution loop will be aborted and the result will be returned.
+     * @param {function(TraceablePeerConnection)} callback the function to be
+     * executed
+     * @return {*} any first non-null nor undefined value returned by
+     * the callback.
+     * @private
+     */
+    _iteratePeerConnections(callback) {
```
paweldomas: this method is going to trash. I forgot to remove it once I moved to Map

modules/RTC/TraceablePeerConnection.js
paweldomas: at line 114 above it defines TPCSSRCInfo as an {Object} with the properties defined below. Does it answer you question ?

modules/RTC/TraceablePeerConnection.js
paweldomas: I think it should be thrown where there is attempt to add more than 1 track and as far as I know there is such check in JitsiConference (it actually logs an error, but does not throw.).

modules/RTC/TraceablePeerConnection.js
paweldomas: Sounds reasonable, but given that there are only 2 arguments and hardly ever one of them is undefined I would rather keep it as is. It will be only extra effort when calling this method to declare each argument inside of an object.

modules/RTC/TraceablePeerConnection.js
paweldomas: Because it's called outside of TPC. But it's still "RTC module internal" hence the "_" sign. Not sure if it's the right usage though...

modules/RTC/TraceablePeerConnection.js
paweldomas: hmm technically yes, I will check if we can get rid of this method

paweldomas: I think it makes sense to use detach when doing video mute, but because there's a risk of breaking stuff I would rather keep it as a one of "to be fixed" thing for after the initial PR is merged.

modules/RTC/TraceablePeerConnection.js
paweldomas: it was a bug, no longer in existing code

modules/RTC/TraceablePeerConnection.js
paweldomas: Yes, but SDPConsistency only handles video and 1 track. When I added support for storing local tracks in TPC there was a need to store that for any track. Maybe some cleanup around that will be required.

paweldomas: I'm adding that to the list of "to be fixed"

modules/connectivity/ParticipantConnectionStatus.js
paweldomas: like I wrote in one of the comments above it should say "is P2P connection in use/being used/the active one"

modules/xmpp/ChatRoom.js
paweldomas: For me -1 means does not exists in the array. Having to compare that to >= 0 will make me think and trying to decode what does it mean. I would prefer to keep it this way.

modules/xmpp/JingleSessionPC.js
paweldomas: It's not to synchronise, but to delay until the first sRD/sLD cycle occurred. Once that happens they are added as they come.

paweldomas: I will rename the field to `delayedIceCandidates`

modules/xmpp/JingleSessionPC.js
paweldomas: It's that the invitee will send 'session-initiate' and then immediately do a burst of ice candidates. There's no guarantee on how fast and when we do the initial sRD/sLD, so they have to be delayed/stored.

modules/xmpp/JingleSessionPC.js
paweldomas: `_doRenegotiate` pushes a task to the modification queue and it also contains common/duplicated logic.

modules/xmpp/JingleSessionPC.js
paweldomas: please check on my answer above

modules/xmpp/RtxModifier.js
paweldomas: well, but it does the same except that takes different arguments ? so I see that as ver2 (because JS does not distinguish between argument types)

modules/xmpp/JingleSessionPC.js
paweldomas: it's to remove duplicate logic

paweldomas: maybe it is clearer when looking at single commit changes, but might be not:
https://github.com/jitsi/lib-jitsi-meet/pull/413/commits/c657ddbefd7a73dd4886481ab7925452f3d89be7

JitsiConference.js
```diff
@@ -1623,7 +1795,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+
+    // Remove remote tracks
+    this._removeRemoteJVBTracks();
+
+    // Add remote tracks
+    this._addRemoteP2PTracks();
+
+    // Remove local tracks from JVB PC
+    // But only if it has started
+    if (this.jvbJingleSession) {
```
paweldomas: 1. It will extend the process
2. It will make the transition more noticeable (the black video flash) mostly because if the P2P is established first there will be no effect at all
3. I don't see too much improvements this would bring. There are not so many checks around that and I guess those will still be required to some extent

JitsiConference.js
```diff
@@ -1623,7 +1795,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+    // Accept the offer
+    this.p2pJingleSession = jingleSession;
+
+    this.p2pJingleSession.initialize(
+        false /* initiator */, this.room, this.rtc);
+
+    const localTracks = this.getLocalTracks();
+
+    logger.debug(`Adding ${localTracks} to P2P...`);
+    this.p2pJingleSession.addLocalTracks(localTracks).then(
```
paweldomas: I don't get what you mean here. Why would you want to catch ? Can you be more specific ? What code needs to be executed if something fails ?

There is reject handler already that prints an error, no ?

paweldomas: If we fail to add something is messed up and we don't accept P2P. P2P is best effort currently.

JitsiConference.js
```diff
@@ -1623,7 +1795,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+
+/**
+ * Attaches local tracks back to the JVB connection.
+ * @private
+ */
+JitsiConference.prototype._attachLocalTracksToJvbSession = function() {
+    const localTracks = this.getLocalTracks();
+
+    logger.info(`Attaching ${localTracks} to JVB`);
+    this.jvbJingleSession.attachLocalTracks(localTracks).then(
```
paweldomas: I'm afraid there's not much we can do if that fails (unless you have any ideas)

JitsiConference.js
```diff
@@ -1623,7 +1795,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+
+/**
+ * Detaches local tracks from the JVB connection.
+ * @private
+ */
+JitsiConference.prototype._detachLocalTracksFromJvbSession = function() {
+    const localTracks = this.getLocalTracks();
+
+    logger.info(`Detaching local tracks from JVB: ${localTracks}`);
+    this.jvbJingleSession.detachLocalTracks(localTracks).then(
```
paweldomas: but what can be done in catch ?

JitsiConference.js
```diff
@@ -1623,7 +1795,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+};
+
+/**
+ * Sets new P2P status and updates some events/states hijacked from
+ * the <tt>JitsiConference</tt>.
+ * @param {boolean} newStatus the new P2P status value, <tt>true</tt> means that
+ * P2P is now in use, <tt>false</tt> means that the JVB connection is now in use
+ * @private
+ */
+JitsiConference.prototype._setP2PStatus = function(newStatus) {
```
paweldomas: I will clean that up together with field name

JitsiConferenceEvents.js
```diff
@@ -133,6 +133,11 @@ export const PARTICIPANT_PROPERTY_CHANGED
     = 'conference.participant_property_changed';
 
 /**
+ * Indicates that the conference has switched between JVB and P2P connections.
+ */
+export const P2P_STATUS = 'conference.p2pConnectionStatus';
```
paweldomas: We refer to that as sessions only inside of JitsiConference. In the outside world we care only whether we're running P2P or not. I will rename the string.

JitsiConference.js
```diff
@@ -1623,7 +1795,519 @@ JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
...
+};
+
+/**
+ * Returns the current ICE state of the P2P connection.
+ * NOTE: method is used by the jitsi-meet-torture tests.
+ * @return {string|null} an ICE state or <tt>null</tt> if there's currently
+ * no P2P connection.
+ */
+JitsiConference.prototype.getP2PConnectionState = function() {
+    if (this.p2pEstablished && this.p2pJingleSession) {
```
paweldomas: it should not be possible, removing it

modules/RTC/JitsiLocalTrack.js
```diff
@@ -144,6 +158,34 @@ function JitsiLocalTrack(
...
+
+/**
+ * Checks whether or not this instance is attached to given
+ * <tt>TraceablePeerConnection</tt>. See
+ * {@link TraceablePeerConnection.attachTrack} for more info.
+ * @param {TraceablePeerConnection.attachTrack} tpc
+ * @return {boolean} <tt>true</tt> if this tracks is currently attached to given
+ * peer connection or <tt>false</tt> otherwise.
+ */
+JitsiLocalTrack.prototype._isAttachedToPC = function(tpc) {
```
paweldomas: none of these things are really public, the problem is that all the fields in all of the classes are like that currently. It will take too much time to fix everything now

JitsiConference.js
```diff
@@ -91,14 +106,57 @@ function JitsiConference(options) {
...
+
+    /* P2P related fields below: */
+
+    /**
+     * Stores reference to deferred start P2P task. It's created when 3rd
+     * participant leaves the room in order to avoid ping pong effect (it
+     * could be just a page reload).
+     * @type {number|null}
+     */
+    this.deferredStartP2P = null;
```
paweldomas: None of the fields of any class are really public. It's just the public/private doc (and _ ) were not adopted yet and I don't want to make confusion here with one or two fields. All of them should be converted at once.

JitsiConference.js
paweldomas: renamed to "maybe start or stop"

JitsiConference.js
paweldomas: no, it starts only the so called "remote stats" which are used for audio levels and resolutions

modules/xmpp/JingleSessionPC.js
paweldomas: I've tried, but it's easier said than done :P Probably promises could help. I have left a FIXME note

modules/xmpp/JingleSessionPC.js
paweldomas: I will add that to the "to be fixed" list

JitsiConference.js
```diff
@@ -91,14 +107,57 @@ function JitsiConference(options) {
...
+
+    /**
+     * A JingleSession for the direct peer to peer connection.
+     * @type {JingleSessionPC}
+     */
+    this.p2pJingleSession = null;
 }
 
+// FIXME convert JitsiConference to ES6 - ASAP !
+JitsiConference.prototype.constructor = JitsiConference;
```
paweldomas: it's not needed today, but I don't see any problem with leaving it as is

JitsiConference.js
```diff
@@ -113,6 +172,21 @@ JitsiConference.prototype._init = function(options = {}) {
...
+        = this._onIceConnectionInterrupted.bind(this);
+    this.room.addListener(
+        XMPPEvents.CONNECTION_INTERRUPTED, this._onIceConnectionInterrupted);
+
+    this._onIceConnectionRestored = this._onIceConnectionRestored.bind(this);
+    this.room.addListener(
+        XMPPEvents.CONNECTION_RESTORED, this._onIceConnectionRestored);
+
+    this._onIceConnectionEstablished
+        = this._onP2PConnectionEstablished.bind(this);
```
paweldomas: will fix that

modules/RTC/RTC.js
```diff
@@ -253,34 +254,41 @@ export default class RTC extends Listenable {
...
      * defined in the WebRTC specification.
+     * @param {boolean} isP2P indicates whether or not the new TPC will be used
+     * in a peer to peer type of session
      * @param {Object} options the config options
      * @param {boolean} options.disableSimulcast if set to 'true' will disable
      * the simulcast
      * @param {boolean} options.disableRtx if set to 'true' will disable the RTX
      * @param {boolean} options.preferH264 if set to 'true' H264 will be
      * preferred over other video codecs.
+     * @param {boolean} options.disableRtx <tt>true</tt> to disable RTX
```
paweldomas: thanks ! I must have messed that up during rebase

modules/RTC/RTC.js
```diff
@@ -15,6 +14,8 @@ import VideoType from '../../service/RTC/VideoType';
 
 const logger = getLogger(__filename);
 
+let rtcTrackIdCounter = 0;
```
paweldomas: it is used from static method of RTC, is it possible to make this var static ? I don't know how to make it better

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -0,0 +1,107 @@
...
+    /**
+     * Sets the <tt>ChatRoom</tt> instance used and binds presence listeners.
+     * @param {ChatRoom} room
+     */
+    setChatRoom(room) {
+        const oldChatRoom = this.chatRoom;
+
+        this.chatRoom = room;
+        if (oldChatRoom) {
+            // FIXME ChatRoom removes all listeners, will not be capable of
```
paweldomas: that was fixed - removing it

modules/xmpp/strophe.jingle.js
```diff
@@ -128,15 +138,28 @@ class JingleConnectionPlugin extends ConnectionPlugin {
...
                         $(iq).find('jingle').attr('sid'),
                         $(iq).attr('to'),
                         fromJid,
                         this.connection,
                         this.mediaConstraints,
-                        this.iceConfig, this.xmpp.options);
+
+                        // Only P2P makes use of the ICE config
+                        isP2P ? this.p2pIceConfig : this.jvbIceConfig,
+                        isP2P /* P2P */, false /* initiator */,
```
paweldomas: done this time !

modules/RTC/TraceablePeerConnection.js
```diff
@@ -631,7 +990,14 @@ const getters = {
                 dumpSDP(desc));
         }
 
-        return desc;
+        if (RTCBrowserType.doesVideoMuteByStreamRemove()) {
+            this.localSdpMunger.maybeMungeLocalSdp(desc);
+            logger.debug(
+                'getLocalDescription::postTransform '
+                    + '(munge local SDP)', desc);
```
paweldomas: yes, I have moved it to single line though

modules/RTC/TraceablePeerConnection.js
```diff
@@ -658,12 +1024,45 @@ Object.keys(getters).forEach(prop => {
...
+    this.localTracks.set(rtcId, track);
+    track._addPeerConnection(this);
+
+    const webrtcStream = track.getOriginalStream();
+
+    if (webrtcStream) {
+        this._addStream(webrtcStream);
+
+    // It's not ok for a track to not have a WebRTC stream if:
+    } else if (!RTCBrowserType.doesVideoMuteByStreamRemove()
```
paweldomas: no, it will pass only if it has no stream

modules/RTC/TraceablePeerConnection.js
```diff
@@ -689,12 +1088,294 @@ TraceablePeerConnection.prototype.addStream = function(stream, ssrcInfo) {
...
+        logger.info(
+            `Removing ${localTrack} as mute from ${this}`);
+        this._removeStream(webRtcStream);
+
+        return true;
+    }
+
+    logger.error(`removeStreamMute - no WebRTC stream for ${localTrack}`);
+
+    return false;
```
paweldomas: Probably the normal one should also return a boolean. It optimises the case where no change to the peer connection was made, so the SRD/sLD cycle is not necessary. The method will probably go away, because Brian suggested to use detach and I think it makes sense. I'm adding a FIXME for now.

paweldomas: I mean that I will clean that up when it comes to those changes. It's not a big deal right now, because normal addTrack and removeTrack is called before there the conference starts, so there is no sRD/sLD cycle anyway.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -705,89 +1386,106 @@ TraceablePeerConnection.prototype.createDataChannel = function(label, opts) {
...
-                this.trace('setRemoteDescriptionOnFailure', err);
-                this.eventEmitter.emit(
-                    RTCEvents.SET_REMOTE_DESCRIPTION_FAILED,
-                    err,
-                    this.peerconnection);
-                failureCallback(err);
-            });
-    };
+    /*
+     if (this.statsinterval === null && this.maxstats > 0) {
```
paweldomas: removed

JitsiConference.js
bbaldino: what additional safety checks?  here it would pass `this.jvbJingleSession.remoteTracks` (or whatever the member is), no need to check for anything.  and in the called function it would check `if (remoteTracks)` (which i think it probably does anyway?)

JitsiConference.js
bbaldino: sure, if you think it doesn't make things more confusing to omit `Session`.  i'm in favor of making variables descriptive over keeping them short

modules/RTC/TraceablePeerConnection.js
bbaldino: i thought the leading "_" signified that a method was intended to be 'private', and shouldn't be called from outside the class?

modules/RTC/TraceablePeerConnection.js
bbaldino: ok, yeah sounds like it would be good to either augment or replace sdpconsistency with this logic that can handle more.

modules/xmpp/JingleSessionPC.js
bbaldino: not sure i'm following this.  if we need to wait, then why not use the existing webrtc mechanics?  if we want to send all ice candidates at once (instead of trickle), usually you just set the 'ice gathering finished' handler and, once that's done, re-create the offer so it includes all the candidates.  is that what you mean?

modules/xmpp/RtxModifier.js
bbaldino: but by reading `modifyRtxSsrcs2` it's not clear how that method differs from `modifyRtxSsrcs`, which i think is weird.  js doesn't support method overloading, right?  it'd actually be clearer if they could have the same name i think.  maybe this should be 'modifyVideoMLineRtxSsrcs` or something.  or, this could be called `modifyRtxSsrcs` and the original one could be renamed.

JitsiConference.js
```diff
@@ -142,10 +148,11 @@ function JitsiConference(options) {
 
     /**
      * Flag set to <tt>true</tt> when P2P session has been established
-     * (ICE has been connected).
+     * (ICE has been connected) and this conference is currently in the peer to
+     * peer mode (P2P connection is the active one).
      * @type {boolean}
      */
-    this.p2pEstablished = false;
+    this.p2p = false;
```
bbaldino: i'm a little confused that this represents both whether the p2p connection is established (from an ice perspective) AND that the conference is in p2p mode.  there will be periods of time where that is not necessarily true, right? (when transitioning to/from p2p, there will be a period of time where the p2p ice connection is made, but we're not currently using it?)

JitsiConference.js
```diff
@@ -1012,7 +1019,7 @@ JitsiConference.prototype.onMemberJoined = function(jid, nick, role, isHidden) {
         },
         error => logger.error(`Failed to discover features of ${jid}`, error));
 
-    this._startStopP2PSession();
+    this._maybeStartOrStopP2P();
```
bbaldino: 👍

JitsiConference.js
```diff
@@ -2072,12 +2076,12 @@ JitsiConference.prototype._removeRemoteTracks
  * @private
  */
 JitsiConference.prototype._setP2PStatus = function(newStatus) {
-    if (this.p2pEstablished === newStatus) {
+    if (this.p2p === newStatus) {
```
bbaldino: some places we use the getter and some places we access this field directly

JitsiConference.js
```diff
@@ -2166,7 +2170,7 @@ JitsiConference.prototype._startStopP2PSession = function(userLeftEvent) {
             }, peerCount: ${peerCount} => ${shouldBeInP2P}`);
 
     // Clear deferred "start P2P" task
-    if (!shouldBeInP2P && this.deferredStartP2P) {
+    if (!shouldBeInP2P && this.deferredStartP2PTask) {
```
bbaldino: `_maybeClearDeferredStartP2P` already checks if `this.deferredStartP2PTask` is set, right? that's why it has the 'maybe'?  so we shouldn't need to check it here.

modules/xmpp/SDPUtil.js
```diff
@@ -547,6 +547,20 @@ const SDPUtil = {
...
+     * Extracts the ICE username fragment from an SDP string.
+     * @param {string} sdp the SDP in raw text format
+     */
+    getUfrag(sdp) {
+        const ufragLines
+            = sdp.split('\n').filter(
+                    line => line.startsWith('a=ice-ufrag:'));
+
+        if (ufragLines.length > 0) {
+            return ufragLines[0].substr('a=ice-ufrag:'.length);
```
bbaldino: given a line, there's already a function to parse the ufrag in sdputils: `parseICEUfrag`

JitsiConference.js
paweldomas: So the only purpose of this field is to say "now we're using P2P", but that can happen only once the ICE for the P2P connection has been established (reached "connected" for the first time). It doesn't work in the opposite direction tough - when it's true there's no guarantee that the P2P ICE is currently in "connected" state. It can be in "disconnected" for a moment when there is some network disruption. We won't switch back to JVB until P2P ice goes to FAILED. This flag should never be confused with ICE state.

paweldomas: Is the description of this field wrong ? For me it looks good, but if you have suggestions please give something specific how it can be improved.

JitsiConference.js
paweldomas: no, it should be only this method that accesses it directly and the getter for this state. But for me it's fine to access it here directly (it's a setter for this field).

JitsiConference.js
bbaldino: ok, i was thrown off by the explicit mention of the ice state in the comment, viewing it as 2 separate conditions.  maybe then it's sufficient to just say that the flag represents whether or not the p2p session is the one actively being used by this conference?

JitsiConference.js
paweldomas: there is a check if "this.jvbJingleSession" is defined which print a log in case it's not:
https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L1950

Anyway I don't really understand what's wrong with passing JingleSessionPC instance there and how passing remote tracks makes it any better.

PR #415, opened 2017-03-17
Remove references to undefined dispose method

dwilson6:
The xmpp kicked event wasn't being emitted because of an undefined method being referenced.


damencho:
Jenkins, test this please.

PR #416, opened 2017-03-17
On electron, call into JitsiMeet to open a desktop picker

virtuacoplenny:


modules/RTC/ScreenObtainer.js
```diff
@@ -116,13 +116,24 @@ const ScreenObtainer = {
                     });
             };
         } else if (RTCBrowserType.isElectron()) {
-            obtainDesktopStream = (_, onSuccess, onFailure) =>
-                window.JitsiMeetElectron.obtainDesktopStream(
+            obtainDesktopStream = (_, onSuccess, onFailure) => {
+
```
saghul: extra newline

modules/RTC/ScreenObtainer.js
```diff
@@ -116,13 +116,24 @@ const ScreenObtainer = {
...
+                let obtainStreamId
+                    = window.JitsiMeetElectron.obtainDesktopStream;
```
saghul: let's drop the deprecated way of doing this and just use the new picker always.

virtuacoplenny: I've removed the call to obtainDesktopStream and I added a couple electron specific error messages.

PR #419, opened 2017-03-21
Last n event change

damencho:
Changes lastN event params to be leaving and entering endpoint IDs.


JitsiConferenceEvents.js
```diff
@@ -84,17 +84,17 @@ export const DTMF_SUPPORT_CHANGED = 'conference.dtmfSupportChanged';
...
-
-/**
  * You are kicked from the conference.
  */
 export const KICKED = 'conferenece.kicked';
 
 /**
  * The Last N set is changed.
+ *
+ * @param {Array<string>|null} leavingEndpointIds the ids of all the endpoints
```
hristoterezov: We should update the arguments of this event in API.md file.

modules/RTC/RTC.js
```diff
@@ -168,9 +169,23 @@ export default class RTC extends Listenable {
                 this._dataChannelOpenListener);
 
             // Add Last N change listener.
-            this._lastNChangeListener = lastNEndpoints => {
-                this._lastNEndpoints = lastNEndpoints;
-            };
+            this._lastNChangeListener
+                = (lastNEndpoints, enteringLastNEndpoints) => {
```
hristoterezov: Correct me if I'm wrong but I was thinking the goal is to remove `enteringLastNEndpoints` completely (from the lib and from the bridge). 

If this is the case this implementation should not rely on `enteringLastNEndpoints` argument and should construct entering and leaving from the old and last sets of lastN participants.

Also don't you think it's better to move this function definition as a method or function?

PR #420, opened 2017-03-22
Adds a config parameter to set the m-line direction.

gpolitis:


saghul:
@gpolitis Can you elaborate about why this is needed? I can see what it does, but I don't understand why :-)

gpolitis:
yeah, some context is missing here ;) so I wanted to be able to launch chrome instances (from the command line) that are send only for testing purposes (so I needed a config option that can be set from the URL).

saghul:
IMHO this should use a private a more descriptive name, such as "_sdpAnswerDirection" + some documentation in config.js.

gpolitis:
@saghul  I can document the parameter in config.js; however, I'm not sure I understand what you would like me to change. Would you like me to rename '_setAnswerDirection` to `_sdpAnswerDirection`?

saghul:
No, I mean the option name itself. "direction" is not very descriptive... :-)

modules/RTC/TraceablePeerConnection.js
```diff
@@ -945,6 +967,11 @@ TraceablePeerConnection.prototype.createAnswer
                 _fixAnswerRFC4145Setup(remoteDescription, localDescription);
                 answer.sdp = localDescription.raw;
 
+                if (answer && this.options._sdpAnswerDirection) {
```
saghul: you don't need to check for `answer` it will be there or the live above would have thrown

saghul:
Also, you'll need to rebase, now that P2P landed.

gpolitis:
As explained in https://github.com/jitsi/jicofo/pull/181 this breaks stuff: If a participant joins video muted, then the m-line direction is set to recvonly and the colibri channel in the bridge gets configured as sendonly and it's not getting updated.

PR #427, opened 2017-03-27
fix(lastN): handle first lastN message correctly

saghul:
The first time a lastN was received it was ignored because the old lastN was
null, handle that case properly.

Furthermore, use default function parameters to avoid a conditional.


saghul:
@damencho PTAL, since you were the last one to touch these parts of the code. Thanks!

PR #428, opened 2017-03-27
Introduces new config ignoreStartMuted

damencho:
Ignores start audio/video muted requests received by jicofo.

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -588,6 +588,10 @@ JitsiConferenceEventManager.prototype.setupXMPPListeners = function() {
 
     conference.xmpp.addListener(XMPPEvents.START_MUTED_FROM_FOCUS,
         (audioMuted, videoMuted) => {
+            if (conference.options.config.ignoreStartMuted) {
+                return;
+            }
+
             conference.startAudioMuted = audioMuted;
             conference.startVideoMuted = videoMuted;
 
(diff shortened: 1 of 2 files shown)


saghul:
Can we call this something more generic like "role" perhaps? And have it be an enum of sorts, like "sip_gateway" or maybe something else in the future...

damencho:
@saghul This was added similar to the thing we have in jitsi-meet. We have there iAmRecorder config parameter. So this params are added as url params #config.iAmSipGateway and we start the conference with both settings enabled. I'm not sure how this alines with your comment, how we will pass an array in url params. Till now we were only adding particular configs as those above one by one.

bgrozev:
Non blocking: I'd rather have specific options configured by the different roles (i.e. have the sip gateway set a "ignoreStartMuted" option) instead. Otherwise we might up with SIP-gateway specific logic in different parts of the code.

damencho:
Just changed the option to ignoreStartMuted.

JitsiConferenceEventManager.js
```diff
@@ -588,6 +588,11 @@ JitsiConferenceEventManager.prototype.setupXMPPListeners = function() {
 
     conference.xmpp.addListener(XMPPEvents.START_MUTED_FROM_FOCUS,
         (audioMuted, videoMuted) => {
+            if (conference.options.config.ignoreStartMuted) {
+
```
saghul: extra newline

saghul:
I see where @bgrozev comes from. My line of thought is that a role can potentially require more than one setting so setting the role to "recorder" or "sip_gateway" would do the needful. If we ever needed more than one role we could use a an array and JSON-encode-then-percent-encode it and that works. Or just take a comma separated value string.

At any rate, no strong opinions, so I approved the PR with a minor nit.

JitsiConferenceEventManager.js
damencho: Ups, too protective against lint complaining :) Fixed.

PR #429, opened 2017-03-28
Cleanup around doRenegotiate

paweldomas:
work in progress.... just want to run the jenkins tests

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -22,6 +22,148 @@ const logger = getLogger(__filename);
 const IQ_TIMEOUT = 10000;
 
 /**
+ * The task which is executed on {@link JingleSessionPC.modificationQueue}
+ * which synchronizes modifications to the underlying
+ * {@link TraceablePeerConnection} state. All work that could change how either
+ * local or remote description will look should be executed through that queue.
+ *
+ * A task instance should be queued with {@link JingleSessionPC._doRenegotiate}
+ * which wraps common checks around renegotiation process (sRD/sLD cycle) and
+ * makes sure to send 'source-add/'source-remove' notifications in order to
+ * broadcast local stream description to all conference participants
+ * (through Jicofo).
+ *
+ * Class takes as an argument a worker function which should modify the
+ * instance state in a way which would reflect the result of the job performed.
+ * The instance is passed as an argument to the worker function. Once the task
+ * was executed the {@link ModificationQueueTask} state will be examined by
+ * taking into account the following:
+ *
+ * 1. If a worker function encounters critical error which should prevent from
+ * triggering renegotiation it should call
+ * {@link ModificationQueueTask.setError} and pass meaningful error description
+ * as a <tt>string</tt>. Further processing will stop at this point.
+ * 2. If a worker function succeeds and wants to provide new remote description
+ * for the renegotiation it should call
+ * {@link ModificationQueueTask.setRemoteDescription} and pass it as
+ * an argument. In case remote description is provided step 3 is omitted in
+ * the processing.
+ * 3. If a worker function finishes it's job it should set
+ * {@link ModificationQueueTask.setShouldRenegotiate} to <tt>true</tt> if any
+ * changes requiring renegotiation were done or <tt>false</tt> if there's no
+ * need to renegotiate.
+ *
+ * Given that the renegotiation will happen the local description from before
+ * and after the task execution are compared and 'source-remove'/'source-add'
+ * notifications will be triggered accordingly.
+ */
+class ModificationQueueTask {
+
+    /**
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -22,6 +22,148 @@ const logger = getLogger(__filename);
...
+ * local or remote description will look should be executed through that queue.
+ *
+ * A task instance should be queued with {@link JingleSessionPC._doRenegotiate}
+ * which wraps common checks around renegotiation process (sRD/sLD cycle) and
+ * makes sure to send 'source-add/'source-remove' notifications in order to
+ * broadcast local stream description to all conference participants
+ * (through Jicofo).
+ *
+ * Class takes as an argument a worker function which should modify the
+ * instance state in a way which would reflect the result of the job performed.
```
bbaldino: could you clarify what instance is being referred to here (and in the following sentence)?

modules/xmpp/JingleSessionPC.js
```diff
@@ -22,6 +22,148 @@ const logger = getLogger(__filename);
...
+ * 1. If a worker function encounters critical error which should prevent from
+ * triggering renegotiation it should call
+ * {@link ModificationQueueTask.setError} and pass meaningful error description
+ * as a <tt>string</tt>. Further processing will stop at this point.
+ * 2. If a worker function succeeds and wants to provide new remote description
+ * for the renegotiation it should call
+ * {@link ModificationQueueTask.setRemoteDescription} and pass it as
+ * an argument. In case remote description is provided step 3 is omitted in
+ * the processing.
+ * 3. If a worker function finishes it's job it should set
```
bbaldino: typo: `it's` -> `its`

modules/xmpp/JingleSessionPC.js
```diff
@@ -22,6 +22,148 @@ const logger = getLogger(__filename);
...
+    }
+
+    /**
+     * Indicates whether or not the renegotiation should follow worker function
+     * execution. This value is examined ony after both {@link getError} and
+     * {@link getRemoteDescription} return <tt>null</tt> values.
+     * @return {boolean}
+     */
+    shouldRenegotiate() {
+        return this._renegotiate;
```
bbaldino: shouldn't this do `return this._error === null && (this._remoteDescription !== null || this._renegotiate)` ?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1457,19 +1598,19 @@ export default class JingleSessionPC extends JingleSession {
...
+                thisTask.setError(
+                    'Error: tried adding stream with no active peerconnection');
             }
+        });
 
-            return true;
-        };
-
-        return this._doRenegotiate('addStreams', workFunction);
+        return this._doRenegotiate('addStreams', queueTask);
```
bbaldino: it's a bit confusing to me to call this method `_doRenegotiate` when we then have a separate variable (set inside the worker function) that is tested inside of `_doRenegotiate` saying whether or not we should renegotiate.  maybe it was misleading before, too; i can't remember.

modules/xmpp/JingleSessionPC.js
bbaldino: if you do this i think you could simplify your test logic in the actual task execution method

modules/xmpp/JingleSessionPC.js
```diff
@@ -1508,61 +1650,48 @@ export default class JingleSessionPC extends JingleSession {
...
         const workFunctionWrap = finishedCallback => {
             // Remember localSDP from before any modifications are done, by
             // the worker function
             let oldSdp = this.peerconnection.localDescription.sdp;
-            let remoteSdp = null;
 
-            let modifySources = workFunction();
+            modificationQueueTask.execute();
+
+            const taskError = modificationQueueTask.getError();
```
bbaldino: if you make the change above, this could be:
```
  if (modificationQueueTask.shouldRenegotiate) {
    // renegotiation logic
  } else if (modificationQueueTask.getError()) {
    // handle error
  }
```
which is maybe cleaner?  i dunno.  the multi-step process of calling `shouldRenegotiate` and then `getRemoteDescription` seems a little weird

modules/xmpp/JingleSessionPC.js
```diff
@@ -1508,61 +1650,48 @@ export default class JingleSessionPC extends JingleSession {
...
-                finishedCallback(modifySources);
+            if (taskError) {
+                finishedCallback(taskError);
 
                 return;
-            } else if (typeof modifySources === 'object') {
-                remoteSdp = modifySources;
+            }
+
+            let modifySources = modificationQueueTask.shouldRenegotiate();
```
bbaldino: the name of this variable seems a bit odd

paweldomas:
https://github.com/jitsi/lib-jitsi-meet/pull/436

PR #430, opened 2017-03-28
Adds peer connection statuses (active/inactive/interrupted/restoring).

damencho:
The peer connection used to be check with is connection active. Adding new states like active, inactive, when the connection was intentionally stopped by the bridge, interrupted due to network problem and restoring, which is the state of a peer which was inactive as being out of lastN and is now entering lastM and eventually will become active.

### DIFF
--- a/JitsiParticipant.js
+++ b/JitsiParticipant.js
@@ -1,5 +1,7 @@
 /* global Strophe */
 import * as JitsiConferenceEvents from './JitsiConferenceEvents';
+import { ParticipantConnectionStatus }
+    from './modules/connectivity/ParticipantConnectionStatus';
 import * as MediaType from './service/RTC/MediaType';
 
 /**
@@ -33,7 +35,7 @@ export default class JitsiParticipant {
             video: undefined
         };
         this._hidden = hidden;
-        this._connectionStatus = null;
+        this._connectionStatus = ParticipantConnectionStatus.ACTIVE;
         this._properties = {};
     }
 
--- a/JitsiParticipant.js
+++ b/JitsiParticipant.js
@@ -33,7 +33,7 @@ export default class JitsiParticipant {
             video: undefined
         };
         this._hidden = hidden;
-        this._isConnectionActive = true;
+        this._connectionStatus = null;
         this._properties = {};
     }
 
@@ -72,22 +72,23 @@ export default class JitsiParticipant {
 
     /**
      * Updates participant's connection status.
-     * @param {boolean} isActive true if the user's connection is fine or false
-     * when the user is having connectivity issues.
+     * @param {string} state the current participant connection state.
+     * {@link ParticipantConnectionStatus}.
      * @private
      */
-    _setIsConnectionActive(isActive) {
-        this._isConnectionActive = isActive;
+    _setConnectionStatus(status) {
+        this._connectionStatus = status;
     }
 
     /**
-     * Checks participant's connectivity status.
+     * Return participant's connectivity status.
      *
-     * @returns {boolean} true if the connection is currently ok or false when
-     * the user is having connectivity issues.
+     * @returns {string} the connection status
+     * <tt>ParticipantConnectionStatus</tt> of the user.
+     * {@link ParticipantConnectionStatus}.
      */
-    isConnectionActive() {
-        return this._isConnectionActive;
+    getConnectionStatus() {
+        return this._connectionStatus;
     }
 
     /**
--- a/modules/connectivity/ParticipantConnectionStatus.js
+++ b/modules/connectivity/ParticipantConnectionStatus.js
@@ -18,6 +18,16 @@ const logger = getLogger(__filename);
 const DEFAULT_RTC_MUTE_TIMEOUT = 2000;
 
 /**
+ * The time to wait a track to be restored. Track which was out of lastN
+ * should be inactive and when entering lastN it becomes restoring and when
+ * data is received from bridge it will become active, but if no data is
+ * received for some time we set status of that participant connection to
+ * interrupted.
+ * @type {number}
+ */
+const DEFAULT_RESTORING_TIMEOUT = 5000;
+
+/**
  * Participant connection statuses.
  *
  * @type {{
@@ -121,6 +131,26 @@ export default class ParticipantConnectionStatusHandler {
          */
         this.rtcMutedTimestamp = { };
         logger.info(`RtcMuteTimeout set to: ${this.rtcMuteTimeout}`);
+
+        /**
+         * This map holds the timestamps indicating when participant's video
+         * entered lastN set. Participants entering lastN will have connection
+         * status restoring and when we start receiving video will become
+         * active, but if video is not received for certain time
+         * {@link DEFAULT_RESTORING_TIMEOUT} that participant connection status
+         * will become interrupted.
+         *
+         * @type {Object.<string, number>}
+         */
+        this.enteredLastNTimestamp = {};
+
+        /**
+         * A map of the "endpoint ID"(which corresponds to the resource part
+         * of MUC JID(nickname)) to the restoring timeout callback IDs
+         * scheduled using window.setTimeout.
+         * @type {Object.<string, number>}
+         */
+        this.restoringTimers = [];
     }
 
     /**
@@ -429,20 +459,24 @@ export default class ParticipantConnectionStatusHandler {
         if (isConnActiveByJvb) {
             if (isInLastN) {
                 if (isVideoTrackFrozen) {
-                    // if participant was inactive, and is in lastN,
-                    // it is restoring
-                    const isInactive = participant.getConnectionStatus()
-                        === ParticipantConnectionStatus.INACTIVE;
-
-                    newState = isInactive
-                        ? ParticipantConnectionStatus.RESTORING
-                            : ParticipantConnectionStatus.INTERRUPTED;
+                    newState = this._isRestoringTimedout(id)
+                        ? ParticipantConnectionStatus.INTERRUPTED
+                            : ParticipantConnectionStatus.RESTORING;
                 } else {
                     newState = ParticipantConnectionStatus.ACTIVE;
                 }
             }
-        } else if (isInLastN && isVideoTrackFrozen) {
+        } else {
+            // when there is a connection problem signaled from jvb
+            // it means no media was flowing for at least 15secs, so everything
+            // should be interrupted, when in p2p mode we will never end up here
             newState = ParticipantConnectionStatus.INTERRUPTED;
+        }
+
+        // if the new state is not restoring clear timers and timestamps
+        // that we use to track the restoring state
+        if (newState !== ParticipantConnectionStatus.RESTORING) {
+            this._clearRestoringTimer(id);
         }
 
         logger.debug(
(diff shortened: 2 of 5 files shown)


modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +424,53 @@ export default class ParticipantConnectionStatus {
             isConnActiveByJvb = true;
         }
 
-        const isConnectionActive
-            = isConnActiveByJvb
-                && (isVideoMuted || (isInLastN && !isVideoTrackFrozen));
+        let newState = ParticipantConnectionStatus.INACTIVE;
+
+        if (isConnActiveByJvb) {
+            if (isInLastN) {
```
paweldomas: if the person is not in last N it will immediately be marked as INACTIVE ? I'd say it's not safe to assume connectivity issues this way. We may be showing "connectivity issues" for brief moments for good video if the lastN msg will not arrive in time.

paweldomas: last N should be only an excuse to not going to INACTIVE if video is reported as frozen

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +424,53 @@ export default class ParticipantConnectionStatus {
...
+                if (isVideoTrackFrozen) {
+                    // if participant was inactive, and is in lastN,
+                    // it is restoring
+                    const isInactive = participant.getConnectionStatus()
+                        === ParticipantConnectionStatus.INACTIVE;
+
+                    newState = isInactive
```
paweldomas: Previously the method was refactored to calculate output status for given "input" values when called at any point of time. Like a function of arguments. Right now this introduces state machine where it depends on the previous state (unless I'm reading it wrong). We want to avoid that, because it makes things much more complex.

With this code if 'figureOutConnectionStatus' is called soon after it enters RESTORING it will go to INTERRUPTED, but we don't know where is this update coming from (it could be P2P status change).

I'd say this should be changed to store the time when user entered last N and if the timestamp is too old (it's taking too long to go to active) then it means it's interrupted, but before that if the video is frozen it means it's RESTORING. This logic could be encapsulated into some utility method "isParticipantRestoring" or whatever you find suitable.

Btw. probably lastN makes no sense when in P2P, so that should be taken into account as well.

modules/connectivity/ParticipantConnectionStatus.js
damencho: Can you explain what you mean. When a participant is not in lastN it is inactive, and no video should be shown, the UI will show avatar and appropriate message. Inactive is not a connectivity issue.

paweldomas: oh yes, interrupted is connectivity issues

modules/connectivity/ParticipantConnectionStatus.js
damencho: The definition of RESTORING is being inactive and entering lastN, so I do not see a way of calculating this without checking previous state. 
I agree with your point about calling it twice and that it can stay in RESTORING in certain cases, will fix those.
I will check the p2p, this was written before p2p was merged. I was assuming that p2p will work as and when lastN is disabled, all participants will be in lastN, is this correct?

paweldomas: well restoring can be something like:
RESTORING = isInLastN & isVideoTrackFrozen & timeSinceEneteredLastN < 5000;
else INTERRUPTED
No ?
This is a simplification of course I need to think this throughly, but it is possible to figure out restoring without previous state (the timestamp is kind of tracking the previous state).

Also it's important that isVideoTrackFrozen can be true only on Chrome, so we'd have to think about that some more (not sure if that changes anything here).

paweldomas: About lastN and P2P: lastN should be ignored when in P2P, but I guess the lastN event will be triggered by the JVB.

JitsiParticipant.js
```diff
@@ -72,22 +72,23 @@ export default class JitsiParticipant {
...
-     * @returns {boolean} true if the connection is currently ok or false when
-     * the user is having connectivity issues.
+     * @returns {string} the connection status
+     * <tt>ParticipantConnectionStatus</tt> of the user.
+     * {@link ParticipantConnectionStatus}.
      */
-    isConnectionActive() {
-        return this._isConnectionActive;
+    getConnectionStatus() {
+        return this._connectionStatus;
```
paweldomas: It seems that this value can be null initially.

JitsiParticipant.js
```diff
@@ -33,7 +33,7 @@ export default class JitsiParticipant {
             video: undefined
         };
         this._hidden = hidden;
-        this._isConnectionActive = true;
+        this._connectionStatus = null;
```
paweldomas: We may want to start from ACTIVE until anything happens (lastN or any other event happens). WDYT ?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +454,117 @@ export default class ParticipantConnectionStatus {
...
+                 } => ${newState}`);
+
+        this._changeConnectionStatus(participant, newState);
+    }
+
+    /**
+     * On change in Last N set check all leaving and entering participants to
+     * change their corresponding statuses.
+     *
+     * @param leavingLastN
```
paweldomas: missing type declarations

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +454,117 @@ export default class ParticipantConnectionStatus {
...
+                this.figureOutConnectionStatus(id);
+            });
+        }
+    }
+
+    /**
+     * Clears the restoring timer for participant's video track and the
+     * timestamp for entering lastN.
+     *
+     * @param participantId the id of the conference participant which is
```
paweldomas: missing type declarations

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +454,117 @@ export default class ParticipantConnectionStatus {
...
+        }
+    }
+
+    /**
+     * Checks whether a track had stayed enough in restoring state, compares
+     * current time and the time the track entered in lastN. If it hasn't
+     * timedout and there is no timer added, add new timer in order to give it
+     * more time to become active or mark it as interrupted on next check.
+     *
+     * @param participantId the id of the conference participant which is
```
paweldomas: missing type declarations

modules/connectivity/ParticipantConnectionStatus.js
paweldomas: I still find it valid that it could happen that there will be some timing issues with "lastN" events and it may happen we display working video as inactive, but maybe let's see how it behaves in practice.

paweldomas:
Other than some missing type declarations and the initial status value it LGTM

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -18,10 +18,55 @@ const logger = getLogger(__filename);
...
+ *      INACTIVE: string,
+ *      INTERRUPTED: string,
+ *      RESTORING: string
+ * }}
+ */
+export const ParticipantConnectionStatus = {
+    /**
+     * Status indicating that connection is currently active.
+     */
+    ACTIVE: 'active',
```
saghul: Can we use Symbol instead? I find it better because if you ever make a typo it will cause a ReferenceError instead of a silent mismatch.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -86,6 +131,27 @@ export default class ParticipantConnectionStatus {
...
+         * This map holds the timestamps indicating when participant's video
+         * entered lastN set. Participants entering lastN will have connection
+         * status restoring and when we start receiving video will become
+         * active, but if video is not received for certain time
+         * {@link DEFAULT_RESTORING_TIMEOUT} that participant connection status
+         * will become interrupted.
+         *
+         * @type {Object.<string, number>}
+         */
+        this.enteredLastNTimestamp = {};
```
saghul: use a Map ?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -86,6 +131,27 @@ export default class ParticipantConnectionStatus {
...
+
+        /**
+         * A map of the "endpoint ID"(which corresponds to the resource part
+         * of MUC JID(nickname)) to the restoring timeout callback IDs
+         * scheduled using window.setTimeout.
+         *
+         * @type {Object.<string, number>}
+         */
+        this.restoringTimers = [];
```
saghul: You say it's a mapping but it's an array? use a Map maybe?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +455,117 @@ export default class ParticipantConnectionStatus {
...
+        }
 
         logger.debug(
-            `Figure out conn status, is video muted: ${isVideoMuted
+            `Figure out conn status for ${id}, is video muted: ${isVideoMuted
                  } is active(jvb): ${isConnActiveByJvb
                  } video track frozen: ${isVideoTrackFrozen
                  } is in last N: ${isInLastN
-                 } => ${isConnectionActive}`);
+                 } current2NewStatus: ${participant.getConnectionStatus()
```
saghul: connection status? current2Status reads a bit weird doesn't it?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +455,117 @@ export default class ParticipantConnectionStatus {
...
+    /**
+     * On change in Last N set check all leaving and entering participants to
+     * change their corresponding statuses.
+     *
+     * @param {Array<string>} leavingLastN array of ids leaving lastN.
+     * @param {Array<string>} enteringLastN array of ids entering lastN.
+     * @private
+     */
+    _onLastNChanged(leavingLastN, enteringLastN) {
+        if (leavingLastN) {
```
saghul: use default parameters and a for..of loop and you'll save a level of indentation here

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +455,117 @@ export default class ParticipantConnectionStatus {
...
+     * @returns {boolean} <tt>true</tt> if the track was in restoring state
+     * more than the timeout ({@link DEFAULT_RESTORING_TIMEOUT}.) in order to
+     * set its status to interrupted.
+     * @private
+     */
+    _isRestoringTimedout(participantId) {
+        const enteredLastNTimestamp
+            = this.enteredLastNTimestamp[participantId];
+
+        if (typeof enteredLastNTimestamp === 'number'
```
saghul: since you are setting this it can only be a number or undefined, right? Then I'd check if it's not undefined.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -380,18 +455,117 @@ export default class ParticipantConnectionStatus {
...
+            >= DEFAULT_RESTORING_TIMEOUT) {
+            return true;
+        }
+
+        // still haven't reached timeout, if there is no timer scheduled,
+        // schedule one so we can track the restoring state and change it after
+        // reaching the timeout
+        if (!this.restoringTimers[participantId]) {
+            this.restoringTimers[participantId] = setTimeout(
+                this.figureOutConnectionStatus.bind(this, participantId),
```
saghul: use an arrow function to avoid the bind

modules/connectivity/ParticipantConnectionStatus.js
paweldomas: wow now that is interesting how it works, was this tested for real ?

paweldomas: or maybe it's possible to use Array as a Map I don't know :P

saghul: It does work, because arrays are objects:

```
> var x = []
undefined
> x['foo'] = 'bar'
'bar'
> x
[ foo: 'bar' ]
> typeof x
'object'
> Array.isArray(x)
true
>
```

modules/connectivity/ParticipantConnectionStatus.js
paweldomas: well, not sure if that makes sense here

saghul: that's one of the reasons why arrow functions exist, see the first paragraph here: https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Arrow_functions

If that doesn't convince you: MORE NEWER MORE BETTER :-P

paweldomas: so how would you write this arrow function here ?

paweldomas: > These function expressions are best suited for non-method functions
We're passing a class method here, no ?

paweldomas: but I agree with "more newer more better" lol :)

saghul: > We're passing a class method here, no ?

My reading is that the arrow function itself is better suited not to be a class method.

You would write it as follows:

```
this.restoringTimers[participantId] = setTimeout(() => {
      this.figureOutConnectionStatus(participantId);
}, DEFAULT_RESTORING_TIMEOUT);
```

paweldomas: I vote for bind in such use cases

paweldomas: but since it's done it's done

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -18,10 +18,55 @@ const logger = getLogger(__filename);
...
+ *      INACTIVE: string,
+ *      INTERRUPTED: string,
+ *      RESTORING: string
+ * }}
+ */
+export const ParticipantConnectionStatus = {
+    /**
+     * Status indicating that connection is currently active.
+     */
+    ACTIVE: Symbol('active'),
```
saghul: I hope you don't hate me... turns out pollyfilling Symbol globally for Jitsi Meet is a bit cumbersome at the moment. Can you go back to strings? Sorry! 😢

paweldomas:
@damencho it is helpful when PR which depends on this one is mentioned either here or in jitsi-meet - makes life much easier

damencho:
@paweldomas Note taken, sorry.

PR #431, opened 2017-03-29
Set the currentBrowser if Internet Explorer is detected

virtuacoplenny:
The setting of currentBrowser in RTCBrowserType was
accidentally removed.

### DIFF
--- a/modules/RTC/RTCBrowserType.js
+++ b/modules/RTC/RTCBrowserType.js
@@ -319,6 +319,11 @@ function detectIE() {
         version = parseInt(ua.substring(rv + 3, ua.indexOf('.', rv)), 10);
     }
 
+    if (version) {
+        currentBrowser = RTCBrowserType.RTC_BROWSER_IEXPLORER;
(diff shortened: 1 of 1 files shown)


modules/RTC/RTCBrowserType.js
```diff
@@ -319,6 +319,11 @@ function detectIE() {
         version = parseInt(ua.substring(rv + 3, ua.indexOf('.', rv)), 10);
     }
 
+    if (version) {
+        currentBrowser = RTCBrowserType.RTC_BROWSER_IEXPLORER;
```
paweldomas: LGTM, How is it even possible that it was not here before :) ?

saghul: My bad, I didn't catch this in the Edge review, sorry folks!

ibc:
My fault in fact. Sorry.

modules/RTC/RTCBrowserType.js
paweldomas: no worries !

PR #432, opened 2017-03-29
ref(JitsiConference): modify add/remove tracks

paweldomas:
Do not pass whole JingleSessionPC, but only the tracks to be added/removed.


bbaldino:
lgtm, should i merge?

PR #434, opened 2017-03-29
Delayed ice candidates

paweldomas:
Get rid of 'delayedIceCandidates' buffer.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1296,22 +1296,23 @@ JitsiConference.prototype.onIncomingCall
 
     // Add local tracks to the session
     try {
-        jingleSession.addLocalTracks(this.getLocalTracks()).then(() => {
-            jingleSession.acceptOffer(jingleOffer, null,
-                error => {
-                    GlobalOnErrorHandler.callErrorHandler(error);
-                    logger.error(
-                        'Failed to accept incoming Jingle session', error);
-                }
-            );
+        jingleSession.acceptOffer(
+            jingleOffer,
+            null /* success */,
+            error => {
+                GlobalOnErrorHandler.callErrorHandler(error);
+                logger.error(
+                    'Failed to accept incoming Jingle session', error);
+            },
+            this.getLocalTracks()
+        );
 
-            // Start callstats as soon as peerconnection is initialized,
-            // do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never
-            // happen in case if user doesn't have or denied permission to
-            // both camera and microphone.
-            this.statistics.startCallStats(jingleSession);
-            this._startRemoteStats();
-        });
+        // Start callstats as soon as peerconnection is initialized,
+        // do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never
+        // happen in case if user doesn't have or denied permission to
+        // both camera and microphone.
+        this.statistics.startCallStats(jingleSession);
+        this._startRemoteStats();
     } catch (e) {
         GlobalOnErrorHandler.callErrorHandler(e);
         logger.error(e);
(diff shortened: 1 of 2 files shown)


JitsiConference.js
```diff
@@ -1296,22 +1296,23 @@ JitsiConference.prototype.onIncomingCall
...
-            );
-
-            // Start callstats as soon as peerconnection is initialized,
-            // do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never
-            // happen in case if user doesn't have or denied permission to
-            // both camera and microphone.
-            this.statistics.startCallStats(jingleSession);
-            this._startRemoteStats();
-        });
+        jingleSession.acceptOffer(
```
bbaldino: just wondering, why is it we no longer add the local tracks here (and the similar one below)?

JitsiConference.js
```diff
@@ -2106,16 +2098,7 @@ JitsiConference.prototype._startP2PSession = function(peerJid) {
...
-    this.p2pJingleSession.addLocalTracks(localTracks).then(
-        () => {
-            logger.info(`Added ${localTracks} to P2P`);
-            logger.info('About to send P2P \'session-initiate\'...');
-            this.p2pJingleSession.invite();
-        },
-        error => {
-            logger.error(`Failed to add ${localTracks} to P2P`, error);
-        });
+    this.p2pJingleSession.invite(localTracks);
```
bbaldino: why change this from the previous method where we add the tracks first, and then do the invite?

paweldomas:
I tried to describe the need in FIXMEs and commit messages. Please check here:

> Refactor the current code to add local tracks together with initial
offer/answer to make this atomic/single operation on
the modificationQueue.

>This is required to be able to get rid of 'delayedIceCandidates' list
and to execute 'addIceCandidate' task on the queue. If local track
addition is not bundled with initial offer it will often happen that ICE
candidates are queued, before the offer/answer which is queued only once
the local tracks task is done.

Also addLocalTrack was library internal and it turns out there's no need to have them separate as we add them always on initial offer/answer

modules/xmpp/JingleSessionPC.js
```diff
@@ -695,9 +707,20 @@ export default class JingleSessionPC extends JingleSession {
      * @param success callback called when sRD/sLD cycle finishes successfully.
      * @param failure callback called with an error object as an argument if we
      *        fail at any point during setRD, createAnswer, setLD.
+     * @param {Array<JitsiLocalTrack>} [localTracks] the optional list of
+     * the local tracks that will be added, before the offer/answer cycle
+     * executes (for the local track addition to be an atomic operation together
+     * with the offer/answer).
```
bbaldino: i'm thinking we add here:
```
We allow the localTracks to optionally be passed in so that the addition of the local tracks and the processing of the initial offer can all be done atomically.  The existing addLocalTracks method queues up its own modificationQueue operation, meaning we can't call it AND process the initial offer as part of the same atomic operation.
```

modules/xmpp/JingleSessionPC.js
```diff
@@ -607,16 +609,25 @@ export default class JingleSessionPC extends JingleSession {
...
+    /* eslint-enable max-params */
+
     /**
      * Creates an offer and sends Jingle 'session-initiate' to the remote peer.
+     * @param {Array<JitsiLocalTrack>} localTracks the local tracks that will be
+     * added, before the offer/answer cycle executes (for the local track
+     * addition to be an atomic operation together with the offer/answer).
      */
-    invite() {
+    invite(localTracks) {
```
bbaldino: `doSessionInitiate`?

PR #435, opened 2017-03-29
New methods to check mutiple mic support and local stats collecting

virtuacoplenny:


JitsiMediaDevices.js
```diff
@@ -73,6 +74,13 @@ const JitsiMediaDevices = {
     },
 
     /**
+     * Returns true if it is possible to have multiple audio input streams.
```
saghul: This documentation string could use a bit of an improvement. I guess what we mean here is *simultaneous* input devices open, right?

JitsiMediaDevices.js
```diff
@@ -73,6 +74,13 @@ const JitsiMediaDevices = {
...
+     */
+    isMultipleAudioInputSupported() {
+        return !RTCBrowserType.isFirefox();
```
saghul: 😞

JitsiMediaDevices.js
virtuacoplenny: Yes, you are correct. I've updated the comment.

modules/statistics/LocalStatsCollector.js
```diff
@@ -139,4 +138,8 @@ LocalStatsCollector.prototype.stop = function() {
     }
 };
 
+LocalStatsCollector.isLocalStatsSupported = function() {
```
paweldomas: missing JSDoc here ?

PR #438, opened 2017-03-31
build: run webpack on postinstall

saghul:
The net effect is the same, with the bonus that it avoids problems when running
yarn.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,8 +49,8 @@
     "webpack": "1.14.0"
   },
   "scripts": {
-    "install": "webpack -p",
     "lint": "jshint . && eslint . && flow",
+    "postinstall": "webpack -p",
     "test": "karma start karma.conf.js",
     "validate": "npm ls"
   },


webintoit:
Guys, you should move all your "devDependencies" to "dependencies" to make it possible "webpack -p" works. 
When i try `npm install jitsi/lib-jitsi-meet` it loads only few dependencies and then crashes with "command webpack not found".

saghul:
Au contraire, I want to move everything to devDependencies, since this is a browser library. But please, let's keep issues on topic.

How is that related tho this PR? I did get the error you mention with yarn, but this PR fixes that. 🤷‍♂️

webintoit:
@saghul, it`s not a issue, i was never able  to add lib-jitsi-meet to package.json of my project. npm never installs  dev-dependencies of lib-jitsi-meet with normal flow. `webpack-dev-server`, `eslint` it is `DEV` dependencies, i think (project can be used without them), but if some of dependencies required for standard using, i think it  not dev. It just my opinion. My comment not really related to this PR, sorry.

PR #445, opened 2017-04-08
Report P2P to  CallStats

paweldomas:
The goal here is to have both JVB and P2P connection reported to CallStats.

Please check each commit msg for more info.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -220,6 +220,7 @@ JitsiConference.prototype._init = function(options = {}) {
                     || window.location.hostname,
             callStatsCustomScriptUrl:
                 this.options.config.callStatsCustomScriptUrl,
+            callStatsAliasName: this.myUserId(),
             roomName: this.options.name
         });
     }
@@ -273,6 +274,16 @@ JitsiConference.prototype.leave = function() {
         this.statistics.dispose();
     }
 
+    // Close both JVb and P2P JingleSessions
+    if (this.jvbJingleSession) {
+        this.jvbJingleSession.close();
+        this.jvbJingleSession = null;
+    }
+    if (this.p2pJingleSession) {
+        this.p2pJingleSession.close();
+        this.p2pJingleSession = null;
+    }
+
     // leave the conference
     if (this.room) {
         const room = this.room;
@@ -296,12 +307,6 @@ JitsiConference.prototype.leave = function() {
             // ChatRoom instance.
             this.getParticipants().forEach(
                 participant => this.onMemberLeft(participant.getJid()));
-
-            // Close the JingleSession
-            if (this.jvbJingleSession) {
-                this.jvbJingleSession.close();
-                this.jvbJingleSession = null;
-            }
         });
     }
 
@@ -1307,8 +1312,10 @@ JitsiConference.prototype.onIncomingCall
         // do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never
         // happen in case if user doesn't have or denied permission to
         // both camera and microphone.
+        logger.info('Starting CallStats for JVB connection...');
         this.statistics.startCallStats(
-            this.jvbJingleSession.peerconnection, this.myUserId());
+            this.jvbJingleSession.peerconnection,
+            'jitsi' /* Remote user ID for JVB is 'jitsi' */);
         this._startRemoteStats();
     } catch (e) {
         GlobalOnErrorHandler.callErrorHandler(e);
@@ -1399,7 +1406,9 @@ JitsiConference.prototype.onCallEnded
         // Stop the stats
         if (this.statistics) {
             this.statistics.stopRemoteStats();
-            this.statistics.stopCallStats();
+            logger.info('Stopping JVB CallStats');
+            this.statistics.stopCallStats(
+                this.jvbJingleSession.peerconnection);
         }
 
         // Current JVB JingleSession is no longer valid, so set it to null
@@ -1738,14 +1747,15 @@ JitsiConference.prototype.isCallstatsEnabled = function() {
 JitsiConference.prototype._onTrackAttach = function(track, container) {
     const isLocal = track.isLocal();
     let ssrc = null;
+    const isP2P = track.isP2P;
+    const remoteUserId = isP2P ? track.getParticipantId() : 'jitsi';
+    const peerConnection
+        = isP2P
+            ? this.p2pJingleSession && this.p2pJingleSession.peerconnection
+            : this.jvbJingleSession && this.jvbJingleSession.peerconnection;
 
     if (isLocal) {
-        // FIXME currently we do CallStats only for JVB, but the same logic will
-        // need to happen for the P2P as well once the support is added.
         // Local tracks have SSRC stored on per peer connection basis
-        const peerConnection
-            = this.jvbJingleSession && this.jvbJingleSession.peerconnection;
-
         if (peerConnection) {
             ssrc = peerConnection.getLocalSSRC(track);
         }
@@ -1755,8 +1765,14 @@ JitsiConference.prototype._onTrackAttach = function(track, container) {
     if (!container.id || !ssrc) {
         return;
     }
+
     this.statistics.associateStreamWithVideoTag(
-        ssrc, isLocal, track.getUsageLabel(), container.id);
+        peerConnection,
+        ssrc,
+        isLocal,
+        remoteUserId,
+        track.getUsageLabel(),
+        container.id);
 };
 
 /**
@@ -1878,6 +1894,11 @@ JitsiConference.prototype._acceptP2PIncomingCall
 
     this.p2pJingleSession.initialize(
         false /* initiator */, this.room, this.rtc);
+
+    logger.info('Starting CallStats for P2P connection...');
+    this.statistics.startCallStats(
+        this.p2pJingleSession.peerconnection,
+        Strophe.getResourceFromJid(this.p2pJingleSession.peerjid));
 
     const localTracks = this.getLocalTracks();
 
@@ -2090,6 +2111,11 @@ JitsiConference.prototype._startP2PSession = function(peerJid) {
 
     this.p2pJingleSession.initialize(true /* initiator */, this.room, this.rtc);
 
+    logger.info('Starting CallStats for P2P connection...');
+    this.statistics.startCallStats(
+        this.p2pJingleSession.peerconnection,
+        Strophe.getResourceFromJid(this.p2pJingleSession.peerjid));
+
     // NOTE one may consider to start P2P with the local tracks detached,
     // but no data will be sent until ICE succeeds anyway. And we switch
     // immediately once the P2P ICE connects.
@@ -2218,6 +2244,9 @@ JitsiConference.prototype._stopP2PSession
     // Stop P2P stats
     logger.info('Stopping remote stats with P2P connection');
     this.statistics.stopRemoteStats();
+    logger.info('Stopping CallStats for P2P connection');
+    this.statistics.stopCallStats(
+        this.p2pJingleSession.peerconnection);
 
     if (JingleSessionState.ENDED !== this.p2pJingleSession.state) {
         this.p2pJingleSession.terminate(
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1307,7 +1307,8 @@ JitsiConference.prototype.onIncomingCall
         // do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never
         // happen in case if user doesn't have or denied permission to
         // both camera and microphone.
-        this.statistics.startCallStats(jingleSession);
+        this.statistics.startCallStats(
+            this.jvbJingleSession.peerconnection, this.myUserId());
         this._startRemoteStats();
     } catch (e) {
         GlobalOnErrorHandler.callErrorHandler(e);
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2248,7 +2248,7 @@ JitsiConference.prototype._stopP2PSession
     }
 
     // Stop P2P stats
-    logger.info('Stopping remote stats with P2P connection');
+    logger.info('Stopping remote stats for P2P connection');
     this.statistics.stopRemoteStats();
     logger.info('Stopping CallStats for P2P connection');
     this.statistics.stopCallStats(
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2071,6 +2071,12 @@ JitsiConference.prototype._setP2PStatus = function(newStatus) {
         logger.info('Peer to peer connection closed!');
     }
 
+    // Put the JVB connection on hold/resume
+    if (this.jvbJingleSession) {
+        this.statistics.sendConnectionResumeOrHoldEvent(
+            this.jvbJingleSession.peerconnection, !newStatus);
+    }
+
     // Clear dtmfManager, so that it can be recreated with new connection
     this.dtmfManager = null;
 
--- a/modules/statistics/CallStats.js
+++ b/modules/statistics/CallStats.js
@@ -185,7 +185,6 @@ export default class CallStats {
 
         // there is no lib, nothing to report to
         if (err !== 'success') {
-
             return;
         }
 
@@ -206,7 +205,6 @@ export default class CallStats {
         }
 
         if (!atLeastOneFabric) {
-
             return;
         }
 
--- a/modules/statistics/CallStats.js
+++ b/modules/statistics/CallStats.js
@@ -273,7 +273,7 @@ export default class CallStats {
 
         /* eslint-disable max-params */
         theBackend.reportError
-        = function(pc, cs, type, ...otherArguments) {
+        = function(pc, cs, type, ...args) {
             // Logs from the logger are submitted on the applicationLog event
             // "type". Logging the arguments on the logger will create endless
             // loop, because it will put all the logs to the logger queue again.
@@ -284,11 +284,10 @@ export default class CallStats {
                 // by the logger implementation)
                 console && console.debug('reportError', pc, cs, type);
             } else {
-                logger.debug('reportError', pc, cs, type, ...otherArguments);
+                logger.debug('reportError', pc, cs, type, ...args);
             }
             try {
-                originalReportError.call(
-                    theBackend, pc, cs, type, ...otherArguments);
+                originalReportError.call(theBackend, pc, cs, type, ...args);
             } catch (exception) {
                 if (type === wrtcFuncNames.applicationLog) {
                     console && console.error('reportError', exception);
@@ -307,10 +306,10 @@ export default class CallStats {
      * @param {object} options
      * @param {String} options.callStatsID CallStats credentials - ID
      * @param {String} options.callStatsSecret CallStats credentials - secret
-     * @param {string} options.userName the <tt>userName</tt> part of
-     * the <tt>userID</tt> aka display name, see CallStats docs for more info.
      * @param {string} options.aliasName the <tt>aliasName</tt> part of
      * the <tt>userID</tt> aka endpoint ID, see CallStats docs for more info.
+     * @param {string} options.userName the <tt>userName</tt> part of
+     * the <tt>userID</tt> aka display name, see CallStats docs for more info.
      *
      */
     static initBackend(options) {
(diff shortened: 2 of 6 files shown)


modules/statistics/statistics.js
```diff
@@ -230,13 +232,31 @@ Statistics.prototype.stopRemoteStats = function() {
...
+ * @param {string} aliasName an alias name for the local endpoint which will be
+ * used for reporting the new CallStats session.
  */
-Statistics.prototype.startCallStats = function(session) {
+Statistics.prototype.startCallStats = function(tpc, aliasName) {
     if (this.callStatsIntegrationEnabled && !this.callStatsStarted) {
+        /* eslint-disable prefer-template */
+        // The confID is case sensitive!!!
+        const confID
+            = this.options.callStatsConfIDNamespace
```
saghul: I know you didn't add this, but can you check if it's truthy? On RN we get undefined there.

modules/statistics/CallStats.js
```diff
@@ -47,7 +47,7 @@ const fabricEvent = {
     activeDeviceList: 'activeDeviceList'
 };
 
-let callStats = null;
+let callStatsBackend = null;
```
saghul: ❤️ I would even go as far as calling it "backend", since the file is called CallStats.

modules/statistics/CallStats.js
```diff
@@ -262,22 +262,23 @@ CallStats.sendDominantSpeakerEvent = tryCatch(cs => {
...
  */
 CallStats.sendActiveDeviceListEvent = tryCatch((devicesData, cs) => {
-    CallStats._reportEvent.call(cs, fabricEvent.activeDeviceList, devicesData);
+    CallStats._reportEvent(cs, fabricEvent.activeDeviceList, devicesData);
 });
 
 /**
  * Reports an error to callstats.
  *
+ * @param {CallStats} cs
```
saghul: is this the backend or our wrapper?

JitsiConference.js
```diff
@@ -1307,8 +1312,10 @@ JitsiConference.prototype.onIncomingCall
         // do not wait for XMPPEvents.PEERCONNECTION_READY, as it may never
         // happen in case if user doesn't have or denied permission to
         // both camera and microphone.
+        logger.info('Starting CallStats for JVB connection...');
         this.statistics.startCallStats(
-            this.jvbJingleSession.peerconnection, this.myUserId());
+            this.jvbJingleSession.peerconnection,
+            'jitsi' /* Remote user ID for JVB is 'jitsi' */);
```
saghul: doesn't it make more sense to call it "jvb" since it's actually the JVB?

JitsiConference.js
```diff
@@ -1738,14 +1747,15 @@ JitsiConference.prototype.isCallstatsEnabled = function() {
 JitsiConference.prototype._onTrackAttach = function(track, container) {
     const isLocal = track.isLocal();
     let ssrc = null;
+    const isP2P = track.isP2P;
+    const remoteUserId = isP2P ? track.getParticipantId() : 'jitsi';
+    const peerConnection
+        = isP2P
+            ? this.p2pJingleSession && this.p2pJingleSession.peerconnection
```
saghul: would it make sense to have some "activeJingleSession" or "activePeerConnection" getter?

modules/statistics/CallStats.js
```diff
@@ -165,25 +289,27 @@ CallStats.prototype.pcCallback = tryCatch((err, msg) => {
...
+        streamEndpointId,
         usageLabel,
         containerId) {
     if (!callStatsBackend) {
         return;
     }
 
     // 'jitsi' is default remote user ID for now
-    const callStatsId = isLocal ? this.userID : DEFAULT_REMOTE_USER;
+    const callStatsId = isLocal ? CallStats.userID : streamEndpointId;
```
saghul: does the comment above still apply here?

modules/statistics/CallStats.js
```diff
@@ -481,55 +608,70 @@ function initCallback(err, msg) {
...
-        logger.log('callstats fabric not initilized', ret.message);
+    if (!atLeastOneFabric) {
 
         return;
     }
 
     CallStats.initialized = true;
-    CallStats.feedbackEnabled = true;
+
+    // There is no conference ID nor a PeerConnection available when some of
```
saghul: Is this really useful?

modules/statistics/statistics.js
```diff
@@ -239,50 +246,87 @@ Statistics.prototype.stopRemoteStats = function() {
...
-                + '/' + this.options.roomName;
-
-        /* eslint-enable prefer-template */
-        const options = {
-            confID,
-            aliasName,
-            userName: Settings.getCallStatsUserName(),
+Statistics.prototype.startCallStats = function(tpc, remoteUserID) {
+    if (!this.callStatsIntegrationEnabled) {
+
```
saghul: extra newline

modules/statistics/statistics.js
```diff
@@ -239,50 +246,87 @@ Statistics.prototype.stopRemoteStats = function() {
...
+        })) {
+
+            // Backend initialization failed bad
+            return;
+        }
     }
+
+    logger.info(`Starting CallStats for ${tpc}...`);
+
+    /* eslint-enable prefer-template */
```
saghul: where is this disabled?

modules/statistics/CallStats.js
```diff
@@ -204,6 +204,92 @@ CallStats.initBackend = function(options) {
...
+ *
+ * @param err
+ * @param msg
+ */
+function initCallback(err, msg) {
+    logger.log(`CallStats Status: err=${err} msg=${msg}`);
+
+    // there is no lib, nothing to report to
+    if (err !== 'success') {
+
```
saghul: extra newline

modules/statistics/CallStats.js
```diff
@@ -204,6 +204,92 @@ CallStats.initBackend = function(options) {
     }
 };
 
+/* eslint-disable no-invalid-this */
+
+/**
+ *
+ * @param err
```
saghul: any chance you can document these while you're here?

modules/statistics/CallStats.js
```diff
@@ -204,6 +204,92 @@ CallStats.initBackend = function(options) {
...
+                atLeastOneFabric = true;
+                if (!defaultInstance) {
+                    defaultInstance = callStatsInstance;
+                }
+            }
+        }
+    }
+
+    if (!atLeastOneFabric) {
+
```
saghul: extra newline

JitsiConference.js
paweldomas: I agree and it was 'jvb' at the very beginning, but then Emil wanted to have it called 'jitsi'.

paweldomas:
I was thinking about converting to ES6 and I will give it a try

saghul:
> I was thinking about converting to ES6 and I will give it a try

❤️

JitsiConference.js
saghul: 😞

modules/statistics/CallStats.js
paweldomas: it kind of does, but does not make sense here - removing

modules/statistics/CallStats.js
paweldomas: Camel case CallStats is our wrapper

modules/statistics/statistics.js
paweldomas: it is now here: https://github.com/jitsi/lib-jitsi-meet/blob/p2p_callstats/modules/statistics/statistics.js#L316

I'm not sure what should we do about undefined case. Log an error at least ?

paweldomas: Or provide some default ? Any suggestions ?

paweldomas: I see that in JitsiConference we try to default it to window.location.hostname, so maybe it should be an error.

modules/statistics/statistics.js
paweldomas: messed up something when refactoring - removing

modules/statistics/CallStats.js
paweldomas: I know that the logs needs to be queued, but not sure about the other things. I'm just trying to keep the old behaviour.

JitsiConference.js
paweldomas: There is a getter, but it's based on the conference state rather than the flag set on the remote track. I think it's more clear this way and more "safe", but eventually I can change it if you really insist (I'd rather keep it this way though).

saghul: No biggie. Leave it as it is.

modules/statistics/CallStats.js
```diff
@@ -47,7 +45,11 @@ const fabricEvent = {
     activeDeviceList: 'activeDeviceList'
 };
 
-let callStats = null;
+/**
+ * The CallStats API backend instance
+ * @type {callstats}
+ */
+let backend = null;
```
saghul: This can now become a static variable in the CallStats class, right?

modules/statistics/CallStats.js
```diff
@@ -68,164 +70,330 @@ const reportType = {
...
-const CallStats = tryCatch(function(jingleSession, options) {
-    try {
-        CallStats.feedbackEnabled = false;
-        callStats = new callstats($, io, jsSHA); // eslint-disable-line new-cap
+            return false;
+        }
+    }
 
-        this.peerconnection = jingleSession.peerconnection.peerconnection;
+    /* eslint-disable no-invalid-this */
```
saghul: I don't think you need this anymore.

modules/statistics/CallStats.js
```diff
@@ -68,164 +70,330 @@ const reportType = {
...
+        CallStats.initialized = true;
+
+        // There is no conference ID nor a PeerConnection available when some of
+        // the events are scheduled on the reportsQueue, so those will be
+        // reported on the first initialized fabric.
+        const defaultConfID = defaultInstance.confID;
+        const defaultPC = defaultInstance.peerconnection;
+
+        // notify callstats about failures if there were any
+        if (CallStats.reportsQueue.length) {
```
saghul: You can use a for..of and get rid of the extra level of indentation (no need to check for the length).

modules/statistics/CallStats.js
```diff
@@ -68,164 +70,330 @@ const reportType = {
...
+            // Logs from the logger are submitted on the applicationLog event
+            // "type". Logging the arguments on the logger will create endless
+            // loop, because it will put all the logs to the logger queue again.
+            if (type === wrtcFuncNames.applicationLog) {
+                console && console.debug('reportError', allArguments);
+            } else {
+                logger.debug('reportError', allArguments);
+            }
+            try {
+                originalReportError.apply(theBackend, allArguments);
```
saghul: Instead of using an array you can use `.call` and `...args`. Also, you don't seem to use the error parameter, so you can leave it as part of `...args`.

modules/statistics/CallStats.js
```diff
@@ -235,330 +403,300 @@ CallStats.prototype.associateStreamWithVideoTag = function(
...
-};
 
-/* eslint-disable no-invalid-this */
-/**
- * Notifies CallStats for connection setup errors
- */
-CallStats.prototype.sendTerminateEvent = tryCatch(function() {
-    if (!CallStats.initialized) {
-        return;
+    /* eslint-disable no-invalid-this */
```
saghul: I don't think you need this anymore.

modules/statistics/CallStats.js
```diff
@@ -235,330 +403,300 @@ CallStats.prototype.associateStreamWithVideoTag = function(
...
-});
+    /* eslint-enable no-invalid-this */
+
+    /* eslint-disable max-params */
+    /**
+     * Reports an error to callstats.
+     *
+     * @param {CallStats} [cs]
+     * @param type the type of the error, which will be one of the wrtcFuncNames
+     * @param e the error
```
saghul: let's just call this error

modules/statistics/CallStats.js
```diff
@@ -235,330 +403,300 @@ CallStats.prototype.associateStreamWithVideoTag = function(
...
-    if (!error) {
-        logger.warn('No error is passed!');
-        error = new Error('Unknown error');
+    /**
+     * Notifies CallStats that getUserMedia failed.
+     *
+     * @param {Error} e error to send
+     * @param {CallStats} cs callstats instance related to the error (optional)
+     */
+    static sendGetUserMediaFailed(e, cs) {
```
saghul: I think all static methods should be together, sorted amongst themselves, then instance methods, also sorted.

modules/statistics/CallStats.js
```diff
@@ -235,330 +403,300 @@ CallStats.prototype.associateStreamWithVideoTag = function(
...
+                console.error('sendApplicationLog failed', error);
+            }
+        }
+    }
+}
+
+// some errors/events may happen before CallStats init
+// in this case we accumulate them in this array
+// and send them to callstats on init
+CallStats.reportsQueue = [];
```
saghul: I think all these can be made into static variables.

modules/statistics/statistics.js
```diff
@@ -104,23 +106,32 @@ function Statistics(xmpp, options) {
             && (Statistics.disableThirdPartyRequests !== true);
     if (this.callStatsIntegrationEnabled) {
         loadCallStatsAPI(this.options.callStatsCustomScriptUrl);
+
+        if (!this.options.callStatsConfIDNamespace) {
```
saghul: You probably added this because I told you in the previous review. I have since noticed another problem: we don't have a proper window.location in React Native, but actually, it makes sense to use `this.options.config.hosts.domain` in `JitsiConference`, so it's always set and to the right value.

modules/statistics/statistics.js
```diff
@@ -230,34 +247,89 @@ Statistics.prototype.stopRemoteStats = function() {
...
+
+/**
+ * Constructs the CallStats conference ID based on the options currently
+ * configured in this instance.
+ * @return {string}
+ * @private
+ */
+Statistics.prototype._getCallStatsConfID = function() {
+    // The conference ID is case sensitive!!!
+    return this.options.callStatsConfIDNamespace
```
saghul: see my comment above

modules/statistics/CallStats.js
paweldomas: True, but since I have this 'allArguments' array anyway, isn't it better to use it ?

Will remove the error arg, thanks !

paweldomas: or do you mean to not create this array ?

saghul: Exactly, no array, no problem! :-)

paweldomas: well, let's stick with "more never more better", so I'm getting rid of an array in favour of full REST args

modules/statistics/CallStats.js
paweldomas: it was not me who added this methods in the first place, but I agree

paweldomas: but should static come before or after the constructor ?

saghul: If we don't have any written rule, I'd say before.

modules/statistics/statistics.js
paweldomas: yes, but are you ok with logging this or should this go away ? I think that fixing the jitsi-meet part is another PR

paweldomas:
do not merge I am seeing problems and still need to add resume/hold for p2p

paweldomas:
Should be ok now. Requires https://github.com/jitsi/jitsi-meet/pull/1487

JitsiConference.js
```diff
@@ -2217,6 +2250,9 @@ JitsiConference.prototype._stopP2PSession
     // Stop P2P stats
     logger.info('Stopping remote stats with P2P connection');
     this.statistics.stopRemoteStats();
+    logger.info('Stopping CallStats for P2P connection');
```
saghul: right above we have "with P2P connection" vs here "for P2P connection" let's pick one and be consistent. I'd say "for".

modules/statistics/CallStats.js
```diff
@@ -65,167 +61,483 @@ const reportType = {
...
+ * reporting various events. A fabric is considered disposed when
+ * {@link CallStats.sendTerminateEvent} is executed.
  *
- * @param f the function to invoke in a try/catch block
- * @return a function which invokes f in a try/catch block, logs any exception
- * to the console, and then swallows it
+ * Currently only one backend instance can be created ever and it's done using
+ * {@link CallStats.initBackend}. At the time of this writing there is no way to
+ * explicitly shutdown the backend, but it's supposed to close it's connection
+ * automatically, after all fabrics have been terminated.
```
saghul: I think I might have seen this fail, as in, stuff still happening when the call ends on mobile. Any chance we can forcefully terminate it?

modules/statistics/CallStats.js
```diff
@@ -65,167 +61,483 @@ const reportType = {
...
+            theBackend[methodName] = function(...theArguments) {
+                logger.debug(methodName, theArguments);
+                originalMethod.apply(theBackend, theArguments);
+            };
+        }
+        const originalReportError = theBackend.reportError;
+
+        /* eslint-disable max-params */
+        theBackend.reportError
+        = function(pc, cs, type, ...otherArguments) {
```
saghul: I believe the usual convention is `...args`. No big deal though.

modules/statistics/CallStats.js
```diff
@@ -65,167 +61,483 @@ const reportType = {
...
+    /**
+     * Initializes the CallStats backend. Should be called only if
+     * {@link CallStats.isBackendInitialized} returns <tt>false</tt>.
+     * @param {object} options
+     * @param {String} options.callStatsID CallStats credentials - ID
+     * @param {String} options.callStatsSecret CallStats credentials - secret
+     * @param {string} options.userName the <tt>userName</tt> part of
+     * the <tt>userID</tt> aka display name, see CallStats docs for more info.
+     * @param {string} options.aliasName the <tt>aliasName</tt> part of
+     * the <tt>userID</tt> aka endpoint ID, see CallStats docs for more info.
```
saghul: arent these the other way around? also, style: alphabetical sorting

modules/statistics/statistics.js
```diff
@@ -230,34 +247,89 @@ Statistics.prototype.stopRemoteStats = function() {
...
+Statistics.prototype.startCallStats = function(tpc, remoteUserID) {
+    if (!this.callStatsIntegrationEnabled) {
+        return;
+    } else if (this.callsStatsInstances.has(tpc.id)) {
+        logger.error('CallStats instance for ${tpc} exists already');
+
+        return;
+    }
+
+    if (!CallStats.isBackendInitialized()) {
```
saghul: So, I'm thinking, this should be moved into CallStats itself. As in, we pass all the options to the constructor, and if there is no backend initialized it will do it on its own internally, that way the backend is "hidden" from the statistics module. WDYT?

modules/statistics/CallStats.js
paweldomas: As far as I know prior to this PR fabrics were not terminated. The "terminateFabric" event was never sent, so there's no way this could work. Try searching for usages of "sendTerminateEvent" on the master - none.

modules/statistics/CallStats.js
paweldomas: It may seem, so but according to the CallStats dashboard it's this misleading way around. I'll double check to be sure.

modules/statistics/statistics.js
paweldomas: It was like that in the first place, but then the public API of the CallStats module is misleading. It will create false impression that you can assign different parameters like local ID or callStatsID/secret to each CallStats instance, but that's not really true.

saghul: I see. We could throw if you change them, perhaps, since it's not supported?

paweldomas: Not sure if I understand what you mean, but throwing from a constructor only because it's called with different arguments seems a bit weird to me.

I think that it may be possible to initialise backend another time, but then the previous one will be closed (there will be new WebSocket connection to the Callstats server ?). But I'm not sure about that. If that would be the case still I would prefer to add method 'disposeBackend' and have that called from outside, so that this action would be explicit, because it would overwrite the previous backend. It should be "informed decision" and only on the outside of CallStats we have enough information to take such decision.

saghul: Okidoki!

PR #448, opened 2017-04-12
Add callstats support for mobile

saghul:


saghul:
I set a WIP label because I have only tested with a debug build callstats provided me with, I'll update once we have the definitive build.

Note that despite the naming, react-native-callstats does not depend on react-native. I'll try to convince them to rename the package.

modules/statistics/statistics.js
```diff
@@ -7,6 +7,8 @@ import JitsiTrackError from '../../JitsiTrackError';
 const logger = require('jitsi-meet-logger').getLogger(__filename);
 const LocalStats = require('./LocalStatsCollector.js');
 const RTPStats = require('./RTPStatsCollector.js');
+
+import RTCBrowserType from '../RTC/RTCBrowserType';
```
paweldomas: is that in order ? shouldn't it come before RTPStats ?

modules/statistics/CallStats.js
```diff
@@ -1,9 +1,9 @@
-/* global $, callstats */
+/* global callstats */
 const logger = require('jitsi-meet-logger').getLogger(__filename);
-const GlobalOnErrorHandler = require('../util/GlobalOnErrorHandler');
 
-const jsSHA = require('jssha');
-const io = require('socket.io-client');
+import RTCBrowserType from '../RTC/RTCBrowserType';
+import GlobalOnErrorHandler from '../util/GlobalOnErrorHandler';
```
paweldomas: not in order ?

I'm not sure whether we order by class names of file names ?

modules/statistics/statistics.js
```diff
@@ -104,7 +106,7 @@ function Statistics(xmpp, options) {
             // of callstats.io may be disabled because of globally-disallowed
             // requests to any third parties.
             && (Statistics.disableThirdPartyRequests !== true);
-    if (this.callStatsIntegrationEnabled) {
+    if (this.callStatsIntegrationEnabled && !RTCBrowserType.isReactNative()) {
```
paweldomas: if we skip this if for RN then the check at 112 should be moved outside

paweldomas: or only the loading part should be under isReactNative condition

saghul: oh, I see. That wasn't there before so I missed it during the rebase. Will fix.

modules/statistics/statistics.js
saghul: Right you are sir!

modules/statistics/CallStats.js
saghul: file names

saghul:
@paweldomas Thanks for the review! I think I addressed all your comments, can you PTAL?

PR #456, opened 2017-04-19
fix: Ensures simulcast is activated in the browser.

gpolitis:
Ensures that the simulcast ssrc-group appears after any other
ssrc-groups in the SDP so that simulcast is properly activated.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1318,6 +1318,43 @@ TraceablePeerConnection.prototype.createDataChannel = function(label, opts) {
 };
 
 /**
+ * Ensures that the simulcast ssrc-group appears after any other ssrc-groups
+ * in the SDP so that simulcast is properly activated.
+ *
+ * @param {Object} localSdp the WebRTC session description instance for
+ * the local description.
+ * @private
+ */
+TraceablePeerConnection.prototype._ensureSimulcastGroupIsLast
+= function(localSdp) {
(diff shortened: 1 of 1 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1318,6 +1318,43 @@ TraceablePeerConnection.prototype.createDataChannel = function(label, opts) {
...
+    }
+
+    const simEndIndex = sdpStr.indexOf('\r\n', simStartIndex);
+    const simStr = sdpStr.substring(simStartIndex, simEndIndex + 2);
+
+    sdpStr = sdpStr.replace(simStr, '');
+    otherStartIndex = sdpStr.lastIndexOf('a=ssrc-group');
+    const otherEndIndex = sdpStr.indexOf('\r\n', otherStartIndex);
+    const sdpHead = sdpStr.slice(0, otherEndIndex);
+    const simStrTrimmed = simStr.trim();
```
bgrozev: Why do we first copy the "\r\n" into simStr, and then trim it? Isn't this equivalent to not copying in the first place?

gpolitis: We can re-arrange the code so that we don't call trim(), but it is not equivalent to not copying it in the first place.

We would also have to change this `sdpStr = sdpStr.replace(simStr, '');` to this `sdpStr = sdpStr.replace(simStr + '\r\n', '');`.

Would you prefer this?

PR #465, opened 2017-04-29
fix(stats): lazy initialization for statistics Sets

virtuacoplenny:
Internet Explorer is throwing an error when doing a for...of over
Sets. This is because IE needs a Set polyfill, which may be applied
after lib-jitsi-meet has loaded. However, lib-jitsi-meet on load
creates two Sets, one for CallStats.fabrics and another for
Statistics.instances. When the polyfill is then applied, the two
existing Sets do not get new methods. The fix is to lazily
initialize the Sets to allow for polyfills to be applied.

~~Internet Explorer is throwing an error when doing a for...of over
sets. Set.prototype.values() is also throwing an error...so the
compromise was ensuring the sets being iterated over are indeed
arrays.~~

~~I'm open to changing this fix or more opinions on what to try out.
I don't completely understand right now why iterating over sets
would be broken on IE and why values would also be broken.~~

### DIFF
--- a/modules/statistics/CallStats.js
+++ b/modules/statistics/CallStats.js
@@ -62,6 +62,12 @@ const reportType = {
 };
 
 /**
+ * Set of currently existing {@link CallStats} instances.
+ * @type {Set<CallStats>}
+ */
+let _fabrics;
+
+/**
  * An instance of this class is a wrapper for the CallStats API fabric. A fabric
  * reports one peer connection the the CallStats backend and is allocated with
  * {@link callstats.addNewFabric}. It has a bunch of instance methods for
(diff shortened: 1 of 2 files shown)


virtuacoplenny:
Looks like this is an issue with js loading order. babel-polyfill will provide the necessary Set methods, but that gets loaded as part of app.bundle.min.js for jitsi-meet, which is after lib-jitsi-meet gets loaded. CallStats.js and statistics.js, both in lib-jitsi-meet, create Sets on load, before the polyfill occurs. A couple new solutions present themselves to me: add back babel-polyfill (although the babel docs say it isn't intended for a lib) or move Set construction to occur within some lib-jitsi-meet initialization method the user has to call.

saghul:
Thanks for digging to the bottom of this @virtuacoplenny! IMHO the best approach here would be to create those sets lazily, I'd really like to avoid bringing babel-polyfill back.

PR #467, opened 2017-05-04
fix(permission-prompt): Use labels to predict if the prompt will not …

virtuacoplenny:
…display

In Firefox, getUserMedia for audio and video can take longer than
USER_MEDIA_PERMISSION_PROMPT_TIMEOUT, causing an event to fire stating the
permission prompt is displayed even when it is not. On my machine,
getUserMedia for audio and video would take between 200ms and 700ms. Due to
the variance, relying on the timeout alone to know about permission prompt
display might not be sufficient. One way to know if permission has been
granted, thereby knowing the prompt will not display, is by executing
enumerateDevices and seeing if devices have labels. The call to enumerate
is consistently and significantly faster that getUserMedia. This will
not work on temasys though because enumerateDevices will always return
labels.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -555,6 +555,49 @@ function enumerateDevicesThroughMediaStreamTrack(callback) {
 }
 
 /**
+ * Returns a promise to check whether or not the permission prompt for user
+ * media might display. The promise returns a boolean but the boolean value is
+ * not a definitive indicator, except in the case of false as then it is certain
+ * no prompt will display.
+ *
+ * @param {Function} enumerateDevices - The native enumerate devices function.
+ * @returns {Promise<boolean>}
+ */
+function mightShowPermissionPrompt(enumerateDevices) {
+    // The focus check is necessary for browsers (Firefox) that do not complete
+    // getUserMedia when the browser tab is not in focus. In those cases, it is
+    // not known with certainty if the permission prompt will display or not.
+    if (!document.hasFocus()) {
+        return Promise.resolve(true);
+    }
+
+    return enumerateDevices()
+        .then(mediaDevices => {
+            let hasAudioPermission = false;
+            let hasVideoPermission = false;
+
+            // Check for the existence of a label on an audio and video device.
+            // Labels should only be present if permission has been granted.
+            for (let i = 0; i < mediaDevices.length; i++) {
+                const device = mediaDevices[i];
+
+                if (device.kind === 'audioinput'
+                    && device.label) {
+                    hasAudioPermission = true;
+                } else if (device.kind === 'videoinput'
+                    && device.label) {
+                    hasVideoPermission = true;
+                }
+            }
+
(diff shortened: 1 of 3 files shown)


paweldomas:
Have you noticed that Firefox will block get user media even if you have permission in case FF window is not currently focused ? I think this optimisation may not work in this case, but haven't tried that.

virtuacoplenny:
That's a great catch. You are correct, Firefox blocks getUserMedia when the window is not focused. I've added a focus check and rebased on top of latest master.

virtuacoplenny:
@hristoterezov, could I get your opinion on the current implementation of this? I think you mentioned maybe doing the enumerateDevices check and then GUM, but now I'm thinking that might add a perceptible delay so doing both in parallel might be better. enumerateDevices on my machine on Firefox can take 100-200ms so that would be additional time on top of a slower GUM. It could also be possible you didn't mention that.

In the PR's current state, enumerateDevices is fired and the timeout is also started, and then GUM is called. You preferred timeout firing only for temasys?

hristoterezov:
@virtuacoplenny Executing enumerateDevices and GUM in parallel sounds better! 

I don't think I meant to wait for enumerateDevices in order to call GUM but I maybe haven't expressed myself well. I'm sorry for the confusion!

I'd prefer to fire the timeout only for temasys if we think that enumerateDevices is stable enough and if there aren't any other problems.

modules/RTC/RTCUtils.js
```diff
@@ -555,6 +555,49 @@ function enumerateDevicesThroughMediaStreamTrack(callback) {
...
+ * no prompt will display.
+ *
+ * @param {Function} enumerateDevices - The native enumerate devices function.
+ * @returns {Promise<boolean>}
+ */
+function mightShowPermissionPrompt(enumerateDevices) {
+    // The focus check is necessary for browsers (Firefox) that do not complete
+    // getUserMedia when the browser tab is not in focus. In those cases, it is
+    // not known with certainty if the permission prompt will display or not.
+    if (!document.hasFocus()) {
```
hristoterezov: In this case what will be the result of enumerateDevices call. It will not finish too?

It's not clear for me what is the connection between the prompt detection via `enumerateDevices` and the focused tab. In the comment you are mentioning GUM and I don't see how GUM is related to this function.

virtuacoplenny: In this case enumerateDevices would finish and would be able to see labels. However, for firefo as long as the tab is not in focus the call to getusermedia will not complete. The workaround was put in place because the following will occur with the current implementation:
1. mightShowPermissionPrompt will return false.
2. This prevents PERMISSION_PROMPT_IS_SHOWN from firing, which prevents the permissions overlay from display. Only a gray screen is displayed.

virtuacoplenny:
Testing around this more, it looks like listPermissions is actually given when either microphone or camera has been given permission. This means it's possible for camera to have permission set to allow but microphone with not permission set yet and the labels will display. This makes me think using enumerateDevices isn't the reliable solution.

PR #479, opened 2017-05-11
Analytics for ICE checks and gathering durations

paweldomas:
Will emit analytics events for ICE gathering and ICE checks duration,
separately for initiator/responder and p2p/jvb connections. Initiator
and responder have to be separated, because the flow and the values have
significant differences.

XMPPEvents.CONNECTION_ESTABLISHED is emitted outside of 'is stable'
condition, because for the JVB connection the signalling state is often
not in 'stable' ('have-remote-offer') state when ICE goes to
'connected'.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -69,6 +69,26 @@ export default class JingleSessionPC extends JingleSession {
             options) {
         super(sid, me, peerjid, connection, mediaConstraints, iceConfig);
 
+        /**
+         * Stores result of {@link window.performance.now()} at the time when
+         * ICE enters 'checking' state.
+         * @type {number|null} null if no value has been stored yet
+         * @private
+         */
+        this._iceCheckingStartedTimestamp = null;
+
+        /**
+         * Stores result of {@link window.performance.now()} at the time when
+         * first ICE candidate is spawned by the peerconnection to mark when
+         * ICE gathering started. That's, because ICE gathering state changed
+         * events are not supported by most of the browsers, so we try something
+         * that will work everywhere. It may not be as accurate, but given that
+         * 'host' candidate usually comes first, the delay should be minimal.
+         * @type {number|null} null if no value has been stored yet
+         * @private
+         */
+        this._gatheringStartedTimestamp = null;
+
         this.lasticecandidate = false;
         this.closed = false;
 
@@ -217,8 +237,13 @@ export default class JingleSessionPC extends JingleSession {
 
             // XXX this is broken, candidate is not parsed.
             const candidate = ev.candidate;
+            const now = window.performance.now();
 
             if (candidate) {
+                if (this._gatheringStartedTimestamp === null) {
+                    this._gatheringStartedTimestamp = now;
+                }
+
                 // Discard candidates of disabled protocols.
                 let protocol = candidate.protocol;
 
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -69,6 +69,26 @@ export default class JingleSessionPC extends JingleSession {
...
+         * @private
+         */
+        this._iceCheckingStartedTimestamp = null;
+
+        /**
+         * Stores result of {@link window.performance.now()} at the time when
+         * first ICE candidate is spawned by the peerconnection to mark when
+         * ICE gathering started. That's, because ICE gathering state changed
+         * events are not supported by most of the browsers, so we try something
+         * that will work everywhere. It may not be as accurate, but given that
```
gpolitis: It may not be as accurate as what? What would be the most accurate solution for Chrome, for example.

paweldomas: https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onicegatheringstatechange

modules/xmpp/JingleSessionPC.js
```diff
@@ -290,18 +324,32 @@ export default class JingleSessionPC extends JingleSession {
                 this,
                 this.peerconnection.iceConnectionState);
             switch (this.peerconnection.iceConnectionState) {
+            case 'checking':
+                this._iceCheckingStartedTimestamp = now;
```
gpolitis: Do we know how they deal with reconnects (i.e. if they trigger checking state again)?

modules/xmpp/JingleSessionPC.js
```diff
@@ -290,18 +324,32 @@ export default class JingleSessionPC extends JingleSession {
...
                     }
+                }
+
+                if (!this.wasConnected && this.wasstable) {
+                    let eventName = this.isP2P ? 'p2p.ice.' : 'ice.';
+
+                    eventName += this.isInitiator ? 'initiator.' : 'responder.';
+                    eventName += 'checksDuration';
+                    Statistics.analytics.sendEvent(
+                        eventName,
```
gpolitis: Probably isn't worth it, but how about we also put the total connection connection time? In theory we should have AVG(connectionEstablishmentDuration) = AVG(gatheringDuration) + AVG(checkingDuration), so this would serve only as a check.

modules/xmpp/JingleSessionPC.js
gpolitis: got it, thanks Pawel 👍

modules/xmpp/JingleSessionPC.js
paweldomas: I don't think this equation is valid. Doesn't in trickle ICE gathering overlap with checking ?

modules/xmpp/JingleSessionPC.js
paweldomas: I haven't seen it going back to checking, but that's probably, because we do not try to send new candidates. I will check what happens with 'transport-replace' (JVB conference migration).

modules/xmpp/JingleSessionPC.js
gpolitis: You're probably right, in which case the need for a total connection establishment duration is necessary? Otherwise how can we get the total connection establishment duration?

paweldomas: yes, probably, will add that to TODO list

modules/xmpp/JingleSessionPC.js
paweldomas: do you think I should change that comment to make it more clear ?

PR #481, opened 2017-05-12
Removes obsolete config.token.

damencho:
Token is passed as parameter when creating JitsiConnection and we no longer use it from the options passed to the connection.


damencho:
Fails just testSwitchBackToJVB which is a known issue. Merging.

PR #485, opened 2017-05-16
Store SSRCs as numbers

paweldomas:
Converts all the places where SSRCs where stored as string to use numbers.

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -53,32 +53,25 @@ export default function JitsiConferenceEventManager(conference) {
  * @param resolutions map of resolutions by ssrc
  */
 function mapResolutionsByUserId(conference, resolutions) {
-
     const id2resolution = {};
 
     // preprocess resolutions: group by user id, skip incorrect
     // resolutions etc.
-    Object.keys(resolutions).forEach(ssrc => {
-        const resolution = resolutions[ssrc];
-
-        if (!resolution.width || !resolution.height
-            || resolution.width === -1 || resolution.height === -1) {
-            return;
-        }
-
+    for (const ssrc of resolutions.keys()) {
+        const resolution = resolutions.get(ssrc);
         const id = conference.rtc.getResourceBySSRC(ssrc);
 
-        if (!id) {
-            return;
+        if (id
+            && resolution.width && resolution.height
+            && resolution.width !== -1 && resolution.height !== -1) {
+            // ssrc to resolution map for user id
+            const idResolutions = id2resolution[id] || {};
+
+            idResolutions[ssrc] = resolution;
+
+            id2resolution[id] = idResolutions;
         }
-
-        // ssrc to resolution map for user id
-        const idResolutions = id2resolution[id] || {};
-
-        idResolutions[ssrc] = resolution;
-
-        id2resolution[id] = idResolutions;
-    });
+    }
 
     return id2resolution;
 }
@@ -89,30 +82,22 @@ function mapResolutionsByUserId(conference, resolutions) {
  * @param framerates map of framerates by ssrc
  */
 function mapFrameratesByUserId(conference, framerates) {
-
     const id2framerate = {};
 
     // preprocess framerates: group by user id
-    Object.keys(framerates).forEach(ssrc => {
-        const framerate = framerates[ssrc];
-
-        if (framerate === 0) {
-            return;
-        }
-
+    for (const ssrc of framerates.keys()) {
+        const framerate = framerates.get(ssrc);
         const id = conference.rtc.getResourceBySSRC(ssrc);
 
-        if (!id) {
-            return;
+        if (framerate !== 0 && id) {
+            // ssrc to framerate map for user id
+            const id2framerates = id2framerate[id] || {};
+
+            id2framerates[ssrc] = framerate;
+
+            id2framerate[id] = id2framerates;
         }
-
-        // ssrc to framerate map for user id
-        const id2framerates = id2framerate[id] || {};
-
-        id2framerates[ssrc] = framerate;
-
-        id2framerate[id] = id2framerates;
-    });
+    }
 
     return id2framerate;
 }
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -57,8 +57,7 @@ function mapResolutionsByUserId(conference, resolutions) {
 
     // preprocess resolutions: group by user id, skip incorrect
     // resolutions etc.
-    for (const ssrc of resolutions.keys()) {
-        const resolution = resolutions.get(ssrc);
+    for (const [ ssrc, resolution ] of resolutions) {
         const id = conference.rtc.getResourceBySSRC(ssrc);
 
         if (id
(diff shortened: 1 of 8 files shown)


JitsiConferenceEventManager.js
```diff
@@ -53,32 +53,25 @@ export default function JitsiConferenceEventManager(conference) {
...
     // resolutions etc.
-    Object.keys(resolutions).forEach(ssrc => {
-        const resolution = resolutions[ssrc];
-
-        if (!resolution.width || !resolution.height
-            || resolution.width === -1 || resolution.height === -1) {
-            return;
-        }
-
+    for (const ssrc of resolutions.keys()) {
```
saghul: you can simplify this maybe:

```
for (var [key, value] of myMap) {
  // ...
}
```

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -50,6 +50,11 @@ export default function JitsiRemoteTrack(
         mediaType,
         videoType);
     this.rtc = rtc;
+
+    // Prevent from mixing up type of SSRC which should be a number
+    if (typeof ssrc !== 'number') {
+        throw new Error(`SSRC ${ssrc} is not a number`);
```
saghul: maybe `TypeError` is more appropriate?

saghul:
Not a full review, but I left a couple of drive-by comments. Let me know if you want me to fully review it @paweldomas

modules/statistics/RTPStatsCollector.js
```diff
@@ -401,6 +404,27 @@ StatsCollector.prototype._defineGetStatValueMethod = function(keys) {
     };
 };
 
+/**
+ * Obtains a stat value from given stat and converts it to a non-negative
+ * number. If the value is invalid <tt>NaN</tt> is returned.
```
virtuacoplenny: I think this comment needs updating because NaN isn't returned.

paweldomas: Oh yes it returns 0, will fix that

paweldomas:
@saghul yes, it would be great if you could take a look

paweldomas:
I had to rebase and force push, because of a conflict.

modules/statistics/RTPStatsCollector.js
```diff
@@ -401,6 +404,27 @@ StatsCollector.prototype._defineGetStatValueMethod = function(keys) {
...
 
+/**
+ * Obtains a stat value from given stat and converts it to a non-negative
+ * number. If the value is either invalid or negative then 0 will be returned.
+ * @param report
+ * @param {string} name
+ * @return {number}
+ */
+StatsCollector.prototype.getNonNegativeStat = function(report, name) {
+    let value = this._getStatValue(report, name);
```
saghul: use const

saghul: You can simplify all this as follows:

```
const value = this._getStatValue(report, name);

return Math.max(0, value | 0);
```

`value | 0` is a neat trick to coerce anything to a number or 0:

```
> undefined | 0
0
> '123' | 0
123
> null | 0
0
> -1 | 0
-1
> NaN | 0
0
> 123 | 0
123
>
```

modules/statistics/RTPStatsCollector.js
```diff
@@ -401,6 +404,27 @@ StatsCollector.prototype._defineGetStatValueMethod = function(keys) {
...
 };
 
+/**
+ * Obtains a stat value from given stat and converts it to a non-negative
+ * number. If the value is either invalid or negative then 0 will be returned.
+ * @param report
+ * @param {string} name
+ * @return {number}
+ */
+StatsCollector.prototype.getNonNegativeStat = function(report, name) {
```
saghul: should this be "public" ?

modules/statistics/RTPStatsCollector.js
```diff
@@ -650,34 +663,31 @@ StatsCollector.prototype.processStatsReport = function() {
...
     let bitrateUpload = 0;
-    const resolutions = {};
-    const framerates = {};
+    const resolutions = new Map();
+    const framerates = new Map();
 
-    Object.keys(this.ssrc2stats).forEach(
-        function(ssrc) {
-            const ssrcStats = this.ssrc2stats[ssrc];
+    for (const ssrc of this.ssrc2stats.keys()) {
```
saghul: `for (const [ ssrc, ssrcStats ] of this.ssrc2stats) {`

modules/xmpp/JingleSessionPC.js
```diff
@@ -542,7 +542,11 @@ export default class JingleSessionPC extends JingleSession {
                         + 'source[xmlns="urn:xmpp:jingle:apps:rtp:ssma:0"]');
 
             ssrcs.each((i2, ssrcElement) => {
-                const ssrc = ssrcElement.getAttribute('ssrc');
+                const ssrc = Number(ssrcElement.getAttribute('ssrc'));
+
+                if (isNaN(ssrc)) {
+                    return;
```
saghul: maybe a warning? If we get something not a number here it would mean there is a problem in Jicofo, right?

saghul: also, we should check if the ssrc is < 0, since they are 32bit unsigned

modules/RTC/TraceablePeerConnection.js
```diff
@@ -575,10 +575,20 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track) {
 
     // FIXME the length of ssrcLines[0] not verified, but it will fail
     // with global error handler anyway
-    const trackSsrc = ssrcLines[0].substring(7).split(' ')[0];
+    const ssrcStr = ssrcLines[0].substring(7).split(' ')[0];
+    const trackSsrc = Number(ssrcStr);
     const ownerEndpointId = this.signalingLayer.getSSRCOwner(trackSsrc);
 
-    if (!ownerEndpointId) {
+    if (isNaN(trackSsrc)) {
```
saghul: also check if it's < 0, the ssrc is supposed to be unsigned

modules/statistics/RTPStatsCollector.js
paweldomas: given that it's only used in this file I would make it private unless we need it elsewhere

modules/statistics/RTPStatsCollector.js
paweldomas: I am not a big fan of "neat tricks", because they are hard to understand for not neat tricks experts, but I will leave a comment

modules/statistics/RTPStatsCollector.js
saghul: agreed

modules/statistics/RTPStatsCollector.js
saghul: Me neither, but I have seen this one enough times around to have memorized it :-)

paweldomas: It does not seem to work for me:
> 2290353494 | 0
> -2004613802

saghul: That number is larger than 2147483647 (2^32 -1) so the code is actually working even better, since that's not a valid SSRC number, isn't it?

bgrozev: > That number is larger than 2147483647 (2^32 -1) so the code is actually working even better, since that's not a valid SSRC number, isn't it?

The number you quote is actually 2^31 - 1, and 2290353494 is a valid value for an SSRC . Also, there's no mention that this necessarily only reads SSRC values, and the documentation explicitly says it returns 0 if the value is invalid.

saghul: Yeah, you're right, sorry for the noise... leave it as it is @paweldomas

paweldomas: yes, it should be the old way now

PR #486, opened 2017-05-16
fix(video-label): expose feature check for video resize events

virtuacoplenny:


saghul:
@virtuacoplenny LGTM, but is this commit meant to be evaluated standalone? Why not make it part of the PR which uses it? Or is it not part of lib-jitsi-meet? If so, can you reference it in the commit log?

PR #487, opened 2017-05-19
More p2p analytics

paweldomas:


JitsiConference.js
```diff
@@ -595,13 +595,21 @@ JitsiConference.prototype.addTrack = function(track) {
...
+ * Fires TRACK_AUDIO_LEVEL_CHANGED change conference event (for local tracks).
+ * @param {TraceablePeerConnection|null} tpc
  * @param audioLevel the audio level
  */
-JitsiConference.prototype._fireAudioLevelChangeEvent = function(audioLevel) {
-    this.eventEmitter.emit(
-        JitsiConferenceEvents.TRACK_AUDIO_LEVEL_CHANGED,
-        this.myUserId(), audioLevel);
+JitsiConference.prototype._fireAudioLevelChangeEvent
+= function(tpc, audioLevel) {
```
bgrozev: Is the lack of indentation intentional?

JitsiConference.js
```diff
@@ -595,13 +595,21 @@ JitsiConference.prototype.addTrack = function(track) {
...
+    const activeTpc = this.getActivePeerConnection();
+
+    // There can be no TraceablePeerConnection if audio levels do not come from
+    // a peerconnection (see LocalStatsCollector.js)
+    if (tpc === null || activeTpc === tpc) {
```
bgrozev: I don't understand the comment (LocalStatsCollector.js doesn't call this method directly and I can't follow the entire chain which leads to it).

modules/RTC/RTC.js
```diff
@@ -647,70 +647,32 @@ export default class RTC extends Listenable {
...
-     * @param {number} ssrc the ssrc to check.
-     * @return {JitsiRemoteTrack|undefined} return the first remote track that
-     * matches given SSRC or <tt>undefined</tt> if no such track was found.
-     * @private
-     */
-    _getRemoteTrackBySSRC(ssrc) {
-        return this.getRemoteTracks().find(
-            remoteTrack => ssrc === remoteTrack.getSSRC());
-    }
+    /* eslint-enable max-params */
```
bgrozev: I'm just curious why "eslint-disable max-params" encompasses the whole function and not just the first line.

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -75,6 +75,205 @@ class AverageStatReport {
...
+                = (id, data) => this._processRemoteStats(id, data);
+            conference.on(
+                ConnectionQualityEvents.REMOTE_STATS_UPDATED,
+                this._onRemoteStatsUpdated);
+        }
+    }
+
+    /**
+     * Processes next batch of stats.
+     * @param {go figure} data
```
bgrozev: LOL

bgrozev: I thought the convention was that the names alternate -- "data" from one function goes to "object" in the next function, and then back to "data" ;)

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -75,6 +75,205 @@ class AverageStatReport {
...
+            logger.error('No stats');
+
+            return;
+        }
+
+        if (RTCBrowserType.supportsRTTStatistics()) {
+            if (data.transport && data.transport.length) {
+                this._avgRTT.addNext(data.transport[0].rtt);
+            } else {
+                this._avgRTT.reset();
```
bgrozev: Why do we reset in this case? It looks like we just throw away some values, since this._sampleIdx is incremented regardless.

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -452,8 +565,6 @@ export default class AvgRTPStatsReporter {
         this._avgRemoteFPS.reset();
         this._avgLocalFPS.reset();
         this._avgCQ.reset();
-        this._avgRTT.reset();
-        this._avgRemoteRTTMap.clear();
```
bgrozev: It looks like this function fails to reset _avgPacketLossTotal. Should it also reset jvbStatsMonitor and p2pStatsMonitor?

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -533,16 +532,23 @@ export default class AvgRTPStatsReporter {
...
+                        peerSsrcCount += 1;
                     }
+                }
 
-                    peerAvg /= ssrcs.length;
+                if (peerSsrcCount > 0) {
+                    peerFpsSum /= peerSsrcCount;
 
-                    subFrameAvg += peerAvg;
+                    subFrameAvg += peerFpsSum / peerSsrcCount;
```
bgrozev: Why do we divide by peerSsrcCount twice?

modules/statistics/AvgRTPStatsReporter.js
bgrozev: This is fixed in a later commit (and the answer to my question appears to be "no").

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -519,42 +554,92 @@ export default class AvgRTPStatsReporter {
...
+                        avgRemoteFps += peerAvgFPS;
+                        peerCount += 1;
+                    }
+                }
+            }
+        }
 
-                let peerFpsSum = 0;
-                let peerSsrcCount = 0;
+        return avgRemoteFps / peerCount;
```
bgrozev: It would be easier to read if "avgRemoteFps" was named something like "peersFpsSum" (the same as elsewhere in this file).

modules/statistics/AvgRTPStatsReporter.js
bgrozev: This is also fixed in a later commit.

JitsiConference.js
paweldomas: It means that if audio levels are emitted by LocalStatsCollector.js there will be no TPC instance, because it does not need one to measure them (web audio API is used). I'll try to describe it better.

JitsiConference.js
paweldomas: yes, because lint will complain about wrong indentation otherwise (for the lines below it will want double indentation instead of single).

modules/RTC/RTC.js
paweldomas: I can try make it only for the first line, but I think it makes JSDoc look messy. I have to check if the IDE will still match JSDoc with the method.

modules/statistics/AvgRTPStatsReporter.js
paweldomas: I would expect that lack of transport will mean that transport changed or something is wrong (more likely the latter), so I thought it should be reset to make sense for the new transport. Each stat carries it's own sum and count. But maybe it doesn't really make sense. I'll remove -
 it's only causing unnecessary confusion.

modules/RTC/RTC.js
paweldomas: nope, doesn't work. Eslint will complain about missing JSDoc if I try put "eslint-disable-next-line max-params"

modules/statistics/AvgRTPStatsReporter.js
paweldomas: in this file it's consistent with "data"

paweldomas:
I've tried to address all of the comments. Please squash and merge.

PR #488, opened 2017-05-19
feat(JSPC): ICE establishment time

paweldomas:
Will report total ICE establishment time under
'ice.initiator.establishmentDuration' and
'ice.responder.establishmentDuration' ('p2p.' prefix added for P2P).
It's the amount of time between the time when either checking or
gathering started (whichever starts first) and when ICE entered
'connected' for the first time.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -352,7 +352,7 @@ export default class JingleSessionPC extends JingleSession {
 
                     eventName += this.isInitiator ? 'initiator.' : 'responder.';
                     Statistics.analytics.sendEvent(
-                        `${eventName}checkDuration`,
+                        `${eventName}checksDuration`,
                         {
                             value: now - this._iceCheckingStartedTimestamp
                         });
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -351,12 +351,26 @@ export default class JingleSessionPC extends JingleSession {
                     let eventName = this.isP2P ? 'p2p.ice.' : 'ice.';
 
                     eventName += this.isInitiator ? 'initiator.' : 'responder.';
-                    eventName += 'checksDuration';
                     Statistics.analytics.sendEvent(
-                        eventName,
+                        `${eventName}checkDuration`,
```
bgrozev: Is the change from "checks" to "check" intentional?

modules/xmpp/JingleSessionPC.js
```diff
@@ -351,12 +351,26 @@ export default class JingleSessionPC extends JingleSession {
...
                         {
                             value: now - this._iceCheckingStartedTimestamp
                         });
+
+                    // Switch between ICE gathering and ICE checking whichever
+                    // started first (scenarios are different for initiator
+                    // vs responder)
+                    const iceStarted
+                        = this._gatheringStartedTimestamp
```
bgrozev: I think it would be clearer to use Math.min here.

paweldomas:
fixed both comments, thanks !

PR #489, opened 2017-05-22
Allow override of the infrastructure level channel LastN value.

gpolitis:
For debugging purposes, it's sometimes useful to override the
infrastructure level channel LastN value via the URL (i.e. by launching
https://meet.example.org/room#config.channelLastN=X). This commit
achieves this by calling the JitsiConference.setLastN method during the
Jitsi conference initialization phase.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -52,6 +52,9 @@ const logger = getLogger(__filename);
  * @param {number} [options.config.backToP2PDelay=5] a delay given in seconds,
  * before the conference switches back to P2P, after the 3rd participant has
  * left the room.
+ * @param {number} [options.config.channelLastN=-1] The requested amount of
+ * videos are going to be delivered after the value is in effect. Set to -1 for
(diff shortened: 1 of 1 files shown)


JitsiConference.js
```diff
@@ -248,6 +248,10 @@ JitsiConference.prototype._init = function(options = {}) {
             () =>
                 this.eventEmitter.emit(JitsiConferenceEvents.TALK_WHILE_MUTED));
     }
+
+    if (options.config.channelLastN) {
```
paweldomas: options.config.channelLastN is not documented in int arguments. I think it may fit better to be configured in constructor and described like other options:
https://github.com/jitsi/lib-jitsi-meet/pull/489/commits/42253ee101c7ef0c1062d21dc678a11b4912e938#diff-40e35c919311b508dc5b45a7bf028f54L47

gpolitis: I'll follow your advice, thanks @paweldomas !

PR #490, opened 2017-05-23
remove jitsiRegionInfo from lib-jitsi-meet

aaronkvanmeerten:
Use new global property window.jitsiAnalyticsPermanentProperties to report deployment-specific properties
no longer need to set jitsiRegionInfo from external_connect, instead rely on new global property from Jitsi meet local.html, can be customized by deployment

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1308,8 +1308,10 @@ JitsiConference.prototype.onIncomingCall
     // add info whether call is cross-region
     let crossRegion = null;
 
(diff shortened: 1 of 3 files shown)


aaronkvanmeerten:
This is related to https://github.com/jitsi/jitsi-meet/pull/1591 where this same global properties are used in analytics.js, and where the local.html is SSI included from index.html, where these properties are intended to be set in the page.

PR #499, opened 2017-06-02
Fix OverconstrainedError handling in Firefox

saghul:
Firefox uses "constraint" as the key in the error:

```
[modules/RTC/RTCUtils.js] <value/<>:  Failed to get access to local media. Error
MediaStreamError { name: "OverconstrainedError", message: "Constraints could be
not satisfied.", constraint: "width", stack: "" } Object { audio: false, video:
Object }  lib-jitsi-meet.min.js:6:10419
```

### DIFF
--- a/JitsiTrackError.js
+++ b/JitsiTrackError.js
@@ -88,7 +88,7 @@ function JitsiTrackError(error, options, devices) {
             break;
         case 'ConstraintNotSatisfiedError':
         case 'OverconstrainedError': {
-            const constraintName = error.constraintName;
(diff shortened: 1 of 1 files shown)


saghul:
I should have an old webcam at home, I'll try to test it, but the fix looks technically correct based on the reporter's traceback.

lyubomir:
Merged into jitsi/lib-jitsi-meet#master as 9cf65ce2fe1e04f7ab44aab63f938c7beeafdb56.

PR #460, opened 2017-04-23
P2P react native (work in progress)

paweldomas:
Requires https://github.com/oney/react-native-webrtc/pull/261 and new release

PR #504, opened 2017-06-06
Misc session fixes

paweldomas:
Extract fixes from https://github.com/jitsi/lib-jitsi-meet/pull/460 discovered while working on P2P for react-native. These commits are not directly related to p2p+react-native and can be useful in other cases.

### DIFF
--- a/modules/RTC/DataChannels.js
+++ b/modules/RTC/DataChannels.js
@@ -49,7 +49,7 @@ DataChannels.prototype.onDataChannel = function(event) {
     const self = this;
 
     dataChannel.onopen = function() {
-        logger.info('Data channel opened by the Videobridge!', dataChannel);
+        logger.info('Data channel opened by the Videobridge!');
 
         // Code sample for sending string and/or binary data
         // Sends String message to the bridge
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -827,7 +827,7 @@ export default class JingleSessionPC extends JingleSession {
             const oldLocalSdp
                 = this.peerconnection.localDescription.sdp;
 
-            this._renegotiate(newRemoteSdp)
+            this._renegotiate(newRemoteSdp.raw)
                 .then(() => {
                     if (this.state === JingleSessionState.PENDING) {
                         this.state = JingleSessionState.ACTIVE;
@@ -1250,7 +1250,7 @@ export default class JingleSessionPC extends JingleSession {
                     ? this._processRemoteAddSource(addOrRemoveSsrcInfo)
                     : this._processRemoteRemoveSource(addOrRemoveSsrcInfo);
 
-            this._renegotiate(newRemoteSdp)
+            this._renegotiate(newRemoteSdp.raw)
                 .then(() => {
                     const newLocalSdp
                         = new SDP(this.peerconnection.localDescription.sdp);
@@ -1355,7 +1355,7 @@ export default class JingleSessionPC extends JingleSession {
 
     /**
      * Do a new o/a flow using the existing remote description
-     * @param {SDP object} optionalRemoteSdp optional remote sdp
+     * @param {string} [optionalRemoteSdp] optional, raw remote sdp
      *  to use.  If not provided, the remote sdp from the
      *  peerconnection will be used
      * @returns {Promise} promise which resolves when the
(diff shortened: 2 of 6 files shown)


modules/RTC/DataChannels.js
```diff
@@ -49,7 +49,7 @@ DataChannels.prototype.onDataChannel = function(event) {
     const self = this;
 
     dataChannel.onopen = function() {
-        logger.info('Data channel opened by the Videobridge!', dataChannel);
+        logger.info('Data channel opened by the Videobridge!');
```
saghul: ❤️ ❤️ ❤️

modules/xmpp/JingleSessionPC.js
```diff
@@ -1050,45 +1050,51 @@ export default class JingleSessionPC extends JingleSession {
...
-            sessionTerminate.up().c('text').t(text);
+    terminate(success, failure, options) {
+        if (this.state === JingleSessionState.ENDED) {
+            return;
         }
 
-        // Calling tree() to print something useful
-        sessionTerminate = sessionTerminate.tree();
-        logger.info('Sending session-terminate', sessionTerminate);
+        if (!options || Boolean(options.sendSessionTerminate) !== false) {
```
saghul: `if (!options || Boolean(options.sendSessionTerminate))` ?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1050,45 +1050,51 @@ export default class JingleSessionPC extends JingleSession {
...
+                    .t(options.reasonDescription);
+            }
 
-        this.connection.sendIQ(
-            sessionTerminate,
-            success,
-            this.newJingleErrorHandler(sessionTerminate, failure), IQ_TIMEOUT);
+            // Calling tree() to print something useful
+            sessionTerminate = sessionTerminate.tree();
+            logger.info('Sending session-terminate', sessionTerminate);
```
saghul: info? maybe debug?

paweldomas: session-initiate is logged on info, so I wanted to have it consistent. But I can change both eventually

paweldomas: source-add and source-remove are also logged on info

paweldomas:
@saghul I will do a force push for the comments as I think it makes sense to "rebase and merge" instead of "squash and merge" in this case. It will be good to have each of these fixes in separate commit.

paweldomas:
@saghul will you have some time to take a look on the fixes and merge eventually ?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1310,9 +1356,19 @@ export default class JingleSessionPC extends JingleSession {
      *  rejects with an error {string}
      */
     _renegotiate(optionalRemoteSdp) {
+        const defaultRemoteSdp = this.peerconnection.remoteDescription.sdp;
```
saghul: Can we rename this to `currentRemoteSdp` instead? I think it reads a bit better

modules/xmpp/JingleSessionPC.js
```diff
@@ -1310,9 +1356,19 @@ export default class JingleSessionPC extends JingleSession {
...
+
+        // FIXME new SDP() may not be necessary - raw SDP is needed anyway
```
saghul: can we check this and get rid of it if necessary since we're here?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1461,7 +1517,11 @@ export default class JingleSessionPC extends JingleSession {
                 this.peerconnection.addTrack(newTrack);
             }
 
-            if ((oldTrack || newTrack) && oldLocalSdp) {
+            // FIXME it makes sense to check for ACTIVE state in other places,
```
saghul: I don't really get what this fixme says, can you rephrase?

modules/xmpp/JingleSessionPC.js
```diff
@@ -713,7 +713,8 @@ export default class JingleSessionPC extends JingleSession {
                         error,
                         this.mediaConstraints);
                     finishedCallback(error);
-                });
+                },
+                this.mediaConstraints);
```
saghul: This one looks like a "leftover bug" from "fix(JingleSessionPC): execute 'invite' on the queue" right? If so, can you please combine the commits?

modules/xmpp/JingleSessionPC.js
saghul: ok

modules/xmpp/JingleSessionPC.js
paweldomas: I'm removing it

PR #506, opened 2017-06-07
P2P dominant speaker

damencho:
Makes speaker stats to work in p2p mode.

### DIFF
--- a/modules/P2PDominantSpeakerDetection.js
+++ b/modules/P2PDominantSpeakerDetection.js
@@ -1,0 +1,56 @@
+import * as JitsiConferenceEvents from '../JitsiConferenceEvents';
+
+/**
+ * The value which we use to say, every sound over this threshold
+ * is talking on the mic.
+ * @type {number}
+ */
+const SPEECH_DETECT_THRESHOLD = 0.6;
+
+/**
+ * The <tt>P2PDominantSpeakerDetection</tt> is activated only when p2p is
+ * currently used.
+ * Listens for changes in the audio level changes of the local p2p audio track
+ * or remote p2p one and fires dominant speaker events to be able to use
+ * features depending on those events (speaker stats), to make them work without
+ * the video bridge.
+ */
+export default class P2PDominantSpeakerDetection {
+    /**
+     * Creates P2PDominantSpeakerDetection
+     * @param conference the JitsiConference instance that created us.
+     * @constructor
+     */
+    constructor(conference) {
+        this.conference = conference;
+
+        conference.addEventListener(
+            JitsiConferenceEvents.TRACK_AUDIO_LEVEL_CHANGED,
+            this._audioLevel.bind(this));
+
+        this.myUserID = this.conference.myUserId();
+    }
+
+    /**
+     * Receives audio level events for all streams in the conference.
+     *
+     * @param {String} id - The participant id
+     * @param {number} audioLevel - The audio level.
+     */
+    _audioLevel(id, audioLevel) {
+
+        // we do not process if p2p is not active
+        // or audio level is under certain threshold
+        // or if the audio level is for local audio track which is muted
+        if (!this.conference.isP2PActive()
+            || audioLevel <= SPEECH_DETECT_THRESHOLD
+            || (id === this.myUserID
+                    && this.conference.getLocalAudioTrack().isMuted())) {
+            return;
+        }
+
+        this.conference.eventEmitter.emit(
+            JitsiConferenceEvents.DOMINANT_SPEAKER_CHANGED,
+            id);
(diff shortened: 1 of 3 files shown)


modules/P2PDominantSpeakerDetection.js
```diff
@@ -0,0 +1,99 @@
...
+     * endpoint/participant/stream being reported.
+     * @param {number} audioLevel - The audio level of <tt>ssrc</tt>.
+     * @param {boolean} isLocal - <tt>true</tt> if <tt>ssrc</tt> represents a
+     * local/send stream or <tt>false</tt> for a remote/receive stream.
+     */
+    _audioLevel(pc, ssrc, audioLevel, isLocal) {
+
+        // if audio level is lower than this we consider noise and no one is
+        // talking. The same value is used in TalkMutedDetection
+        if (audioLevel <= 0.6) {
```
saghul: can we extract this to a constant at the top of the file? It would make is easier to read / share.

damencho: Yep, I want to use that value from both P2PDominantSpeaker and TalkWhileMuted, but we don't have a common place for constants. I'm wondering should I introduce one, just for this thing ...? WDTYT @saghul ?

saghul: You mean in modules/ ? That might look a bit odd. Even if the value is the same, defining the constant at the file level helps with reading, and also if the value is used more than once within the same file. We can move it elsewhere later on perhaps.

damencho: Just fixed those.

saghul: Yeah, looks good! Can't comment on the rest of the PR, sorry!

modules/P2PDominantSpeakerDetection.js
```diff
@@ -0,0 +1,106 @@
...
+    /**
+     * Notified when p2p state changes for current conference.
+     *
+     * @param {JitsiConference} conference - the current conference
+     * @param {bool} isP2P - are we going in p2p mode or exiting that mode
+     * @private
+     */
+    _onP2PChanged(conference, isP2P) {
+        if (isP2P) {
+            conference.addEventListener(
```
paweldomas: have you checked if listeners will be removed when conference ends (while still in P2P mode) ?

modules/P2PDominantSpeakerDetection.js
```diff
@@ -0,0 +1,106 @@
...
+                JitsiConferenceEvents.TRACK_ADDED,
+                this._trackAdded);
+        }
+    }
+
+    /* eslint-disable max-params */
+    /**
+     * Receives audio level events for all send and receive streams.
+     *
+     * @param pc - WebRTC PeerConnection object of the
```
paweldomas: the type should be {TraceablePeerConnection}, also it would be good to name the variable 'tpc' to make it easier to figure out that it's not the WebRTC PeerConnection, but our wrapper

modules/P2PDominantSpeakerDetection.js
```diff
@@ -0,0 +1,106 @@
...
+     * @param ssrc - The synchronization source identifier (SSRC) of the
```
paweldomas: the type should be {number} or {string} ? what's the type there ?

modules/P2PDominantSpeakerDetection.js
```diff
@@ -0,0 +1,106 @@
...
+     * @param {JitsiConference} conference - the current conference
+     * @param {bool} isP2P - are we going in p2p mode or exiting that mode
+     * @private
+     */
+    _onP2PChanged(conference, isP2P) {
+        if (isP2P) {
+            conference.addEventListener(
+                JitsiConferenceEvents.TRACK_ADDED,
+                this._trackAdded);
+            conference.statistics.addAudioLevelListener(this._audioLevel);
```
paweldomas: If find storing remote audio track's SSRC  not the cleanest way of doing it. Maybe it would be better if this module would listen to the top level audio level event which will filter out the events coming from different TraceablePeerConnections:
JitsiConferenceEvents.TRACK_AUDIO_LEVEL_CHANGED
There you will get the user's ID and the actual audio level. You can compare user's Id to figure out if it's the local or remote audio level. Also disable the module or ignore event when conference is not in P2P mode (JitsiConference.isP2PActive())

paweldomas: https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L1187

damencho:
@paweldomas Can you take a look again, I just force pushed it.

PR #507, opened 2017-06-08
P2P audio only

paweldomas:


paweldomas:
jenkins test this please

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1456,16 +1474,28 @@ TraceablePeerConnection.prototype.setLocalDescription
...
+ * no effect.
+ * @return {boolean} <tt>true</tt> if the value has changed and sRD/sLD cycle
+ * needs to be executed in order for the changes to take effect or
+ * <tt>false</tt> if the given value was the same as the previous one.
  * @public
  */
-TraceablePeerConnection.prototype.setMediaTransferActive = function(active) {
-    this.mediaTransferActive = active;
+TraceablePeerConnection.prototype.setAudioTransferActive = function(active) {
+    logger.debug(`${this} audio transfer active: ${active}`);
```
hristoterezov: Why do we print `this`?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1456,16 +1474,28 @@ TraceablePeerConnection.prototype.setLocalDescription
...
+    let changed = false;
+
+    if (typeof active === 'boolean') {
```
hristoterezov: Do we pass somewhere something different than boolean? If yes probably we should stop doing that.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1529,6 +1559,31 @@ TraceablePeerConnection.prototype.setRemoteDescription
...
+ * @param {boolean} [active] <tt>true</tt> to enable video media transmission or
+ * <tt>false</tt> to disable. If the value is not a boolean the call will have
+ * no effect.
+ * @return {boolean} <tt>true</tt> if the value has changed and sRD/sLD cycle
+ * needs to be executed in order for the changes to take effect or
+ * <tt>false</tt> if the given value was the same as the previous one.
+ * @public
+ */
+TraceablePeerConnection.prototype.setVideoTransferActive = function(active) {
+    logger.debug(`${this} video transfer active: ${active}`);
```
hristoterezov: This function is copy / paste of `setAudioTransferActive`. I have the same questions here.

Also why we need 2 functions that are the same? Can't we merge them into 1 function?

hristoterezov:
There are a lot of places where we are printing `this`. Do we really need to print it? It seems like a bad practice to me. I think in the past we have had some issues with this kind of prints because of converting `this` to string ( maybe when we are trying to send the logs somewhere ).

I think in most cases and probably for all cases this can be replaced with just printing a property.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1747,31 +1817,73 @@ export default class JingleSessionPC extends JingleSession {
...
+     * @param {boolean} [audioActive] <tt>true</tt> to enable audio media
+     * transfer or <tt>false</tt> to suspend audio media transmission.
+     * @param {boolean} [videoActive] <tt>true</tt> to enable video media
+     * transfer or <tt>false</tt> to suspend video media transmission.
      * @return {Promise} a <tt>Promise</tt> which will resolve once
      * the operation is done. It will be rejected with an error description as
      * a string in case anything goes wrong.
      */
-    setMediaTransferActive(active) {
+    setMediaTransferActive(audioActive, videoActive) {
```
hristoterezov: Why do we need to check for the types here?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1799,6 +1907,113 @@ export default class JingleSessionPC extends JingleSession {
...
+     * Parses 'senders' attribute of the video content and figures out if remote
+     * peer is interested in receiving our video.
+     * @param {jQuery} jingleContents
+     * @return {boolean|null} <tt>true</tt> if the remote peer is interested in
+     * receiving our video or <tt>false</tt> otherwise (which means that we
+     * should stop sending video). <tt>null</tt> is returned when it failed to
+     * parse the senders attribute of Jingle video contents.
+     * @private
+     */
+    _parseVideoSenders(jingleContents) {
```
hristoterezov: Isn't `isRemoteVideoActive ` better name for this method? 

Also I think currently you are treating the `null`  result as `false`. You can return just `false` in the case when now you are returning `null`. This will simplify the code.

hristoterezov:
Is there any way to remove all this type checks and use only booleans? Do we really need the `null` value in all this methods?

If we do this the code will be so much simpler!

modules/RTC/TraceablePeerConnection.js
paweldomas: We use it, because TraceablePeerConnection has 'toString' method implemented for the debugging purpose. We need to know on which peer connection things are happening to figure out whether it's JVB or P2P.

modules/RTC/TraceablePeerConnection.js
paweldomas: Note that 'active' argument is optional, see [] syntax in the JSDoc. So this is basically a check for 'undefined'. I know it feels weird, but it makes things simpler for the code that uses this method (which also has optional arguments). I can try to get rid of that.

modules/RTC/TraceablePeerConnection.js
paweldomas: Well, they set different property. Not sure how do you imagine it being a one function.

modules/xmpp/JingleSessionPC.js
paweldomas: because those arguments are optional

paweldomas: hmm, but actually after I checked again on the usages they are always passed, so maybe no need to make them optional

modules/xmpp/JingleSessionPC.js
paweldomas: it would sound like a getter for 'remoteVideoActive', but it's not what it does. It parses Jingle 'contents' element to figure out what information is there

paweldomas:
I will take a look into removing type checks

paweldomas:
I don't remember why I made those args optional lol, I'm sure there was some reason for that... but it does not seem to be the case 🤦‍♂️

modules/xmpp/JingleSessionPC.js
```diff
@@ -1799,6 +1894,113 @@ export default class JingleSessionPC extends JingleSession {
...
+     * Will put and execute on the queue a session modify task. Currently it
+     * only checks the senders attribute of the video content in order to figure
+     * out if the remote peer has video in the inactive state (stored locally
+     * in {@link _remoteVideoActive} - see field description for more info).
+     * @param {jQuery} jingleContents jQuery selector pointing to the jingle
+     * element of the session modify IQ.
+     * @see {@link _remoteVideoActive}
+     * @see {@link _localVideoActive}
+     */
+    modifyContents(jingleContents) {
```
hristoterezov: I'd implement this function:
```
modifyContents(jingleContents) {
    logger.debug('Queued modify contents task');

    this.modificationQueue.push(
        finishedCallback => {
            if (this._assertNotEnded('modify contents')
                && this._parseContentSenders(jingleContents)) {
                this._renegotiate()
                    .then(finishedCallback, finishedCallback /* (error) */);
            } else {
                finishedCallback();
            }
        },
        error => logger.error('modifyContents failed', error)
    );
}
```

modules/xmpp/JingleSessionPC.js
hristoterezov: Hmm true. But at the end it returns false/true/null which is used only to check remoteVideoActive, right? 

I don't insist on isRemoteVideoActive. My problem with _parseVideoSenders is that I expect from the function to return some senders but it returns boolean|null.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1799,6 +1894,113 @@ export default class JingleSessionPC extends JingleSession {
...
+     */
+    _parseVideoSenders(jingleContents) {
+
+        const videoContents = jingleContents.find('>content[name="video"]');
+        let newDirection = null;
+
+        if (videoContents.length) {
+            const senders = videoContents[0].getAttribute('senders');
+
+            if (senders) {
```
hristoterezov: I don't think that you actually need `newDirection` actual value. From senders attribute and isInitiator you can calculate the result.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1799,6 +1894,113 @@ export default class JingleSessionPC extends JingleSession {
...
+            workFunction,
+            error => {
+                if (error) {
+                    logger.error('modifyContents failed', error);
+                }
+            });
+    }
+
+    /**
+     *
```
hristoterezov: Missing JSDoc.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1799,6 +1894,113 @@ export default class JingleSessionPC extends JingleSession {
...
+     * @param jingleContents
+     * @param finishedCallback
+     * @return {boolean}
+     * @private
+     */
+    _parseContentSenders(jingleContents) {
```
hristoterezov: Here I have the same problem with the name as _parseVideoSenders

modules/xmpp/JingleSessionPC.js
paweldomas: I will apply the simplification part, but not the work function part, because in all other 'modificationQueue' usages we're doing it this way and I don't want to make it different in one place now. There was another discussion about that with Brian when he was writing that initially, so I don't want to discuss it again.

modules/xmpp/JingleSessionPC.js
paweldomas: so how about 'parseRemoteVideoActive' ?

modules/xmpp/JingleSessionPC.js
paweldomas: are you saying that you want to see there nested ternary return at the end ?

modules/xmpp/JingleSessionPC.js
paweldomas: I will probably have to merge this 2 methods into 1. Otherwise I won't find any suitable name

modules/xmpp/JingleSessionPC.js
paweldomas: I looked into simplifying it, but we need to know if senders was one of valid value and if it was present. We don't want to suspend video transfer if 'senders' is missing or invalid.

modules/xmpp/strophe.jingle.js
```diff
@@ -172,6 +172,10 @@ class JingleConnectionPlugin extends ConnectionPlugin {
                 XMPPEvents.CALL_ACCEPTED, sess, $(iq).find('>jingle'));
             break;
         }
+        case 'session-modify': {
```
hristoterezov: It should be content-modify.

paweldomas: 🤦‍♂️

modules/xmpp/JingleSessionPC.js
hristoterezov: Awesome!

modules/xmpp/JingleSessionPC.js
```diff
@@ -974,6 +1006,44 @@ export default class JingleSessionPC extends JingleSession {
...
+        const newSendersValue = videoTransferActive ? 'both' : 'none';
+
+        const sessionModify
+            = $iq({
+                to: this.peerjid,
+                type: 'set'
+            })
+            .c('jingle', {
+                xmlns: 'urn:xmpp:jingle:1',
+                action: 'session-modify',
```
hristoterezov: It should be content-modify

modules/xmpp/JingleSessionPC.js
paweldomas: but I can simplify it by returning on switch instead of adding this "SDP" direction

modules/RTC/TraceablePeerConnection.js
hristoterezov: There are couple of ways of doing it. But for example:
```
setMediaTransferActive(type, active) {
    const prop = type === 'audio' ? 'audioTransferActive' : 'audioTransferActive';
    const changed = this[prop] !== active;
    this[prop] = active;
    return changed;
}
```

or you can store the media active state by type initially.

I bet there are other ways too. But I'm OK even with the current implementation.

modules/xmpp/JingleSessionPC.js
```diff
@@ -974,6 +1006,44 @@ export default class JingleSessionPC extends JingleSession {
...
     /**
+     * Will send 'session-modify' IQ in order to ask the remote peer to
+     * either stop or resume sending video media.
+     * @param {boolean} videoTransferActive <tt>false</tt> to let the other peer
+     * know that we're not sending nor interested in receiving video contents.
+     * When set to <tt>true</tt> remote peer will be asked to resume video
+     * transfer.
+     * @private
+     */
+    sendSessionModify(videoTransferActive) {
```
hristoterezov: Maybe we should rename this method to `sendContentModify `.

modules/RTC/TraceablePeerConnection.js
paweldomas: I'd rather keep the current implementation, because even though this trick is interesting and removes some duplication, it will make it hard to track usages of these field.

modules/xmpp/JingleSessionPC.js
```diff
@@ -25,6 +25,33 @@ const IQ_TIMEOUT = 10000;
...
+     * invalid then <tt>null</tt> will be returned.
+     * @private
+     */
+    static parseVideoSenders(jingleContents) {
+        const videoContents = jingleContents.find('>content[name="video"]');
+
+        if (videoContents.length) {
+            const senders = videoContents[0].getAttribute('senders');
+
+            switch (senders) {
```
hristoterezov: Isn't it better if you do this with `if` instead of switch? It looks a little bit weird to me with the switch because it is binary decision.

paweldomas: yes, whatever...

PR #508, opened 2017-06-13
ref(JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED): reorder args

paweldomas:
Reorders JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED event arguments by
putting TraceablePeerConnection at the end. This way it's easier to
treat it as "library internal".


paweldomas:
Must be merged with https://github.com/jitsi/jitsi-meet/pull/1667

PR #514, opened 2017-06-19
fix(lint): add space before comment

virtuacoplenny:


virtuacoplenny:
@damencho, could I get your opinion on this? I haven't verified which commit introduced a change that now causes this linting error.

damencho:
Yeah I was looking at the same problem cause other PRs were also seeing this.

virtuacoplenny:
I went back to commit e903d, which is a merge from April 20, and the linting still fails. Could be caused by a dependency of a dependency...

PR #515, opened 2017-06-21
P2P video track added later

paweldomas:
The P2P will break and never receive video in case a video track is not added, before the initial offer/answer. That's because we'll sent the media direction as recvonly and it is never updated on the remote side after initial offer/answer. We need to always advertise 'sendrecv' through Jingle. This used to work fine in JVB connection, because it's Jicofo that creates the offer and the media direction is always 'sendrecv'.

Also simplifies signalling for SSRC ownership in P2P where we can always assume that the remote peer owns all SSRCs. Without the fix P2P connection will fail to process 'source-add' notifications.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1035,6 +1035,33 @@ const normalizePlanB = function(desc) {
 };
 
 /**
+ * Makes sure that both audio and video directions are configured as 'sendrecv'.
+ * @param {Object} localDescription the SDP object as defined by WebRTC.
+ */
+const enforceSendRecv = function(localDescription) {
+    if (localDescription && localDescription.sdp) {
+        const transformer = new SdpTransformWrap(localDescription.sdp);
+        const audioMedia = transformer.selectMedia('audio');
+        let changed = false;
+
+        if (audioMedia && audioMedia.direction !== 'sendrecv') {
+            audioMedia.direction = 'sendrecv';
+            changed = true;
+        }
+        const videoMedia = transformer.selectMedia('video');
+
+        if (videoMedia && videoMedia.direction !== 'sendrecv') {
+            videoMedia.direction = 'sendrecv';
+            changed = true;
+        }
+
+        if (changed) {
+            localDescription.sdp = transformer.toRawSDP();
+        }
+    }
+};
+
+/**
  *
  * @param {JitsiLocalTrack} localTrack
  */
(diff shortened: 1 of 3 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1035,6 +1035,33 @@ const normalizePlanB = function(desc) {
 };
 
 /**
+ * Makes sure that both audio and video directions are configured as 'sendrecv'.
+ * @param {Object} localDescription the SDP object as defined by WebRTC.
+ */
+const enforceSendRecv = function(localDescription) {
```
bbaldino: so if there are no ssrcs in the mline, will chrome complain about setting it to sendrecv?

paweldomas: https://github.com/jitsi/lib-jitsi-meet/blob/6efe7de93d2042e171756f2018cc223e8c855563/modules/RTC/TraceablePeerConnection.js#L1475

bbaldino: makes sense but not sure i get that from the comment.  if i understand you right we're not setting it as 'sendrecv' when we actually do setLocalDescription in chrome, just making it 'sendrecv' when we get it out for sending to the other peer, yeah?  maybe the comment could be a bit clearer. (you're talking about the comment on line 1100 below, right?)

paweldomas: sure, I will try to come up with something better, thanks !

paweldomas: I've just force pushed, please let me know if it's better now

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1070,6 +1097,16 @@ const getters = {
...
         }
 
+        // What comes out of this getter will be signalled over Jingle to
+        // the other peer and we need to make sure that the media direction is
+        // 'sendrecv', because no changes are not signaled, after the initial
+        // offer/answer exchange.
+        //
+        // Note the this is not the effective media direction which will be set
+        // on the peerconnection - it's being adjusted to the correct value in
+        // the local description setter.
```
bbaldino: How about:

> 
> What comes out of this getter will be signalled over Jingle to the other peer, so we need to make sure the media direction is 'sendrecv' because we won't change the direction later and don't want the other peer to think we can't send or receive.
> 
> Note that the description we set in chrome does have the accurate direction (e.g. 'recvonly'), since that is technically what is happening.

PR #516, opened 2017-06-21
Expose the screen sharing source id

hristoterezov:


virtuacoplenny:
Would it be possible to split the jshint removal and the feature change itself into separate commits?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -15,696 +15,697 @@ import VideoType from '../../service/RTC/VideoType';
...
-    if (this._noDataFromSourceTimeout) {
-        clearTimeout(this._noDataFromSourceTimeout);
-        this._noDataFromSourceTimeout = null;
+    /**
+     * Sets the stream property of JitsiLocalTrack object and sets all stored
+     * handlers to it.
+     * @param {MediaStream} stream the new stream.
+     */
+    _setStream(stream) {
+        JitsiTrack.prototype._setStream.call(this, stream);
```
virtuacoplenny: I think you're also changing the functions a bit for es6. Am I correct assuming so? Could this be changed to a super._setStream call?

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -40,222 +44,236 @@ export default function JitsiRemoteTrack(
...
+            this.hasBeenMuted = true;
+        }
+
+        // we can have a fake video stream
+        if (this.stream) {
+            this.stream.muted = value;
+        }
+
+        this.muted = value;
+        this.eventEmitter.emit(JitsiTrackEvents.TRACK_MUTE_CHANGED, this);
```
virtuacoplenny: Do these these have a reference to this.eventEmitter? Would these calls be to this.emit?

.jshintignore
```diff
@@ -1,11 +0,0 @@
...
-# source code.
-doc/
-node_modules/
-lib-jitsi-meet.js
-lib-jitsi-meet.js.map
-lib-jitsi-meet.min.js
-
-# The following are checked by ESLint and take advantage of advanced language
-# features such as Facebook Flow which are not supported by JSHint.
-flow-typed/
```
virtuacoplenny: Is jshintrc still needed?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -46,28 +46,34 @@ function createMuteUnmutePromise(track, mute) {
...
+     * @param trackInfo.stream WebRTC MediaStream, parent of the track
+     * @param trackInfo.track underlying WebRTC MediaStreamTrack for new
+     * JitsiRemoteTrack
+     * @param trackInfo.mediaType the MediaType of the JitsiRemoteTrack
+     * @param trackInfo.videoType the VideoType of the JitsiRemoteTrack
+     * @param trackInfo.resolution the video resolution if it's a video track
+     * @param trackInfo.deviceId the ID of the local device for this track
+     * @param trackInfo.facingMode the camera facing mode used in getUserMedia
+     * call
      * @constructor
```
virtuacoplenny: You might want to add sourceId to the doc.

modules/RTC/RTCUtils.js
```diff
@@ -628,14 +628,16 @@ function handleLocalStream(streams, resolution) {
             videoStream = streams.video;
         }
 
-        // Again, different choices on different types of browser.
-        desktopStream = streams.desktopStream || streams.desktop;
+        desktopStream = streams.desktop;
```
virtuacoplenny: Why was "streams.desktopStream" needed in the past?

modules/RTC/JitsiRemoteTrack.js
hristoterezov: No. Will fix it!

.jshintignore
hristoterezov: Not really.

modules/RTC/RTCUtils.js
hristoterezov: We were naming the prop `desktopStream` for some browsers and for others `desktop`. I don't believe this was intentional.

hristoterezov:
@virtuacoplenny I think I addressed all comments!

package.json
```diff
@@ -39,7 +39,6 @@
     "eslint-plugin-import": "2.3.0",
     "flow-bin": "0.38.0",
     "jasmine-core": "2.5.2",
-    "jshint": "2.9.4",
```
virtuacoplenny: I noticed some inline jshint disabling still left behind. Maybe it's good to just remove those now?

hristoterezov: Fixed!

virtuacoplenny: I still see some references in adapter.screenshare. Were those part of the third-party library?

hristoterezov: Yes. It's provided by Temasys. We are updating it from time to time. It's interesting why it is part of our source code but unfortunately I don't know. 

Anyway this is whole different story :)

PR #517, opened 2017-06-22
fix(localStorage): exception if localStorage is disabled

hristoterezov:


modules/settings/Settings.js
```diff
@@ -13,8 +13,15 @@ function getLocalStorage() {
 
     // eslint-disable-next-line no-invalid-this
     const global = typeof window === 'undefined' ? this : window;
+    let storage;
 
-    return global.localStorage;
+    try {
+        storage = global.localStorage;
```
virtuacoplenny: I don't think this will catch the non-existence of localStorage across all browsers, specifically FireFox.

paweldomas: what was the exception here ? was it 'global' undefined or localStorage ?

paweldomas: Also this Exception will be thrown every time getLocalStorage is called.
Probably not a big deal at this point, but seems suboptimal to me.

hristoterezov: @paweldomas :

> what was the exception here ? was it 'global' undefined or localStorage ?
The exception is:
`Uncaught DOMException: Failed to read the 'localStorage' property from 'Window': Access is denied for this document.`

And it happens when you do `window.localStorage`.

> I'm guessing that it's 'global' then I don't see why try catch instead making a check

It is not "something is undefined exception" as you can see. That's why we cannot do a check. The only solution that I have in mind is the try...catch.

> Also this Exception will be thrown every time getLocalStorage is called.
Probably not a big deal at this point, but seems suboptimal to me.

This exception will be thrown every time getLocalStorage is called but only when the localStorage is disable. Are you proposing to store in a variable whether we can access localStorage or not and use it? That's not a bad idea actually!

paweldomas: ok I see, thanks for the explanation. About the variable I don't mind, only if you have time to add it.

PR #521, opened 2017-06-23
ref: Use STUN/TURN for the P2P connection.

gpolitis:
Please let me know if this breaks any previous functionality. I don't think we're using anywhere STUN/TURN for the JVB connection, but I may be wrong.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -356,7 +356,15 @@ class JingleConnectionPlugin extends ConnectionPlugin {
                     }
                     }
                 });
-                this.jvbIceConfig.iceServers = iceservers;
+
+                if (this.xmpp.options.useStunTurn) {
+                    this.jvbIceConfig.iceServers = iceservers;
+                }
+
+                if (this.xmpp.options.p2p.useStunTurn) {
+                    this.p2pIceConfig.iceServers = iceservers;
+                }
+
             }, err => {
                 logger.warn('getting turn credentials failed', err);
                 logger.warn('is mod_turncredentials or similar installed?');
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -133,7 +133,7 @@ export default class XMPP extends Listenable {
             now);
         if (status === Strophe.Status.CONNECTED
             || status === Strophe.Status.ATTACHED) {
-            if (this.options.useStunTurn) {
+            if (this.options.useStunTurn || this.options.p2p.useStunTurn) {
                 this.connection.jingle.getStunAndTurnCredentials();
             }
 


paweldomas:
I think Emil mentioned that someone is using TURN for JVB, so maybe it should be configurable.

gpolitis:
Thanks Pawel, I'll ask him offline.

gpolitis:
@paweldomas I introduced a new config.js variable called `useStunTurn4P2P` instead of re-using the old one.

Please do not merge yet as it's not ready, I need to update the connection info to show TURN or something like that.

modules/xmpp/strophe.jingle.js
```diff
@@ -36,6 +36,8 @@ class JingleConnectionPlugin extends ConnectionPlugin {
         if (Array.isArray(p2pStunServers)) {
             logger.info('Configured STUN servers: ', p2pStunServers);
             this.p2pIceConfig.iceServers = p2pStunServers;
+
+            // this.p2pIceConfig.iceTransportPolicy = 'relay';
```
saghul: leftover from debug?

saghul:
This PR introduces 2 more p2p specific settings, at the top level. This is getting out of hand. Since P2P is recent I'd **really** **really** like to see the following configuration structure:

```
p2p: {
    enabled: true,
    preferH264: true,
    stunServers: [...],
    useStunTurn: true
}
```

Fixing the entire config will take a while, but we can at least make the new settings better, now.

In addition, `useStunTurn` is a terrible name for a setting. Can we use `discoverStunTurn` for the new setting instead?

saghul:
One more thing. Once this lands, this needs to be updated: https://github.com/jitsi/jitsi-meet/blob/master/react/features/base/conference/actions.js#L272

saghul:
Also: DOCS. 📖

paweldomas:
I agree about the config structure, we should definitely put the new ones there. Ideally we would also mark the old options as deprecated and redirect them to the new ones for the time being.

saghul:
> Ideally we would also mark the old options as deprecated and redirect them to the new ones for the time being.

For old options, sure! For new ones such as these P2P related ones, I'd just move them right now.

saghul:
Folks, I'm going to make a PR replacing that preferH2644P2P with a P2P settings group. Once done I'll remove that part from this PR so @gpolitis can continue it forwards.

modules/xmpp/xmpp.js
```diff
@@ -133,7 +133,7 @@ export default class XMPP extends Listenable {
             now);
         if (status === Strophe.Status.CONNECTED
             || status === Strophe.Status.ATTACHED) {
-            if (this.options.useStunTurn) {
+            if (this.options.useStunTurn || this.options.useStunTurn4P2P) {
```
saghul: can you please move this to the p2p section?

gpolitis:
@saghul could you PTAL?

saghul:
Also, can you make a PR to jitsi-meet with the new option?

gpolitis:
por supuesto! https://github.com/jitsi/jitsi-meet/pull/1732

gpolitis:
I've asked Jenkins to test again, it looks like it's failing for some unrelated reason.

PR #524, opened 2017-06-27
feat(video): default to 720p

saghul:


saghul:
It's time. Some people who might not have updated config files would default to some non-glorious 360p. That's unacceptable.

PR #526, opened 2017-06-28
feat(p2p): refactor p2p configuration options

saghul:
Also add a preferH264 option for P2P.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -45,14 +45,14 @@ const logger = getLogger(__filename);
  * @param {number} [options.config.avgRtpStatsN=15] how many samples are to be
  * collected by {@link AvgRTPStatsReporter}, before arithmetic mean is
  * calculated and submitted to the analytics module.
- * @param {boolean} [options.config.enableP2P] when set to <tt>true</tt>
+ * @param {boolean} [options.config.p2p.enabled] when set to <tt>true</tt>
  * the peer to peer mode will be enabled. It means that when there are only 2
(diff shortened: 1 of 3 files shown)


JitsiConference.js
```diff
@@ -151,7 +151,8 @@ export default function JitsiConference(options) {
      */
     this.deferredStartP2PTask = null;
 
-    const delay = parseInt(options.config.backToP2PDelay, 10);
+    const delay
+        = parseInt(options.config.p2p && options.config.p2p.backToP2PDelay, 10);
```
paweldomas: I wish there was a nicer way to access these properties, but I can't think of any which would be simple so let it be...

saghul:
Will do! Can anyone merge when tests have run? Thanks!

paweldomas:
I suppose that the tests may actually fail, because those options are used there...

saghul:
Yep, let me fixup those....

damencho:
We had already pushed p2p into stable. Why not make the code to work with both configurations?
We had always try not to brake previous deployments. If we are braking it we try to announce it on the mailinglist, but as we are moving away from it ... maybe we should add a news update on the main site :)

saghul:
Pushed a new commit enabling P2P in absence of new settings, PTAL.

saghul:
Jenkins, test this please.

saghul:
Results are in! 💚 http://ci.jitsi.org/job/jitsi-meet-PR-test/2075/

PR #533, opened 2017-07-05
Refactor deployment info

bgrozev:


JitsiConference.js
```diff
@@ -1340,8 +1340,11 @@ JitsiConference.prototype.onIncomingCall
 
     // TODO: remove deprecated cross region property from this specific event
     // once all existing deployments include analytics changes
-    if (window.jitsiDeploymentInfo) {
-        crossRegion = window.jitsiDeploymentInfo.CrossRegion;
+    if (this.options.config.deploymentInfo) {
```
virtuacoplenny: Is it guaranteed this.options.config will be defined?

JitsiConference.js
```diff
@@ -1340,8 +1340,11 @@ JitsiConference.prototype.onIncomingCall
...
+        crossRegion
+            = this.options.config.deploymentInfo.region
+                    !== this.options.config.deploymentInfo.userRegion;
+        crossRegion = crossRegion ? 0 : 1;
```
bbaldino: why convert it to 0/1?

JitsiConference.js
bgrozev: Nope, good catch!

modules/xmpp/ChatRoom.js
```diff
@@ -148,6 +148,16 @@ export default class ChatRoom extends Listenable {
         // Jicofo makes decisions based on that. Initialize it with 'false'
         // here.
         this.addVideoInfoToPresence(false);
+
+        if (options.deploymentInfo && options.deploymentInfo.userRegion) {
+            this.presMap.nodes.push(
+                { 'tagName': 'region',
```
bbaldino: i don't know if we have a style for this, i feel like i typically see a newline after the opening brace of the object, but i don't know if we're consistent.  @virtuacoplenny ?

JitsiConference.js
bgrozev: Added checks (it's a bit ugly, let me know if you see a cleaner way to do it), and changed it to use the crossRegion property directly.

JitsiConference.js
bgrozev: Not sure, I'm just preserving the previous behavior.

modules/xmpp/ChatRoom.js
virtuacoplenny: What you said sounds correct. Normally when passing in a multi-attribute object it's something like:
```javascript
functionName({
  key1: val1,
  key2: val2
});
```

bgrozev: Fixed

JitsiConference.js
virtuacoplenny: I'm sorry I didn't see your last reply... It's fine the way it is. Here's an alternative way that I've used a lot (and can be used with destructuring) when I don't have a util library handy or don't want to create my own util:
```javascript
const obj2 = obj1 && obj1.obj2;
const obj3 = obj2 && obj3;
```
The main difference is you don't have to access obj1.obj2.obj3 at the end.

PR #535, opened 2017-07-06
ref: Refactors average statistics reporting.

gpolitis:
This commit changes the way the average stats are being reported to the
analytics backend. Instead of prefixes for the different scenarios
("p2p_", "jvb121_" and no prefix for multi-party calls) we now have
properties that allow to distinguish which scenario an event pertains
to. The query language of the analytics backend can be used to build
appropriate queries that group by those properties.

Additionally, we now report whether the average stats event concerns a
relayed call (a call that's routed via a TURN server) or not.

### DIFF
--- a/modules/statistics/RTPStatsCollector.js
+++ b/modules/statistics/RTPStatsCollector.js
@@ -453,6 +453,33 @@ StatsCollector.prototype.getNonNegativeStat = function(report, name) {
     return Math.max(0, value);
 };
 
+/**
+ * Determines whether the local ICE candidate address is a relayed one.
+ *
+ * @param {TraceablePeerConnection} rtcPeerConnection the peer connection that
+ * has the local ICE candidate address.
+ * @param {Array} localAddressArray an array in the form
+ * [ ip address, port ].
+ * @return {boolean} true if the local ICE candidate is a relayed one, otherwise
+ * false.
+ */
+function isRelayed(rtcPeerConnection, localAddressArray) {
+    const localSdpLines
+        = rtcPeerConnection.localDescription.sdp.match(/[^\r\n]+/g);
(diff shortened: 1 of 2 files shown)


modules/statistics/RTPStatsCollector.js
```diff
@@ -453,6 +453,27 @@ StatsCollector.prototype.getNonNegativeStat = function(report, name) {
     return Math.max(0, value);
 };
 
+/**
+ * @param rtcPeerConnection
```
paweldomas: Could you please describe this method in more details and add type declarations to JSDoc ?

modules/statistics/RTPStatsCollector.js
```diff
@@ -453,6 +453,31 @@ StatsCollector.prototype.getNonNegativeStat = function(report, name) {
     return Math.max(0, value);
 };
 
+/**
+ * Determines whether the local ICE candidate address is a relayed one.
+ *
+ * @param rtcPeerConnection the peer connection that has the local ICE candidate
```
paweldomas: @param {TraceablePeerConnection} rtcPeerConnection

modules/xmpp/strophe.jingle.js
```diff
@@ -36,6 +36,8 @@ class JingleConnectionPlugin extends ConnectionPlugin {
         if (Array.isArray(p2pStunServers)) {
             logger.info('Configured STUN servers: ', p2pStunServers);
             this.p2pIceConfig.iceServers = p2pStunServers;
+
+            // this.p2pIceConfig.iceTransportPolicy = 'relay';
```
paweldomas: is it intended or some debug leftover ?

PR #539, opened 2017-07-10
Treats local/remote relayed addresses differently.

gpolitis:
A call isn't just relayed. It may be the case that the local participant is using a relayed address, while the remote one is directly connected to the TURN server. We add a 'remoterelayed' and 'localrelayed' in the stats that holds this piece of information.

### DIFF
--- a/modules/statistics/AvgRTPStatsReporter.js
+++ b/modules/statistics/AvgRTPStatsReporter.js
@@ -17,8 +17,10 @@ import * as VideoType from '../../service/RTC/VideoType';
  *
  * {
  *   p2p: true,
- *   conferenceSize: 1,
- *   relayed: true,
+ *   conferenceSize: 2,
+ *   localCandidateType: "relay",
+ *   remoteCandidateType: "relay",
+ *   transportType: "udp",
  *
  *   "stat_avg_rtt": {
  *     value: 200,
@@ -92,18 +94,12 @@ class AverageStatReport {
      * Appends the report to the analytics "data" object. The object will be
      * set under <tt>prefix</tt> + {@link this.name} key.
      * @param {Object} report the analytics "data" object
-     * @param {Object} extra properties to append to the analytics "data" object
      */
-    appendReport(report, props) {
-
-        const reportValue = {
+    appendReport(report) {
+        report[this.name] = {
             value: this.calculate(),
             samples: this.samples
         };
-
-        Object.assign(reportValue, props);
-
-        report[this.name] = reportValue;
     }
 
     /**
@@ -125,12 +121,12 @@ class AverageStatReport {
 class ConnectionAvgStats {
     /**
      * Creates new <tt>ConnectionAvgStats</tt>
-     * @param {JitsiConference} conference
+     * @param {AvgRTPStatsReporter} avgRtpStatsReporter
      * @param {boolean} isP2P
      * @param {number} n the number of samples, before arithmetic mean is to be
      * calculated and values submitted to the analytics module.
      */
-    constructor(conference, isP2P, n) {
+    constructor(avgRtpStatsReporter, isP2P, n) {
         /**
          * Is this instance for JVB or P2P connection ?
          * @type {boolean}
@@ -175,13 +171,26 @@ class ConnectionAvgStats {
          * @type {JitsiConference}
          * @private
          */
-        this._conference = conference;
+        this._avgRtpStatsReporter = avgRtpStatsReporter;
+
+        /**
+         * The latest average E2E RTT for the JVB connection only.
+         *
+         * This is used only when {@link ConnectionAvgStats.isP2P} equals to
+         * <tt>false</tt>.
+         *
+         * @type {number}
+         */
+        this._avgEnd2EndRTT = undefined;
 
         this._onConnectionStats = (tpc, stats) => {
             if (this.isP2P === tpc.isP2P) {
                 this._calculateAvgStats(stats);
             }
         };
+
+        const conference = avgRtpStatsReporter._conference;
+
         conference.statistics.addConnectionStatsListener(
             this._onConnectionStats);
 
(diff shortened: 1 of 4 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -307,6 +307,32 @@ const dumpSDP = function(description) {
...
+ *
+ * @param {RTCSessionDescription} rtcSessionDescription the session description
+ * that has the local ICE candidate address.
+ * @param {Array} addressArray an array in the form [ ip address, port ].
+ * @param {String} candidateType the candidate type: One of "host", "srflx",
+ * "prflx", or "relay"
+ * @return {boolean} true if the ICE candidate is a relayed one, otherwise
+ * false.
+ */
+function isRelay(rtcSessionDescription, addressArray, candidateType) {
```
paweldomas: For me it seems that this method verifies if given address exists in given SDP. Why is it called isRelay and why does it take candidateType if it's supposed to check whether given ICE candidate is a relay one ? Am I missing something ?

gpolitis: You're right Pawel, I'll fix that

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2008,6 +2034,36 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function(track) {
...
+ * @param {Array} addressArray an array in the form [ ip address, port ].
+ * @param {String} candidateType the candidate type: One of "host", "srflx",
+ * "prflx", or "relay"
+ * @return {boolean} true if the ICE candidate is a relayed one, otherwise
+ * false.
+ */
+TraceablePeerConnection.prototype.checkRemoteCandidateType
+= function(addressArray, candidateType) {
+    return isRelay(
+        this.peerconnection.remoteDescription, addressArray, candidateType);
```
paweldomas: As per our discussion offline, it would be good to add a comment here which will say that we explicitly go around 'remoteDescription' getter ('localDescription' in the other method) in order to go around SDP transformation chain since local addresses are not modified currently anyway. Just to warn others to not change it the other way by mistake.

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -61,6 +61,7 @@ class AverageStatReport {
         this.count = 0;
         this.sum = 0;
         this.samples = [];
+        this.latestAvg = undefined;
```
paweldomas: What's the purpose of this field ? Do you use it for debugging ?

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -230,21 +244,32 @@ class ConnectionAvgStats {
...
+                            value: avgRTTDiff
                         };
+                    }
+                } else {
+                    // Report end to end RTT only for JVB.
+                    const avgRemoteRTT = this._calculateAvgRemoteRTT();
+                    const avgLocalRTT = this._avgRTT.calculate();
 
-                        Object.assign(reportValue, props);
+                    this._avgEnd2EndRTT = avgLocalRTT + avgRemoteRTT;
```
gpolitis: It's a little bit tricky because it's populated by the `jvbStatsMonitor` but it's accessed by the `p2pStatsMonitor`.

modules/statistics/AvgRTPStatsReporter.js
gpolitis: It's also used at line 254 in order to calculate the average RTT difference between the P2P connection and the JVB RTT (end-to-end).

paweldomas: why don't you just call calculate it's the same thing

gpolitis: You're right, I'll adjust this!

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -230,21 +244,32 @@ class ConnectionAvgStats {
 
                 this._avgRTT.appendReport(batchReport, props);
 
-                // Report end to end RTT only for JVB
-                if (!this.isP2P) {
-                    const avgRemoteRTT = this._calculateAvgRemoteRTT();
-                    const avgLocalRTT = this._avgRTT.calculate();
+                if (this.isP2P) {
+                    // Report RTT diff only for P2P.
+                    const jvbEnd2EndRTT = this._conference.avgRtpStatsReporter
```
paweldomas: do we have to access it through the conference object ?
Maybe we could make 'AvgRTPStatsReporter' a parent (pass in the constructor) then this will give access to _conference and other stuff we may need in future

gpolitis: I see what you mean, I finally opted for the less invasive (but dirtier) approach, but I can gladly change it, if that's what you'd rather see here. Would you like me to proceed with the change?

paweldomas: yes, please

gpolitis: 👍

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -95,16 +96,11 @@ class AverageStatReport {
      * @param {Object} report the analytics "data" object
      * @param {Object} extra properties to append to the analytics "data" object
```
paweldomas: this argument has been removed, right ?

gpolitis: Correct, thanks Pawel, I'll remove the JSDoc.

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -223,26 +219,33 @@ class ConnectionAvgStats {
             return;
         }
 
+        if (!data.transport || !data.transport.length) {
```
paweldomas: in what case there is no transport ? I wonder if the stats should not be cleared in such case, otherwise the avg may not be accurate (if there's no transport for too long)

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -223,26 +219,33 @@ class ConnectionAvgStats {
...
         if (this._sampleIdx >= this._n) {
             if (RTCBrowserType.supportsRTTStatistics()) {
-                const batchReport = { };
                 const conference = this._avgRtpStatsReporter._conference;
-                const props = {
-                    localrelayed: data.localrelayed,
-                    remoterelayed: data.remoterelayed,
+
+                const batchReport = {
+                    localrelayed: transport.localrelayed,
```
paweldomas: is localrelayed/remoterelayed an IP or boolean ?

gpolitis: it's a boolean, I should probably document this somewhere.

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -564,14 +567,22 @@ export default class AvgRTPStatsReporter {
...
     _calculateAvgStats(data) {
 
+        if (!data) {
+            logger.error('No stats');
+
+            return;
+        }
+
+        if (!data.transport || !data.transport.length) {
+            logger.error('No transport.');
```
paweldomas: There are things other than transport being calculated here, in such case only transport related stuff should be omitted (values will be NaN and not reported).
This code is unaware when the transport will be unavailable - can not assume that lack of transport means lack of all other stats.

modules/statistics/AvgRTPStatsReporter.js
gpolitis: not only that, but it's not clear why the transport needs to be an array.. In my experiments I haven't observed any cases where the transport is not set to something.

paweldomas: it is assumed an array, because it's created as such in RTPStatsCollector

modules/statistics/AvgRTPStatsReporter.js
paweldomas: What I mean is that some stats are available on one browser while missing on another

modules/statistics/AvgRTPStatsReporter.js
paweldomas: having this return here makes me feel that if we add something else below that does not rely on transport we may have a bug

modules/statistics/AvgRTPStatsReporter.js
gpolitis: I see, I'll remove that then. The motivation behind this error/return was that the transport is a determining factor for all the stats that are being logged and we shouldn't mix TCP with UDP. We can adjust the queries to explicitly include the transport type, so even if we log stats without transport, we can ignore them (for now at least).

modules/statistics/AvgRTPStatsReporter.js
gpolitis: Right, I was just wondering why that's the case. Probably for the non-bundle case, which is no longer used.

paweldomas:
jenkins test this please

PR #542, opened 2017-07-11
feat: Always enables googSuspendBelowMinBitrate for p2p.

bgrozev:


modules/RTC/RTCUtils.js
```diff
@@ -872,6 +865,21 @@ class RTCUtils extends Listenable {
                         return this.audioTracks;
                     };
                 }
+
+                this.p2pPcConstraints
+                    = JSON.parse(JSON.stringify(this.pcConstraints));
+
+                // Allows sending of video to be suspended if the bandwidth
+                // estimation is too low.
+                if (!options.disableSuspendVideo) {
```
saghul: shouldn't this setting now go to the p2p section? or be removed altogether, if we'd like to have it on always?

bgrozev: No, the option affects the bridge connection. For p2p we enable it unconditionally.

saghul: Doh, ok, I missread, sorry!

saghul:
Hopefully atlaskit is fixed soon so the CI works, I'll merge this then.

bgrozev:
jenkins test this please

PR #544, opened 2017-07-12
fix(JingleSessionPC): avoid failures if replaceTrack runs late

saghul:
Due to it being ran asyncronously, it's possible a replaceTrack operation is
scheduled while there is a valid PC, but by the time it runs was closed.

Accommodate to this situation by pretending everything is just peachy.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -1576,6 +1576,18 @@ export default class JingleSessionPC extends JingleSession {
      */
     replaceTrack(oldTrack, newTrack) {
         const workFunction = finishedCallback => {
+            // Check if the connection was closed and pretend everything is OK.
+            // This can happen if a track removal is scheduled but takes place
+            // after the connection is closed.
+            if (this.peerconnection.signalingState === 'closed'
+                || this.peerconnection.connectionState === 'closed'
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -1576,6 +1576,18 @@ export default class JingleSessionPC extends JingleSession {
      */
     replaceTrack(oldTrack, newTrack) {
         const workFunction = finishedCallback => {
+            // Check if the connection was closed and pretend everything is OK.
+            // This can happen if a track removal is scheduled but takes place
+            // after the connection is closed.
+            if (this.peerconnection.signalingState === 'closed'
```
paweldomas: Maybe it would be better to use _assertNotEnded in such cases instead of checking on the closed state.

paweldomas: But probably 'onTerminated' needs to be executed on the queue, so maybe your version is better for now

bbaldino: it seems to me that we might have this same problem in multiple places, @saghul did you happen to notice it anywhere else?

saghul: Nope, just here.

PR #546, opened 2017-07-12
add method to send a receiver video constraint message

bbaldino:


modules/RTC/BridgeChannel.js
```diff
@@ -170,6 +170,20 @@ export default class BridgeChannel {
     }
 
     /**
+     * Sends a "receiver video constraint" message via the channel.
+     * @param {Number} maxFrameHeightPixels the maximum fram height,
+     * in pixels, this receiver is willing to receive
+     */
+    sendReceiverVideoConstraintMessage(maxFrameHeightPixels) {
+        logger.log('sending a ReceiverVideoConstraint message with '
```
virtuacoplenny: Where would the code be that calls this?

bbaldino: the way i see it, there would be a UI element (the quality slider) that would call some logic upon changing which would translate the value of the slider into a constraint

virtuacoplenny: Would you need more code then to make sure the jitsi-meet UI can easily access this method through a public api? I'm not sure if at this point BridgeChannel is well exposed.

bbaldino: potentially, though i guess i saw that as part of the ui plumbing work (separate from this)

modules/RTC/BridgeChannel.js
```diff
@@ -170,6 +170,20 @@ export default class BridgeChannel {
     }
 
     /**
+     * Sends a "receiver video constraint" message via the channel.
+     * @param {Number} maxFrameHeightPixels the maximum fram height,
+     * in pixels, this receiver is willing to receive
+     */
+    sendReceiverVideoConstraintMessage(maxFrameHeightPixels) {
```
gpolitis: Would you mind removing the "receiver" part from the message name? I can't imagine having a `SendVideoConstraintMessage`.

modules/RTC/BridgeChannel.js
```diff
@@ -170,6 +170,20 @@ export default class BridgeChannel {
     }
 
     /**
+     * Sends a "receiver video constraint" message via the channel.
+     * @param {Number} maxFrameHeightPixels the maximum fram height,
```
gpolitis: s/fram/frame

modules/RTC/BridgeChannel.js
bbaldino: i actually think being explicit that this is a receiver constraint is useful (even if we don't ever plan to have a send video constraint message, i still think it's nice to be more descriptive).  do you feel strongly about removing it?

modules/RTC/BridgeChannel.js
bbaldino: will fix this

modules/RTC/BridgeChannel.js
gpolitis: Kind of strong, unless you feel even stronger about not removing it :) We don't prepend lastN, nor selected endpoint, for instance, and the idea there is exactly the same.

bbaldino: but for those messages there are no possible ambiguities, are there?  i think 'VideoConstraintMessage' is too vague when, with things around o/a, we have 'constraints' being expressed from both sides.

bbaldino: i guess i'm not sure why we'd want to add ambiguity here.  the method is for sending a receive-side video constraint, so why not name it as such?

PR #548, opened 2017-07-13
feat: add 'disableDesktopSharing' option

paweldomas:


doc/API.md
```diff
@@ -39,23 +39,24 @@ You can access the following methods and objects trough ```JitsiMeetJS``` object
...
+    10. desktopSharingFirefoxExtensionURL - The URL to the Firefox extension for desktop sharing. "null" if no extension is required.
+    11. disableAudioLevels - boolean property. Enables/disables audio levels.
+    12. disableSimulcast - boolean property. Enables/disables simulcast.
+    13. enableWindowOnErrorHandler - boolean property (default false). Enables/disables attaching global onerror handler (window.onerror).
+    14. disableThirdPartyRequests - if true - callstats will be disabled and the callstats API won't be included.
+    15. enableAnalyticsLogging - boolean property (default false). Enables/disables analytics logging.
+    16. callStatsCustomScriptUrl - (optional) custom url to access callstats client script
+    17. callStatsConfIDNamespace - (optional) a namespace to prepend the callstats conference ID with. Defaults to the window.location.hostname
+    18. disableRtx - (optional) boolean property (default to false).  Enables/disable the use of RTX.
+    19. preferH264 - (optional) boolean property (default to false).  Enables/disable preferring the first instance of an h264 codec in an offer by moving it to the front of the codec list.
```
virtuacoplenny: In markdown the numbers don't have to be in order to get a numbered list. They could all be "1." for example. Do you think that's worthwhile? Maybe it would be confusing when looking at the non-formated markdown.

paweldomas: I didn't know that, that's interesting. I can change it - yes

paweldomas: I've tried that and it appears as I, II, III

Actually I don't get why numbers... order is irrelevant there. Better change it to dots ?

modules/RTC/ScreenObtainer.js
```diff
@@ -76,11 +76,9 @@ const ScreenObtainer = {
         this.options = options = options || {};
         gumFunction = gum;
 
-        if (RTCBrowserType.isFirefox()) {
-            initFirefoxExtensionDetection(options);
-        }
-
-        if (RTCBrowserType.isNWJS()) {
+        if (this.options.disableDesktopSharing) {
```
yanas: It's probably very subjective, but we only have those lines at the end of the if-else sequence : 
if (!obtainDesktopStream) {
logger.info('Desktop sharing disabled');
}
this.obtainStream = obtainDesktopStream;
so I think it would improve code readability to handle the the disable case right here, log whatever we need to, assign null to obtain stream if needed and return. It's just weird to me to have if () { nothing }.

paweldomas: but then all weird extension detection code will execute unnecessary

paweldomas: I will think about something tomorrow

yanas: I was actually thinking of having something like:

if (this.options.disableDesktopSharing) {
logger.info('Desktop sharing disabled');
this.obtainStream = null;
return;
}

before all the ifs starting with if (RTCBrowserType.isNWJS()) ...

paweldomas: I had the same thoughts as you when trying to add this initially, but it's not as simple as it seems.

If I add this at the very beginning with some return statement then I will print the message about desktop sharing disabled twice, because it can be disabled during this 'if else if' chain execution (disabled on one browser).

If I add this at the end then all extension detection will be executed for nothing (possibly registering some handlers - I don't know if it's done now, but can be added later and we'll have bugs).

So I found it's the easies to abort this long else if chain, but I can wrap that long "if else if" into separate method and maybe something better will come out of that.

PR #549, opened 2017-07-14
Currently available devices fix

bbaldino:


bbaldino:
this is technically a change in the api (as it will no longer return undefined), but i checked the jitsi usages and this doesn't break anything (and in fact will fix a few places which assume `undefined` will never be returned).  the only logic this does change is if we only want `pollForAvailableMediaDevices` to fill out `currentlyAvailableMediaDevices` **once**, it seems to me that we wouldn't want that, but i dunno.

bbaldino:
i think it's inconvenient that this exposes an `undefined` state (and in fact there are many callers of the getter for currentlyAvailableMediaDevices that do not handle `undefined`) and this code looks like it is in need of a larger refactor (to use the event-based device-changed notification instead of polling), but i'm not going to touch it for now.

PR #550, opened 2017-07-15
Allows unavailable presences to contain a ignore tag.

damencho:
If we detect a custom <ignore tag in presence unavailable we ignore those presences and do not process them.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -618,6 +618,11 @@ export default class ChatRoom extends Listenable {
      * @param from
      */
     onPresenceUnavailable(pres, from) {
+        // ignore presence
+        if ($(pres).find('>ignore[xmlns="http://jitsi.org/jitmeet/"]').length) {
+            return true;
+        }
+
         // room destroyed ?
         if ($(pres).find('>x[xmlns="http://jabber.org/protocol/muc#user"]'
             + '>destroy').length) {


bbaldino:
just curious, what's the use case for this?

damencho:
Poltergeist, a participant joins the room and stays there just before the real participant joins. As both use the same jid in the xmpp server, the nicks we construct to join mucs are the same. So in order to avoid unnecessary leave/join messages in the UI, the unavailable presence for the poltergeist is marked to be ignored by the client and this presence is sent in the moment of the real user authenticates into the server, a moment before joining.

bbaldino:
gotcha, thanks.  does the poltergeist have to join the muc at all?

modules/xmpp/ChatRoom.js
```diff
@@ -618,6 +618,11 @@ export default class ChatRoom extends Listenable {
      * @param from
      */
     onPresenceUnavailable(pres, from) {
+        // ignore presence
+        if ($(pres).find('>ignore[xmlns="http://jitsi.org/jitmeet/"]').length) {
```
paweldomas: do we also use this tag to ignore user in Jicofo ?

damencho:
No but we don't need that logic in jocofo. This is just for the ui. The fact that we do not ignore it make it work, without any modifications in jicofo. Otherwise we need to not invite poltergeist and wait for its advertised features to change with supporting jingle in order to start inviting it.

paweldomas:
So there are no changes in Jicofo for that or do you have a PR ? I don't see how this can work without any changes in Jicofo, but maybe I'm missing something.

damencho:
The fact that someone leaves and then enters again makes it work in jicofo. It is just a prosody module handling this logic with minimum changes in the client (this PR). This PR just makes the UI experience better, just skips the notifications about someone leaving/entering.

paweldomas:
ok I see. Yes it sounds like it should work in Jicofo this way

PR #551, opened 2017-07-19
fix: Disable simulcast when H264 is preferred.

gpolitis:


modules/RTC/TraceablePeerConnection.js
```diff
@@ -371,8 +371,7 @@ TraceablePeerConnection.prototype.isSimulcastOn = function() {
         // doesn't have a reference to the config options, so we have
         // to do it here
         && (!RTCBrowserType.isFirefox()
-            || this.options.enableFirefoxSimulcast)
-        && !this.isP2P;
+            || this.options.enableFirefoxSimulcast);
```
bbaldino: why get rid of the check for p2p in `isSimulcastOn`? we should not be doing simulcast in p2p, no?

gpolitis: You're right, we should not be doing simulcast in p2p. I moved the check in JingleSessionPC (in this commit).

PR #552, opened 2017-07-20
feat(FirefoxSimulcast): move option to testing group

saghul:


saghul:
Documented in https://github.com/jitsi/jitsi-meet/pull/1806

@gpolitis @bbaldino can you PTAL?

saghul:
The code is already in, so I guess we might as well have it proper, right?

bbaldino:
well i was about to revert it all today, but i think i'm gonna push forward with it.  but still not sure it's worth documenting in its current state.

saghul:
But you already merged the PR which documents this option 😮

bbaldino:
yeah, didn't think about this until after.  i agree we should either also merge this or not merge this and revert the other.  what do you think with regards to the policy of when a feature flag should be documented, even in the 'testing' section?  do you think something in this state qualifies?

saghul:
I'm not 100% sure of the state, but IIRC I read something about it being experimental (please do correct me if I'm wrong here) hence why I thought it belongs in testing.

PR #554, opened 2017-07-21
firefox 54 crashes if we insert the rid simulcast receive lines in th…

bbaldino:
…e remote description, so only do that if simulcast is turned on (which is still experimental for firefox and not enabled by default)


bbaldino:
@damencho can we test this on beta?  for some reason even if i turned firefox sim on i couldn't reproduce the firefox crash on my dev server.

damencho:
Well, testing on beta is after merging it. 
But as I have tested this one a deployment made to catch this problem, I will merge it.

PR #555, opened 2017-07-24
feat(quality-slider): expose video constraint method and data channel…

virtuacoplenny:
… opening

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -438,7 +438,7 @@ JitsiConferenceEventManager.prototype.setupRTCListeners = function() {
         Statistics.analytics.sendEvent('conference.dataChannel.open',
             { value: now });
 
-        conference.eventEmitter.emit(JitsiConferenceEvents.DATA_CHANNEL_OPEN);
+        conference.eventEmitter.emit(JitsiConferenceEvents.DATA_CHANNEL_OPENED);
     });
 
     rtc.addListener(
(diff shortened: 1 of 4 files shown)


JitsiConferenceEventManager.js
```diff
@@ -437,6 +437,8 @@ JitsiConferenceEventManager.prototype.setupRTCListeners = function() {
         conference.room.connectionTimes['data.channel.opened'] = now;
         Statistics.analytics.sendEvent('conference.dataChannel.open',
             { value: now });
+
+        conference.eventEmitter.emit(JitsiConferenceEvents.DATA_CHANNEL_OPEN);
```
bbaldino: is there already an event for the websocket version?

virtuacoplenny: They should go through the same abstraction. Both peerconnection and websocket go through BridgeChannel's _handleChannel(), which sets the channel.onopen callback, which should fire RTCEvents.DATA_CHANNEL_OPEN, which conference event manager reacts to.

bbaldino: ah that's a bummer...this event should've been renamed then it seems like.

JitsiConferenceEvents.js
```diff
@@ -57,6 +57,11 @@ export const CONNECTION_INTERRUPTED = 'conference.connectionInterrupted';
 export const CONNECTION_RESTORED = 'conference.connectionRestored';
 
 /**
+ * A connection to the video bridge's data channel has been established.
+ */
+export const DATA_CHANNEL_OPEN = 'conference.dataChannelOpen';
```
yanas: We can rename this to DATA_CHANNEL_OPENED for consistency.

virtuacoplenny: That would make it inconsistent with RTCEvents.DATA_CHANNEL_OPEN. I'm not opposed to changing it, though.

yanas: I see. I still would prefer to make it consistent with the rest of the events in this class. Even though it's unfortunate we haven't followed the consistency in RTCEvents.

PR #556, opened 2017-07-25
Start with audio/video muted

paweldomas:
Required for https://github.com/jitsi/jitsi-meet/pull/1819

### DIFF
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -214,11 +214,18 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         console.log(`(TIME) Render ${type}:\t`, now);
         this.conference.getConnectionTimes()[`${type}.render`] = now;
 
+        // The conference can be started without calling GUM
+        // FIXME if there would be a module for connection times this kind
+        // of logic (gumDuration or ttfm) should end up there
+        const gumStart = window.connectionTimes['obtainPermissions.start'];
+        const gumEnd = window.connectionTimes['obtainPermissions.end'];
+        const gumDuration
+            = !isNaN(gumEnd) && !isNaN(gumStart) ? gumEnd - gumStart : 0;
+
         const ttfm = now
             - (this.conference.getConnectionTimes()['session.initiate']
             - this.conference.getConnectionTimes()['muc.joined'])
-            - (window.connectionTimes['obtainPermissions.end']
-            - window.connectionTimes['obtainPermissions.start']);
+            - gumDuration;
 
         this.conference.getConnectionTimes()[`${type}.ttfm`] = ttfm;
         console.log(`(TIME) TTFM ${type}:\t`, ttfm);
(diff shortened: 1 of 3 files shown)


modules/RTC/JitsiRemoteTrack.js
```diff
@@ -214,11 +214,18 @@ export default class JitsiRemoteTrack extends JitsiTrack {
...
 
+        // The conference can be started without calling GUM
+        // FIXME if there would be a module for connection times this kind
+        // of logic (gumDuration or ttfm) should end up there
+        const gumStart = window.connectionTimes['obtainPermissions.start'];
+        const gumEnd = window.connectionTimes['obtainPermissions.end'];
+        const gumDuration
+            = !isNaN(gumEnd) && !isNaN(gumStart) ? gumEnd - gumStart : 0;
+
         const ttfm = now
```
bbaldino: i know this was existing, but what does `ttfm` stand for?  could it have a more descriptive name?

paweldomas: time to first media

modules/RTC/RTCUtils.js
```diff
@@ -811,17 +764,9 @@ class RTCUtils extends Listenable {
                 const getUserMedia
                     = navigator.webkitGetUserMedia.bind(navigator);
 
-                if (navigator.mediaDevices) {
-                    this.getUserMedia = wrapGetUserMedia(getUserMedia);
-                    this.enumerateDevices
-                        = wrapEnumerateDevices(
-                            navigator.mediaDevices.enumerateDevices.bind(
-                                navigator.mediaDevices));
-                } else {
```
hristoterezov: We are not supporting anymore browsers where GUM is with callbacks or this is removed by mistake?

modules/RTC/RTCUtils.js
```diff
@@ -888,10 +833,7 @@ class RTCUtils extends Listenable {
                         navigator.mediaDevices.getUserMedia.bind(
                             navigator.mediaDevices),
                             true);
-                this.enumerateDevices
-                    = wrapEnumerateDevices(
-                        navigator.mediaDevices.enumerateDevices.bind(
-                            navigator.mediaDevices));
+                this.enumerateDevices = rawEnumerateDevicesWithCallback;
```
hristoterezov: `rawEnumerateDevicesWithCallback` can be `undefined`.  Should we handle this somehow?

The old code will fail too if enumerateDevices is not defined. So this is not a regression for sure. I've just noticed it and decided to ask.

modules/RTC/RTCUtils.js
paweldomas: That's what wrapGetUserMedia is doing isn't it ?

modules/RTC/RTCUtils.js
paweldomas: It can be undefined but that's not what this PR is about. It was the same before, unless I'm missing something ?

PR #558, opened 2017-07-27
Strophe.js dependency removed

webintoit:
AFAIK, Strophe.js is required by  `jitsi-meet`, and never imported here, so it can be removed.


saghul:
No, it's the other way around.

PR #561, opened 2017-08-01
Send resolution (in pixels) to the analytics backend.

gpolitis:


modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -516,6 +516,40 @@ export default class AvgRTPStatsReporter {
...
+        this._avgRemotePixels
+            = new AverageStatReport('stat_avg_pixels_remote');
+
+        /**
+         * Average pixels for remote screen streaming videos (reported only if
+         * not a <tt>NaN</tt>).
+         * @type {AverageStatReport}
+         * @private
+         */
+        this._avgRemoteScreenPixels
```
bbaldino: does this refer to screenshare vs the other being camera?  looks like the comment needs to be tweaked for one of them.  also, i'd prefer naming the other one something more descriptive (`_avgRemoteCameraPixels`? and maybe this one `_avgRemoteScreensharePixels`? though this one isn't too bad as is)

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -516,6 +516,40 @@ export default class AvgRTPStatsReporter {
...
+            = new AverageStatReport('stat_avg_pixels_screen_remote');
+
+        /**
+         * Average pixels for local video (camera)
+         * @type {AverageStatReport}
+         * @private
+         */
+        this._avgLocalPixels = new AverageStatReport('stat_avg_pixels_local');
```
bbaldino: `_avgLocalCameraPixels`? it'd make it more obvious when reading it in code (instead of having to go back to the comment)

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -695,6 +761,102 @@ export default class AvgRTPStatsReporter {
     }
 
     /**
+     * Calculates average number of pixels for the report
+     *
+     * @param {go figure} resolution
```
bbaldino: `go figure`?

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -695,6 +761,102 @@ export default class AvgRTPStatsReporter {
...
+     * @param {boolean} isLocal if the average is to be calculated for the local
+     * video or <tt>false</tt> if for remote videos.
+     * @param {VideoType} videoType
+     * @return {number|NaN} average number of pixels or <tt>NaN</tt> if there
+     * are no samples.
+     * @private
+     */
+    _calculateAvgVideoPixels(resolution, isLocal, videoType) {
```
bbaldino: it looks like `resolution` is a map?  named this way i assumed it was a single resolution.  maybe the arg should be `peerResolutions` or something?

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -695,6 +761,102 @@ export default class AvgRTPStatsReporter {
...
+     * @param {object} videos maps resolution per video SSRC
+     * @param {JitsiParticipant|null} participant remote participant or
+     * <tt>null</tt> for local video pixels calculation.
+     * @param {VideoType} videoType the type of the video for which an average
+     * will be calculated.
+     * @return {number|NaN} average video pixels of all participant's videos or
+     * <tt>NaN</tt> if currently not available
+     * @private
+     */
+    _calculatePeerAvgVideoPixels(videos, participant, videoType) {
```
bbaldino: if this is used for both local and remote, it seems odd to name it something with "peer" in it (since it handles local, non-peer average as well)

bbaldino:
also, is the average number of total pixels as useful as, say, the discrete frame heights/widths over time (a la webrtc-internals)?

gpolitis:
@bbaldino logging independently makes sense and I like the idea but I would leave that as a future improvement.

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -695,6 +762,102 @@ export default class AvgRTPStatsReporter {
...
+     *
+     * @param {map} peerResolutions a map of peer resolutions
+     * @param {boolean} isLocal if the average is to be calculated for the local
+     * video or <tt>false</tt> if for remote videos.
+     * @param {VideoType} videoType
+     * @return {number|NaN} average number of pixels or <tt>NaN</tt> if there
+     * are no samples.
+     * @private
+     */
+    _calculateAvgVideoPixels(peerResolutions, isLocal, videoType) {
```
bbaldino: so this method is basically just filtering the set of resolutions to process, right?  it's the other method that is doing the actual averaging?

also, why do we even need the filtering here?  every time we call it for remote, we also call it for local...so why not just do the processing all at once?

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -646,6 +686,24 @@ export default class AvgRTPStatsReporter {
                     frameRate, true /* local */, VideoType.DESKTOP));
         }
 
+        if (resolution) {
+            this._avgRemoteCameraPixels.addNext(
```
bbaldino: what is the scope of this...is it always tracking whichever camera stream is on the main stage?

gpolitis: It's computing an average over all remote participants. We're going to use it for drawing conclusions in 1:1 calls (so it doesn't really matter whether or not the average is a good aggregator function or not).

bbaldino: oh, so these are only used for 1:1?

gpolitis:
Also, about logging height and widths instead of pixels, it won't work because avg(height) * avg(width) != avg(height*width), for example:

(360+720) / 2 * (640+1280)/2 = 518400
(360*640 + 720*1280) / 2 = 576000

The first formula is not the average resolution, which is what we want to measure.

modules/statistics/AvgRTPStatsReporter.js
gpolitis: This method is an adaption of what's already there for the FPS. I'm not so enthusiast to change it, since what's there appears to be working well, unless of course you see a problem with the current implementation.

bbaldino:
@gpolitis sorry, what i meant was logging discrete values in height and width over time, like webrtc-internals does.  (so no averaging).

gpolitis:
The thing is that in the end we want to produce resolution averages.

PR #564, opened 2017-08-03
feat(JitsiLocalTrack): make mute/unmute more resilient

saghul:
Instead of failing with an error if another operation is in progress, which is
impossible to know as a user, chain them using the primises API.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -16,29 +16,20 @@ import VideoType from '../../service/RTC/VideoType';
 const logger = getLogger(__filename);
 
 /**
- * Creates Promise for mute/unmute operation.
+ * Creates Promise for mute/unmute operation. Mute operations are chained, so
+ * this function can be called as many times as needed in a row and operations
+ * will be executed in a serialized fashion.
  *
  * @param {JitsiLocalTrack} track - The track that will be muted/unmuted.
  * @param {boolean} mute - Whether to mute or unmute the track.
  * @returns {Promise}
  */
 function createMuteUnmutePromise(track, mute) {
-    if (track.inMuteOrUnmuteProgress) {
-        return Promise.reject(
-            new JitsiTrackError(JitsiTrackErrors.TRACK_MUTE_UNMUTE_IN_PROGRESS)
-        );
-    }
+    const doMute = () => track._setMute(mute);
 
-    track.inMuteOrUnmuteProgress = true;
+    track._mutePromise = track._mutePromise.then(doMute, doMute);
 
-    return track._setMute(mute)
-        .then(() => {
-            track.inMuteOrUnmuteProgress = false;
-        })
-        .catch(status => {
-            track.inMuteOrUnmuteProgress = false;
-            throw status;
-        });
+    return track._mutePromise;
 }
 
 /**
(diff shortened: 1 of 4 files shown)


saghul:
@virtuacoplenny I dragged you here because @paweldomas told me you had thought of / attempted this and found some issues. Can you PTAL?

saghul:
Needed for https://github.com/jitsi/jitsi-meet/pull/1832

paweldomas:
About Lenny's attempts I think it was more about chaining stuff in react - not here.

saghul:
Jenkins test this please.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -23,22 +23,24 @@ const logger = getLogger(__filename);
  * @returns {Promise}
  */
 function createMuteUnmutePromise(track, mute) {
-    if (track.inMuteOrUnmuteProgress) {
-        return Promise.reject(
-            new JitsiTrackError(JitsiTrackErrors.TRACK_MUTE_UNMUTE_IN_PROGRESS)
-        );
+    // If there is a mute / unmute operation in progress, chain this operation
+    // so it runs afterwards. We do it both on `then` and `catch` because
+    // primises don't have a `finally` yet.
```
virtuacoplenny: "promises don't have a"

virtuacoplenny:
I didn't attempt making things more promisified. I think Pawel is referring to when I was looking at one of his PRs and I mentioned how promises should be used more instead of just throwing and how promises in general should be chained onto instead of firing and forgetting.

saghul:
@virtuacoplenny ack, at any rate, thanks for taking a look!

modules/RTC/JitsiLocalTrack.js
```diff
@@ -23,22 +23,24 @@ const logger = getLogger(__filename);
...
-        return Promise.reject(
-            new JitsiTrackError(JitsiTrackErrors.TRACK_MUTE_UNMUTE_IN_PROGRESS)
-        );
+    // If there is a mute / unmute operation in progress, chain this operation
+    // so it runs afterwards. We do it both on `then` and `catch` because
+    // promises don't have a `finally` yet.
+    if (track._mutePromise) {
+        return track._mutePromise
+            .then(() => track._setMute(mute))
+            .catch(() => track._setMute(mute));
```
lyubomir: 1. I'm afraid I don't see that the above does "chain" fully. `inMuteOrUnmuteProgress` was making sure that one doesn't enter `JitsiLocalTrack._setMute` while a previous method call hasn't exited it. `_mutePromise` does not do that. If one invokes `JitsiLocalTrack.mute`/`unmute` 3 or more times quickly enough - A, B & C, then A will enter `_setMute`, B will wait for A to enter `_setMute`, but C will wait for A or nothing depending on whether A has been fulfilled/rejected by that time and in either of the two cases C will be allowed to enter `_setMute` while B is not guaranteed to have exited it.

2. Promise.then(onFulfilled).catch(onRejected) is not the same as Promise.then(onFulfilled, onRejected). And to me it looks like the two uses `createMuteUnmutePromise` rather want the second construct really.

Let's see what we have now:

`track._setMute` is fulfilled, then `track._mutePromise = undefined` throws so we catch that and do `track._mutePromise = undefined` again. Really? Anyway, `track._mutePromise = undefined` will probably not throw so we only get source code that's strange but should practically not happen.

But the problem in the new source code is more realistic: `track._,mutePromise` is fulfilled, so we do `track._setMute(mute)` which throws, but we catch that and we do `track._setMute(mute)` again. Was that really the intention of the source code?

3. I think the comment about `finally` should be removed. I don't really know what you mean by `finally` in the context of Promises but I guess you want something like `Promise.then(finally, finally)`.

saghul:
@lyubomir Thanks for the thorough analysis. The chaining was indeed incorrect. I reworked the logic, and made it even simpler (and correct, I hope!), PTAL.

PR #569, opened 2017-08-09
fix(JitsiLocalTrack): restore 'inMuteOrUnmuteProgress'

paweldomas:
Restores 'inMuteOrUnmuteProgress' flag used by the LocalSdpMunger.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -128,6 +128,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
         this._realDeviceId = this.deviceId === '' ? undefined : this.deviceId;
 
         /**
+         * Set to <tt>true</tt> when there's ongoing "mute/unmute" operation in
+         * progress. Used by {@link LocalSdpMunger}.
+         * @type {boolean}
+         */
+        this.inMuteOrUnmuteProgress = true;
+
+        /**
          * Indicates that we have called RTCUtils.stopMediaStream for the
          * MediaStream related to this JitsiTrack object.
          */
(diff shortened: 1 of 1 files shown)


saghul:
What happens if we don't do that? I'm not familiar with that part.

paweldomas:
LocalSdpMunger will not fake video SSRCs correctly and it may trigger 'source-add' when unmuting or 'source-remove' when muting.

saghul:
So when would this happens? If a participant joins while we are muting?

paweldomas:
So actually it will not rigger 'source-add'/'source-remove' because the code in JingleSessionPC._verifyNoSSRCChanged only detects incorrect state and logs an error.

paweldomas:
Keep in mind that current lib-jitsi-meet is in broken state and I'm not going to refactor the whole approach at this point.

paweldomas:
I wonder why this error was ignored during video mute/unmute refactoring

saghul:
> So actually it will not rigger 'source-add'/'source-remove' because the code in JingleSessionPC._verifyNoSSRCChanged only detects incorrect state and logs an error.

When will it trigger then? I'm sorry for delaying this, but I don't really like bringing that back and I'd like to fully understand if we have other options.

> Keep in mind that current lib-jitsi-meet is in broken state and I'm not going to refactor the whole approach at this point.

What is broken?

> I wonder why this error was ignored during video mute/unmute refactoring

It was my mistake, sorry. I didn't notice it was used outside of the muting related code. I'll grep next time. 🙏

debugmaster:
@paweldomas can you give more details why it may be broken right now? I am having issues and I found this PR very interesting to what I was looking for.

lyubomir:
I merged it in master https://github.com/jitsi/lib-jitsi-meet/commit/8c1283c7fdbb1df69eabd529be5ea9428e23ee67 as part of another PR which contained the same commit.

paweldomas:
@debugmaster honestly I couldn't find a case where it would cause issues right now, but if you describe yours maybe it would be easier to connect the two

PR #571, opened 2017-08-10
Fix local video disappearing from local thumbnail

paweldomas:
Local video was being detached when rapidly muting/unmuting, because of the "track ended" event being fired and that was not handled correctly.


modules/RTC/JitsiLocalTrack.js
```diff
@@ -128,6 +124,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
         this._realDeviceId = this.deviceId === '' ? undefined : this.deviceId;
 
         /**
+         * Set to <tt>true</tt> when there's ongoing "mute/unmute" operation in
+         * progress. Used by {@link LocalSdpMunger}.
+         * @type {boolean}
+         */
+        this.inMuteOrUnmuteProgress = true;
```
bbaldino: should this be `muteOrUnmuteInProgress`?

PR #572, opened 2017-08-10
Add check for iOS WebViews in screen sharing adapter

bostondv:
Loading lib-jitsi-meet on an iOS WebView (UIWebView & WKWebView) currently fails due to its user agent string not matching any of the existing checks. This causes the script to fail due to undefined `webrtcDetected...` variables.

This adds a check for those webviews and sets detection variables similar to Safari.

### DIFF
--- a/modules/RTC/adapter.screenshare.js
+++ b/modules/RTC/adapter.screenshare.js
@@ -373,6 +373,16 @@ AdapterJS.parseWebrtcDetectedBrowser = function () {
     webrtcDetectedType      = isMobile.length === 0 ? 'plugin' : null;
     webrtcDetectedDCSupport = isMobile.length === 0 ? 'SCTP' : null;
 
+  // Detect WebView on iOS (does not support WebRTC yet)
+  } else if (/(iPhone|iPod|iPad).*AppleWebKit(?!.*Safari)/i.test(navigator.userAgent)) {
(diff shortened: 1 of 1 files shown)


bostondv:
Thanks, I've signed the CLA :)

saghul:
@bbaldino can you PTAL?

bbaldino:
what's the desired behavior here, to get a more graceful failure?

bostondv:
@bbaldino Yes, currently Jitsi will fail with a JavaScript error.

bbaldino:
cool, lgtm...i didn't double check the specific useragent or anything but seems reasonable based on what i've seen after a quick google search.

paweldomas:
Hey guys, we should not modify this file in our repo. PR should be against Temasys / AdapterJS and then updated here. It will be hard to pull new version after we make changes here.
https://github.com/Temasys/AdapterJS/blob/master/publish/adapter.screenshare.js

saghul:
Great catch @paweldomas! I hadn't noticed the patch was against that file 😞

bbaldino:
ah, i didn't realize this was a file we copied from there either.  maybe it should be kept in a thirdparty dir or something?

bostondv:
Would you guys like me to open a PR with Temasys or do you want to pass it
upstream yourselves? We unfortunately didn't realize it came from a third
party either.

PR #575, opened 2017-08-17
Fire JitsiConferenceEvents.ENDPOINT_MESSAGE_RECEIVED events via JSON messages received via the Chatroom

nikvaessen:


JitsiConferenceEventManager.js
```diff
@@ -300,6 +300,23 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
                 JitsiConferenceEvents.USER_STATUS_CHANGED, id, status);
         });
 
+    chatRoom.addListener(XMPPEvents.JSON_MESSAGE_RECEIVED,
+        (from, payload) => {
+            const participant = conference.getParticipantById(
+                from.split('/')[1]);
```
bgrozev: I think we have a utility function for this (Strophe.getResourceFromJid)

JitsiConferenceEventManager.js
```diff
@@ -300,6 +300,23 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
...
+
+            if (participant) {
+                conference.eventEmitter.emit(
+                    JitsiConferenceEvents.ENDPOINT_MESSAGE_RECEIVED,
+                    participant, payload);
+            } else {
+                logger.warn(
+                    'Ignored ENDPOINT_MESSAGE_RECEIVED for not existing '
+                    + `participant: ${from}`,
```
bgrozev: Technically we ignored an XMPPEvents.JSON_MESSAGE_RECEIVED. I think it might be useful to change the message so that logs from the two places can be distinguished.

bgrozev:
@jitsi/developers can some of you library people also take a look, please?

modules/xmpp/ChatRoom.js
```diff
@@ -79,6 +79,33 @@ function filterNodeFromPresenceJSON(pres, nodeName) {
     return res;
 }
 
+/**
+ * Check if the given argument is a valid JSON string by parsing it.
+ * If it successfully parses, the JSON object is returned.
+ *
+ * @param jsonString check if this string is a valid json string
```
paweldomas: jsonString type should be documented:
`@param {string} jsonString`

modules/xmpp/ChatRoom.js
```diff
@@ -742,6 +769,19 @@ export default class ChatRoom extends Listenable {
             this.discoRoomInfo();
         }
 
+        // todo add a JSON layer such that not every json
+        // passed around in the chat is automatically assumed to be
+        // json used for messaging
+        const json = tryParseJSON(txt);
+
+        if (json) {
+            logger.log('chat json message', from, json);
```
paweldomas: is it going to be many messages during regular session ? maybe debug level would be more appropriate here

paweldomas:
LGTM just few really minor things

modules/xmpp/ChatRoom.js
nikvaessen: Yes, it might actually be more than a few per seconds... maybe we should get rid of it altogether?

nikvaessen:
As I'm already considering how to send the json from Jigasi to the react module, I was thinking about how to fix the todo in this PR (to not make every json object passed in the chat suddenly an endpoint message). 

Would the following work? The JSON in the chat message needs to contain 2 keys.

- **jitsi-meet-muc-chatroom-endpoint-message-name**: Which **needs** to have a string as a value. I don't think it's necessary to keep a list of strings and check if the given string is in that list, but it might be a possibility.
- **payload**: Should we require the key to be an object? or only check if it's given?

Example json:
```
{
  "jitsi-meet-muc-chatroom-endpoint-message-name":"transcription-result" // or whatever you want
  "payload": {
      the intering json object
  }
}
```
Any feedback?

modules/xmpp/ChatRoom.js
paweldomas: In that case it would be better to get id of it especially that our logging_config.js is not working very well with deployment upgrades

paweldomas:
Good idea, but maybe something shorter than 'jitsi-meet-muc-chatroom-endpoint-message-name' like 'jitsi-meet-muc-msg-name' or 'jitsi-meet-muc-msg-topic' or 'jitsi-meet-muc-msg-subject'

modules/xmpp/ChatRoom.js
nikvaessen: Fixed

modules/xmpp/ChatRoom.js
nikvaessen: fixed

modules/xmpp/ChatRoom.js
```diff
@@ -79,6 +79,54 @@ function filterNodeFromPresenceJSON(pres, nodeName) {
...
+            const topic = json[VALID_FIELD_NAMES[0]];
+            const payload = json[VALID_FIELD_NAMES[1]];
+
+            if ((typeof topic === 'string' || topic instanceof String)
+                && payload) {
+
+                return json;
+            }
+
+            console.debug('parsing valid json but does not have correct '
```
paweldomas: probably logger should be used instead of console

bgrozev:
In the longer term I think we should signal that the message contains JSON in XMPP (e.g. put the JSON in a separate element, not in "message>body"). But this is reasonable for now and doesn't make it any harder to switch later.

PR #578, opened 2017-08-18
fix(presence): clear presence status on join

virtuacoplenny:
- Send an empty status tag in the presence when joining the muc.
- Check for the existence of a status tag on the onPresence
  handler. Accept its contents if the tag exists, even if the
  contents are empty. This will clear any presence left behind
  from a poltergeist participant.


virtuacoplenny:
I'm unsure if this is the correct approach for having status leftover from poltergeist be cleared when the real participant joins.

virtuacoplenny:
https://github.com/jitsi/lib-jitsi-meet/pull/580 will be used instead.

PR #579, opened 2017-08-21
feat(recording): expose recording status and type constants

virtuacoplenny:


modules/xmpp/recording.js
```diff
@@ -61,6 +61,7 @@ Recording.status = {
     UNAVAILABLE: 'unavailable',
     PENDING: 'pending',
     RETRYING: 'retrying',
+    ERROR: 'error',
```
yanas: Isn't FAILED status covering the error as well?

yanas: Where do we use this error status here in the code?

virtuacoplenny: It's not used in this file but jitsi-meet uses it for one of its cases with Recording.js. jitsi-meet js had these constants, plus error, in Recording.js with a todo to use the constants in lib.

PR #580, opened 2017-08-23
Fire status message update when there is no status message.

damencho:
In case there was a status message and later that message is removed from the presence, fire an event so the UI can also clear it.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -341,7 +341,14 @@ export default class ChatRoom extends Listenable {
         const member = {};
 
         member.show = $(pres).find('>show').text();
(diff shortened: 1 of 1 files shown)


modules/xmpp/ChatRoom.js
```diff
@@ -447,6 +456,11 @@ export default class ChatRoom extends Listenable {
             if (member.displayName) {
                 memberOfThis.displayName = member.displayName;
             }
+
+            if (memberOfThis.status !== member.status) {
```
virtuacoplenny: Do you think it's worthwhile to leave a comment about this or would it be obvious to someone who is working in this code?

damencho: Added.

PR #582, opened 2017-08-30
Adds callstatsname to the permanent analytics properties.

damencho:


saghul:
Why is this needed? Some explanation in the commit description would be nice to have.

PR #583, opened 2017-08-31
ref(JitsiLocalTrack): get rid of 'setMutedInProgress'

paweldomas:
The 'setMutedInProgress' flag is unreliable and it can lead to weird
situations when mute operation is in progress, but the removal from
JingleSessionPC is waiting on the queue. In such case LocalSdpMunger
will fake SDP even if the track is in the PeerConnection which may
confuse some parts of the SDP transformation chain.

For example RtxModifier will not attempt to inject Rtx group description
if it sees that the SSRC are already there. But it can happen that even
though the SSRCs are there they do not contain valid track ID, because
the track has changed.

The LocalSdpMunger is supposed to fake the SDP only when the video
MediaStream is not in the PeerConnection. So this commit implements just
that: a way to tell if JitsiLocalTrack's stream is currently in
the PeerConnection (added through TraceablePeerConnection).

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -321,15 +321,6 @@ export default class JitsiLocalTrack extends JitsiTrack {
 
         let promise = Promise.resolve();
 
-        /**
-         * Set to <tt>true</tt> when there's ongoing "mute/unmute" operation in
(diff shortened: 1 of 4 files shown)


paweldomas:
yes

saghul:
👍 I'll give it a test and merge if it's all kosher. Great work Pawel!

PR #584, opened 2017-08-31
Improve logging

paweldomas:
This PR fixes few cases when on Jingle error whole objects with cyclic references were logged which could sometimes crash the app on iOS react-native.

Also adds PrefixedLogger which will append JingleSessionPC/PeerConnection's identifier to log messages of it's subcomponents. This makes debugging much easier when there are both P2P and JVB sessions active.

### DIFF
--- a/modules/RTC/LocalSdpMunger.js
+++ b/modules/RTC/LocalSdpMunger.js
@@ -2,6 +2,7 @@
 
 import { getLogger } from 'jitsi-meet-logger';
 import * as MediaType from '../../service/RTC/MediaType';
+import PrefixedLogger from '../util/PrefixedLogger';
 import { SdpTransformWrap } from '../xmpp/SdpTransformUtil';
 
 const logger = getLogger(__filename);
@@ -23,6 +24,7 @@ export default class LocalSdpMunger {
      */
     constructor(tpc) {
         this.tpc = tpc;
+        this.log = new PrefixedLogger(logger, tpc.toString());
     }
 
     /**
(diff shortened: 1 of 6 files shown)


modules/util/PrefixedLogger.js
```diff
@@ -0,0 +1,76 @@
...
+        this.prefix = prefix;
+    }
+
+    /**
+     * Logs on TRACE logging level.
+     * @param text
+     * @param args
+     * @constructor
+     */
+    t(text, ...args) {
```
saghul: (please don't hate me) Why don't you use the same names as Logger? That would make it easier to interchange.

paweldomas: It's for the expression to be as short as possible, because wrapping template strings is PITA. Also note that it would not replace 'logger' which can be used from static methods.

saghul: > t's for the expression to be as short as possible, because wrapping template strings is PITA.

Hum, `this.log.e('There is more than 1 video track !', localVideos);` is not that much of an improvement over `this.log.error('There is more than 1 video track !', localVideos);` IMHO, and removed the need for the reader to remember they are different.



> Also note that it would not replace 'logger' which can be used from static methods.

Sure, I mean their APIs are interchangeable (mostly).

modules/RTC/LocalSdpMunger.js
```diff
@@ -2,6 +2,7 @@
 
 import { getLogger } from 'jitsi-meet-logger';
 import * as MediaType from '../../service/RTC/MediaType';
+import PrefixedLogger from '../util/PrefixedLogger';
```
saghul: I wonder if there is any value in putting this in jitsi-meet-logger ?

paweldomas: Yes, maybe that would work better. Not sure when I get to that though...

paweldomas: But do you mean just moving this class or integrating this somehow into the logger ?

I've noticed that the log method name now always comes from PrefixedLogger. Having that integrated would probably allow to overcome this limitation.

saghul: I meant moving the class. Maybe a Proxy object could help? The bonus nachos is that you really only need to override a single method (log) in that case, right?

PR #585, opened 2017-08-31
enforce preferh264 when setting the local description

bbaldino:
for p2p, we have to take preferh264 into account when setting the local description, since
that is the offer that will be received by the far side.  if we only do it on the remote description,
like before this change, we'll never actually end up preferring h264 in p2p


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1536,6 +1536,17 @@ TraceablePeerConnection.prototype.setLocalDescription
 
     this.trace('setLocalDescription::preTransform', dumpSDP(localSdp));
 
+    if (this.options.preferH264) {
```
saghul: Do you think it's worth it to extract this to an auxiliary function and use it on both local and remote?

PR #588, opened 2017-09-08
RN: add support for XMPP authentication on mobile

paweldomas:


JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
+import * as JitsiConnectionEvents from './JitsiConnectionEvents';
+import XMPP from './modules/xmpp/xmpp';
+
+/**
+ * Class creates separate XMPP connection and tries to connect using given
```
saghul: THAT creates A separate... using THE given

JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
...
+ * credentials. Once connected will contact Jicofo to obtain and store session
```
saghul: IT will contact ... THE session ID

JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
...
+ * ID which is then used by the parent conference to upgrade user's role to
+ * moderator. It's also used to join the conference when starting from anonymous
+ * domain and only authenticated users are allowed to create new rooms.
+ */
+export default class JitsiAuthConnection {
+    /**
+     * Creates new <tt>JitsiAuthConnection</tt> for given conference.
```
saghul: A new ... THE given

JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
...
+                        this.xmpp.disconnect();
+
+                        reject({
+                            authenticationError: error.error,
+                            message: error.message
+                        });
+                    });
+            };
+
+            this.xmpp.addListener(
```
saghul: when are these listeners disposed?

JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
...
+    }
+
+    /**
+     * Cancels the authentication if it's currently in progress. The promise
+     * returned by {@link authenticateAndUpgradeRole} will be rejected with
+     * an empty object (none of the error fields set).
+     */
+    cancel() {
+        this.canceled = true;
+        this.xmpp.disconnect();
```
saghul: isn't this unnecessary when we never attempted to connect?

JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
...
+     * the XMPP server.
+     * @property {String} [authenticationError] - One of XMPP error conditions
+     * returned by Jicofo on authentication attempt. See
+     * https://xmpp.org/rfcs/rfc3920.html#streams-error.
+     * @property {String} [message] - More details about the error.
+     *
+     * NOTE If neither one of the errors is present it means that the operation
+     * has been canceled.
+     */
+    /**
```
saghul: Maybe leave an empty line here in between the large comment blocks?

JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
...
+                        this.conference.options.config);
+                this.room.moderator.authenticate()
+                    .then(() => {
+                        if (this.canceled) {
+                            return;
+                        }
+
+                        this.xmpp.disconnect();
+
+                        // At this point we should have new session ID stored in
```
saghul: Question not related to the code: where does this storage of the session ID happen?

modules/xmpp/moderator.js
```diff
@@ -465,11 +465,13 @@ Moderator.prototype.authenticate = function() {
             result => {
                 this.parseSessionId(result);
                 resolve();
-            }, error => {
-                // eslint-disable-next-line newline-per-chained-call
-                const code = $(error).find('>error').attr('code');
-
-                reject(error, code);
+            }, errorIq => {
+                reject({
```
saghul: I don't know what this does, so I'm trusting It Works (TM) :-)

JitsiAuthConnection.js
```diff
@@ -0,0 +1,141 @@
...
+     * in order to obtain a session ID (which is then stored in the local
+     * storage). User's role of the parent conference will be upgraded to
+     * moderator (by Jicofo).
+     *
+     * @param {Object} options
+     * @param {string} options.id - XMPP user's ID (user@xmpp-server.com).
+     * @param {string} options.password - User's password used to login.
+     * @param {string} [options.roomPassword] - Room password required to join
+     * the MUC room.
+     * @param {Function} [options.onLoginSuccessful] - Callback called when
```
saghul: Why is this callback needed? Is there code that needs to run between the login and the role upgrade?

JitsiAuthConnection.js
```diff
@@ -0,0 +1,162 @@
...
+                (connectionError, message) => reject({
+                    connectionError,
+                    message
+                }));
+            this.xmpp.addListener(
+                CONNECTION_DISCONNECTED,
+                () => {
+                    if (this.canceled) {
+                        reject({});
+                    }
```
lyubomir: And what if the connection disconnects without it being canceled? Handle it please or at least write a comment.

JitsiAuthConnection.js
```diff
@@ -0,0 +1,162 @@
...
+    }
+
+    /**
+     * Cancels the authentication if it's currently in progress. The promise
+     * returned by {@link authenticateAndUpgradeRole} will be rejected with an
+     * empty object (i.e. none of the error fields set).
+     */
+    cancel() {
+        this.canceled = true;
+        this.xmpp.disconnect();
```
lyubomir: Like @saghul said, what if this.xmpp is not connected at all? The documentation says that, if `cancel` is called, the `Promise` will be rejected with an empty object but... What if I call `cancel` before the event `CONFERENCE_FAILED` arrives or before the `catch` of `this.room.moderator.authenticate()` executes?

JitsiAuthConnection.js
```diff
@@ -0,0 +1,162 @@
...
+            const connectionEstablished = () => {
+                if (this.canceled) {
+                    return;
+                }
+
+                // Let the caller know that the XMPP login was successful.
+                onLoginSuccessful && onLoginSuccessful();
+
+                // Now authenticate with Jicofo and get a new session ID.
+                this.room
```
lyubomir: * Why is `room` a property given that it's only written and read locally?
* Is it public or non-public?

JitsiAuthConnection.js
```diff
@@ -0,0 +1,162 @@
...
+                    .then(() => {
+                        if (this.canceled) {
+                            return;
+                        }
+
+                        this.xmpp.disconnect();
+
+                        // At this point we should have the new session ID
+                        // stored in the settings. Jicofo will allow to join the
+                        // room.
```
lyubomir: * @saghul asked where that's "stored in the settings". I think that's `this.room.moderator.authenticate` which does `Settings.setSessionId`.

* I'm not really sure why `sessionId` is stored in `localStorage` given that `localStorage` is expected to (maybe) survive across browser sessions and the piece of data suggests that it's a session-related piece of data.

JitsiAuthConnection.js
```diff
@@ -0,0 +1,162 @@
...
+
+                        resolve();
+                    })
+                    .catch(({ error, message }) => {
+                        this.xmpp.disconnect();
+
+                        reject({
+                            authenticationError: error,
+                            message
+                        });
```
lyubomir: I'd suggest `reject` first and only then `this.xmpp.disconnect`. If `this.xmpp.disconnect` throws, it seems like the important piece is that the rejection of `_promise` was because of `authenticationError`.

JitsiConference.js
```diff
@@ -304,6 +305,16 @@ JitsiConference.prototype.join = function(password) {
 };
 
 /**
+ * Creates a new {@link JitsiAuthConnection} which authenticates and upgrades
+ * the user's role to moderator.
+ *
+ * @returns {JitsiAuthConnection}
+ */
+JitsiConference.prototype.createAuthenticationConnection = function() {
```
lyubomir: 1. I do not support multiple names for one and the same thing because that makes it difficult for me comprehend the source code. Is the abstraction called `AuthenticationConnection` or `JitsiAuthConnection`?

2. I find the abstraction too complex:

* As an API client/consumer, I have to `jitsiConference.createAuthenticationConnection()`, then call `authenticateAndUpgradeRole` on the return value, and then wait on the returned `Promise`.

* I understand I need an object to call `cancel` on.

* `cancel` renders the `JitsiAuthConnection` instance un-re-usable because `cancel` only sets `canceled` and `authenticateAndUpgradeRole` only returns `_promise`.

* I'd suggest having `JitsiConference.prototype.authenticateAndUpgradeRole` which returns a Promise with a `cancel` method (which should be documented, of course). This way there will be no need for creating an object and then using it to only get another object. Then `cancel` may be internally implemented nicely with `reject` in order to satisfy the documentation that `cancel` always rejects with an empty object.

* Alternatively, I'd drop the whole `JitsiAuthConnection ` and `Promise` business because of the weirdness with returning a `Promise` and having a `onLoginSuccessful` callback which @saghul pointed out. If there's going to be callback involved anyway, then I'd have `JitsiConference.prototype.authenticateAndUpgradeRole({ onLoginSuccessful, onError/Rejected, onFullfilled/Success})` i.e. a function that's deals with callbacks only. Then if the caller wants cancellation as well, they may as well add a `cancelled` callback in there, internally we can invoke it whenever we currently do `this.canceled`, and then we don't even have to promise that we'll do `reject` with an empty object because it's the API client/consumer who's doing the cancellation anyway.

PR #589, opened 2017-09-11
feat(conference): Add stop/start transcriber methods

hristoterezov:


JitsiConference.js
```diff
@@ -1742,6 +1742,19 @@ JitsiConference.prototype.hangup = function() {
 };
 
 /**
+ * Starts the transcription service.
+ */
+JitsiConference.prototype.startTranscriber = function() {
+    return this.dial('jitsi_meet_transcribe');
```
virtuacoplenny: Where does this.dial get defined?

virtuacoplenny: Ah okay, I see it now. It's calling into another existing method.

PR #590, opened 2017-09-12
fix(recording): call toggleRecording with no args

paweldomas:


modules/xmpp/recording.js
```diff
@@ -334,10 +334,15 @@ Recording.prototype.setRecording = function(...args) {
...
+ * @param {Object} options
+ * @param {string} options.token token for authentication
+ * @param {string} options.streamId the stream ID to be used with Jibri in
+ * the streaming mode.
  * @param statusChangeHandler {function} receives the new status as argument.
  */
-Recording.prototype.toggleRecording = function(options, statusChangeHandler) {
+Recording.prototype.toggleRecording = function(
+    options = { },
+    statusChangeHandler) {
```
saghul: I see `statusChangeHandler` is used unconditionally. Do we really support calling this function without args, or is it just working out of luck?

paweldomas: Sorry, it should say calling JitsiConference.toggleRecording with no args

saghul: Then how are we calling it? `conference.toggleRecording(undefined, statusChangeHandler);` ?

saghul: If so, LGTM.

paweldomas: We're calling it like this:
https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L1696
The purpose of this PR is to allow to pass undefined options.

PR #591, opened 2017-09-12
feat(transcription): Add transcription status

hristoterezov:


JitsiConference.js
```diff
@@ -631,6 +631,15 @@ JitsiConference.prototype.getTranscriber = function() {
 };
 
 /**
+ * Returns the status transcription.
```
yanas: "the transcription status" sounds better I think.

JitsiConferenceEvents.js
```diff
@@ -230,6 +230,15 @@ export const TRACK_MUTE_CHANGED = 'conference.trackMuteChanged';
 export const TRACK_REMOVED = 'conference.trackRemoved';
 
 /**
+ * Notifies for transcription status changes. The event provides the
+ * following parameters to its listeners:
+ *
+ * @param {String} status - The new status.
+ */
+export const TRANSCRIPTION_STATUS_CHANGED = 'xmpp.transcription_status_changed';
```
yanas: Is there any reason why this isn't called "conference.transcriptionStatusChanged", the same way we call the other JitsiConferenceEvents?

modules/xmpp/ChatRoom.js
```diff
@@ -175,6 +175,7 @@ export default class ChatRoom extends Listenable {
         this.participantPropertyListener = null;
 
         this.locked = false;
+        this.transcriptionStatus = 'off';
```
yanas: I think we should export the transcription status in the JitsiMeetJS.constants, the same way we do for recording and SIP GW. This will allow using those constants in Jitsi Meet to check the status, instead of comparing strings.

PR #593, opened 2017-09-18
Move some devDependencies to dependencies

adambowles:
I've been unable to install `lib-jitsi-meet` in a project that does not have `webpack` in its dependencies.

The error was in `lib-jitsi-meet`'s `postinstall` script (`webpack -p`):

```shell
~/P/project   master *+…  rm -rf node_modules/lib-jitsi-meet; npm i lib-jitsi-meet

> lib-jitsi-meet@0.0.0 postinstall /home/adambowles/Projects/project/node_modules/lib-jitsi-meet
> webpack -p

module.js:471
    throw err;
    ^

Error: Cannot find module 'webpack'
```
Moving webpack dev dependency to dependencies fixes this (along with a couple other packages webpack is employing).

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,9 @@
   "readmeFilename": "README.md",
   "dependencies": {
     "async": "0.9.0",
+    "babel-loader": "7.1.2",
+    "babel-preset-es2015": "6.24.1",
+    "babel-preset-stage-1": "6.24.1",
     "current-executing-script": "0.1.3",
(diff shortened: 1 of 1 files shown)


adambowles:
I'm unable to sign your agreement:

image

saghul:
Hi Adam! Sorry for the delay. I'm afraid we cannot take your PR. webpack **is** a development dependency, it should not be needed for runtime, that's why the split is in such a way.

PR #594, opened 2017-09-19
request 1080p for chrome >= 61

bbaldino:
there was a bug in chrome that caused video to stop sending when we requested 1080p,
that bug has been fixed so start requesting 1080p video for chrome versions >= 61

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -269,8 +269,15 @@ export default {
         window.connectionTimes['obtainPermissions.start']
             = window.performance.now();
 
+        if (!this._inResolutionRollback
+                && RTCBrowserType.isChrome()
+                && RTCBrowserType.getChromeVersion() >= 61) {
+            options.resolution = '1080';
+        }
+
         return RTC.obtainAudioAndVideoPermissions(options || {})
             .then(tracks => {
+                this._inResolutionRollback = false;
                 promiseFulfilled = true;
 
                 window.connectionTimes['obtainPermissions.end']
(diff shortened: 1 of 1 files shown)


bbaldino:
so this PR is very simple and does solve the problem, but I'm not crazy about it as it feels a bit awkward and adds some state.  other options I could think of:

1. Add an argument to createLocalTracks that signals whether or not this is the first try or a retry.  _I don't really like this, as createLocalTracks is part of the main API, and I think it would be confusing to add this parameter since it's only used internally_

2. Add the isRetry flag in the options dict.  _This is similar to the one above, but doesn't include it as part of the API so I think it's slightly better.  I still find it awkward, as this isn't really an 'option' so it's kinda abusing the options dict_

3. Add another method below createLocalTracks (e.g. `_createLocalTracks`).  The internal `_createLocalTracks` would take in another argument: `isRetry`.  `createLocalTracks` would call this with `false` and the retry logic would re-call the internal `_createLocalTracks` with `true`, denoting it's now retrying.  _This feels better to me, but I didn't want to code it up without getting some feedback as to what others thought_

4. Change the way we call getUserMedia.  Instead of passing a single resolution for both the required minimum and maximum, instead request a required aspect ratio, but leave the resolution as flexible.  _I think this would be a good direction for us to head long term.  It'd need some testing to ensure there weren't any corner cases it would have problems with, but this is much more flexible than what we have now_

bbaldino:
I'm gonna try coding up option 3 and see how it looks, so don't merge this yet.

bbaldino:
so this gets tricky due to the following scenario:
1. Join a call with a camera that supports the requested resolution (say 1080)
2. In the device selector dialog, switch to a camera that only supports 720

Expected: we switch to the selected camera in a lower resolution
Actual: we remain on the current camera
Reason: the requested resolution is set in the 'required' section, whereas the device is set in the 'optional' section, so we'll prefer any camera that can support the requested resolution in favor of the selected camera which can only support a lower resolution.  This bug exists even before this PR, but probably isn't very likely to be hit (since we only request 720 now and most cameras will support that).

virtuacoplenny:
Superseded by https://github.com/jitsi/lib-jitsi-meet/pull/617.

PR #597, opened 2017-09-22
feat(xmpp): include credentials in CONNECTION_FAILED

paweldomas:
Will include user's credentials in the CONNECTION_FAILED/PASSWORD_REQUIRED error event.


lyubomir:
Merged as 60ad3e6.

PR #600, opened 2017-09-29
feat(remote_mute): Add analytics event

hristoterezov:


JitsiConferenceEventManager.js
```diff
@@ -72,6 +72,8 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
 
     chatRoom.addListener(XMPPEvents.AUDIO_MUTED_BY_FOCUS,
         value => {
+            Statistics.analytics.sendEvent('conference.remotelymuted');
```
virtuacoplenny: Looks like the other events will use camel case, so maybe it should be "remotelyMuted".

hristoterezov: Fixed!

PR #602, opened 2017-10-02
feat(analytics): available devices

hristoterezov:


modules/RTC/RTCUtils.js
```diff
@@ -1457,6 +1486,9 @@ function onReady(options, GUM) {
         rawEnumerateDevicesWithCallback(ds => {
             currentlyAvailableMediaDevices = ds.splice(0);
 
+            logger.info('Available devices: ', currentlyAvailableMediaDevices);
+            sendDeviceListToAnalytics(currentlyAvailableMediaDevices);
```
bbaldino: why does it need to be here too? won't the call in `onMediaDevicesListChanged` cover it?

damencho: Cause we want and the initial list of devices.

bbaldino: `onMediaDevicesListChanged` doesn't get fired the first time?

hristoterezov: I don't think onMediaDevicesListChanged is called initially.

bbaldino: seems like maybe it'd be good to change it to do that, then we have a single place where anything that cares about the device list changing can listen?  i also think most people would expect `onMediaDevicesListChanged` to fire even for the initial filling of the list, so seems more consistent.

hristoterezov: Sounds reasonable! 

Although currently we have `DEVICE_LIST_CHANGED` event fired from `onMediaDevicesListChanged` and `DEVICE_LIST_AVAILABLE` event fired when we initially receive the available devices. The format of the devices passed with the event is different. Also we handle  `DEVICE_LIST_AVAILABLE` only for logging purposes in different place. So I think we should review and refactor the whole implementation about the devices.

This doesn't sounds to me like 2 line change. We have to check jitsi-meet code to make sure that it'll handle properly the `DEVICE_LIST_CHANGED` event if it is fired initially. Maybe we have to rename it.

Let's do this change with another PR, I'll write it down? @bbaldino WDYT?

damencho: The code was written on purpose this way, cause at the time of coding onMediaDevicesListChanged was not supported by some or all browsers and we are currently emulating it (polling for changes).
There is an API that browser will implement and then we can easily switch with current code. I'm not very familiar with the details about this API.

damencho: Here is the discussion, maybe I remember it wrong: https://github.com/jitsi/lib-jitsi-meet/pull/148/files/20754b7dd093a644f72ad0793682315ed6155ba4#r68390049

bbaldino: @hristoterezov yeah that sounds good to me

bbaldino: @damencho i liked your initial setup better...i don't get why we would special case to avoid firing the event the first time, but maybe there is a scenario i'm missing.  anyway, agree with @hristoterezov that it would be good to revisit this and think about the big picture.  adapterjs work may affect this a little as well.

PR #606, opened 2017-10-09
feat(RTCUtils): add options for suspend video evaluation

hristoterezov:


modules/RTC/RTCUtils.js
```diff
@@ -869,17 +869,41 @@ class RTCUtils extends Listenable {
...
-                // Allows sending of video to be suspended if the bandwidth
-                // estimation is too low.
-                if (!options.disableSuspendVideo) {
-                    this.pcConstraints.optional.push(
-                        { googSuspendBelowMinBitrate: true });
+                const {
+                    forceSuspendVideo,
+                    disableSuspendVideo
+                } = options;
+                const suspendVideoEnabled
```
bbaldino: i'm finding the meanings of `forceSuspendVideo` and `disableSuspendVideo` a bit confusing to figure out, i assume they'll be documented in config.js but is there any more explanation we could add here in a comment or something?

modules/RTC/RTCUtils.js
```diff
@@ -869,17 +869,41 @@ class RTCUtils extends Listenable {
...
+                    = typeof forceSuspendVideo === 'undefined'
+                        || (typeof forceSuspendVideo === 'number'
+                            && Math.random() < forceSuspendVideo);
```
bbaldino: looks like `forceSuspendVideo` might be better named as `forceSuspendVideoProbability`?

bbaldino: and am i reading this right that, if `forceSuspendVideo` isn't defined at all (typeof === undefined), we equate that to '100% probability'?

bbaldino: whatever default we want for `forceSuspendVideo`, maybe we could do it once at the top like: `forceSuspendVideo = forceSuspendVideo || 1` or something (since we have to check it isn't undefined in multiple places)

modules/RTC/RTCUtils.js
```diff
@@ -869,17 +869,41 @@ class RTCUtils extends Listenable {
...
+                const disableSuspendVideoForJVBPC
+                    = Boolean(disableSuspendVideo);
```
bbaldino: do we need both variables `forceSuspendVideo` and `disableSuspendVideo`? can't we just say that, to disable, `forceSuspendVideo` should be `0` ?  with both of them we have to handle all the valid/invalid interactions between the 2 (if i'm understanding this correctly)

modules/RTC/RTCUtils.js
```diff
@@ -869,17 +869,41 @@ class RTCUtils extends Listenable {
...
+                        disableSuspendVideoForJVBPC
+                    });
                 }
 
-                // There's no reason not to use this for p2p
-                this.p2pPcConstraints.optional.push({
-                    googSuspendBelowMinBitrate: true
-                });
+                if (suspendVideoEnabled) {
+                    if (!disableSuspendVideoForJVBPC) {
```
bgrozev: If we use the "force" property we get the following behavior:
If it succeeds (Math.random() < forceSuspendVideo), then we add the suspend constraint for p2p and honor the existing configuration for the jvb pc.
If it doesn't succeed, we ignore the configuration and do not add the constraint to either PC. This seems strangely inconsistent. 

Can we just add a second flag for p2p, and then have the "force" property work as an actual force (i.e. override the config in both cases)? 

Or, in case what we want to implement is a forced override only for p2p, then add a second flag for p2p and have the "force" property only override the p2p?

modules/RTC/RTCUtils.js
```diff
@@ -869,17 +869,41 @@ class RTCUtils extends Listenable {
...
+                        // Allows sending of video to be suspended if the
+                        // bandwidth estimation is too low.
+                        this.pcConstraints.optional.push(
+                            { googSuspendBelowMinBitrate: true });
+                    }
+
+                    // There's no reason not to use this for p2p
```
bgrozev: The comment is now obsolete

modules/RTC/RTCUtils.js
hristoterezov: I believe `disableSuspendVideo` is already documented. I was left with the impression that `forceSuspendVideo` will be something temporary for internal use only, that's why I prefer to not add it into the config. WDYT?

modules/RTC/RTCUtils.js
hristoterezov: Thanks for the proposal! 

I'm usually very bad with naming variables, etc. I'll change the name! 

Also I like the idea about `forceSuspendVideo = forceSuspendVideo || 1`!

modules/RTC/RTCUtils.js
hristoterezov: My understanding about the task is the following:
We want to implement a way to enable the suspend video parameter for 1/2 of the users and disable it for the other half. I have the impression that this parameter will be temporary, only for internal use. @emcho please correct me if I'm wrong.

I'm sorry that I haven't added this to the description to avoid confusion.

I wouldn't say that the "strangely inconsistent" behavior is introduced by the  `forceSuspendVideo` option (maybe only the naming is a little "strangely inconsistent" and the lack of comments :) ). The situation before the PR is the following:
We have config option `disableSuspendVideo` that will disable the suspend video only for the JVB peer connection but in the same time the p2p peer connection will always use the suspend video. I think this is the cause for the inconsistent behavior.

Basically my implementation is preserving the previous logic and makes sure that if somebody is using a config without `forceSuspendVideo` the behavior will be exactly the same as before.

I agree that if we keep the 'force' naming it make sense to override `disableSuspendVideo` option. Although both of them are specified in the config and I don't think it actually matters from technical perspective or especially if we are going to remove this option soon.

@bgrozev I think you have a preference for fixing the situation by adding one more option for the p2p peer connection, right? Isn't this too complicated and hard to document? What would be the use case where somebody will want to enable suspend video only for one of the peer connections? I'm also worried that if we introduce new option for disabling the p2p peer connection, we would want to change the `disableSuspendVideo` option naming which will break the backward compatibility. My proposal is to use `disableSuspendVideo` option for all peer connections. WDYT?

bbaldino: > We want to implement a way to enable the suspend video parameter for 1/2 of the users 
> and disable it for the other half.

is this for both peer connections (p2p and bridge) or only the bridge?

hristoterezov: My understanding is for both. I can't think of a reason why would we split them. @emcho could you please confirm?

bgrozev: We discussed this in person. Let's wait for Hristo's new version and comment on that.

hristoterezov:
@bgrozev @bbaldino I've just pushed the new version. It turned out we want to do this only for the p2p peer connection!

modules/RTC/RTCUtils.js
```diff
@@ -876,10 +876,34 @@ class RTCUtils extends Listenable {
...
+                 * the option is the ratio:
+                 * (users with suspended video enabled)/(all users).
+                 *
+                 * Note: The option is not documented because it is temporary
+                 * and only for internal testing purpose.
+                 *
+                 * @type {number}
+                 */
+                const forceP2PSuspendVideoRatio
+                    = options.testing.forceP2PSuspendVideoRatio;
```
bbaldino: we can change this to
`const forceP2PSuspendVideoRatio = options.testing.forceP2PSuspendVideoRatio || 1`

and then get rid of the `typeof forceP2PSuspendVideoRatio !== 'number'` part of the if below

modules/RTC/RTCUtils.js
```diff
@@ -876,10 +876,34 @@ class RTCUtils extends Listenable {
...
+                 * part of the users on the P2P peer connection. The value of
+                 * the option is the ratio:
+                 * (users with suspended video enabled)/(all users).
+                 *
+                 * Note: The option is not documented because it is temporary
+                 * and only for internal testing purpose.
+                 *
+                 * @type {number}
+                 */
+                const forceP2PSuspendVideoRatio
```
bbaldino: i prefer `forceP2PSuspendVideoProbability` to ratio, but maybe just my preference.

bbaldino:
👍 thanks hristo, couple comments but looks much cleaner!

bgrozev:
LGTM

modules/RTC/RTCUtils.js
```diff
@@ -876,10 +876,34 @@ class RTCUtils extends Listenable {
...
+
+                // If <tt>forceP2PSuspendVideoRatio</tt> is invalid (not a
+                // number) fallback to the default behavior (enabled for every
+                // user).
+                if (typeof forceP2PSuspendVideoRatio !== 'number'
+                        || Math.random() < forceP2PSuspendVideoRatio) {
+                    logger.info(`Enable suspend video mode for p2p (ratio=${
+                        forceP2PSuspendVideoRatio})`);
+                    Statistics.analytics.addPermanentProperties({
+                        forceP2PSuspendVideo: true
```
bgrozev: Do we want to log a message if we disable the suspension (since this is when we actually introduce a change)?

modules/RTC/RTCUtils.js
hristoterezov: Now we have only one place where we check `typeof forceP2PSuspendVideoRatio !== 'number' `. That's why I included it in the if condition. 

Also `const forceP2PSuspendVideoRatio = options.testing.forceP2PSuspendVideoRatio || 1` will handle only the case when the option is undefined. So it won't be the same.

I prefer to put the check in the if condition instead of assigning a value because this way we will save a Math.random call in the cases where the option has invalid type. It's not a big deal but in this case it is very easy to do and I don't think the code suffers from it.

modules/RTC/RTCUtils.js
hristoterezov: Hmm. I don't think it is necessary because if you have the message it is enabled and if not - it should be disabled. I think this is enough to check what was the state.

modules/RTC/RTCUtils.js
bbaldino: are you saying we need to handle the case where someone configures `forceP2PSuspendVideoRatio` to a value that isn't a number?  that doesn't seem like something we'd want to worry about...i doubt we do that for all the configuration values, do we?  even if we don't want to cover that, i still think it'd be better to handle it in the initialization of `forceP2PSuspendVideoRatio` rather than in the if.

hristoterezov: @bbaldino I think it's good practice to do the type checking. Personally for me it's hard to predict the result from `0.5 < false` or `0.5 < {}` :) I can't tell about the other config props. Do you prefer not to check for this? Do you think it's bad to check for this?

The benefit of not doing this in the initialization is that we won't execute Math.random for some cases and the code doesn't look uglier. I really don't see any benefits of handling this in the initialization. I'm sorry if I'm missing something! Why do you prefer to do this in the initialization?

bbaldino: few thoughts on this:

1. type-checking can be great, but there are plenty of places where don't do it...why start doing it here?  do we do it consistently for all values in config?  if we want to start with this one, that's also ok...but i think it'd be better to do type-checking and setting of default-values for config variables in a central place rather than all over.

2. i definitely don't think we should do `0.5 < {}`, i was saying we should specifically avoid that by getting `forceP2PSuspendVideoRatio` to its expected type (a number between 0 and 1) as soon as possible.  in general, the earlier we do that, the more code below can safely assume that `forceP2PSuspendVideoRatio` will be a number between 0 and 1, and not have to do other error-checking (i realize it's only accessed in one place below now, but in the future that could change).

3. for someone else reading the code, i think the following if makes the intent much clearer:
```
    if (Math.random() < forceP2PSuspendVideoRatio)
```
plus, why do we consider a value that isn't a number to be equal to '100% probability'?

4. i don't think optimizing for avoiding a call to `Math.random()` is worth sacrificing (in my opinion) readability (without data to show it's causing performance problems).

hristoterezov: Thanks for the explanation!

1. I totally agree. But for now we are not doing this in the central place and we don't have a rule for it. So until we come up with something I think it is better for this feature to handle it where I use it instead of not handling it at all.
2. I believe your idea from 1) will address this, right?
3. Probably but if you compare:
```
const forceP2PSuspendVideoRatio = (typeof forceP2PSuspendVideoRatio !== 'number') ? 1 : forceP2PSuspendVideoRatio;
 if (Math.random() < forceP2PSuspendVideoRatio)
...
```
versus
```
 if (typeof forceP2PSuspendVideoRatio !== 'number' || Math.random() < forceP2PSuspendVideoRatio)
...
```

I definitely prefer the second one.

Also I consider invalid values as 100% probability because this is the default behavior and deployments that don't specify this parameter for example should continue working as they were working before. Does that make sense?

4. Sure! I agree, but as I explained in 3) I don't think we are sacrificing anything. I'd implement it your way before because we were using it in more than one place but now I definitely like the current version more. Also I don't believe that in future we'll use this parameter anywhere else because this is temporary code only for internal testing purpose that will probably be removed very soon(in the following weeks...).

PR #608, opened 2017-10-11
Uses optional statsId to report to callstats and push it to presence.

damencho:
The feature is behind a flag which is disabled by default.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -254,8 +254,14 @@ JitsiConference.prototype._init = function(options = {}) {
         // bellow).
         const windowLocation = window.location;
 
+        let callStatsAliasName = this.myUserId();
+
+        if (config.enableDisplayNameInStats && config.displayName) {
+            callStatsAliasName = config.displayName;
+        }
+
         this.statistics = new Statistics(this.xmpp, {
-            callStatsAliasName: this.myUserId(),
+            callStatsAliasName,
             callStatsConfIDNamespace:
                 config.callStatsConfIDNamespace
                     || (windowLocation && windowLocation.hostname)
@@ -263,7 +269,8 @@ JitsiConference.prototype._init = function(options = {}) {
             callStatsCustomScriptUrl: config.callStatsCustomScriptUrl,
             callStatsID: config.callStatsID,
             callStatsSecret: config.callStatsSecret,
-            roomName: this.options.name
+            roomName: this.options.name,
+            swapUserNameAndAlias: config.enableStatsID
         });
     }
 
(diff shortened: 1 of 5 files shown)


modules/statistics/statistics.js
```diff
@@ -318,8 +318,10 @@ Statistics.prototype.startCallStats = function(tpc, remoteUserID) {
         if (!CallStats.initBackend({
             callStatsID: this.options.callStatsID,
             callStatsSecret: this.options.callStatsSecret,
-            userName,
-            aliasName: this.options.callStatsAliasName
+            userName: this.options.swapUserNameAndAlias
+                ? this.options.callStatsAliasName : userName,
+            aliasName: this.options.swapUserNameAndAlias
+                ? userName : this.options.callStatsAliasName
```
virtuacoplenny: What benefit does the swap provide? I'm guessing for easier searching/filtering but I'm unsure.

JitsiConference.js
```diff
@@ -2392,9 +2413,20 @@ JitsiConference.prototype._startP2PSession = function(peerJid) {
...
+
+    let remoteID = Strophe.getResourceFromJid(this.p2pJingleSession.peerjid);
+
+    if (this.options.config.enableStatsID) {
+        const participant = this.participants[remoteID];
+
+        if (participant) {
+            remoteID = participant.getStatsId() || remoteID;
+        }
+    }
```
virtuacoplenny: I'm not gonna be strict on this: do you see a simple way to remove the duplicated logic for getting the remoteID?

JitsiParticipant.js
```diff
@@ -20,8 +20,9 @@ export default class JitsiParticipant {
      * @param displayName
      * @param {Boolean} hidden - True if the new JitsiParticipant instance is to
      * represent a hidden participant; otherwise, false.
+     * @param {string} starsId - optional participant statsId
      */
-    constructor(jid, conference, displayName, hidden) {
+    constructor(jid, conference, displayName, hidden, statsId) {
```
virtuacoplenny: Trivial nits: typo in "starsId" and maybe it should be "statsID" to be consistent with the capitalization in the config?

modules/statistics/statistics.js
damencho: Yes, so till now we are sending callstatsUsername as alias and username is the endpoint ID which is taken from the jid and changes every call. To view aliases you need to be admin.
With this option that I'm adding we send callstatsUsername as username, and eventually the display name (behind a flag) as an alias.
As callstatsUsername is not changing between calls in callstats you can search and find all calls for this user and only admins can see its real name if this option is enabled.

JitsiParticipant.js
damencho: Ok, will change it. Its confusing cause switching between project (javascript and java) and in one place is statsID and in the other statsID and in the colibri messages its stats-id. I should check all PRs :)

JitsiConference.js
damencho: Well will not make it any better ...
        if (participant && participant.getStatsId()) {
            remoteID = participant.getStatsId();
        }
Do you think this is better?

virtuacoplenny:
Jenkins test this please

damencho:
The tests will not run, as this feature is a set of 5 PRs and we need to first merge the jitsi one, deploy maven artifacts and fix jicofo and jvb pom files.

damencho:
Jenkins, test this please.

damencho:
Jenkins, test this please.

PR #609, opened 2017-10-12
fix(sdp): group ssrcs when converting to jingle

virtuacoplenny:
Ssrcs not grouped together in the jingle element
are not recognized by the bridge. When parsing an
sdp with ssrcs that are not listed consecutively
it is important to still group them together
in the jingle element.


bbaldino:
also i'll note here that this is an issue we ran into when doing testing with edge

modules/xmpp/SDP.js
```diff
@@ -267,26 +267,18 @@ SDP.prototype.toJingle = function(elem, thecreator) {
...
-                        if (linessrc !== ssrc) {
-                            elem.up();
-                            ssrc = linessrc;
-                            elem.c('source', { ssrc,
-                                xmlns: 'urn:xmpp:jingle:apps:rtp:ssma:0' });
-                        }
-                        const kv = line.substr(idx + 1);
+                const ssrcMaps = SDPUtil.parseSSRC(this.media[i]);
+
+                Object.keys(ssrcMaps).forEach(availableSsrc => {
```
saghul: drive-by comment :-) can you please use for..in or for..of where appropriate?, we have been writing all new code with those for a while, thanks!

virtuacoplenny:
Rebased onto latest master to get the package-lock changes, squashed while I was at it, and force pushed.

PR #610, opened 2017-10-12
feat(h264): add ability to disable H.264

saghul:
This is performed by stripping all traces of H.264 from the local SDP before
passing it to `setLocalDescription`.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -45,6 +45,8 @@ const SIM_LAYER_RIDS = [ SIM_LAYER_1_RID, SIM_LAYER_2_RID, SIM_LAYER_3_RID ];
  * @param {boolean} options.disableRtx if set to 'true' will disable the RTX
  * @param {boolean} options.enableFirefoxSimulcast if set to 'true' will enable
  * experimental simulcast support on Firefox.
+ * @param {boolean} options.disableH264 If set to 'true' H264 will be
+ *      disabled by removing it from the SDP.
  * @param {boolean} options.preferH264 if set to 'true' H264 will be preferred
  * over other video codecs.
  *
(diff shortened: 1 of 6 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1572,14 +1574,19 @@ TraceablePeerConnection.prototype.setLocalDescription = function(
 
     this.trace('setLocalDescription::preTransform', dumpSDP(localSdp));
 
-    if (this.options.preferH264) {
+    if (this.options.disableH264 || this.options.preferH264) {
```
bbaldino: having these two settings being discrete looks like it's made things a bit of a pain.  what about a single 'h264' configuration value that could be set to something like 'enabled', 'preferred' or 'disabled'? that way we wouldn't have to worry about all the erroneous interactions between the two settings.

(this could be done as a string in config.js or a value (e.g. 0, 1, 2) either of which would map to a constant or enum in the js itself)

modules/xmpp/SDPUtil.js
```diff
@@ -595,6 +595,61 @@ const SDPUtil = {
...
+    /**
+     * Strips the given codec from the given mline. All related RTX payload
+     * types are also stripped. If the resulting mline would have no codecs,
+     * it's disabled.
+     *
+     * @param {object} videoMLine the video mline object from an sdp as parsed
+     * by transform.parse.
+     * @param {string} codecName the name of the codec which will be stripped.
+     */
+    stripVideoCodec(videoMLine, codecName) {
```
bbaldino: this is a good candidate for a unit test, i think.  i have one for the preferCodec already you could follow.

modules/RTC/TraceablePeerConnection.js
saghul: TBH I don't like either. I went with this due to the urgency of the matter, but personally I'd like something like `videoCodecs = ['h264', 'vp8']` which would filter all codecs, and only allow those in that order.

Also, adding a new h264 section right now would me more of a pain because of backwards compat, added to the fact that we want this today.

modules/xmpp/SDPUtil.js
saghul: True that! Will do.

saghul:
Updated with 100% more tests!

modules/RTC/TraceablePeerConnection.js
bbaldino: ok, i will reluctantly relent :) 

(i didn't mean a new h264 section, btw, just a single value) like:
old:
```
    preferH264: {true|false}
    disableH264: {true|false}
```
new:
```
    h264: {'enabled'|'preferred'|'disabled'}
```

bgrozev: One more place where the lack of overridable config parameters makes things harder.

PR #613, opened 2017-10-17
Enables desktop sharing when using opera.

damencho:
To install the chrome webstore extension, the user must already have installed "Download Chrome Extension". On missing extension we open a popup window which allows the user to add the extension and then install it.
screen shot 2017-10-17 at 00 01 41
screen shot 2017-10-17 at 00 01 51
screen shot 2017-10-17 at 00 02 13
screen shot 2017-10-17 at 00 02 34
screen shot 2017-10-17 at 00 02 49

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -198,8 +198,9 @@ const ScreenObtainer = {
             logger.info('Using Temasys plugin for desktop sharing');
 
             return obtainWebRTCScreen;
-        } else if (RTCBrowserType.isChrome()) {
-            if (RTCBrowserType.getChromeVersion() < 34) {
+        } else if (RTCBrowserType.isChrome() || RTCBrowserType.isOpera()) {
+            if ((RTCBrowserType.getChromeVersion()
+                    || RTCBrowserType.getOperaVersion()) < 34) {
                 logger.info('Chrome extension not supported until ver 34');
 
                 return null;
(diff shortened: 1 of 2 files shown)


saghul:
Question: since Bave uses Electron, does something like this also work for it?

damencho:
I'm not familiar with Bave @saghul. Not sure what do they have, for our jitsi-meet-electron for example we needed to provide obtainDesktopStreams implementation, not sure how do they manage that.

modules/RTC/ScreenObtainer.js
```diff
@@ -389,6 +392,19 @@ const ScreenObtainer = {
...
+            // for opera there is no inline install
+            // extension "Download Chrome Extension" allows us to open
+            // the chrome webstore and install from there and then activate our
+            // extension
+            if (RTCBrowserType.isOpera()) {
+                options.listener('waitingForExtension',
+                    getWebStoreInstallUrl(this.options));
+                this.checkForChromeExtensionOnInterval(
+                    options, streamCallback, failCallback);
+
```
hristoterezov: Maybe it will be nice if we create a function ( handleExternalInstall or something like that) for this 2 lines.

modules/RTC/ScreenObtainer.js
```diff
@@ -198,8 +198,11 @@ const ScreenObtainer = {
             logger.info('Using Temasys plugin for desktop sharing');
 
             return obtainWebRTCScreen;
-        } else if (RTCBrowserType.isChrome()) {
-            if (RTCBrowserType.getChromeVersion() < 34) {
+        } else if (RTCBrowserType.isChrome() || RTCBrowserType.isOpera()) {
+            if ((RTCBrowserType.isChrome()
+                    && RTCBrowserType.getChromeVersion() < 34)
+                || (RTCBrowserType.isOpera()
+                        && RTCBrowserType.getOperaVersion() < 34)) {
```
hristoterezov: Since we have already checked for the browser type let's do the check:
```
(RTCBrowserType.getChromeVersion() || RTCBrowserType.getOperaVersion()) < 34
```

WDYT?

damencho:
Does it look better now? :)

saghul:
@damencho Sorry, I mistyped Brave, but I think you got it ;-)

Being based on Chrome means it will almost always follow the Chrome code path. They launched Chrome extensions support recently: https://www.brave.com/loading-chrome-extensions-in-brave/ I think it warrants at least a cursory look :-)

PR #614, opened 2017-10-17
Update strophe and use it via imports.

hristoterezov:


modules/xmpp/Caps.js
```diff
@@ -41,7 +44,7 @@ export default class Caps extends Listenable {
         if (!this.disco) {
             throw new Error(
                 'Missing strophe-plugins '
-                + '(disco and caps plugins are required)!');
+                + '(disco plugins are required)!');
```
saghul: "disco plugin is required"

modules/xmpp/strophe.emuc.js
```diff
@@ -2,15 +2,17 @@
  * can only handle a single MUC room
  */
 
-/* global $, Strophe */
+/* global $ */
```
saghul: shouldn't this be the first comment in the file?

modules/xmpp/xmpp.js
```diff
@@ -1,7 +1,10 @@
-/* global $, Strophe */
+/* global $ */
 
 import { getLogger } from 'jitsi-meet-logger';
-const logger = getLogger(__filename);
+import { Strophe } from 'strophe.js';
+
+// Strophe plugins:
```
saghul: pointless comment?

hristoterezov: I wanted to separate the plugins because we are just importing them without using exported value. If you think this is not useful I don't mind removing it.

hristoterezov: OK. I agreed! We have also our plugins and the comment is confusing. I'll remove it!

PR #617, opened 2017-10-18
feat(1080p): support on chrome using adapter

virtuacoplenny:
- Add a new browser check so adapter shim usage can be gated.
- Get track resolution for stats from the track itself to account
  for browser resolution fallback logic. Do this only if
  we can be sure adapter has shimmed it in.
- Create a new getUserMediaFlow, with RTC being the orchestration
  for various RTCUtils calls.
- Remove connection quality stat "resolution" which was being
  emitted but not used but listeners.

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -249,7 +249,7 @@ export default {
      *     A promise that returns an array of created JitsiTracks if resolved,
      *     or a JitsiConferenceError if rejected.
      */
-    createLocalTracks(options, firePermissionPromptIsShownEvent) {
+    createLocalTracks(options = {}, firePermissionPromptIsShownEvent) {
         let promiseFulfilled = false;
 
         if (firePermissionPromptIsShownEvent === true) {
@@ -268,7 +268,7 @@ export default {
         window.connectionTimes['obtainPermissions.start']
             = window.performance.now();
 
-        return RTC.obtainAudioAndVideoPermissions(options || {})
+        return RTC.obtainAudioAndVideoPermissions(options)
             .then(tracks => {
                 promiseFulfilled = true;
 
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -69,10 +69,20 @@ export default class JitsiLocalTrack extends JitsiTrack {
         this.sourceId = sourceId;
         this.sourceType = sourceType;
 
-        // FIXME Currently, Firefox is ignoring our constraints about
-        // resolutions so we do not store it, to avoid wrong reporting of local
-        // track resolution.
-        this.resolution = RTCBrowserType.isFirefox() ? null : resolution;
+        if (RTCBrowserType.usesNewGumFlow()) {
+            // Get the resolution from the track itself because it cannot be
+            // certain which resolution webrtc has fallen back to using.
+            this.resolution = track.getSettings().height;
+
+            // Cache the constraints of the track in case of any this track
+            // model needs to call getUserMedia again, such as when unmuting.
+            this._constraints = track.getConstraints();
+        } else {
+            // FIXME Currently, Firefox is ignoring our constraints about
+            // resolutions so we do not store it, to avoid wrong reporting of
+            // local track resolution.
+            this.resolution = RTCBrowserType.isFirefox() ? null : resolution;
+        }
 
         this.deviceId = deviceId;
 
@@ -356,38 +366,52 @@ export default class JitsiLocalTrack extends JitsiTrack {
                 facingMode: this.getCameraFacingMode()
             };
 
-            if (this.resolution) {
-                streamOptions.resolution = this.resolution;
+            if (RTCBrowserType.usesNewGumFlow()) {
+                promise
+                    = RTCUtils.newObtainAudioAndVideoPermissions(Object.assign(
+                        {},
+                        streamOptions,
+                        { constraints: { video: this._constraints } }));
+            } else {
+                if (this.resolution) {
+                    streamOptions.resolution = this.resolution;
+                }
+
+                promise
+                    = RTCUtils.obtainAudioAndVideoPermissions(streamOptions);
             }
 
-            promise = RTCUtils.obtainAudioAndVideoPermissions(streamOptions)
-                .then(streamsInfo => {
-                    const mediaType = this.getType();
-                    const streamInfo
-                        = streamsInfo.find(
+
+            promise.then(streamsInfo => {
+                const mediaType = this.getType();
+                const streamInfo
+                    = RTCBrowserType.usesNewGumFlow()
+                        ? streamsInfo.find(
+                            info => info.track.kind === mediaType)
+                        : streamsInfo.find(
                             info => info.mediaType === mediaType);
 
-                    if (streamInfo) {
-                        this._setStream(streamInfo.stream);
-                        this.track = streamInfo.track;
+                if (streamInfo) {
+                    this._setStream(streamInfo.stream);
+                    this.track = streamInfo.track;
 
-                        // This is not good when video type changes after
-                        // unmute, but let's not crash here
-                        if (this.videoType !== streamInfo.videoType) {
-                            logger.warn(
-                                `${this}: video type has changed after unmute!`,
-                                this.videoType, streamInfo.videoType);
-                            this.videoType = streamInfo.videoType;
-                        }
-                    } else {
-                        throw new JitsiTrackError(TRACK_NO_STREAM_FOUND);
+                    // This is not good when video type changes after
+                    // unmute, but let's not crash here
+                    if (this.videoType !== streamInfo.videoType) {
+                        logger.warn(
+                            `${this}: video type has changed after unmute!`,
+                            this.videoType, streamInfo.videoType);
+                        this.videoType = streamInfo.videoType;
                     }
+                } else {
+                    throw new JitsiTrackError(TRACK_NO_STREAM_FOUND);
+                }
 
-                    this.containers = this.containers.map(
-                        cont => RTCUtils.attachMediaStream(cont, this.stream));
+                this.containers = this.containers.map(
+                    cont => RTCUtils.attachMediaStream(cont, this.stream));
 
-                    return this._addStreamToConferenceAsUnmute();
-                });
+                return this._addStreamToConferenceAsUnmute();
+            });
         }
 
         return promise
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -10,6 +10,7 @@ import JitsiTrackError from '../../JitsiTrackError';
 import * as JitsiTrackErrors from '../../JitsiTrackErrors';
 import Listenable from '../util/Listenable';
 import * as MediaType from '../../service/RTC/MediaType';
+import RTCBrowserType from './RTCBrowserType';
 import RTCEvents from '../../service/RTC/RTCEvents';
 import RTCUtils from './RTCUtils';
 import TraceablePeerConnection from './TraceablePeerConnection';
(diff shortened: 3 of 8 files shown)


virtuacoplenny:
This pull request whittles down adapterjs work to a minimum for getting chrome 1080p. The idea is to get chrome working and then move onto other browsers to use adapter.

JitsiMeetJS.js
```diff
@@ -268,7 +268,11 @@ export default {
         window.connectionTimes['obtainPermissions.start']
             = window.performance.now();
 
-        return RTC.obtainAudioAndVideoPermissions(options || {})
+        const obtainMediaPromise = RTCBrowserType.usesNewGumFlow()
+            ? RTC._newObtainAudioAndVideoPermissions(options)
```
bbaldino: i think i added this leading `_` and we should probably get rid of it since it's called from outside that file.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -74,6 +74,17 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // track resolution.
         this.resolution = RTCBrowserType.isFirefox() ? null : resolution;
 
+        if (RTCBrowserType.usesNewGumFlow()) {
+            // Not all environments support getSettings, however it's important
+            // to get the resolution from the track itself because it cannot be
+            // certain which resolution webrtc has fallen back to using.
+            try {
```
bbaldino: as discussed offline, let's get rid of the try/catch blocks around getSettings

modules/RTC/RTC.js
```diff
@@ -49,6 +49,83 @@ function createLocalTracks(tracksInfo, options) {
...
+function _newCreateLocalTrack({
+    sourceId,
+    sourceType,
+    stream,
+    track,
+    videoType = null
+}, options) {
+    let settingsFromTrack;
+
+    try {
```
bbaldino: no try/catch

modules/RTC/RTC.js
```diff
@@ -143,6 +220,150 @@ export default class RTC extends Listenable {
...
+        // they are not being used anywhere else, so only this function needs to
+        // know about them.
+
+        /**
+         * Executes a request for desktop media if specified in options.
+         *
+         * @returns {Promise}
+         */
+        function maybeRequestDesktopDevice() {
+            const devices = options.devices || [ 'audio', 'video' ];
```
bbaldino: we don't use `audio` or `video` in this block, so can just default it to an empty list if `options.devices` isn't set

modules/RTC/RTC.js
```diff
@@ -143,6 +220,150 @@ export default class RTC extends Listenable {
...
+            const requestedDesktopDevices = devices.filter(
+                device => device === 'desktop');
```
bbaldino: since we only filter for a single type, i wonder if it might be clearer to do:

```
const isDesktopDeviceRequsted = devices.indexOf('desktop') !== -1;

return isDesktopDeviceRequested ? ... : ...
```

modules/RTC/RTC.js
```diff
@@ -143,6 +220,150 @@ export default class RTC extends Listenable {
...
+        /**
+         * Instantiates a {@code JitsiLocalTrack} with the passed in desktop
+         * stream and pushes the model instance to the internal array
+         * localJitsiTracks to be returned later.
+         *
+         * @param {MediaStreamTrack} desktopStream - A track for a desktop
+         * capture.
+         * @returns {void}
+         */
+        function maybeCreateDesktopModel(desktopStream) {
```
bbaldino: i find 'model' here to be a bit confusing.  what about `maybeCreateAndAddDesktopTrack`?

modules/RTC/RTC.js
```diff
@@ -143,6 +220,150 @@ export default class RTC extends Listenable {
...
+        /**
+         * Instantiates a {@code JitsiLocalTrack} with the passed in desktop
+         * stream and pushes the model instance to the internal array
+         * localJitsiTracks to be returned later.
+         *
+         * @param {MediaStreamTrack} avStream - A track for with audio and/or
+         * video track.
+         * @returns {void}
+         */
+        function maybeCreateAVModels(avStream) {
```
bbaldino: `maybeCreateAndAddAVTracks`?

modules/RTC/RTCUtils.js
```diff
@@ -749,7 +763,49 @@ class RTCUtils extends Listenable {
         initRawEnumerateDevicesWithCallback();
 
         return new Promise((resolve, reject) => {
-            if (RTCBrowserType.isFirefox()) {
+            if (RTCBrowserType.usesNewGumFlow()) {
+                this.RTCPeerConnectionType = window.RTCPeerConnection;
+
+                this.getUserMedia
```
bbaldino: i don't remember why we did this, since i think adapter will do either the callback or the promise (based on how many args were passed), so we should double check

modules/RTC/RTCUtils.js
```diff
@@ -749,7 +763,49 @@ class RTCUtils extends Listenable {
...
+
+                                return stream;
+                            })
+                            .catch(err => {
+                                errorCallback && errorCallback(err);
+
+                                return Promise.reject(err);
+                            });
+
+                this.enumerateDevices = callback =>
```
bbaldino: same as above

modules/RTC/RTCUtils.js
```diff
@@ -1024,6 +1034,81 @@ class RTCUtils extends Listenable {
...
+            || RTCBrowserType.isReactNative()) {
+            this.pcConstraints = { optional: [] };
+
+            if (options.useIPv6) {
+                // https://code.google.com/p/webrtc/issues/detail?id=2828
+                this.pcConstraints.optional.push({ googIPv6: true });
+            }
+
+            this.p2pPcConstraints
+                    = JSON.parse(JSON.stringify(this.pcConstraints));
```
bbaldino: wrong indent?

modules/RTC/RTCUtils.js
```diff
@@ -1074,6 +1159,176 @@ class RTCUtils extends Listenable {
...
+     * to capture a desktop sharing video.
+     * @param {string} options.facingMode - Which direction the camera is
+     * pointing to.
+     * @param {string} options.micDeviceId - The device id for the audio capture
+     * device to get audio from.
+     * @private
+     * @returns {Object}
+     */
+    _newGetConstraints(um, options) {
+        const constraints = options.constraints
```
bbaldino: i think adding a comment to call out the copy here would be good.  or (as discussed) a util function maybe?

modules/RTC/RTCUtils.js
virtuacoplenny: It looks to me like one reason it's done this way is to avoid touching screen obtainer, which takes in this.getUserMedia and uses the callback style.

virtuacoplenny: Plus I do not see in chrome/getusermedia.js it looking at the arguments to determine which type of gum to use. It looks to me whatever is on navigator.mediaDevices expects constraints to be passed in and a promise to be returned.

modules/RTC/RTCUtils.js
virtuacoplenny: enumerateDevices looks to always return a promise in the getusermedia.js shim. I don't see it looking for a callback and calling it.

modules/RTC/RTC.js
virtuacoplenny: I just tried chrome 58 on electron by making it be detected as chrome and it does not have getSettings. Whenever older chrome (or electron) gets put into the new gum flow it'll need truthy checks for getSettings. The try/catches were remnants from wip when we didn't take a browser-specific approach with adapter and needed to get around native throwing an error.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -74,6 +74,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // track resolution.
         this.resolution = RTCBrowserType.isFirefox() ? null : resolution;
 
+        if (RTCBrowserType.usesNewGumFlow()) {
```
saghul: why not combine this with the above check in some if-elseif-else ?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -74,6 +74,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            // For new gum flow, getSettings should be available... Get the
+            // resolution from the track itself because it cannot be certain
+            // which resolution webrtc has fallen back to using.
+            this.resolution = track.getSettings().height;
```
saghul: shouldn't this be the smallest of height or width? (I'm thinking a portrait orientation)

modules/RTC/RTC.js
```diff
@@ -143,6 +184,149 @@ export default class RTC extends Listenable {
...
+        function maybeCreateAndAddAVTracks(avStream) {
+            if (!avStream) {
+                return;
+            }
+
+            if (avStream.getAudioTracks().length) {
+                const audioStream = new MediaStream();
+
+                avStream.getAudioTracks().forEach(
+                    t => audioStream.addTrack(t));
```
saghul: can't addTrack get an array?

package.json
```diff
@@ -25,6 +25,7 @@
     "sdp-transform": "2.3.0",
     "strophe": "1.2.4",
     "strophejs-plugins": "0.0.7",
+    "webrtc-adapter": "webrtc/adapter#1eec19782b4058d186341263e7d049cea3e3290a",
```
saghul: why this specific commit? is there no release we can use?

virtuacoplenny: webrtc-adapter didn't do a version bump after merging in a constraints fix that was desired for 1080p. I'm assuming it didn't happen because it'll be a major version bump so they're not rushing it.

modules/RTC/JitsiLocalTrack.js
virtuacoplenny: Essentially I was having fun and trying to reduce the amount of red in the changeset. My needless desire to reduce red was because the previous attempt at using adapter applied adapter globally, causing a scramble situation to fix all environments. That spurred me to avoid changing existing code as little as possible to be iterative in applying adapter. I'll combine the checks.

virtuacoplenny:
force pushed to resolve conflicts in package.json

modules/RTC/JitsiLocalTrack.js
bbaldino: is this what the consumers of this field expect?  if so we could change that.  though right now this flow only applies to chrome...in general we've tried to defer handling of scenarios introduced by other clients until we move those clients over to the new flow, so maybe we can hold off on changing this until portrait-able devices get moved over.

modules/RTC/RTC.js
virtuacoplenny: Doesn't look like it, at least based on mdn docs for addTrack. However, the constructor might be able to take an array of tracks, which is maybe what you were remembering.

modules/RTC/JitsiLocalTrack.js
saghul: I can put my 24 inch monitor in portrait mode :-)

bbaldino: this isn't about display, though, this is about capture

virtuacoplenny: One problem here is how the instance variable resolution is being used. Its primary usage is in ConnectionQuality, where it is used to calculate a connection quality value. The width (jitsiLocalTrack.resolution) is used to look up a height in the hardcoded hash Resolutions, which will return an assumed height and width combo. From there calculation are done based on the height and width. ConnectionQuality and Resolutions does not account for portrait. Later on AvgRTPStatsReporter will take the height and width combo and do some average pixel calculation. I haven't looked at what exactly is being calculated at any point in the stats, just how resolution is being used.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -69,10 +69,20 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+
+            // Cache the constraints of the track in case of any this track
+            // model needs to call getUserMedia again, such as when unmuting.
+            this._constraints = track.getConstraints();
+        } else {
+            // FIXME Currently, Firefox is ignoring our constraints about
+            // resolutions so we do not store it, to avoid wrong reporting of
+            // local track resolution.
+            this.resolution = RTCBrowserType.isFirefox() ? null : resolution;
```
bbaldino: we could do a property for `this.resolution`? maybe a future improvement

modules/RTC/RTCUtils.js
```diff
@@ -346,6 +355,102 @@ function getConstraints(um, options) {
...
+        constraints.audio = false;
+    }
+
+    if (um.indexOf('desktop') >= 0) {
+        if (!constraints.video || typeof constraints.video === 'boolean') {
+            constraints.video = {};
+        }
+
+        constraints.video = {
+            // TODO(brian): think this is chrome only, when we
```
bbaldino: we can yank this comment, don't think it's adding anything

virtuacoplenny:
While these changes have been approved, I think it's worthwhile to think about the scenario where this adapterjs work fails to be completed and what state that leaves the codebase in.

PR #621, opened 2017-10-25
fix(JitsiLocalTrack): Microphone not working detection.

hristoterezov:
This PR should fix the issue with the false "Unable to access your microphone" error dialogs.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -102,16 +102,16 @@ export default class JitsiLocalTrack extends JitsiTrack {
         this._trackEnded = false;
 
         /**
-         * The value of bytes sent received from the statistics module.
+         * Indicates whether data has been sent or not.
          */
-        this._bytesSent = null;
+        this._hasSentData = false;
 
         /**
          * Used only for detection of audio problems. We want to check only once
-         * whether the track is sending bytes ot not. This flag is set to false
+         * whether the track is sending data ot not. This flag is set to false
          * after the check.
          */
-        this._testByteSent = true;
+        this._testDataSent = true;
 
         // Currently there is no way to determine with what device track was
         // created (until getConstraints() support), however we can associate
@@ -553,19 +553,20 @@ export default class JitsiLocalTrack extends JitsiTrack {
     }
 
     /**
-     * Sets the value of bytes sent statistic.
+     * Handles bytes sent statistics.
      * @param {TraceablePeerConnection} tpc the source of the "bytes sent" stat
      * @param {number} bytesSent the new value
      * NOTE: used only for audio tracks to detect audio issues.
      */
-    _setByteSent(tpc, bytesSent) {
-        this._bytesSent = bytesSent;
+    _onByteSentStatsReceived(tpc, bytesSent) {
+        if (bytesSent > 0) {
+            this._hasSentData = true;
+        }
         const iceConnectionState = tpc.getConnectionState();
 
-        if (this._testByteSent && iceConnectionState === 'connected') {
+        if (this._testDataSent && iceConnectionState === 'connected') {
             setTimeout(() => {
-                if (this._bytesSent <= 0) {
-                    // FIXME: Remove ${this}
+                if (!this._hasSentData) {
                     logger.warn(`${this} 'bytes sent' <= 0: \
                         ${this._bytesSent}`);
 
@@ -573,7 +574,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
                     this._fireNoDataFromSourceEvent();
                 }
             }, 3000);
-            this._testByteSent = false;
+            this._testDataSent = false;
         }
     }
 
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -615,7 +615,7 @@ JitsiConferenceEventManager.prototype.setupStatisticsListeners = function() {
                 return;
             }
 
-            track._setByteSent(tpc, stats[ssrc]);
+            track._onByteSentStatsReceived(tpc, stats[ssrc]);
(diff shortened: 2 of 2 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -104,7 +104,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
         /**
          * The value of bytes sent received from the statistics module.
          */
-        this._bytesSent = null;
+        this._bytesSentPerPC = new Map();
```
bbaldino: if we're just going to add up all the bytes sent for each peer connection, why make this a map? it could still just be a number which we add to every time, no?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -559,13 +559,18 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * NOTE: used only for audio tracks to detect audio issues.
      */
     _setByteSent(tpc, bytesSent) {
```
bbaldino: does this method get called multiple times?  then we'd call setTimeout each time, no? is that what we want?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -559,13 +559,18 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
-        this._bytesSent = bytesSent;
+        this._bytesSentPerPC.set(tpc.id, bytesSent);
         const iceConnectionState = tpc.getConnectionState();
 
         if (this._testByteSent && iceConnectionState === 'connected') {
             setTimeout(() => {
-                if (this._bytesSent <= 0) {
-                    // FIXME: Remove ${this}
```
bbaldino: is this fix still something we want to do? if so maybe we should fix it or keep the comment

bbaldino:
if i understand what the bug was, there's still a potential issue here i think...is it possible we could use peer connection A, but have data only sent on peer connection B? then we would detect bytes being sent, but there wouldn't be anything sending on the connection actually being used.  do we want to cover this scenario?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -559,13 +559,18 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * NOTE: used only for audio tracks to detect audio issues.
      */
     _setByteSent(tpc, bytesSent) {
-        this._bytesSent = bytesSent;
+        this._bytesSentPerPC.set(tpc.id, bytesSent);
```
paweldomas: since we end up adding those values anyway during the check, is there any point in keep a map instead of just a singe var as an accumulator ?

paweldomas:
I basically agree with everything that @bbaldino wrote, so handing it over to him

modules/RTC/JitsiLocalTrack.js
hristoterezov: Hmm. When we receive the new value it should represent the bytes since the beginning of the conference, isn't it? If we just add them every time I believe we won't get the total bytes sent but something else.

modules/RTC/JitsiLocalTrack.js
hristoterezov: It is called multiple times but the setTimeout will be called only ones because of the _testByteSent property.

modules/RTC/JitsiLocalTrack.js
hristoterezov: While I was debugging the issue, printing `this` was quite useful. That's why I removed the comment.

modules/RTC/JitsiLocalTrack.js
bbaldino: `this._testByteSent` will be true any time it's non-zero though?

hristoterezov:
@bbaldino So the issue is that for working microphone we are reporting an issue. In the scenario that you described:

> if i understand what the bug was, there's still a potential issue here i think...is it possible we could use peer connection A, but have data only sent on peer connection B? then we would detect bytes being sent, but there wouldn't be anything sending on the connection actually being used. do we want to cover this scenario?

I don't think we should report problem since we are sending bytes which should mean that the microphone is working. Am I missing something?

modules/RTC/JitsiLocalTrack.js
hristoterezov: @paweldomas See the comment above.

bbaldino:
> @bbaldino So the issue is that for working microphone we are reporting an issue. In the scenario that you described:
> 
>> if i understand what the bug was, there's still a potential issue here i think...is it possible we could use peer connection A, but have data only sent on peer connection B? then we would detect bytes being sent, but there wouldn't be anything sending on the connection actually being used. do we want to cover this scenario?
> I don't think we should report problem since we are sending bytes which should mean that the microphone is working. Am I missing something?

nope, makes sense

modules/RTC/JitsiLocalTrack.js
bbaldino: ok, i didn't know what the value we were receiving was, but here we really only care that the number is > 0, right?  so this could even just be a boolean `haveReceivedDataFromMicrophone` which would be set to true any time `_setBytesSent` was called with a value `> 0`

modules/RTC/JitsiLocalTrack.js
hristoterezov: It is true until we execute `setTimeout` and after that we are setting it to false.

bbaldino: oops, i got confused between the two variables.  i see now.

modules/RTC/JitsiLocalTrack.js
hristoterezov: @bbaldino It is boolean now!

PR #622, opened 2017-10-25
Loads CallStats backend only after the script has finished downloading.

damencho:


modules/statistics/statistics.js
```diff
@@ -33,20 +33,46 @@ let isCallstatsLoaded = false;
  * downloading their API as soon as possible and (2) do the downloading
  * asynchronously.
  *
- * @param customScriptUrl
+ * @param options to use for downloading and initializing callstats backend
```
paweldomas: please describe structure of the 'options' argument and that it contains 'customScriptUrl'

modules/statistics/statistics.js
```diff
@@ -33,20 +33,46 @@ let isCallstatsLoaded = false;
...
+    if (!CallStats.initBackend({
+        callStatsID: options.callStatsID,
+        callStatsSecret: options.callStatsSecret,
+        userName: options.swapUserNameAndAlias
+            ? options.callStatsAliasName : userName,
+        aliasName: options.swapUserNameAndAlias
+            ? userName : options.callStatsAliasName,
+        applicationName: options.applicationName
+    })) {
+        logger.warn('CallStats Backend initialization failed bad');
```
paweldomas: should be logger as an error since it failed so bad

modules/statistics/statistics.js
```diff
@@ -33,20 +33,46 @@ let isCallstatsLoaded = false;
...
+            /* loadCallback */ () => _initCallStatsBackend(options)
+        );
         isCallstatsLoaded = true;
     }
+}
+
+/**
+ * Initializes Callstats backend.
+ *
+ * @param options to use for initializing callstats backend
```
paweldomas: describe options or define type somewhere and use it everywhere

PR #628, opened 2017-11-01
feat(adapter): use with firefox

virtuacoplenny:


modules/RTC/RTCBrowserType.js
```diff
@@ -164,8 +164,9 @@ const RTCBrowserType = {
      * @returns {boolean}
      */
     usesNewGumFlow() {
-        return RTCBrowserType.isChrome()
-            && RTCBrowserType.getChromeVersion() >= 61;
+        return (RTCBrowserType.isChrome()
+            && RTCBrowserType.getChromeVersion() >= 61)
+            || RTCBrowserType.isFirefox();
```
bbaldino: let's double check if we can move all versions of firefox over.

modules/xmpp/SDPUtil.js
```diff
@@ -588,7 +588,10 @@ const SDPUtil = {
         }
         if (payloadType) {
             const payloadTypes
-                = videoMLine.payloads.split(' ').map(p => parseInt(p, 10));
+                = videoMLine.payloads
+                    .toString()
```
bbaldino: i'd put a comment about the issue you found here that requires the `toString` call

modules/RTC/RTCBrowserType.js
saghul: Drive by comment: do we have a policy on what versions of Firefox we support? I guess that supporting >= ESR should be ok, that is, >= 52. Older versions could be considered unsupported.

bbaldino: not that i know of, but having a policy of something like `N versions behind stable` would definitely be nice (and make this sort of work much more straightforward)

bbaldino:
jenkins, test this please

modules/RTC/RTCBrowserType.js
virtuacoplenny: I uploaded this branch to my instance, used Firefox 50 and a quick sanity check passed, including screensharing with the extension. The version displayed in ESR 52.4.1.

saghul:
Great work, less code -> more better! :-)

PR #629, opened 2017-11-01
feat(audio-only): be able to lock a browser into capturing audio only

virtuacoplenny:


JitsiMediaDevices.js
```diff
@@ -138,6 +138,18 @@ const JitsiMediaDevices = {
...
         eventEmitter.emit(event, ...args);
+    },
+
+    /**
+     * Returns whether or not the current browser is limited to supporting only
+     * audio send and receive.
+     *
+     * @returns {boolean}
+     */
+    supportsOnlyAudio() {
```
bbaldino: how about inverting this and doing `supportsVideo`

virtuacoplenny: supportVideoOnly or separate checks for supporting audio and supporting video?

bbaldino: yeah, could make it totally dynamic and have both

JitsiMediaDevices.js
```diff
@@ -138,6 +138,18 @@ const JitsiMediaDevices = {
...
+    /**
+     * Returns whether or not the current browser can support capturing video,
+     * be it camera or desktop, and displaying received video.
+     *
+     * @returns {boolean}
+     */
+    supportsVideo() {
+        // Defer to RTCBrowserType to allow exposure of the api to the consumer
+        // but prevent other files from having to import JitsiMediaDevices.
+        return RTCBrowserType.supportsVideo();
```
bbaldino: can we just expose RTCBrowserType?

damencho: I remember we had a discussion at some point about exporting RTCBrowserType so we can use it in jitsi-meet. We decided not to do that ... but I don't remember details.

modules/RTC/RTCUtils.js
```diff
@@ -1532,7 +1532,15 @@ class RTCUtils extends Listenable {
         const maybeRequestCaptureDevices = function() {
             const umDevices = options.devices || [ 'audio', 'video' ];
             const requestedCaptureDevices = umDevices.filter(
-                device => device !== 'desktop');
+                device => {
```
bbaldino: how about
```
umDevices
  .filter(d => {
    if (d === 'audio') return true;
    if (d === 'video' && RTCVideoBrowserType.supportsVideo) return true;
    return false;
  });
```

virtuacoplenny:
Rebased onto latest master to get package-lock, squashed previous commits for addressing feedback, and force pushed.

PR #630, opened 2017-11-01
Add package-lock.json

hristoterezov:


package.json
```diff
@@ -18,14 +18,14 @@
   "dependencies": {
     "async": "0.9.0",
     "current-executing-script": "0.1.3",
-    "jitsi-meet-logger": "jitsi/jitsi-meet-logger",
+    "jitsi-meet-logger": "github:jitsi/jitsi-meet-logger",
```
lyubomir: Oh, you don't want to lock in package.json? I do want to lock in package.json. Given that we're on different opinions let's lay out the advantages and disadvantages of both approaches.

saghul: Did this discussion take place?

saghul: Ah, I see it's fixed now :-)

PR #631, opened 2017-11-02
fix(peerconnection): create a new description to modify sdp

virtuacoplenny:
Firefox has deprecated modifying a description's sdp through
re-assignment. Safari (without temasys) does not allow it.
As such, wherever the sdp has to be modified instead create
a new description with the modified sdp.


hristoterezov:
@virtuacoplenny Could you please rebase with the latest master in order to get the package-lock file?

virtuacoplenny:
Rebased onto latest master and force pushed.

PR #633, opened 2017-11-06
fix(adapter): inject h264 into safari's remote description

virtuacoplenny:
Safari will fail to set its remote description if no
supported video codecs are found. This can happen with
config.p2p.disableH264 set to true, as browsers will strip
H264 from their offers. To get around this problem in a way
that can be removed once this is fixed on Safari's side,
add a function to inject H265 before setting the remote
description.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1806,10 +1806,15 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(
 };
 
 /**
- * Inserts an H264 payload into the description if not already present.
+ * Inserts an H264 payload into the description if not already present. This is
+ * need for Safari WebRTC, which errors when no supported video codec is found
+ * in the offer. Related bug reports:
+ * https://bugs.webkit.org/show_bug.cgi?id=173141
+ * https://bugs.chromium.org/p/webrtc/issues/detail?id=4957
  *
  * @param {RTCSessionDescription} description - An RTCSessionDescription
  * to inject with an H264 payload.
+ * @private
  * @returns {RTCSessionDescription}
  */
 TraceablePeerConnection.prototype._injectH264IfNotPresent = function(
@@ -1818,7 +1823,7 @@ TraceablePeerConnection.prototype._injectH264IfNotPresent = function(
     const videoMLine = parsedSdp.media.find(m => m.type === 'video');
 
     if (!videoMLine) {
-        logger.debug('No videoMLine found, no need tp inject H264.');
+        logger.debug('No videoMLine found, no need to inject H264.');
 
         return description;
     }
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1773,6 +1773,15 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(
         description = normalizePlanB(description);
     }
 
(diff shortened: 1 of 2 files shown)


saghul:
When disableH264 is used the resulting SDP should contain no codecs and be disabled (port set to zero and direction inactive), there should be no need to add H264 back. Can you try to change this for a number instead? https://github.com/jitsi/lib-jitsi-meet/commit/1387e283d4d27beaeb2845f2be1a26c038c6fd32#diff-40e5e13062ea2a3f41fe2e463cba7d4aR639

Also, can you paste the resulting broken SDP?

virtuacoplenny:
Adding back h264 when safari is setting the remote description is a workaround for safari erroring when a supported video codec is not present in the remote description's video mline. For example, Chrome will send VP8 and VP9, properly stripping H264, but likely a Safari-side issue causes that to throw ERROR_CONTENT. Here's a fiddle with a snippet that we were using to repro the issue in the console: https://jsfiddle.net/fsw2n4an/.

virtuacoplenny:
Related webkit bug reports: https://bugs.webkit.org/show_bug.cgi?id=173141, https://bugs.chromium.org/p/webrtc/issues/detail?id=4957

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1773,6 +1773,15 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(
         description = normalizePlanB(description);
     }
 
+    // Safari WebRTC errors when no supported video codec is found in the offer.
+    // To prevent the error, inject H264 into the video mLine.
+    if (RTCBrowserType.usesNewGumFlow() && RTCBrowserType.isSafari()) {
```
saghul: Does Safari ever not use the new GUM flow?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1797,6 +1806,71 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(
 };
 
 /**
+ * Inserts an H264 payload into the description if not already present.
+ *
```
saghul: Can you mention why this is needed here too? Hopefully we can get rid of it soon. A link to the issues would also help.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1797,6 +1806,71 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(
...
+        logger.debug('H264 codec found in video mLine, no need to inject.');
+
+        return description;
+    }
+
+    const { fmtp, payloads, rtp } = videoMLine;
+    const payloadsArray = payloads.toString().split(' ');
+    let dummyPayloadType;
+
+    for (let i = 127; i >= 96; i--) {
```
saghul: i can be const.

modules/RTC/TraceablePeerConnection.js
virtuacoplenny: Safari isn't part of new gum flow at all right now, waiting on only-audio-only to go in first. It's likely Safari 11 and up will be put into new gum while older version are left with temasys for now.

modules/RTC/TraceablePeerConnection.js
virtuacoplenny: I don't think variable i can be const in this case because then variable i cannot be reassigned a value in the increment expression.

saghul: You're right, of course!

PR #634, opened 2017-11-07
Video suspend abtest

bgrozev:
Re-implements the "suspend video " A/B test. 

The test is enabled if config.abTesting.enableSuspendVideoTest==true.

If the test is disabled the previous behavior is preserved, and no properties are added to the analytics events.

If the test is enabled, the two participants in a p2p call will agree on a boolean value based on the JID of the Jingle initiator (which is essentially selected randomly). If the agreed-upon value is true, the suspend video option will be enabled; otherwise, it will be disabled. The two participants will include this value in analytics events with the "abtestP2PSuspendVideo" property.

### DIFF
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -13,6 +13,7 @@ import * as MediaType from '../../service/RTC/MediaType';
 import RTCBrowserType from './RTCBrowserType';
 import RTCEvents from '../../service/RTC/RTCEvents';
 import RTCUtils from './RTCUtils';
+import Statistics from '../statistics/statistics';
 import TraceablePeerConnection from './TraceablePeerConnection';
 import VideoType from '../../service/RTC/VideoType';
 
@@ -433,12 +434,22 @@ export default class RTC extends Listenable {
      * @return {TraceablePeerConnection}
      */
     createPeerConnection(signaling, iceConfig, isP2P, options) {
+        const pcConstraints = RTC.getPCConstraints(isP2P);
+
+        if (typeof options.forceSuspendVideo !== 'undefined') {
+            RTCUtils.setSuspendVideo(pcConstraints, options.forceSuspendVideo);
+
+            Statistics.analytics.addPermanentProperties(
+                { abtestP2PSuspendVideo: options.forceSuspendVideo });
+        }
+
         const newConnection
             = new TraceablePeerConnection(
                 this,
                 this.peerConnectionIdCounter,
                 signaling,
-                iceConfig, RTC.getPCConstraints(isP2P), isP2P, options);
+                iceConfig, pcConstraints,
+                isP2P, options);
 
         this.peerConnections.set(newConnection.id, newConnection);
         this.peerConnectionIdCounter += 1;
@@ -610,7 +621,10 @@ export default class RTC extends Listenable {
      *
      */
     static getPCConstraints(isP2P) {
-        return isP2P ? RTCUtils.p2pPcConstraints : RTCUtils.pcConstraints;
+        const pcConstraints
+            = isP2P ? RTCUtils.p2pPcConstraints : RTCUtils.pcConstraints;
+
+        return Object.assign({}, pcConstraints);
     }
 
     /**
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -624,7 +624,7 @@ export default class RTC extends Listenable {
         const pcConstraints
             = isP2P ? RTCUtils.p2pPcConstraints : RTCUtils.pcConstraints;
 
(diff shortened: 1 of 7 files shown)


modules/RTC/RTC.js
```diff
@@ -610,7 +621,10 @@ export default class RTC extends Listenable {
      *
      */
     static getPCConstraints(isP2P) {
-        return isP2P ? RTCUtils.p2pPcConstraints : RTCUtils.pcConstraints;
+        const pcConstraints
+            = isP2P ? RTCUtils.p2pPcConstraints : RTCUtils.pcConstraints;
+
+        return Object.assign({}, pcConstraints);
```
bbaldino: why was this needed?  to handle one of the fields not being set at all?

modules/RTC/RTCUtils.js
```diff
@@ -1117,36 +1117,10 @@ class RTCUtils extends Listenable {
...
-                Statistics.analytics.addPermanentProperties({
-                    forceP2PSuspendVideo: true
-                });
-
-                this.p2pPcConstraints.optional.push({
-                    googSuspendBelowMinBitrate: true
-                });
-            }
+            // There's no reason not to use this for p2p
+            this.p2pPcConstraints.optional.push({
```
bbaldino: what about using the `setSuspendVideo` helper for this?

bbaldino:
@virtuacoplenny would be good too

modules/RTC/RTC.js
bgrozev: To keep the instances kept in RTCUtils from being modified (which can now happen in RTC.createPeerConnection).

modules/RTC/RTCUtils.js
bgrozev: We could do that, but I don't think we'd gain much. Besides, I think we should remove setSuspendVideo when we don't need the A/B test anymore.

modules/RTC/RTC.js
saghul: Note that this will copy nested objects and arrays by reference, not by value. You can use https://lodash.com/docs/#cloneDeep or `JSON.parse(JSON.stringify(obj))` to avoid it.

modules/RTC/RTC.js
```diff
@@ -433,12 +434,22 @@ export default class RTC extends Listenable {
      * @return {TraceablePeerConnection}
      */
     createPeerConnection(signaling, iceConfig, isP2P, options) {
+        const pcConstraints = RTC.getPCConstraints(isP2P);
+
+        if (typeof options.forceSuspendVideo !== 'undefined') {
```
virtuacoplenny: Will forceSuspendVideo be used outside of AB testing? Right now it's not clear by name that this is an ab test, but maybe that's intentional?

(The AB testing frameworks I'm used to will also have a separate ab testing helper, which makes it more clear what blocks to remove once the test is over. If this is an AB test, maybe it's worthwhile to consider having all checks go through a helper instead of passing in more options.)

modules/xmpp/JingleSessionPC.js
```diff
@@ -2181,4 +2183,23 @@ export default class JingleSessionPC extends JingleSession {
...
+    /**
+     * If the A/B test for suspend video is disabled according to the room's
+     * configuration, returns undefined. Otherwise returns a boolean which
+     * indicates whether the suspend video option should be enabled or disabled.
+     */
+    _forceSuspendVideo() {
+        if (!this.room.options.abTesting
+            || !this.room.options.abTesting.enableSuspendVideoTest) {
+            return;
+        }
```
virtuacoplenny: Does suspend video only work on chrome? Could other browsers affect whatever is being AB tested?

virtuacoplenny:
How is the success of this AB test being measured?

modules/RTC/RTC.js
bgrozev: > Will forceSuspendVideo be used outside of AB testing?

No, its only use is this AB test.

modules/xmpp/JingleSessionPC.js
virtuacoplenny: Thinking about this more, I suppose whatever is being used for analysis could filter user agents.

bgrozev: I just tested with firefox, and it seems that suspend does NOT work there. Filtering is a bit tricky, because in this case we care about the results on the remote side (i.e. if I enable or disable suspending my video, how does that affect the other participant). I'm trying to find out if we can filter by conferences with 2 chrome participants for example.

modules/RTC/RTC.js
bgrozev: Renamed

PR #637, opened 2017-11-08
fix(analytics): initialize permanentProperties inline

virtuacoplenny:
AnalyticsAdapter is initialized on file load. The constructor
calls a method that calls Object.assign. All this happens
before jitsi-meet has loaded polyfills. So, instead
initialize permanentProperties inline with the expected
values, avoiding Object.assign.

### DIFF
--- a/modules/statistics/AnalyticsAdapter.js
+++ b/modules/statistics/AnalyticsAdapter.js
@@ -70,9 +70,10 @@ class AnalyticsAdapter {
         /**
          * Map of properties that will be added to every event
          */
-        this.permanentProperties = Object.create(null);
-        this.addPermanentProperties(
-            { callstatsname: Settings.callStatsUserName });
+        this.permanentProperties = {
+            callstatsname: Settings.callStatsUserName
+        };
+
         this.analyticsHandlers.add(cacheAnalytics);
     }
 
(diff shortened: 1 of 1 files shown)


virtuacoplenny:
There are at least two more errors in lib-jitsi-meet once a conference is actually started. I wasn't expecting that. I can add more fixes to this PR if preferred.

virtuacoplenny:
I ended up making a new pr for the other fixes because they were conceptually different enough to me: https://github.com/jitsi/lib-jitsi-meet/pull/638

hristoterezov:
@virtuacoplenny Isn't it better to load the polyfills for lib-jitsi-meet too?

virtuacoplenny:
It's possible. I recall last time lib-jitsi-meet broke IE11 there was a side conversation deciding it was better for something else to bring in the polyfill. I think it was based on this line in the babel polyfill docs: This will emulate a full ES2015+ environment and is intended to be used in an application rather than a library/tool.

I forgot about that line in the docs for the external_min change...so I'm being a hypocrite. With the root cause known, I'm fairly lenient to adding a polyfill to the lib if you think it's better.

virtuacoplenny:
Although...now I'm leaning towards reverting the jitsi-meet change to add the polyfill to external_lib to prevent modifications of globals for people embedding jitsi-meet and forcing them to have their own polyfills.

virtuacoplenny:
This also makes me wonder why lib-jitsi-meet is being loaded in a script tag and globally instead of through a node_module import jitsi-meet.

PR #639, opened 2017-11-09
feat(adapter): enable audio-only webrtc safari for >= 11

virtuacoplenny:


virtuacoplenny:
@bbaldino, let's discuss this tomorrow to make sure I didn't skip anything. I think the browser checks could also be improved but I'm not sure just yet about how.

virtuacoplenny:
To make a note for future reference, the following features are known not to work:
- Stats. The current stats formatting flows assume old style stats (in an array) but safari webrtc only supports new style stats. The scope to convert the formatting flow looked too big for the current work and seemed okay to postpone to get safari + webrtc out.
- Device selection. Video and audio device selection would not work when testing locally. When testing on webrtc samples, video device selection worked but audio device selection would not.
- Video. Due to reliance on vp8 on the bridge and safari only supporting h264, safari with webrtc will be left in audio only mode.

PR #640, opened 2017-11-13
Fixes support for chrome versions < 61.

damencho:


bbaldino:
i think we should try and check the compatibility for older versions of chrome using the adapter flow...the choice of 61 for the new flow as somewhat arbitrary (unless i'm forgetting something...@virtuacopylenny?) so we can hopefully have all chrome (that we want to support) go down the new flow and keep whittling away at browsers using the old one.

virtuacoplenny:
There's a ticket on my plate to move older chrome to adapter. From what I remember in the initial implementation of adapter, the only issue older chrome encountered was not having track.getSettings.

PR #642, opened 2017-11-14
fix(audio-output): update sinkIds on load and device list change

virtuacoplenny:
When a mac's default audio output is changed, audio elements
are not being updated with the new default. A default audio
output change in the OS fires a DEVICE_LIST_CHANGED event,
so use that to always update the sinkId.

Also, use wrapAttachMediaStream for newGumFlow to be ensure
the proper sinkId is set on audio elements being coupled
a remote track.

### DIFF
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -178,14 +178,12 @@ export default class RTC extends Listenable {
         // tracks handle this event by themselves.
         if (RTCUtils.isDeviceChangeAvailable('output')) {
             RTCUtils.addListener(RTCEvents.AUDIO_OUTPUT_DEVICE_CHANGED,
-                deviceId => {
-                    const remoteAudioTracks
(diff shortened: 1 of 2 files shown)


bbaldino:
so just to make sure i understand:

1) device_list_changed will fire regardless of which device the user has actually selected, right? (i.e. user could have selected 'system default' or a specific device)
2) if a user selected a specific device, and the system default changes, nothing will happen for the user, right?
3) if a user selected the 'system default' device, and the default is changed on the system, then the user's output device will change to match that in their call, yeah?

virtuacoplenny:
1. DEVICE_LIST_CHANGED will fire regardless of the device the user selected. The event is fired from navigator.mediaDevices firing a "devicechange" event.
2. If the user has selected a specific device, other than default, and the user changes the OS's default, then nothing will happen for the user.
3. If the user selects the default and changes the OS's default, the user's output will remain selected as default but will use the new default output.

PR #643, opened 2017-11-15
Cleanup

bgrozev:


JitsiConference.js
```diff
@@ -1450,18 +1450,9 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
...
-
-    if (this.options.config && this.options.config.deploymentInfo
-            && typeof this.options.config.deploymentInfo.crossRegion
-            !== 'undefined') {
-        crossRegion = this.options.config.deploymentInfo.crossRegion;
-    }
     Statistics.analytics.sendEvent(
         'session.initiate', {
-            value: now - this.room.connectionTimes['muc.joined'],
-            label: crossRegion
```
virtuacoplenny: If cross region is redundant, where else is it getting logged?

modules/statistics/AnalyticsAdapter.js
```diff
@@ -64,14 +64,15 @@ class AnalyticsAdapter {
      */
     constructor() {
         this.disposed = false;
-        this.browserName = RTCBrowserType.getBrowserName();
```
virtuacoplenny: Oh wow, I absolutely appreciate getting rid of extra instance variables (even if this is more of a move).

modules/statistics/AnalyticsAdapter.js
```diff
@@ -29,12 +29,12 @@ class CacheAnalytics extends AnalyticsAbstract {
...
      * Cache analytics event.
-     * @param {String} action the name of the event
+     * @param {String} eventName the name of the event
      * @param {Object} data can be any JSON object
      */
-    sendEvent(action, data = {}) {
+    sendEvent(eventName, data = {}) {
         this.eventCache.push({
-            action,
+            eventName,
```
virtuacoplenny: This change makes so much sense, especially for consistency, I have to point out how much sense it makes.

service/statistics/AnalyticsEvents.js
```diff
@@ -0,0 +1,449 @@
...
+
+    /**
+     * Properties: value
+     *
+     * TODO: document
+     */
+    TRANSPORT_REPLACE_SUCCESS: 'xmpp.transport-replace.success'
+};
+
+module.exports = AnalyticsEvents;
```
virtuacoplenny: Overall I actually like this approach. It's not really a strong commitment to a new implementation but it does the trick for now (and maybe for a long time to come). Where there any alternatives you thought about before choosing this one? When do you suppose those document todos will be done?

service/statistics/AnalyticsEvents.js
```diff
@@ -0,0 +1,447 @@
...
+ * Properties: value
+ *
+ * TODO: document
+ */
+export const TRANSPORT_REPLACE_START = 'xmpp.transport-replace.start';
+
+/**
+ * Properties: value
+ *
+ * TODO: document
```
virtuacoplenny: When do you suppose these TODOs will get done? Can you feel me imagining myself staring you down, trying to guilt you into doing them right now?

JitsiConference.js
bgrozev: It is part of config.deploymentInfo, so the client adds it as a permanent property here: https://github.com/jitsi/jitsi-meet/blob/master/react/features/analytics/functions.js#L72

JitsiConference.js
```diff
@@ -1362,53 +1362,84 @@ JitsiConference.prototype.onRemoteTrackRemoved = function(removedTrack) {
...
-                });
-        } else if (!this.isP2PEnabled() && !this.isP2PTestModeEnabled()) {
             this._rejectIncomingCall(
                 jingleSession, {
-                    reasonTag: 'decline',
-                    reasonMsg: 'P2P disabled',
-                    errorMsg: 'P2P mode disabled in the configuration'
+                    reason: 'security-error',
+                    reasonDescription: description,
+                    errorMsg: description
```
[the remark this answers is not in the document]
bgrozev: Yes. I added a message since the old code didn't have one at all. But I fail to understand why in the other places we use two different strings, with similar length and similar meanings.

service/statistics/AnalyticsEvents.js
bgrozev: Not sure about the meaning of "this approach".

If you mean exporting the AnalyticsEvents object, then see the updated commit -- Lyubo suggested exporting the constants individually using "export const", and I went this way because it made the other code cleaner.

If you mean simply having a bunch of constants, then no, I haven't really considered a different approach. I expect that there are libraries and patterns for things like this in javascript, but I don't know of any of them so I just went with the simple thing. This solves most of my issues (being able to trace back where an event comes from, organizing the code in a consistent(-ish) way, and having documentation close to the code).

service/statistics/AnalyticsEvents.js
bgrozev: I feel zero guilt here :)
It's absolutely impractical for someone to dig and document all these events as part of a cleanup task. I did document the events that I knew about, and that I was working on. I think that's the healthy way to do it: if you work on an event (and therefore know what it does), spend 2 minutes to add a description. I added the TODOs hoping that they will remind people to do this in the future.

JitsiConferenceEventManager.js
```diff
@@ -99,13 +111,15 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
 
             Object.keys(chatRoom.connectionTimes).forEach(key => {
                 const value = chatRoom.connectionTimes[key];
+                const eventName = `conference.${_CONNECTION_TIMES_}.${key}`;
```
virtuacoplenny: Can you please confirm the accuracy of this event name? I see _CONNECTION_TIMES_ defined as an empty string so would this event name be "conference..foo"?

bgrozev: Good catch! Will fix

PR #648, opened 2017-11-21
feat(JitsiLocalTrack): fail mute operation when track is disposed

saghul:
If a mute / unmute operation is queued when the track is disposed, fail with an appropriate error.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -8,7 +8,10 @@ import CameraFacingMode from '../../service/RTC/CameraFacingMode';
 import { getLogger } from 'jitsi-meet-logger';
 import JitsiTrack from './JitsiTrack';
 import JitsiTrackError from '../../JitsiTrackError';
-import { TRACK_NO_STREAM_FOUND } from '../../JitsiTrackErrors';
+import {
+    TRACK_IS_DISPOSED,
+    TRACK_NO_STREAM_FOUND
+} from '../../JitsiTrackErrors';
 import {
     LOCAL_TRACK_STOPPED,
     NO_DATA_FROM_SOURCE,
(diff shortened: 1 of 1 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -329,6 +329,10 @@ export default class JitsiLocalTrack extends JitsiTrack {
             return Promise.resolve();
         }
 
+        if (this.disposed) {
+            return Promise.reject(new Error('Track was already disposed'));
```
paweldomas: We might have a special JitsiTrackError for that case which is then handled in jitsi-meet

paweldomas: JitsiTrackErrors.TRACK_IS_DISPOSED

saghul: Wait, I'm turning that into a resolve, let's pretend everything is fine, the track is dead anyway :-)

paweldomas: yes, I guess that would be fine

paweldomas: although I'm not sure about the unmute case for a disposed track...

saghul:
@paweldomas Thanks for the review! After thinking about it, I think it's better to ignore the operation completely, thus making the API easier to work with.

saghul:
> JitsiTrackErrors.TRACK_IS_DISPOSED

Oh, I hadn't noticed we have that. Hum, in that case maybe it's better to be explicit, and ignore the error in Jitsi Meet then!

saghul:
@paweldomas Updated, PTAL.

paweldomas:
Is there a companion jitsi-meet PR ?

saghul:
Yep, coming up!

PR #651, opened 2017-11-21
feat: Allows setting the transport policy of the 1-1 pc.

gpolitis:


modules/xmpp/xmpp.js
```diff
@@ -470,17 +470,35 @@ export default class XMPP extends Listenable {
...
+            iceConfig.p2p.iceServers = p2pStunServers;
+
+            logger.info(
+                'Configured ICE transport policy: ', p2pIceTransportPolicy);
+            iceConfig.p2p.iceTransportPolicy = p2pIceTransportPolicy;
+        }
+
         initEmuc(this);
-        initJingle(this, this.eventEmitter, p2pStunServers);
+        initJingle(this, this.eventEmitter, iceConfig);
```
virtuacoplenny: Looks okay to me. I like the idea of initializing the default iceConfig outside of the strophe.jingle, although it leaves open the possibility of someone not passing in the proper things. Do you think that could be an issue?

Also, nit, p2pIceTransportPolicy can be scoped inside the if block.

gpolitis: I'm trying to understand what your concern is. Is it because iceConfig is a well known parameter in the RTC world and here we're overloading its structure? e.g. people expect a specific format for the iceConfig param.

virtuacoplenny: Nope, you give me too much credit. I'm wondering if anything/anyone could be using JingleConnectionPlugin on its own and expecting JingleConnectionPlugin to be settings its own defaults.

gpolitis: oh, I see now. I wouldn't worry too much about that since the connection plugin didn't have any configuration of its own nor any defaults -the stun/turn servers were set upon initialization-. Is this a public interface? Should I worry I changed the `p2pStunServers` parameter type?

gpolitis:
jenkins test again

saghul:
@gpolitis is there a companion PR on jitsi-meet which documents this option on config.js? Please make one if
 you didn't already. 🚔 🚔

Issue #653, opened 2017-11-23
Can't set custom resolution for newGUM


Kukunin:
I get a problem with setting a resolution for `getUserMedia`. It just ignores `resolution` option for `createLocalTracks` function. 

as a proof here is a stack trace from the debugger:

* `JitsiMeetJS.createLocalTracks` https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiMeetJS.js#L259
* `RTC.obtainAudioAndVideoPermissions` https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/RTC.js#L203
* `RTCUtils.newObtainAudioAndVideoPermissions` https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/RTCUtils.js#L1448
* `RTCUtils.newGetConstraints` https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/RTCUtils.js#L391

`RTCUtils.newGetConstraints` ignores `options.resolution` completely. 

Yes, I know, I can set `options.constraints` directly (I guess that's not a part of the public interface), but there is the other problem: if constraints fail, it goes into endless loop.

The reason is the check here https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiMeetJS.js#L325 to try again with lower the resolution, which affects nothing.

saghul:
> Yes, I know, I can set options.constraints directly (I guess that's not a part of the public interface)

It is part of the interface, you can set the video constraints you like there.

Ping @bbaldino @virtuacoplenny for the endless loop thing.

PR #654, opened 2017-11-23
fix(gum): skip retry if using new gum flow

virtuacoplenny:
The error handling block for unsupported resolution can be
skipped because native gum should be doing automatic retries
at lower resolutions and leaving in the handling will only
cause an infinite loop. Analytics for the error still get
sent in the error handling directly below the retry logic.

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -322,7 +322,8 @@ export default {
             .catch(error => {
                 promiseFulfilled = true;
 
-                if (error.name === JitsiTrackErrors.UNSUPPORTED_RESOLUTION) {
+                if (error.name === JitsiTrackErrors.UNSUPPORTED_RESOLUTION
+                    && !RTCBrowserType.usesNewGumFlow()) {
                     const oldResolution = options.resolution || '720';
                     const newResolution = getLowerResolution(oldResolution);
 


bbaldino:
Just saw https://github.com/jitsi/lib-jitsi-meet/issues/653 and was gonna comment on this.  I thought we actually split the flows completely and that's how we were avoiding this mixup.  Was there a reason we didn't do that?

virtuacoplenny:
I'm fairly certain this slipped by when refactoring to try and leverage existing error/analytics handling.

bbaldino:
ok, was there a reason we didn't separate the flows more?  i'm wondering if that would be clearer.

virtuacoplenny:
You're referring specifically to JitsiMeetJS#createLocalTracks not being split more? I don't remember why it wasn't split more. From what I remember the early implementations for new gum flow's JitsMeetJS#createLocalTracks returned the promise from the then-existing RTC#newObtainAudioAndVideoPermissions without any error handling. Then it was suggested to use the existing promise flow in JitsiMeetJS#createLocalTracks to get the device-list updating logic and the error handling (which caused the bug) and the analytics reporting. And that's part of the problem of splitting it here...almost all the logic should be reused. One idea could be to split the .then/.catch blocks from JitsiMeetJS#createLocalTracks but there would still be a point where the retry-logic would have to be skipped if using the new gum flow.

bbaldino:
> You're referring specifically to JitsiMeetJS#createLocalTracks not being split more? I don't remember why it wasn't split more. From what I remember the early implementations for new gum flow's JitsMeetJS#createLocalTracks returned the promise from the then-existing RTC#newObtainAudioAndVideoPermissions without any error handling. Then it was suggested to use the existing promise flow in JitsiMeetJS#createLocalTracks to get the device-list updating logic and the error handling (which caused the bug) and the analytics reporting. And that's part of the problem of splitting it here...almost all the logic should be reused. One idea could be to split the .then/.catch blocks from JitsiMeetJS#createLocalTracks but there would still be a point where the retry-logic would have to be skipped if using the new gum flow.

Ok...I was just hoping we could split them enough that, rather than having to do the special check in the catch we could have the catch/loop only apply to the old flow and separate it completely from the new flow...but if that isn't straightforward then we can go with this way.  lgtm then but we can talk a bit more tomorrow and merge it after.

PR #655, opened 2017-11-27
Stores initial participant status and skips the status changed event.

damencho:


JitsiParticipant.js
```diff
@@ -23,16 +23,17 @@ export default class JitsiParticipant {
      * @param {Boolean} hidden - True if the new JitsiParticipant instance is to
      * represent a hidden participant; otherwise, false.
      * @param {string} statsID - optional participant statsID
+     * @param {string }status - the initial status if any.
```
paweldomas: {string } -> {string}

damencho: Thanks :) fixed

PR #657, opened 2017-11-28
Reports ssrc to callstats when screen sharing is started.

damencho:
Adds rtc.LOCAL_TRACK_SSRC_UPDATED event to be emitted when ssrc is updated for a local track.
Fixes adding rtc listeners for CREATE_ANSWER_FAILED, CREATE_OFFER_FAILED, SET_LOCAL_DESCRIPTION_FAILED, SET_REMOTE_DESCRIPTION_FAILED.

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -19,6 +19,7 @@ import * as JitsiConferenceErrors from './JitsiConferenceErrors';
 import * as JitsiConferenceEvents from './JitsiConferenceEvents';
 import * as MediaType from './service/RTC/MediaType';
 import RTCEvents from './service/RTC/RTCEvents';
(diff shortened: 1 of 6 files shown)


JitsiConferenceEventManager.js
```diff
@@ -522,27 +523,36 @@ JitsiConferenceEventManager.prototype.setupRTCListeners = function() {
             }
         });
 
-    if (conference.statistics) {
```
virtuacoplenny: Is there a case when conference.statistics is not defined?

damencho: Nope, it is always initialized in JitsiConference. But before that it was initialized after calling this method, so we were failing to register these listeners.

PR #660, opened 2017-11-29
feat(xmpp): lower max XMPP retries to 3

saghul:
Strophe's builtin retry mechanism uses exponential backoff, and 5 retries
results in a very long time.

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -31,7 +31,12 @@ function createConnection(token, bosh = '/http-bind') {
         bosh += `${bosh.indexOf('?') === -1 ? '?' : '&'}token=${token}`;
     }
 
-    return new Strophe.Connection(bosh);
+    const conn = new Strophe.Connection(bosh);
+
+    // The default maxRetries is 5, which is too long.
+    conn.maxRetries = 3;
+
+    return conn;
 }
 
 /**


modules/xmpp/xmpp.js
```diff
@@ -31,7 +31,12 @@ function createConnection(token, bosh = '/http-bind') {
         bosh += `${bosh.indexOf('?') === -1 ? '?' : '&'}token=${token}`;
     }
 
-    return new Strophe.Connection(bosh);
+    const conn = new Strophe.Connection(bosh);
+
+    // The default maxRetries is 5, which is too long.
+    conn.maxRetries = 3;
```
hristoterezov: @saghul Is `maxRetries` documented somewhere? I found it in Strophe's source code but I'm not able to find it in the documentation!

paweldomas:
So what's the time difference between failing 5 times and 3 times ?
I want to know how this numbers compare to 60 seconds we have on XMPP server for session timeout.

modules/xmpp/xmpp.js
saghul: I don't think it's documented. I saw it referenced in the changelog, some PRs and such.

saghul:
@paweldomas It's exponential, so it grows very quickly. It's higher than 60s even with the 3 retries. With 5 it goes up to 7 or 8 minutes.

modules/xmpp/xmpp.js
hristoterezov: Yeah I saw the same! I'll merge the PR but probably we have to be careful with this!

PR #664, opened 2017-12-01
Passes getWiFiStatsMethod to callstats.

damencho:


damencho:
Thanks, but I'm still waiting for confirmation and update from the callstats side. Will follow updates of package json, both here and in: https://github.com/jitsi/jitsi-meet/pull/2256

PR #668, opened 2017-12-04
exclude fec-fr secondary ssrcs when looking for primary ssrcs

bbaldino:
this was causing us to generate an rtx ssrc for the fec stream

### DIFF
--- a/modules/xmpp/RtxModifier.spec.js
+++ b/modules/xmpp/RtxModifier.spec.js
@@ -299,12 +299,9 @@ describe('RtxModifier', () => {
         });
 
         describe('when given an sdp with a flexfec stream', () => {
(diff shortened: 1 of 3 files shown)


modules/xmpp/RtxModifier.spec.js
```diff
@@ -298,6 +298,20 @@ describe('RtxModifier', () => {
             });
         });
 
+        describe('when given an sdp with a flexfec stream', () => {
+            beforeEach(function() {
+                this.multipleVideoSdp = SampleSdpStrings.flexFecSdp;
```
virtuacoplenny: Is it necessary to put flexFecSdp onto an instance variable? It looks like it might be okay to pass in directly to transform.write.

PR #671, opened 2017-12-15
feat(JitsiConference): lower log severity if caps couldn't be discovered

saghul:


saghul:
This is not a big deal, and users are confused when it comes up, so let's lower the severity.

PR #675, opened 2017-12-20
fix(ie11): do not call JSON.stringify with temasys ice candidate

virtuacoplenny:
Calling JSON.stringify on a temasys object causes a stack overflow.
The result is that the JingleSessionPC's work queue never gets
cleared so future work, like video muting, will not get called.
Instead of passing in the candidate directly, manually do
what chrome's implementation of candidate.toJSON does.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2257,7 +2257,15 @@ TraceablePeerConnection.prototype.addIceCandidate = function(
         successCallback,
         failureCallback) {
     // var self = this;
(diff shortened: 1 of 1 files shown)


virtuacoplenny:
@hristoterezov, could you give this a test as well? My VM is flakey with showing video and I was seeing black large video after video muting but sometimes I see it even when joining the meeting.

virtuacoplenny:
Well I was able to test on a home macbook with bootcamp and could confirm the issues does happen with IE11... After local video muting while a remote participant is on large video, large video turns black. It stays that way until video switching occurs on large video. This issue does not occur on Safari 11 temasys, which I enabled on my environment for debugging.

PR #676, opened 2017-12-21
Analytics refactoring

bgrozev:
Please don't merge this yet (pending an associated PR in jitsi-meet).

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -10,7 +10,7 @@ import {
     ACTION_P2P_SWITCH_TO_JVB,
     ICE_ESTABLISHMENT_DURATION_DIFF,
     createJingleEvent,
-    createP2pEvent
+    createP2PEvent
 } from './service/statistics/AnalyticsEvents';
 import AvgRTPStatsReporter from './modules/statistics/AvgRTPStatsReporter';
 import ComponentsVersions from './modules/version/ComponentsVersions';
@@ -1453,7 +1453,6 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
     this.jvbJingleSession = jingleSession;
     this.room.connectionTimes['session.initiate'] = now;
 
-    // Log "session.restart"
     if (this.wasStopped) {
         Statistics.sendAnalyticsAndLog(
             createJingleEvent(ACTION_JINGLE_RESTART));
@@ -2094,7 +2093,7 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
 
         if (this.p2pJingleSession) {
             Statistics.sendAnalyticsAndLog(
-                createP2pEvent(
+                createP2PEvent(
                     ACTION_P2P_FAILED,
                     {
                         initiator: this.p2pJingleSession.isInitiator
@@ -2279,7 +2278,7 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
     this.statistics.startRemoteStats(this.p2pJingleSession.peerconnection);
 
     Statistics.sendAnalyticsAndLog(
-        createP2pEvent(
+        createP2PEvent(
             ACTION_P2P_ESTABLISHED,
             {
                 initiator: this.p2pJingleSession.isInitiator
@@ -2553,7 +2552,7 @@ JitsiConference.prototype._maybeStartOrStopP2P = function(userLeftEvent) {
         // Log that there will be a switch back to the JVB connection
         if (this.p2pJingleSession.isInitiator && peerCount > 1) {
             Statistics.sendAnalyticsAndLog(
-                createP2pEvent(ACTION_P2P_SWITCH_TO_JVB));
+                createP2PEvent(ACTION_P2P_SWITCH_TO_JVB));
         }
         this._stopP2PSession();
     }
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -3,12 +3,10 @@ import { Strophe } from 'strophe.js';
 
 import {
     ICE_ESTABLISHMENT_DURATION_DIFF,
-    P2P_ESTABLISHED,
-    P2P_FAILED,
-    P2P_SWITCH_TO_JVB,
     SESSION_INITIATE,
     SESSION_RESTART,
-    SESSION_TERMINATE
+    SESSION_TERMINATE,
+    createP2pEvent
 } from './service/statistics/AnalyticsEvents';
 import AvgRTPStatsReporter from './modules/statistics/AvgRTPStatsReporter';
 import ComponentsVersions from './modules/version/ComponentsVersions';
@@ -2086,9 +2084,16 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
         // and "bad" connection
         Statistics.analytics.addPermanentProperties({ p2pFailed: true });
 
-        // Log analytics event, but only for the initiator to not count it twice
-        if (this.p2pJingleSession && this.p2pJingleSession.isInitiator) {
-            Statistics.sendEventToAll(P2P_FAILED);
+        if (this.p2pJingleSession) {
+            const event
+                = createP2pEvent(
+                    'failed',
+                    {
+                        initiator: this.p2pJingleSession.isInitiator
+                    });
+
+            Statistics.sendLog(JSON.stringify(event));
+            Statistics.analytics.sendEvent(event);
         }
         this._stopP2PSession('connectivity-error', 'ICE FAILED');
     }
@@ -2253,7 +2258,6 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
         logger.info('Not removing remote JVB tracks - no session yet');
     }
 
-    // Add remote tracks
     this._addRemoteP2PTracks();
 
     // Stop media transfer over the JVB connection
@@ -2261,14 +2265,18 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
         this._suspendMediaTransferForJvbConnection();
     }
 
-    // Start remote stats
     logger.info('Starting remote stats with p2p connection');
     this.statistics.startRemoteStats(this.p2pJingleSession.peerconnection);
 
-    // Log the P2P established event
-    if (this.p2pJingleSession.isInitiator) {
-        Statistics.sendEventToAll(P2P_ESTABLISHED);
-    }
+    const event
+        = createP2pEvent(
+            'established',
+            {
+                initiator: this.p2pJingleSession.isInitiator
+            });
+
+    Statistics.sendLog(JSON.stringify(event));
+    Statistics.analytics.sendEvent(event);
 };
 
 /**
@@ -2536,7 +2544,10 @@ JitsiConference.prototype._maybeStartOrStopP2P = function(userLeftEvent) {
 
         // Log that there will be a switch back to the JVB connection
         if (this.p2pJingleSession.isInitiator && peerCount > 1) {
-            Statistics.sendEventToAll(P2P_SWITCH_TO_JVB);
+            const event = createP2pEvent('switch.to.jvb');
+
+            Statistics.sendLog(JSON.stringify(event));
+            Statistics.analytics.sendEvent(event);
         }
         this._stopP2PSession();
     }
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1605,9 +1605,7 @@ JitsiConference.prototype.onCallEnded = function(
     if (jingleSession === this.jvbJingleSession) {
         this.wasStopped = true;
 
-        const event = createJingleEvent(ACTION_JINGLE_TERMINATE);
-
-        Statistics.sendEventAndLog(event);
+        Statistics.sendEventAndLog(createJingleEvent(ACTION_JINGLE_TERMINATE));
 
         // Stop the stats
         if (this.statistics) {
@@ -2093,14 +2091,13 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
         Statistics.analytics.addPermanentProperties({ p2pFailed: true });
 
         if (this.p2pJingleSession) {
-            const event
-                = createP2pEvent(
+            Statistics.sendEventAndLog(
+                createP2pEvent(
                     ACTION_P2P_FAILED,
                     {
                         initiator: this.p2pJingleSession.isInitiator
-                    });
+                    }));
 
-            Statistics.sendEventAndLog(event);
         }
         this._stopP2PSession('connectivity-error', 'ICE FAILED');
     }
@@ -2279,14 +2276,13 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
     logger.info('Starting remote stats with p2p connection');
     this.statistics.startRemoteStats(this.p2pJingleSession.peerconnection);
 
-    const event
-        = createP2pEvent(
+    Statistics.sendEventAndLog(
+        createP2pEvent(
             ACTION_P2P_ESTABLISHED,
             {
                 initiator: this.p2pJingleSession.isInitiator
-            });
+            }));
 
-    Statistics.sendEventAndLog(event);
 };
 
 /**
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2105,10 +2105,6 @@ JitsiConference.prototype._onIceConnectionRestored = function(session) {
     } else {
         this.isJvbConnectionInterrupted = false;
     }
-
-    if (session.isP2P === this.isP2PActive()) {
-        this.eventEmitter.emit(JitsiConferenceEvents.CONNECTION_RESTORED);
-    }
 };
 
 /**
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2,6 +2,12 @@
 import { Strophe } from 'strophe.js';
 
 import {
+    ACTION_JINGLE_RESTART,
+    ACTION_JINGLE_SI_RECEIVED,
+    ACTION_JINGLE_TERMINATE,
+    ACTION_P2P_ESTABLISHED,
+    ACTION_P2P_FAILED,
+    ACTION_P2P_SWITCH_TO_JVB,
     ICE_ESTABLISHMENT_DURATION_DIFF,
     createJingleEvent,
     createP2pEvent
@@ -1449,11 +1455,11 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
 
     // Log "session.restart"
     if (this.wasStopped) {
-        Statistics.sendEventAndLog(createJingleEvent('restart'));
+        Statistics.sendEventAndLog(createJingleEvent(ACTION_JINGLE_RESTART));
     }
 
     Statistics.analytics.sendEvent(createJingleEvent(
-        'session-initiate.received',
+        ACTION_JINGLE_SI_RECEIVED,
         {
             value: now
         }));
@@ -1599,7 +1605,7 @@ JitsiConference.prototype.onCallEnded = function(
     if (jingleSession === this.jvbJingleSession) {
         this.wasStopped = true;
 
-        const event = createJingleEvent('terminate');
+        const event = createJingleEvent(ACTION_JINGLE_TERMINATE);
 
         Statistics.sendEventAndLog(event);
 
@@ -2089,7 +2095,7 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
         if (this.p2pJingleSession) {
             const event
                 = createP2pEvent(
-                    'failed',
+                    ACTION_P2P_FAILED,
                     {
                         initiator: this.p2pJingleSession.isInitiator
                     });
@@ -2110,6 +2116,10 @@ JitsiConference.prototype._onIceConnectionRestored = function(session) {
         this.isP2PConnectionInterrupted = false;
     } else {
         this.isJvbConnectionInterrupted = false;
+    }
+
+    if (session.isP2P === this.isP2PActive()) {
+        this.eventEmitter.emit(JitsiConferenceEvents.CONNECTION_RESTORED);
     }
 };
 
@@ -2271,7 +2281,7 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
 
     const event
         = createP2pEvent(
-            'established',
+            ACTION_P2P_ESTABLISHED,
             {
                 initiator: this.p2pJingleSession.isInitiator
             });
@@ -2544,7 +2554,8 @@ JitsiConference.prototype._maybeStartOrStopP2P = function(userLeftEvent) {
 
         // Log that there will be a switch back to the JVB connection
         if (this.p2pJingleSession.isInitiator && peerCount > 1) {
-            Statistics.sendEventAndLog(createP2pEvent('switch.to.jvb'));
+            Statistics.sendEventAndLog(
+                createP2pEvent(ACTION_P2P_SWITCH_TO_JVB));
         }
         this._stopP2PSession();
     }
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2249,7 +2249,7 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
 
         Statistics.sendAnalytics(
             ICE_ESTABLISHMENT_DURATION_DIFF,
-            { 'value': establishmentDurationDiff });
+            { value: establishmentDurationDiff });
     }
 
     if (done) {
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -3,9 +3,7 @@ import { Strophe } from 'strophe.js';
 
 import {
     ICE_ESTABLISHMENT_DURATION_DIFF,
-    SESSION_INITIATE,
-    SESSION_RESTART,
-    SESSION_TERMINATE,
+    createJingleEvent,
     createP2pEvent
 } from './service/statistics/AnalyticsEvents';
 import AvgRTPStatsReporter from './modules/statistics/AvgRTPStatsReporter';
@@ -1451,14 +1449,14 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
 
     // Log "session.restart"
     if (this.wasStopped) {
-        Statistics.sendEventToAll(SESSION_RESTART);
+        Statistics.sendEventToAll(createJingleEvent('restart'));
     }
 
-    Statistics.analytics.sendEvent(
-        SESSION_INITIATE,
+    Statistics.analytics.sendEvent(createJingleEvent(
+        'session-initiate.received',
         {
-            value: now - this.room.connectionTimes['muc.joined']
-        });
+            value: now
+        }));
     try {
         jingleSession.initialize(false /* initiator */, this.room, this.rtc);
     } catch (error) {
@@ -1583,6 +1581,8 @@ JitsiConference.prototype._rejectIncomingCall = function(
 
 /**
  * Handles the call ended event.
+ * XXX is this due to the remote side terminating the Jingle session?
+ *
  * @param {JingleSessionPC} jingleSession the jingle session which has been
  * terminated.
  * @param {String} reasonCondition the Jingle reason condition.
@@ -1599,7 +1599,10 @@ JitsiConference.prototype.onCallEnded = function(
     if (jingleSession === this.jvbJingleSession) {
         this.wasStopped = true;
 
-        Statistics.sendEventToAll(SESSION_TERMINATE);
+        const event = createJingleEvent('terminate');
+
+        Statistics.sendLog(JSON.stringify(event));
+        Statistics.analytics.sendEvent(event);
 
         // Stop the stats
         if (this.statistics) {
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1455,12 +1455,13 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
 
     if (this.wasStopped) {
         Statistics.sendAnalyticsAndLog(
-            createJingleEvent(ACTION_JINGLE_RESTART));
+            createJingleEvent(ACTION_JINGLE_RESTART, { p2p: false }));
     }
 
     Statistics.sendAnalytics(createJingleEvent(
         ACTION_JINGLE_SI_RECEIVED,
         {
+            p2p: false,
             value: now
         }));
     try {
@@ -1606,7 +1607,7 @@ JitsiConference.prototype.onCallEnded = function(
         this.wasStopped = true;
 
         Statistics.sendAnalyticsAndLog(
-            createJingleEvent(ACTION_JINGLE_TERMINATE));
+            createJingleEvent(ACTION_JINGLE_TERMINATE, { p2p: false }));
 
         // Stop the stats
         if (this.statistics) {
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1455,10 +1455,11 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
 
     // Log "session.restart"
     if (this.wasStopped) {
-        Statistics.sendEventAndLog(createJingleEvent(ACTION_JINGLE_RESTART));
+        Statistics.sendAnalyticsAndLog(
+            createJingleEvent(ACTION_JINGLE_RESTART));
     }
 
-    Statistics.analytics.sendEvent(createJingleEvent(
+    Statistics.sendAnalytics(createJingleEvent(
         ACTION_JINGLE_SI_RECEIVED,
         {
             value: now
@@ -1605,7 +1606,8 @@ JitsiConference.prototype.onCallEnded = function(
     if (jingleSession === this.jvbJingleSession) {
         this.wasStopped = true;
 
-        Statistics.sendEventAndLog(createJingleEvent(ACTION_JINGLE_TERMINATE));
+        Statistics.sendAnalyticsAndLog(
+            createJingleEvent(ACTION_JINGLE_TERMINATE));
 
         // Stop the stats
         if (this.statistics) {
@@ -2091,7 +2093,7 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
         Statistics.analytics.addPermanentProperties({ p2pFailed: true });
 
         if (this.p2pJingleSession) {
-            Statistics.sendEventAndLog(
+            Statistics.sendAnalyticsAndLog(
                 createP2pEvent(
                     ACTION_P2P_FAILED,
                     {
@@ -2246,7 +2248,7 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
         const establishmentDurationDiff
             = this.p2pEstablishmentDuration - this.jvbEstablishmentDuration;
 
-        Statistics.analytics.sendEvent(
+        Statistics.sendAnalytics(
             ICE_ESTABLISHMENT_DURATION_DIFF,
             { 'value': establishmentDurationDiff });
     }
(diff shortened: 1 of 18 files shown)


JitsiConference.js
```diff
@@ -1601,7 +1599,10 @@ JitsiConference.prototype.onCallEnded = function(
     if (jingleSession === this.jvbJingleSession) {
         this.wasStopped = true;
 
-        Statistics.sendEventToAll(SESSION_TERMINATE);
+        const event = createJingleEvent('terminate');
+
+        Statistics.sendLog(JSON.stringify(event));
```
bbaldino: looks like we do these 2 things a lot, is it worth adding a helper to do both? `Statistics.logAndSendEvent` or something?

JitsiConference.js
```diff
@@ -2086,9 +2087,16 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
         // and "bad" connection
         Statistics.analytics.addPermanentProperties({ p2pFailed: true });
 
-        // Log analytics event, but only for the initiator to not count it twice
-        if (this.p2pJingleSession && this.p2pJingleSession.isInitiator) {
-            Statistics.sendEventToAll(P2P_FAILED);
+        if (this.p2pJingleSession) {
+            const event
+                = createP2pEvent(
+                    'failed',
```
bbaldino: these bare strings don't feel great to me...i assume we only fire each one of these in a single place (so not a big risk of an accidental typo when trying to send the same event somewhere else?).  even still, using named constants is more descriptive which i think is better. (i.e. P2P_FAILED_EVENT as opposed to 'failed')

bbaldino: also if we ever wanted to change the format (e.g. camel case instead of periods) it'd be easier to do since all the constants would (presumably) be defined one place, as opposed to the strings all over the code

JitsiMeetJS.js
```diff
@@ -335,15 +329,19 @@ export default {
                             newResolution);
 
                         Statistics.analytics.sendEvent(
-                            `${GET_USER_MEDIA_FAIL_}.resolution.${
-                                oldResolution}`);
+                            'warning',
```
bbaldino: what determines whether a `createXXXEvent` helper is used or an event is just manually created?

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,344 @@
...
+ *
+ * For events with type 'operational' and 'ui', the required properties are:
+ * action, actionSubject, source
+ *
+ * For events with type 'page', the required properties are:
+ * action, actionSubject, source, containerType, containerId, objectType,
+ * objectId
  */
-class AnalyticsAbstract {
+class AnalyticsAdapter {
```
bbaldino: isn't this format specific to one backend implementation? should it be named something to match that?

bbaldino: oh i see...this is the thing representing 'our' format...so did we just decide to use the same format for our generic format?

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,344 @@
...
     /**
-     *
+     * Creates new AnalyticsAdapter instance.
      */
-    sendEvent() {} // eslint-disable-line no-empty-function
-}
+    constructor() {
+        /**
+         * Whether this AnalyticsAdapter has been disposed or not.
```
bbaldino: what does it mean to be disposed?

JitsiConference.js
bgrozev: > i assume we only fire each one of these in a single place

No, not necessarily. 

But I'll add constants

JitsiMeetJS.js
bgrozev: Convenience. If I only needed a name, I used a simple constant for the name. If I needed it to set something (e.g. the "source"), then I used a factory function.

modules/statistics/AnalyticsAdapter.js
bgrozev: The format is not specific to one backend implementations. This class dictates how backend implementations work and all backend implementations should be updated to use it. 

I don't quite understand the second question -- what is the difference between "our" and the "generic" format?

modules/statistics/AnalyticsAdapter.js
bgrozev: will reword :)

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -495,15 +498,15 @@ export default class AvgRTPStatsReporter {
...
+            = new AverageStatReport('pixels.screen.remote');
 
         /**
          * Average pixels for local video (camera)
          * @type {AverageStatReport}
          * @private
          */
         this._avgLocalCameraPixels
-            = new AverageStatReport('stat_avg_pixels_local');
+            = new AverageStatReport('pixels_.ocal');
```
bbaldino: should this be `pixels.local` ?

modules/statistics/CallStats.js
```diff
@@ -451,17 +451,17 @@ export default class CallStats {
...
      * @param {string} conferenceID the conference ID for which the feedback
      * will be reported.
-     * @param overallFeedback an integer between 1 and 5 indicating the
+     * @param overall an integer between 1 and 5 indicating the
      * user feedback
-     * @param detailedFeedback detailed feedback from the user. Not yet used
+     * @param comment detailed feedback from the user.
      */
-    static sendFeedback(conferenceID, overallFeedback, detailedFeedback) {
+    static sendFeedback(conferenceID, overall, comment) {
```
bbaldino: if we're gonna rename it, how about something like `score` instead of `overall` ?

modules/statistics/AnalyticsAdapter.js
bbaldino: ok, my understanding is this, let me know if i have something wrong:
1) The events are sent as an object
2) The AnalyticsAdapter adapts those objects into a generic format (the one defined in the comments at the top of AnalyticsAdapter)
3) The handlers will adapt the generic format (from AnalyticsAdapter) into whatever they need.

Is that right?

Here I was just noting that the generic format we used in AnalyticsAdapter seems to be the one from one of the backend handlers we'll be using (I can't remember the name) since, from what I remember, that's what the backend format looked like.

JitsiConference.js
```diff
@@ -1453,14 +1449,14 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
 
     // Log "session.restart"
     if (this.wasStopped) {
-        Statistics.sendEventToAll(SESSION_RESTART);
+        Statistics.sendEventToAll(createJingleEvent('restart'));
```
virtuacoplenny: I still think it would be useful to keep these event names as constants so that all event names are in one place.

JitsiConference.js
```diff
@@ -1453,14 +1449,14 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
...
     }
 
-    Statistics.analytics.sendEvent(
-        SESSION_INITIATE,
+    Statistics.analytics.sendEvent(createJingleEvent(
+        'session-initiate.received',
         {
-            value: now - this.room.connectionTimes['muc.joined']
-        });
+            value: now
```
virtuacoplenny: Why the value change?

JitsiConference.js
```diff
@@ -2086,9 +2087,16 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
...
+        if (this.p2pJingleSession) {
+            const event
+                = createP2pEvent(
+                    'failed',
+                    {
+                        initiator: this.p2pJingleSession.isInitiator
+                    });
+
+            Statistics.sendLog(JSON.stringify(event));
+            Statistics.analytics.sendEvent(event);
```
virtuacoplenny: I noticed that the call to sendEventToAll was removed in some cases but not others. What were your requirements for using sendEventToAll versus not using it?

JitsiConference.js
```diff
@@ -2105,10 +2113,6 @@ JitsiConference.prototype._onIceConnectionRestored = function(session) {
     } else {
         this.isJvbConnectionInterrupted = false;
     }
-
-    if (session.isP2P === this.isP2PActive()) {
-        this.eventEmitter.emit(JitsiConferenceEvents.CONNECTION_RESTORED);
-    }
```
virtuacoplenny: Why was this bit removed?

JitsiConferenceEventManager.js
```diff
@@ -51,14 +48,6 @@ export default function JitsiConferenceEventManager(conference) {
...
                 track.getType());
         });
-    conference.on(
-        JitsiConferenceEvents.CONNECTION_INTERRUPTED,
-        Statistics.sendEventToAll.bind(
-            Statistics, CONNECTION_INTERRUPTED));
-    conference.on(
-        JitsiConferenceEvents.CONNECTION_RESTORED,
-        Statistics.sendEventToAll.bind(
-            Statistics, CONNECTION_RESTORED));
```
virtuacoplenny: Why was this removed? It was firing at the wrong times?

JitsiConferenceEventManager.js
```diff
@@ -85,11 +74,14 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
     });
 
     chatRoom.addListener(XMPPEvents.AUDIO_MUTED_BY_FOCUS,
-        value => {
-            Statistics.analytics.sendEvent(REMOTELY_MUTED);
+        () => {
```
virtuacoplenny: This was a good catch. It's always a positive mute event and the promise below always sets isMutedByFocus true.

modules/RTC/RTCUtils.js
```diff
@@ -547,19 +547,29 @@ function pollForAvailableMediaDevices() {
...
+            videoInputDeviceCount,
+            videoOutputDeviceCount,
+            deviceId: device.deviceId,
+            deviceGroupId: device.groupId,
+            deviceKind: device.kind,
+            deviceLabel: device.label
+        };
 
-    Statistics.analytics.sendEvent(DEVICE_LIST, { devices: devicesPropsArray });
+        Statistics.analytics.sendEvent(AVAILABLE_DEVICE, eventProperties);
```
virtuacoplenny: Is there a reason why this was changed so that each device gets pushed up? Could data like device counts not be calculated/queried easily from the data itself?

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,344 @@
...
+         * @type {boolean}
+         */
+        this.disposed = false;
+
+        /**
+         * The set of handlers to which events will be sent.
+         * @type {Set<any>}
+         */
+        this.analyticsHandlers = new Set();
```
virtuacoplenny: Could you make sure this won't completely break IE11? I have a feeling it might because it doesn't have Set.

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,344 @@
...
-        // some events may happen before init or implementation script download
-        // in this case we accumulate them in this array and send them on init
-        this.eventCache = [];
+        // Note that we disable the cache even if the set of handlers is empty.
+        const cache = this.cache;
+
+        this.cache = null;
+        if (cache) {
+            cache.forEach(event => this._sendEvent(event));
+        }
```
virtuacoplenny: Is there a guarantee or even a desire to make sure all analytics handlers for jitsi-meet get assigned at once so they get all the events? Like if an event is cached, and only one handler gets set, then that handler gets the event, but then a later handler that gets set wouldn't get the event. That's a pretty edge-casey situation and now that I've written it out doesn't seem like something worth handling.

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,344 @@
...
+        this._sendEvent(event);
     }
 
-}
+    /**
+     *
+     * @param event
+     */
+    sendOperationalEvent(event) {
+        event.type = TYPE_OPERATIONAL;
```
virtuacoplenny: It's okay to modify in place the passed in event? It probably is but I want to make sure you're aware of in-place modifications.

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,344 @@
...
 
-const cacheAnalytics = new CacheAnalytics();
+        if (!AnalyticsAdapter._verifyRequiredFields(event)) {
+            logger.warn(
```
virtuacoplenny: I like this. I wonder if it's even worth logging an error so it's more obvious when developing.

modules/statistics/AnalyticsAdapter.js
bgrozev: Yeah, the format is almost the same

modules/statistics/AvgRTPStatsReporter.js
bgrozev: Yes!

modules/statistics/CallStats.js
bgrozev: I was going with "rating" and "comment", but I saw that the callstats API calls it "overall", so I kept it this way.
https://www.callstats.io/api/#callstats-senduserfeedback

JitsiConference.js
bgrozev: To be consistent with the rest of the events. Eventually I think this should be moved to a separate module together with all the messy "connectionTimes" logic

JitsiConference.js
bgrozev: sendEventToAll only works with the analytics.sendEvent(eventName, properties) API. If I have an event object instead I can't use it, so I split the call. I'll add another "send-to-all" kind of function as Brian suggested.

JitsiConferenceEventManager.js
bgrozev: We don't need the CONNECTION_RESTORED analytics event, because we can use the ICE_STATE_CHANGED events instead of it

modules/RTC/RTCUtils.js
bgrozev: The new backend client that we're going to use does not support nesting attributes, so we can not (in a nice way) send a list of devices in a single event. And calculating the number of devices based on a number of events could be tricky as we'd have to find a good way to select the set of events.

JitsiConference.js
bgrozev: Oh, I think that might have been a mistake. Will revert

modules/statistics/AnalyticsAdapter.js
bgrozev: That's not a new addition: https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/statistics/AnalyticsAdapter.js#L67

modules/statistics/AnalyticsAdapter.js
bgrozev: I'm assuming that once the user of the library calls setAnalyticsHandlers(), it is their responsibility to handle events. If they wish to handle the case that you describe, they should implement their own cache :)
But seriously: I don't want to keep the cache enabled just because of this. It's just meant to serve for event sent early on before all the initialization in the client has run

modules/statistics/AnalyticsAdapter.js
bgrozev: Yes, which is the reason factory functions return new objects. But it is definitely worth documenting, so I'll add a note.

modules/statistics/AnalyticsAdapter.js
bgrozev: I don't understand the difference in the way errors and warnings are shown, but I'll trust you on this and make it an error.

service/statistics/AnalyticsEvents.js
```diff
@@ -1,447 +1,339 @@
...
- * Properties: value
+ * Events with source 'sporadic' are sent from various places in the code, and
+ * indicate things which happen any action from the user. These events do not
+ * happen in every conference, and do not happen periodically. They are grouped
+ * together by this value for the 'source' attribute, because they may be of
+ * interest when looking for events in a particular conference.
+ * @type {string}
  */
-export const _ICE_ESTABLISHMENT_DURATION = 'establishmentDuration';
+const SOURCE_SPORADIC = 'sporadic';
```
virtuacoplenny: I'm wondering if it's worthwhile to make separate files for event names and event attributes/types and maybe event factories. What do you think?

JitsiMeetJS.js
virtuacoplenny: For someone who doesn't work in lib-jitsi-meet or even does analytics, what would the thought process be when adding a new event and when to use a string vs factory? How would that person know how to proceed?

bgrozev: I added some tips at the top of AnalyticsEvents.js. Let me know if that's not clear, or if the place is not right

service/statistics/AnalyticsEvents.js
bgrozev: I'm not sure. It would be a little bit harder to find the code given a resulting event. And importing is more of a hassle. I'd keep it in one file.

JitsiConference.js
```diff
@@ -1601,7 +1605,9 @@ JitsiConference.prototype.onCallEnded = function(
     if (jingleSession === this.jvbJingleSession) {
         this.wasStopped = true;
 
-        Statistics.sendEventToAll(SESSION_TERMINATE);
+        const event = createJingleEvent(ACTION_JINGLE_TERMINATE);
+
+        Statistics.sendEventAndLog(event);
```
lyubomir: Previously you wrote:
````javascript
Statistics.sendEventAndLog(createJingleEvent(ACTION_JINGLE_RESTART));`
````

Is there are reason that (pretty much) the same source code now looks differently:
````javascript
const event = createJingleEvent(ACTION_JINGLE_TERMINATE);

Statistics.sendEventAndLog(event);
````

From my perspective as a reader of the source code (and maybe the PR) having one and the same thing written in different ways requires unnecessary effort to comprehend.

File: JitsiMeetJS.js [hunk 21]
```diff
@@ -69,28 +63,29 @@ function getLowerResolution(resolution) {
...
+ * @returns {*} the attributes to attach to analytics events.
  */
-function addDeviceTypeToAnalyticsEvent(name, options) {
-    let ret = name;
+function getAnalyticsAttributesFromOptions(options) {
+    const attributes = {
+        audioRequested:
+            options.devices.indexOf('audio') !== -1,
+        videoRequested:
+            options.devices.indexOf('audio') !== -1,
```
lyubomir: @bgrozev, why do both `audioRequested` and `videoRequested` get assigned one and the same value `options.devices.indexOf('audio') !== -1`? Maybe `videoRequested` should use `'video'`?

JitsiMeetJS.js
```diff
@@ -69,28 +63,29 @@ function getLowerResolution(resolution) {
...
+            options.devices.indexOf('desktop') !== -1
+    };
 
-    if (options.devices.indexOf('audio') !== -1) {
-        ret += '.audio';
-    }
-    if (options.devices.indexOf('desktop') !== -1) {
-        ret += '.desktop';
-    }
     if (options.devices.indexOf('video') !== -1) {
```
lyubomir: Supposedly, we already did `options.devices.indexOf('video') !== -1` and the value is now in `attributes.videoRequested` so maybe we don't need to search again?

JitsiMeetJS.js
```diff
@@ -283,9 +278,8 @@ export default {
                     = window.performance.now();
 
                 Statistics.analytics.sendEvent(
-                    addDeviceTypeToAnalyticsEvent(
-                        GET_USER_MEDIA_SUCCESS_, options),
-                    { value: options });
+                    createGetUserMediaEvent(
+                        'success', getAnalyticsAttributesFromOptions(options)));
```
lyubomir: I don't know about you, it always makes me expend more effort than really necessary to mentally parse such lines than:
````javascript
createGetUserMediaEvent(
    'success',
    getAnalyticsAttributesFromOptions(options)));
````

File: modules/statistics/statistics.js [hunk 24]
```diff
@@ -710,16 +706,26 @@ Statistics.reportGlobalError = function(error) {
...
-Statistics.sendEventToAll = function(eventName, data) {
-    this.analytics.sendEvent(eventName, data);
-    Statistics.sendLog(
-        JSON.stringify(
-            {
-                name: eventName,
-                data
-            }));
+Statistics.sendEventAndLog = function(eventName, properties = {}) {
+    let eventToLog = null;
```
lyubomir: Well, `eventToLog` will be assigned bellow so I don't much see the point in saying that it's `null`. As if it's possible that it's equal to `null`.

modules/statistics/statistics.js
```diff
@@ -710,16 +706,26 @@ Statistics.reportGlobalError = function(error) {
...
+
+    // Also support an API with a single object as an event.
+    if (typeof eventName === 'object') {
```
lyubomir: While we're on the subject of `null`, `typeof null` does equal `'object'`. Is `null` a supposed value?

modules/xmpp/strophe.jingle.js
```diff
@@ -204,17 +202,17 @@ class JingleConnectionPlugin extends ConnectionPlugin {
...
+                { value: now }));
 
             sess.replaceTransport($(iq).find('>jingle'), () => {
                 const successTime = window.performance.now();
 
                 logger.info('(TIME) Transport replace success!', successTime);
-                Statistics.analytics.sendEvent(
-                    TRANSPORT_REPLACE_SUCCESS,
-                    { value: successTime });
+                Statistics.analytics.sendEvent(createJingleEvent(
```
lyubomir: It's very likely not really the scope of this PR but it's definitely exacerbating the following problem for me:

Given that `Statistics` is a "class" and `Statistics.analytics` is static, I'm not sure why we're sprinkling our source code with long lines of `Statistics.analytics.sendEvent` plus its arguments which are not always trivial. I'd suggest having a mere function with a name of less length like `sendAnalytics`.

JitsiConference.js
bgrozev: There's no reason, I've made it consistent by removing the constants.

JitsiMeetJS.js
bgrozev: Yes, good catch!

JitsiMeetJS.js
bgrozev: Good point, fixed

JitsiMeetJS.js
bgrozev: Agreed, fixed

modules/statistics/statistics.js
bgrozev: Removed the assigment

modules/statistics/statistics.js
bgrozev: No, I don't want to accept "null" as an event name (or event object). I've added "if(!eventName) return;" to the top of the function

modules/xmpp/strophe.jingle.js
bgrozev: I don't think we want "analytics" to be static. We have a class already, we might as well instantiate it. But I would leave this for another PR.

As for shortening the name, I would like this, but I'm not sure what the best way would be, because analytics now exports a set of functions: sendEvent, sendUIEvent, sendTrackEvent, sendPageEvent, sendOperationalEvent. Do we expose them all with shorter names, or do we just add sendAnalytics as a shortcut for the (most commonly used) sendEvent?

bgrozev: I will think about this some more, because I'm not convinced anymore that we need to expose these multiple functions from analytics.

bgrozev: I'm removing the sendXXXEvent methods and only leaving sendEvent. And also adding a Statistics.sendAnalytics() shortcut.

JitsiConference.js
```diff
@@ -2257,22 +2268,23 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
...
 
-    // Start remote stats
     logger.info('Starting remote stats with p2p connection');
     this.statistics.startRemoteStats(this.p2pJingleSession.peerconnection);
 
-    // Log the P2P established event
-    if (this.p2pJingleSession.isInitiator) {
-        Statistics.sendEventToAll(P2P_ESTABLISHED);
-    }
+    Statistics.sendAnalyticsAndLog(
```
bbaldino: @virtuacoplenny brought up a point when we were talking yesterday...do we ever want to fire an event and not log it?  should logging an event just be the default?

bbaldino: (that being said, this name is probably still good...just wondering if the plain `sendEvent` should exist)

bgrozev: Well "logging" in this context refers to sending it to callstats, so we actually rarely want to "log" analytics events.

bbaldino: oh, if the `Log` in `sendAnalyticsAndLog` refers to sending to callstats, what does the `sendAnalytics` refer to?

bgrozev: Sending to whatever event-based analytics backend has been configured

modules/statistics/statistics.js
```diff
@@ -710,16 +706,43 @@ Statistics.reportGlobalError = function(error) {
...
+        eventToLog = eventName;
+    } else {
+        eventToLog = {
+            name: eventName,
+            properties
+        };
+    }
+    Statistics.sendLog(JSON.stringify(eventToLog));
```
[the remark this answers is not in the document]
virtuacoplenny: I'm wondering, maybe for the future, if sendLog should be doing the stringifying or the logic directly above.

service/statistics/AnalyticsEvents.js
```diff
@@ -1,447 +1,426 @@
...
- * TODO: document, reformat
- *
- * Known full event names:
- * audio.no_data_from_source
- * video.no_data_from_source
+ * Events with source 'sporadic' are sent from various places in the code, and
+ * indicate things which happen any action from the user. These events do not
+ * happen in every conference, and do not happen periodically. They are grouped
+ * together by this value for the 'source' attribute, because they may be of
+ * interest when looking for events in a particular conference.
```
virtuacoplenny: What kind of interest are you talking about that gets solved by using a sporadic group? Could you give me an example?

JitsiConferenceEventManager.js
```diff
@@ -85,11 +75,14 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
...
+            // TODO: Add a way to differentiate between commands which caused
+            // us to mute and those that did not change our state (i.e. we were
+            // already muted).
+            Statistics.sendAnalytics(createRemotelyMutedEvent());
```
virtuacoplenny: I noticed sendAnalytics takes in eventName and attributes but that calls into analytics.sendEvent(eventName, properties). In some cases you're passing eventName as an object. Can you confirm that analytics.sendEvent properly handle both strings and objects as eventName (and maybe where I can look for its definition)?

virtuacoplenny: Nevermind...I got lost while searching in the changset. It does handle it fine.

service/statistics/AnalyticsEvents.js
```diff
@@ -1,447 +1,426 @@
...
+ * @param errorType ???
+ * @param errorMessage ???
  */
-export const P2P_ESTABLISHED = 'p2p.established';
+export const createConnectionFailedEvent = function(errorType, errorMessage) {
+    return {
+        type: TYPE_OPERATIONAL,
+        name: 'connection.failed',
+        errorType,
+        errorMessage
```
virtuacoplenny: Should errorType and errorMessage be attributes? Other actions seem to follow a strict set of keys and errorType and errorMessage aren't among them.

JitsiMeetJS.js
```diff
@@ -69,28 +63,28 @@ function getLowerResolution(resolution) {
...
-        // we have video add resolution
-        ret += `.video.${options.resolution}`;
+function getAnalyticsAttributesFromOptions(options) {
+    const attributes = {
+        audioRequested:
+            options.devices.indexOf('audio') !== -1,
+        videoRequested:
+            options.devices.indexOf('video') !== -1,
+        screenSharingRequested:
+            options.devices.indexOf('desktop') !== -1
```
virtuacoplenny: You could get away with using Array.prototype.includes here if desired instead of indexOf. The method gets polyfilled by something somewhere.

service/statistics/AnalyticsEvents.js
```diff
@@ -1,447 +1,426 @@
...
- * Properties: none
+ * Indicates that a track was unmuted (?).
  *
- * Known full event names:
- * getUserMedia.userCancel.extensionInstall
- *
- * TODO: document, reformat
+ * Properties:
+ *      mediaType: the media type of the local track ('audio' or 'video').
+ *      type: the type of the track ('local' or 'remote').
```
virtuacoplenny: This is trackType now.

service/statistics/AnalyticsEvents.js
```diff
@@ -1,447 +1,426 @@
...
+ * Creates an event which indicates the Time To First Media (TTFM).
+ * It is measured in milliseconds relative to the beginning of the document's
+ * lifetime (i.e. the origin used by window.performance.now()), and it excludes
+ * the following:
+ * 1. The delay due to getUserMedia()
+ * 2. The period between the MUC being joined and the reception of the Jingle
+ * session-initiate from jicofo. This is because jicofo will not start a Jingle
+ * session until there are at least 2 participants in the room.
+ *
+ * @param attributes the attributes to att the the event. Currently used fields:
```
virtuacoplenny: "att the" typo

modules/statistics/statistics.js
```diff
@@ -710,16 +706,43 @@ Statistics.reportGlobalError = function(error) {
...
  */
-Statistics.sendEventToAll = function(eventName, data) {
-    this.analytics.sendEvent(eventName, data);
-    Statistics.sendLog(
-        JSON.stringify(
-            {
-                name: eventName,
-                data
-            }));
+Statistics.sendAnalyticsAndLog = function(eventName, properties = {}) {
```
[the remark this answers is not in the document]
bgrozev: Renamed to "event"

modules/statistics/statistics.js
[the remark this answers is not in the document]
bgrozev: Yes, the idea is that if "eventName" is an object, then it describes the whole event (and "properties" isn't handled by the analytics code either)

service/statistics/AnalyticsEvents.js
bgrozev: When I started to write this I thought that I would come across more events for this group. But I didn't. I'll remove this.

service/statistics/AnalyticsEvents.js
bgrozev: Yes, nice catch!

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,297 @@
...
+        }
+
+        // Track events have additional required fields.
+        if (type === TYPE_TRACK) {
+            event.objectType = event.objectType || 'generic-object-type';
+            event.containerType
+                = event.containerType || 'generic-container-type';
+
+            if (!event.objectType || !event.objectId
+                || !event.containerType || !event.containerId) {
```
virtuacoplenny: I think maybe the objectType and containerType checks aren't necessary because of the hardcoded fallbacks above.

JitsiMeetJS.js
bgrozev: Sure

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,297 @@
...
      */
-    setAnalyticsHandlers(handlers) {
-        if (this.disposed) {
-            return;
+    _sendEvent(event) {
+        if (this._maybeCacheEvent(event)) {
+            // The event was consumed by the cache.
+        } else {
+            // We append the permanent properties at the time we send the event,
+            // not at the time we receive it.
```
virtuacoplenny: What is the reasoning again for setting permanent properties on send instead of receive?

modules/statistics/AnalyticsAdapter.js
```diff
@@ -1,130 +1,297 @@
...
-        cacheAnalytics.drainCachedEvents().forEach(
-            ev => this.sendEvent(ev.eventName, ev.data));
     }
 
     /**
-     * Adds map of properties that will be added to every event.
-     * @param {Object} properties the map of properties
+     * Extends an event object with the configured permanent properties.
+     * @param event the event to extend with permanent properties.
+     * @returns {any & ({}|*)} the extended event
```
virtuacoplenny: What is this return type? Also, Object.assign will modify in place. Just fyi.

modules/statistics/AnalyticsAdapter.js
bgrozev: I wasn't sure, so I preserved the existing behavior. The difference would manifest when the cache is enabled, so before the instance is initialized. So it makes sense to add the permanent properties after they have been initialized.

service/statistics/AnalyticsEvents.js
```diff
@@ -1,447 +1,414 @@
...
- * TODO: document, reformat
- *
- * Known full event names:
- * ice.initiator.gatheringDuration
- * ice.responder.gatheringDuration
- * p2p.ice.initiator.gatheringDuration
- * p2p.ice.responder.gatheringDuration
- *
- * Properties: value
+ * The constant which identifies an event of type "ui".
```
virtuacoplenny: Maybe these type comments could briefly explain what each type is for.

modules/statistics/AnalyticsAdapter.js
bgrozev: Yes, that's the intention. Except I wanted to assign to event.attributes. Fixed now.

service/statistics/AnalyticsEvents.js
```diff
@@ -1,447 +1,414 @@
...
- *
- * TTMF: Time To First Media
- *
- * Known full event names:
- * audio.ttfm
- * video.ttfm
- * audio.ttfm.muted
- * video.ttfm.muted
+ * The "action" value for Jingle events which indicates that a session-accept
+ * timed out (???).
```
virtuacoplenny: Why the triple question marks? Is it because of an uncertainty? If so, maybe an explicit todo about verifying the action purpose is better.

service/statistics/AnalyticsEvents.js
bgrozev: I'm still trying to figure that out :(

virtuacoplenny: Okay, yeah, I don't understand the difference either fully either...

JitsiConnection.js
```diff
@@ -25,23 +25,28 @@ export default function JitsiConnection(appID, token, options) {
...
+                createConnectionFailedEvent(errType, msg));
         });
 
     this.addEventListener(JitsiConnectionEvents.CONNECTION_DISCONNECTED,
         msg => {
             // we can see disconnects from normal tab closing of the browser
             // and then there are no msgs, but we want to log only disconnects
             // when there is real error
+            // XXX Do we need the difference in handling between the log and
+            // analytics event here?
```
bbaldino: yeah it seems like this work would be when we get this questions answered, are we going to be able to clean this up?

bbaldino:
i think the special casing for callstats still feels weird...if we're going to create this generic analytics backend, shouldn't callstats be incorporated into that rather than treated differently?  i get that some aspects of callstats will be different (like the fact we initialize it with the peer connection) but that doesn't seem to bad.  if we made the callstats event handling match the others, we'd have:

1) no different notion of sending `analytics` and sending `logs` (which feels confusing as it is, as they aren't really different, they're just different code paths for the event handling that's either a) callstats or b) everything else)
2) filtering for any events happens in the adapter for all analytics backends (just like any format transformations) so the call site is always consistent.

is there another reason pushing this stuff to callstats has to be treated differently?  or maybe i'm misremembering something about the whole `sendAnalytics`, `sendLog`, `sendAnalyticsAndLog` thing

PR #678, opened 2018-01-03
feat(browser_type): expose RTCBrowserType

hristoterezov:


JitsiMeetJS.js
```diff
@@ -480,6 +480,7 @@ export default {
     util: {
         AuthUtil,
         RTCUIHelper,
-        ScriptUtil
+        ScriptUtil,
+        RTCBrowserType
```
virtuacoplenny: Will there be follow up work immediately to move RTCBrowserType to its own repo? Is it worthwhile to leave a comment here if that is the case?

hristoterezov: I hope there will be. We discussed it offline and I was left with the impression that many people wanted to have a discussion before creating the repo. I hope next week we can discuss it and come up with the decision. But right now I don't think we can put a meaningful comment there since we don't know which way we are going to go.

PR #679, opened 2018-01-03
move the tpc transform logs to debug

bbaldino:
jitsi-meet has been changed to have tpc.js on info by default, so these
logs still won't show by default.


bbaldino:
logs can be made visible by calling: `JitsiMeetJS.setLogLevelById(JitsiMeetJS.logLevels.DEBUG, "modules/RTC/TraceablePeerConnection.js");` in the console

PR #682, opened 2018-01-08
feat(TPC): don't log the entire datachannel event

saghul:
In Rect Native it will stall the app as it's huge.


paweldomas:
thanks!

bbaldino:
was there no useful information in the event that would've been nice to extract out individually and log?

saghul:
not that I could see. If we ever need anything from there, we can add it individually, instead of in-bulk 😉

PR #683, opened 2018-01-08
Upgrade NPM dependencies/packages: react-native 0.51

saghul:


lyubomir:
Ah, I merged it as baa5d9b9973db66006aab260808ee445f01d326f without first pushing here so it didn't resolve automatically.

PR #687, opened 2018-01-11
doc: fix link to public lib-jitsi-meet JS file

saghul:


brycepj:
I'm getting a 404 with the new url: `(index):39 GET https://meet.jit.si/libs/lib-jitsi-meet.js 404 ()`

damencho:
@brycepj You can try https://meet.jit.si/libs/lib-jitsi-meet.min.js, does this work for you. We need to once again update this doc.

PR #690, opened 2018-01-19
feat(browser_caps): Implement browser capabilities API.

hristoterezov:
In order to include browser capabilities in RTCBrowserType I needed to
refactor RTCBrowserType.

Depends on https://github.com/jitsi/jitsi-meet/pull/2400


doc/browser-capabilities.md
```diff
@@ -0,0 +1,44 @@
+# Jitsi Meet browser capabilities
+
+An API for checking browser capabilities for Jitsi Meet.
+
+## Installation
+
+To use the Jitsi Meet browser capabilities API in your application you need to add a JS file with `BrowserCapabilities` implementation:
+
+```javascript
+<script src="https://meet.jit.si/libs/browser_capabilities.min.js"></script>
```
bgrozev: Perhaps change that to example.com to make it clear that the file is specific to the deployment

modules/browser/ExternalBrowserCapabilities.js
```diff
@@ -0,0 +1,65 @@
+import BrowserCapabilities from './BrowserCapabilities';
+
+/**
+ * API for checking the capabilities for Jitsi Meet for the current or passed
```
bgrozev: Is this the entry point for the new bundle? Can we add a note on what the purpose is?

modules/browser/BrowserCapabilities.js
```diff
@@ -0,0 +1,288 @@
...
+    constructor(isUsingIFrame = false, browserInfo) {
+        super(browserInfo);
+
+        const name = this.getBrowserName();
+        const version = this.getBrowserVersion();
+
+        const browserCapabilities = capabilitiesDB[name] || [];
+
+        const capabilitiesByVersion = browserCapabilities.find(entry =>
+            !entry.version || bowser.compareVersions([
```
bgrozev: Can we use on of the wrappers here (e.g. this.isCurrentVersionGreaterThan)?

modules/statistics/RTPStatsCollector.js
```diff
@@ -386,10 +386,10 @@ StatsCollector.prototype._defineGetStatValueMethod = function(keys) {
...
 
     switch (this._browserType) {
-    case RTCBrowserType.RTC_BROWSER_CHROME:
-    case RTCBrowserType.RTC_BROWSER_OPERA:
-    case RTCBrowserType.RTC_BROWSER_NWJS:
-    case RTCBrowserType.RTC_BROWSER_ELECTRON:
+    case browsers.CHROME:
+    case browsers.OPERA:
+    case browsers.NWJS:
+    case browsers.ELECTRON:
```
bgrozev: It'd be nice to fix the TODO below -- we now have the tools for it!

modules/browser/ExternalBrowserCapabilities.js
hristoterezov: The entry point is index.js and there is a comment. See https://github.com/jitsi/lib-jitsi-meet/blob/423de642b682c5c24b344662970731d001261e16/modules/browser/index.js#L4

bgrozev:
The general approach looks good to me! Some suggestions for simpler and more consistent naming:

BrowserDetection.getBrowserName() -> getName()
BrowserDetection.getBrowserVersion() -> getVersion()
BrowserDetection.isCurrentVersionLowerThan() -> isVersionLessThan()
BrowserDetection.isCurrentVersionGreaterThan() -> isVersionGreaterThan()
BrowserDetection.compareVersionWith() -> compareVersion()
BrowserDetection.isCurrentVersionEqualTo -> isVersionEqualTo()

BrowserCapabilities.isXXXSupported() -> supportsXXX()  (for p2p and VideoMuteOnConnInterrupted)

bowserNames2JitsiNames -> bowserName2JitsiName (it maps a name to a name)

modules/statistics/RTPStatsCollector.js
hristoterezov: Do you mean moving the logic from the switch somewhere into the browser module? Or maybe creating a method supportsStatsMethod()?

hristoterezov: Also I will definitely do it but I'd rather do another PR to avoid complexity in this one.

modules/browser/ExternalBrowserCapabilities.js
bgrozev: The documentation in this file doesn't make it clear why it exists (at least I don't understand it after reading it). It's not obvious why it just re-exposes part of an already available API and nothing else. But perhaps this is a common practice that I'm not familiar with.

modules/statistics/RTPStatsCollector.js
bgrozev: > Do you mean moving the logic from the switch somewhere into the browser module? Or maybe creating a method supportsStatsMethod()?

I don't know whether this use justifies an additional method, and if it does I don't know what the right place for it is (BrowserCapabilities vs kept local to RTPStatsCollector). In any case, it should use browser.isChrome() instead of comparing the name to the constant explicitly.

modules/browser/BrowserCapabilities.js
hristoterezov: Yes. Awesome idea! Thanks! I tried it and it simplifies a lot the code (now bowser import and constants for name and version are gone)!

modules/statistics/RTPStatsCollector.js
hristoterezov: Oh sure! It was looking weird to me when I was refactoring but I didn't figured why :)

modules/browser/ExternalBrowserCapabilities.js
hristoterezov: My idea behind it is the following:
1) The index.js is the only entry point for the module.
2) Everything in the module should be used trough the index.js

Most of the features in Jitsi Meet are implemented this way. I'm not sure if we have it somewhere in a code style guide.

modules/RTC/ScreenObtainer.js
```diff
@@ -262,12 +260,12 @@ const ScreenObtainer = {
 
         if (desktopSharingFirefoxMaxVersionExtRequired === -1
             || (desktopSharingFirefoxMaxVersionExtRequired >= 0
-                && RTCBrowserType.getFirefoxVersion()
-                    <= desktopSharingFirefoxMaxVersionExtRequired)) {
+                && !browser.isVersionGreaterThan(
+                    desktopSharingFirefoxMaxVersionExtRequired))) {
```
virtuacoplenny: To confirm, it's okay that this is checking greater than only and not greater than or equal too?

modules/browser/BrowserDetection.js
```diff
@@ -0,0 +1,305 @@
...
+    REACT_NATIVE
+} from './browsers';
+
+const logger = getLogger(__filename);
+
+/**
+ * Maps the names of the browsers from bowser to the internal names defined in
+ * ./browsers.js
+ */
+const bowserName2JitsiName = {
```
virtuacoplenny: Could we call this bowserNameToJitsiName instead? "to" and "2" are not the same to me.

modules/browser/capabilities.json
```diff
@@ -0,0 +1,121 @@
...
+        }
+    ],
+    "electron": [
+        {
+            "capabilities": {
+                "audioIn": true,
+                "audioOut": true,
+                "videoIn": true,
+                "videoOut": true,
+                "screenSharing": false
```
virtuacoplenny: Why is screensharing false for electron?

modules/RTC/ScreenObtainer.js
hristoterezov: It is checking <= because it is "not grater than". You probably missed `!`.

modules/browser/BrowserDetection.js
hristoterezov: Sure!

modules/browser/capabilities.json
hristoterezov: Good catch! Copy/paste mistake :)

PR #691, opened 2018-01-22
feat(analytics): shield handler execution

saghul:
Avoid throwing an exception if an analytics handler throws one. Emit a warning
though, so the problem can be eventually caught and fixed in the handler.


saghul:
Jenkins, test this please.

PR #692, opened 2018-01-22
feat(TPC): append TPC ID to stream IDs

paweldomas:
This commit will append "-" + tpc.id to every local 'MSID', 'cname',
'label' and 'mslabel', before feeding the local SDP to the Jingle layer.
It will make stream IDs unique across TraceablePeerConnection instances
and prevent from conflicts in some corner cases.

For example this will fix a problem where if the client drops
the conference without leaving the XMPP MUC gracefully and will join
the conference again without recreating the local tracks it would lead
to the MSID conflict, because the stream is still advertised by
"the ghost" participant.

### DIFF
--- a/modules/RTC/LocalSdpMunger.js
+++ b/modules/RTC/LocalSdpMunger.js
@@ -162,6 +162,48 @@ export default class LocalSdpMunger {
     }
 
     /**
+     * Modifies 'cname', 'msid', 'label' and 'mslabel' by appending
+     * the id of {@link LocalSdpMunger#tpc} at the end, preceding by a dash
+     * sign.
+     *
+     * @param {MLineWrap} mediaSection - The media part (audio or video) of the
+     * session description which will be modified in place.
+     * @returns {void}
+     * @private
+     */
+    _transformMediaIdentifiers(mediaSection) {
+        const pcId = this.tpc.id;
+
+        for (const ssrcLine of mediaSection.ssrcs) {
+            switch (ssrcLine.attribute) {
+            case 'cname':
+            case 'label':
+            case 'mslabel':
+                ssrcLine.value = ssrcLine.value && `${ssrcLine.value}-${pcId}`;
+                break;
+            case 'msid': {
+                if (ssrcLine.value) {
+                    const streamAndTrackIDs = ssrcLine.value.split(' ');
+
+                    if (streamAndTrackIDs.length === 2) {
+                        const streamId = streamAndTrackIDs[0];
+                        const trackId = streamAndTrackIDs[1];
+
+                        ssrcLine.value
+                            = `${streamId}-${pcId} ${trackId}-${pcId}`;
+                    } else {
+                        logger.warn(
+                            'Unable to munge local MSID'
+                                + `- weird format detected: ${ssrcLine.value}`);
+                    }
+                }
+                break;
+            }
+            }
+        }
+    }
+
+    /**
      * Maybe modifies local description to fake local video tracks SDP when
      * those are muted.
      *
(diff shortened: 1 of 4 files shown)


modules/RTC/LocalSdpMunger.js
```diff
@@ -161,6 +161,37 @@ export default class LocalSdpMunger {
...
+
+        for (const ssrcLine of mediaSection.ssrcs) {
+            switch (ssrcLine.attribute) {
+            case 'cname':
+            case 'label':
+            case 'mslabel':
+                ssrcLine.value = `${ssrcLine.value}-${pcId}`;
+                break;
+            case 'msid': {
+                const streamAndTrackIDs = /^(.+)\s(.+)$/;
```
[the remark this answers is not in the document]
paweldomas: yes.. let me see

modules/RTC/LocalSdpMunger.js
```diff
@@ -161,6 +161,44 @@ export default class LocalSdpMunger {
...
+            case 'cname':
+            case 'label':
+            case 'mslabel':
+                ssrcLine.value = ssrcLine.value && `${ssrcLine.value}-${pcId}`;
+                break;
+            case 'msid': {
+                if (ssrcLine.value) {
+                    const streamAndTrackIDs = ssrcLine.value.split(' ');
+
+                    if (streamAndTrackIDs.length === 2) {
```
bbaldino: do we not need to worry about the `else` case here?

paweldomas: I was thinking about logging a warning, WDYT ?

paweldomas: I didn't add it in the first place, because it's not like the first entry point for the local SDP where it would be supposed to validate SDP's correctness. I just thought that if we'd be adding checks like that everywhere maybe it would be too much... But I can add it

bbaldino: i think a warning which prints out the msid would be good, yeah.

bbaldino:
> For example this will fix a problem where if the client drops
> the conference without leaving the XMPP MUC gracefully and will join
> the conference again without recreating the local tracks it would lead
> to the MSID conflict, because the stream is still advertised by
> "the ghost" participant.

@paweldomas so the tpc id will not reset in this case, right?

paweldomas:
The case we're worried about is when the app drops without crashing, so that it can preserve the local media stream (which will have the same MSID). On web it will have new stream every time, because we reload. If the app crashes on mobile, then the TPC id counter will reset, but also it will be a new stream and MSID.

bbaldino:
> The case we're worried about is when the app drops without crashing, so that it can preserve the local media stream (which will have the same MSID)

So tpc id will not reset in this case, right?

paweldomas:
yes, it will not

modules/RTC/LocalSdpMunger.js
paweldomas: done

modules/RTC/RTC.js
```diff
@@ -19,6 +19,12 @@ import VideoType from '../../service/RTC/VideoType';
 
 const logger = getLogger(__filename);
 
+/**
+ * The counter used to generated id numbers assigned to peer connections
+ * @type {number}
+ */
+let peerConnectionIdCounter = 1;
```
saghul: Call me paranoid, but in contexts where the app is not reloaded, this could overflow. Maybe we should reset this to 1 once it reaches `Number.MAX_SAFE_INTEGER`: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/MAX_SAFE_INTEGER

paweldomas: I don't think our app is able to stay that long without crashing ;) but yeah I can add that to both counters (there's another one for track IDs).

paweldomas:
@damencho are these test results acceptable ?
Chrome to Chrome:
screen shot 2018-01-25 at 09 03 03
FF to Chrome:
screen shot 2018-01-25 at 09 03 24

damencho:
Have you tested it with Firefox? As those tests are no longer running.
As for the screenshot ringoverlay is broken on FF, I already reported to @bgrozev that the Udp test also is fialing from time to time, as on chrome or on ff we are sometimes connected using tcp.

paweldomas:
It worked on the FF and those results are from when the tests were running. Is the ring overlay broken by this PR or in general ?

damencho:
In general.

PR #693, opened 2018-01-23
fix: Fixes version comparison.

bgrozev:


modules/browser/BrowserDetection.js
```diff
@@ -273,7 +273,7 @@ export default class BrowserDetection {
      */
     isVersionGreaterThan(version) {
         if (this._version) {
-            return this.compareVersion(version) === 1;
+            return this.compareVersion(version) === -1;
         }
```
bbaldino: should the else cases be handled?

bbaldino: or maybe assert that it's valid

bgrozev: I think this is by design, the documentation says it returns undefined if the current version is not set

bbaldino: then should the `compareVersion` method handle it?

bbaldino: oh, and it already is

bgrozev: "the current version" is the version of the current browser, which is this._version. `compareVersion` also handles this part. I don't know what the behavior is when the `version` parameter is undefined (it is passed this way to bowser)

bbaldino: i'm not sure i follow, the check in these places (which wraps the call in `if (this._version)` is identical to what `compareVersion` does, ya?  i'm just saying we should get rid of the `if (this._version)` check anywhere that's calling `compareVersion` since it's redundant.

hristoterezov: The checks are removed!

PR #695, opened 2018-01-26
Minor changes to doc & XEP-0215

guusdk:
Mostly non-functional changes.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -306,8 +306,7 @@ class JingleConnectionPlugin extends ConnectionPlugin {
         this.connection.sendIQ(
             $iq({ type: 'get',
                 to: this.connection.domain })
-                .c('services', { xmlns: 'urn:xmpp:extdisco:1' })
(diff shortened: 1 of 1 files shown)


modules/xmpp/strophe.jingle.js
```diff
@@ -307,8 +306,7 @@ class JingleConnectionPlugin extends ConnectionPlugin {
         this.connection.sendIQ(
             $iq({ type: 'get',
                 to: this.connection.domain })
-                .c('services', { xmlns: 'urn:xmpp:extdisco:1' })
-                .c('service', { host: `turn.${this.connection.domain}` }),
```
damencho: You are removing the turn service?

guusdk: No, not at all. This change modifies the request to an XMPP server to retrieve a list of service addresses. The response will include all service addresses that are configured at the server. This code seems to add a specific server to that request, which doesn't make much sense. This could be intended as a kind of filter ("give me the details of server `turn.example.org`"), but that's not how XEP-0215, nor the Prosody implementation of it, works.

PR #696, opened 2018-01-26
fix(SS): On FF.

hristoterezov:


bgrozev:
What is the reason for not fixing this once and for all by extending the API that we now provide (BrowserDetection) to handle numbers?

saghul:
FWIW, I agree with @bgrozev here. Our API should accept either.

saghul:
> Everywhere else the versions should be specified as strings.

This is very counter intuitive. The version **is** a number. Don't you think that having the library deal with both would shield us from potential problems moving forward? Better be safe than sorry!

hristoterezov:
@saghul 

> This is very counter intuitive. The version is a number. Don't you think that having the library deal with both would shield us from potential problems moving forward? Better be safe than sorry!

My current Chrome version:  64.0.3282.119
I don't see how this could be represented by number! 

Currently we are supporting to specify the version as `'63'` which can be represented by number but anyway as I said before I see more cons than pros on the matter of allowing numbers:

1. It is harder to explain and confusing what you should specify for version and when (obviously if you want to specify 64.0.3282.119 you should use string).
2. Not many benefits
3. Encouraging the developers to write inconsistent code. For example if the half of the calls are executed with number and the other half with string.
4. It is confusing.

Again, this is not a public API and it is used only inside our project. So for now if we don't need number I don't see why should we add it to the API.

Also what do you mean  by

> Better be safe than sorry

It's up to us to pass the correct types to the functions. And if somebody uses it with number, we'll see the error and will change the code to use string. 

I don't accept this argument. In order to be safe we should handle all types, for example objects too. Do you mean that we need to handle objects just to be safe and how will we do that?

saghul:
@hristoterezov 

> My current Chrome version: 64.0.3282.119
> I don't see how this could be represented by number!

All we care is about the major, so that's 64. Which is a number, isn't it? :-)

> It is harder to explain and confusing what you should specify for version and when (obviously if you want to specify 64.0.3282.119 you should use string).

How would you compare that?

> Not many benefits

That's not an argument.

> Encouraging the developers to write inconsistent code. For example if the half of the calls are executed with number and the other half with string.

This I agree with.

> It is confusing.

That's not an argument.

> I don't accept this argument. In order to be safe we should handle all types, for example objects too. Do you mean that we need to handle objects just to be safe and how will we do that?

No. But we have made this mistake ourselves in the past. Not to mention, some of these numbers may come from the user (config.js) so our internal functions should be gentle with what they accept, IMHO.

hristoterezov:
> All we care is about the major, so that's 64. Which is a number, isn't it? :-)

Not really see capabilities.json. In the future maybe somewhere else.

> > It is harder to explain and confusing what you should specify for version and when (obviously if you want to specify 64.0.3282.119 you should use string).

> How would you compare that?

At the end with bowser's method compareVersions.

> > I don't accept this argument. In order to be safe we should handle all types, for example objects too. Do you mean that we need to handle objects just to be safe and how will we do that?

> No. But we have made this mistake ourselves in the past. Not to mention, some of these numbers may come from the user (config.js) so our internal functions should be gentle with what they accept, IMHO.

Yes we have made the mistake ourselves and I think it is better to get an error and change your code to be consistent ('using strings'). 

The config.js use case is handled(in this PR) before the value is passed to the BrowserDetection class. From what I see in the config this one and `desktopSharingChromeMinExtVersion` are the only versions that we specify in config for now. And we already have the inconsistency - `desktopSharingChromeMinExtVersion: '0.1',`.

bgrozev:
> It is harder to explain and confusing what you should specify for version and when (obviously if you want to specify 64.0.3282.119 you should use string).

There's nothing to explain, it's al intuitive. If you use '63' it works, if you use 63 it also works, and there's nothing surprising with the semantics. If we prohibit numbers, that's when we have to exlpain something non obvious.

> Not many benefits

That's what we're discussing, isn't it?

> Encouraging the developers to write inconsistent code. For example if the half of the calls are executed with number and the other half with string.

As I said the other day, I think this is a valid argument. But it is far outweighed by the others.

> It is confusing.

I disagree.



> It's up to us to pass the correct types to the functions. And if somebody uses it with number, we'll see the error and will change the code to use string.

It is up to us to define what is correct. Using it with numbers has a natural interpretation and failing is unexpected.

> In order to be safe we should handle all types, for example objects too.

Why are we discussing objects? No one is suggesting to accept objects. Objects do not naturally translate to a version string, numbers do.

bbaldino:
@bgrozev @saghul do you imagine that the method will convert both numbers and strings into one type (either number or string) as the first thing it does, and then use that type from then on?  or how deep should the handling for both strings and numbers go?

saghul:
@bbaldino Since there are different entry points I think the browser-caps API is the place.

bbaldino:
i think i agree with saul and boris here, let's just handle string & int in one central place and convert ints to strings before we make the bowser call.  from what i can tell, this wouldn't result in a lot of code that would have to deal with both numbers and strings (which was a concern of mine about taking in both), but just one place where we do something like `version = version + ''`.  let me know if any of my assumptions are wrong.

bgrozev:
What is going on with this? Is the bug still present?

hristoterezov:
@bgrozev Since the PR is not merged the bug is still present. I haven't got the time to work on this one.

@bbaldino I'm still thinking that allowing number is a bad idea but since everybody else think that this the correct way to go I'll change the code.

damencho:
Offline we decided that this will be fixed as @hristoterezov already mentioned in the previous comment. For now, I will merge this as this is blocking jitsi-meet stable release.

PR #697, opened 2018-01-26
Adds new parameter details for the connection failed event.

damencho:
Adds suspend and shard_changed details.

### DIFF
--- a/JitsiConnection.js
+++ b/JitsiConnection.js
@@ -23,11 +23,13 @@ export default function JitsiConnection(appID, token, options) {
     this.options = options;
     this.xmpp = new XMPP(options, token);
 
+    /* eslint-disable max-params */
     this.addEventListener(JitsiConnectionEvents.CONNECTION_FAILED,
-        (errType, msg) => {
+        (errType, msg, credentials, details) => {
             Statistics.sendAnalyticsAndLog(
-                createConnectionFailedEvent(errType, msg));
+                createConnectionFailedEvent(errType, msg, details));
         });
+    /* eslint-enable max-params */
 
     this.addEventListener(JitsiConnectionEvents.CONNECTION_DISCONNECTED,
         msg => {
--- a/modules/xmpp/strophe.ping.js
+++ b/modules/xmpp/strophe.ping.js
@@ -24,6 +24,19 @@ const PING_TIMEOUT = 15000;
 const PING_THRESHOLD = 3;
 
 /**
+ * The number of timestamps of send pings to keep.
+ * @type {number} number of timestamps.
+ */
+const PING_TIMESTAMPS_TO_KEEP = 12;
+
+/**
+ * If execution of pings is higher than this threshold, we consider that the
+ * ping was suspended.
+ * @type {number}
+ */
+const SUSPEND_INTERVAL_THRESHOLD = PING_TIMEOUT;
+
+/**
  * XEP-0199 ping plugin.
  *
  * Registers "urn:xmpp:ping" namespace under Strophe.NS.PING.
@@ -38,6 +51,7 @@ class PingConnectionPlugin extends ConnectionPlugin {
         super();
         this.failedPings = 0;
         this.xmpp = xmpp;
+        this.pingExecIntervals = new Array(PING_TIMESTAMPS_TO_KEEP);
     }
 
     /**
@@ -60,6 +74,8 @@ class PingConnectionPlugin extends ConnectionPlugin {
      * timeout <tt>error<//t> callback is called with undefined error argument.
      */
     ping(jid, success, error, timeout) {
+        this._addPingExecutionTimestamp();
+
         const iq = $iq({
             type: 'get',
             to: jid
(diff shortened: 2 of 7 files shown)


modules/xmpp/xmpp.js
```diff
@@ -509,4 +512,45 @@ export default class XMPP extends Listenable {
...
+                const header = parts.shift();
+                const value = parts.join(': ');
+
+                headers[header] = value;
+            });
+
+            /* eslint-disable camelcase */
+            details.shard_changed
+                = this.options.deploymentInfo.shard
+                    !== headers['x-jitsi-shard'];
```
bgrozev: I see "X-Jitsi-Shard" coming from meet.jit.si, and [HTTP headers seem to be case-insensitive](https://tools.ietf.org/html/rfc7230#section-3.2). Is there something in strophe that converts them to all lowercase?

modules/xmpp/strophe.ping.js
```diff
@@ -141,6 +157,52 @@ class PingConnectionPlugin extends ConnectionPlugin {
...
+     * of time.
+     */
+    wasPingSuspended() {
+        const pingIntervals = this.pingExecIntervals.slice();
+
+        // we need current time, as if ping was sent now
+        // if computer sleeps we will get correct interval after next
+        // scheduled ping, bet we sometimes need that interval before waiting
+        // for the next ping, on closing the connection on error.
+        pingIntervals.push(new Date().getTime());
```
bgrozev: Do you intentionally not use _addPingExecutionTimestamp here? It looks like calls to wasPingSuspended can make the array grow (and _addPingExecutionTimestamp will not shrink it back, as it only removes 1 element).

modules/xmpp/xmpp.js
```diff
@@ -509,4 +512,45 @@ export default class XMPP extends Listenable {
...
+            /* eslint-enable camelcase */
+        }
+
+        // check for possible suspend
```
bgrozev: I thought we also wanted the suspend duration somewhere

modules/xmpp/strophe.ping.js
```diff
@@ -23,6 +23,19 @@ const PING_TIMEOUT = 15000;
...
+ * @type {number} number of timestamps.
+ */
+const PING_TIMESTAMPS_TO_KEEP = 12;
+
+/**
+ * If execution of pings is higher than this threshold, we consider that the
+ * ping was suspended.
+ * @type {number}
+ */
+const SUSPEND_INTERVAL_THRESHOLD = PING_TIMEOUT;
```
bgrozev: What is the relationship between SUSPEND_INTERVAL_THRESHOLD and PING_TIMEOUT? 

Wouldn't we want SUSPEND_INTERVAL_THRESHOLD to depend on PING_INTERVAL instead (if PING_INTERVAL is 1000, we could detect suspension sooner; if it is 100_000 we would have false positive detections)?

JitsiConnection.js
```diff
@@ -23,11 +23,13 @@ export default function JitsiConnection(appID, token, options) {
     this.options = options;
     this.xmpp = new XMPP(options, token);
 
+    /* eslint-disable max-params */
     this.addEventListener(JitsiConnectionEvents.CONNECTION_FAILED,
-        (errType, msg) => {
+        (errType, msg, credentials, details) => {
```
bgrozev: We explicitly disable the max-params rule in order to add an extra parameter which is always "undefined" and always ignored :)

bgrozev: Ah, I see now that it is used elsewhere.

modules/xmpp/xmpp.js
damencho: Hum, I suppose the XHR API works that way. All strophe does is: req.xhr.getAllResponseHeaders();
https://github.com/strophe/strophejs/commit/93d63d3f7adc1ea126e6ed57f1b5fabf6a34c765#diff-c8241e10984282f97b0a3814d0bb41e7R619
And I've been testing this with meet.jit.si.

modules/xmpp/strophe.ping.js
damencho: Well my thinking was working on a copy of the array and compute it as if a ping was just sent it, does it make sense. It is a copy of the array, so nothing grows.

modules/xmpp/xmpp.js
damencho: Ok, I will modify it and add that.

modules/xmpp/strophe.ping.js
damencho: Yep, I was thinking to make them dependent ... but if we are always sending the interval maybe I will remove this threshold and we detect this on the analytics side of things and move the threshold there.

damencho:
@bgrozev I pushed the changes and tests finally passed. Does it look ok now?

modules/xmpp/xmpp.js
```diff
@@ -509,4 +512,45 @@ export default class XMPP extends Listenable {
...
+        details.suspendTime = this.connection.ping.getPingSuspendTime();
```
bgrozev: Please use "suspend_time" if this goes to analytics

PR #698, opened 2018-01-27
Fix chromium

hristoterezov:


modules/browser/BrowserDetection.js
```diff
@@ -123,6 +124,8 @@ function _detect() {
     const { name, version } = bowser;
 
     if (name in bowserNameToJitsiName) {
+        logger.info(`This appears to be ${name}, ver: ${version}`);
```
virtuacoplenny: I actually really appreciate this logging getting back added. I hadn't realized how much I had gotten used to seeing it.

hristoterezov: Me too :)

PR #699, opened 2018-01-29
fix(browser-caps): fix .find for ie11 and stop saying temasys is supported

virtuacoplenny:


modules/browser/BrowserCapabilities.js
```diff
@@ -24,8 +24,17 @@ export default class BrowserCapabilities extends BrowserDetection {
         super(browserInfo);
 
         const browserCapabilities = capabilitiesDB[this.getName()] || [];
-        const capabilitiesByVersion = browserCapabilities.find(({ version }) =>
-            !version || !this.isVersionGreaterThan(version));
+        let capabilitiesByVersion;
+
+        for (let i = 0; i < browserCapabilities.length; i++) {
+            const capabilities = browserCapabilities[i];
+            const version = capabilities && capabilities.version;
```
hristoterezov: Do we need this check?

If you want to be safe isn't it going to be better to do:
```
if(typeof capabilities !== 'object') {
    continue;
}
```

paweldomas: why was it better to go with continue and this check ? @virtuacoplenny @hristoterezov just by reading the code it's hard to get why we went with this check

virtuacoplenny: lib-jitsi-meet does not get polyfilled for es6 methods. Using Array.prototype.find will cause breakages in older browsers, like ie11. The quick fix for now is changing the find call to be a for-loop. The long term fix will be discussed soon. The issue was blocking people today that were assuming browser-capabilities was working.

PR #701, opened 2018-01-30
feat(screenshare): allow using a camera as a screenshare source

virtuacoplenny:


modules/RTC/RTCUtils.js
```diff
@@ -1441,6 +1441,26 @@ class RTCUtils extends Listenable {
...
+                                }
+                            }),
+                            error => reject(error),
+                            {
+                                cameraDeviceId:
+                                    options.screenSharingSourceDeviceId
+                            }
+                        );
+
+                        return;
```
virtuacoplenny: Note that the above chunk is for the older gum flow. It does not have guards against exact device id being used because I was unsure of the ideal way to proceed.

bbaldino: man this method sucks, i had to expand it like 3 or 4 times to figure out where `hasDesktop` came from.  i think i need a little help with what the context is when we enter this block (i.e. what things are true/not true)

virtuacoplenny: Sure. This flow is entered through JitsiMeetJS.createLocalTracks. Eventually it gets to here in RTCUtils.obtainAudioAndVideoPermissions to do the actual getUserMediaCall. obtainAudioAndVideoPermissions looks something like this:

```
if (firefox) {
    // special case handling
} else {
    const hasDesktop = if 'desktop' requested in user media array
    if (hasDesktop) {
        // remove 'desktop' from requested user media array
    }
    if (requested audio or video) {
        // do gum
        // some error handling
        // if hasDesktop try to get the desktop stream
    } else if hasDesktop {
        //  try to get the desktop stream
    }
}
```

virtuacoplenny: I didn't handle the request for desktop in the callback for audio/video. Maybe I should? I was hoping that bit of code is not exercised by jitsi-meet but maybe it's something I should cover anyway.

bbaldino: thanks. is `options.screenSharingSourceDeviceId` something entirely new?

modules/RTC/RTCUtils.js
```diff
@@ -1441,6 +1441,26 @@ class RTCUtils extends Listenable {
...
+                    }
+
                     screenObtainer.obtainStream(
```
bbaldino: having an else here (rather than relying on the early return) i think would be more readable/clearer (even if it is redundant).

modules/RTC/RTCUtils.js
```diff
@@ -1482,11 +1502,43 @@ class RTCUtils extends Listenable {
...
             const isDesktopDeviceRequsted = umDevices.indexOf('desktop') !== -1;
 
-            return isDesktopDeviceRequsted
-                ? this._newGetDesktopMedia(
-                    options.desktopSharingExtensionExternalInstallation,
-                    options.desktopSharingSources)
-                : Promise.resolve();
+            let getDesktopPromise;
+
+            if (isDesktopDeviceRequsted) {
```
bbaldino: and what's the scenario we get here instead of the branch above?  why does the code not look quite the same as the `options.screenSharingSourceDeviceId` branch above?

modules/RTC/RTCUtils.js
virtuacoplenny: Yeah, screenSharingSourceDeviceId is the new thing that signals to the lib to use a camera for the desktop sharing input instead of going through normal desktop sharing.

modules/RTC/RTCUtils.js
virtuacoplenny: Do you mean the difference between "old gum flow" and "new gum flow"? The method here is for new gum flow. The biggest difference between the two is I didn't add the "constraints.video.deviceId.exact" constraints above in the old gum flow because I wasn't confident of my implementation here and if I should just copy/paste (or move out) the logic so old gum flow also uses constraints.video.deviceId.exact; the biggest issue there being the old getConstraints has to be modified to handle having deviceId.exact.

PR #707, opened 2018-02-05
ref(gum): try to reduce complexity of obtainAudioAndVideoPermissions

virtuacoplenny:
Even though I walk through the valley of the shadow of death
I will fear no evil... Reduce indenting and repeated calls to
getting desktop streams by creating a promise chain.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -46,6 +46,36 @@ const eventEmitter = new EventEmitter();
 
 const AVAILABLE_DEVICES_POLL_INTERVAL_TIME = 3000; // ms
 
+/**
+ * Default resolution to obtain for video tracks if no resolution is specified.
+ * This default is used for old gum flow only, as new gum flow uses
+ * {@link DEFAULT_CONSTRAINTS}.
+ */
+const OLD_GUM_DEFAULT_RESOLUTION = 720;
+
+/**
+ * Default devices to obtain when no specific devices are specified. This
+ * default is used for old gum flow only.
+ */
+const OLD_GUM_DEFAULT_DEVICES = [ 'audio', 'video' ];
+
+/**
+ * Default MediaStreamConstraints to use for calls to getUserMedia.
+ *
+ * @private
+ */
+const DEFAULT_CONSTRAINTS = {
+    video: {
+        aspectRatio: 16 / 9,
+        height: {
+            ideal: 1080,
+            max: 1080,
+            min: 240
+        }
+    }
+};
+
+
 // TODO (brian): Move this devices hash, maybe to a model, so RTCUtils remains
 // stateless.
 const devices = {
@@ -352,22 +382,6 @@ function getConstraints(um, options) {
 
     return constraints;
 }
-
-/**
- * Default MediaStreamConstraints to use for calls to getUserMedia.
- *
- * @private
- */
-const DEFAULT_CONSTRAINTS = {
-    video: {
-        aspectRatio: 16 / 9,
-        height: {
-            ideal: 1080,
-            max: 1080,
-            min: 240
-        }
-    }
-};
 
 /**
  * Creates a constraints object to be passed into a call to getUserMedia.
@@ -1287,13 +1301,13 @@ class RTCUtils extends Listenable {
      * @returns {*} Promise object that will receive the new JitsiTracks
      */
     obtainAudioAndVideoPermissions(options = {}) {
-        const dsOptions = {
+        const desktopSharingOptions = {
             ...options.desktopSharingExtensionExternalInstallation,
             desktopSharingSources: options.desktopSharingSources
         };
 
-        options.devices = options.devices || [ 'audio', 'video' ];
-        options.resolution = options.resolution || '720';
+        options.devices = options.devices || [ ...OLD_GUM_DEFAULT_DEVICES ];
+        options.resolution = options.resolution || OLD_GUM_DEFAULT_RESOLUTION;
 
         const requestingDesktop = options.devices.includes('desktop');
 
@@ -1353,7 +1367,7 @@ class RTCUtils extends Listenable {
 
                     return new Promise((resolve, reject) => {
                         screenObtainer.obtainStream(
-                            dsOptions,
+                            desktopSharingOptions,
                             desktop => resolve({
                                 audioVideo,
                                 desktop
@@ -1375,7 +1389,8 @@ class RTCUtils extends Listenable {
                     video: (...args) =>
                         this.getUserMediaWithConstraints([ 'video' ], ...args),
                     desktop: (...args) =>
-                        screenObtainer.obtainStream(dsOptions, ...args)
+                        screenObtainer.obtainStream(
+                            desktopSharingOptions, ...args)
                 };
 
                 obtainDevices({
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -1338,13 +1338,13 @@ class RTCUtils extends Listenable {
 
             gumPromise = promiseWrappedGum(options.devices, options)
                 .then(avStream => {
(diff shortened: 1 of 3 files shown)


virtuacoplenny:
This is prerequisite work for https://github.com/jitsi/lib-jitsi-meet/pull/701.

virtuacoplenny:
Also, I have no unit tests to tell me if I broke something. A manual sanity test was performed with chrome going through the old gum flow. If it is preferred, I can try writing unit tests and see if they still pass before and after the changes. I avoided doing so only because the old gum flow...should be temporary.

modules/RTC/RTCUtils.js
```diff
@@ -1287,167 +1287,147 @@ class RTCUtils extends Listenable {
...
+        let gumPromise;
+
+        // XXX The react-native-webrtc at the time of this writing does not
+        // support MediaStream constructors as defined by
+        // https://www.w3.org/TR/mediacapture-streams/#constructors. Instead it
+        // has a single constructor which expects an NSNumber as a MediaStream
+        // ID. Temasys does not support the constructor at all. So we get audio
+        // and video in two distinct GUM calls and pass them back into a single
+        // object.
+        if (browser.isReactNative() || browser.isTemasysPluginUsed()) {
```
paweldomas: can we create a method in "browser" with a meaningful name for this check and put the explanation there ?

modules/RTC/RTCUtils.js
```diff
@@ -1287,167 +1287,147 @@ class RTCUtils extends Listenable {
...
+                        this.getUserMediaWithConstraints(
+                            requestedDevices,
+                            resolve,
+                            reject,
+                            gumOptions);
+                    } else {
+                        resolve(null);
+                    }
+                });
+            }.bind(this);
```
paweldomas: will arrow function work here to not do .bind(this) ? it used to be preferred way

modules/RTC/RTCUtils.js
paweldomas: how about FF ? I may have misread something, but it used to be included in this if. Is it supported now or ?

virtuacoplenny: FF should go through new gum flow only now so the FF shouldn't be needed.

paweldomas: maybe we could check for FF version and return true only when new flow is supported ? that logic could be in the method mentioned above if you decide to add it.

virtuacoplenny: Okay, I tried adding the check. I started getting confused on the boolean flippings so I added an intermediary variable for myself. I'm not sure how useful adding Firefox would be only because all versions of firefox fall under the new gum flow now. But I added it because it's very little work to add it and we can see if it feels good to keep in.

virtuacoplenny:
Force pushed after rebasing onto lastest master. lib-jitsi-meet-PR-test is failing on npm install and maybe that'll fix it...

modules/RTC/RTCUtils.js
```diff
@@ -1287,167 +1287,142 @@ class RTCUtils extends Listenable {
      * @returns {*} Promise object that will receive the new JitsiTracks
      */
     obtainAudioAndVideoPermissions(options = {}) {
-        const self = this;
-
         const dsOptions = {
```
bbaldino: `ds` = `desktopSharing`? i think it'd be clearer if we changed this to be `desktopSharingOptions`

modules/RTC/RTCUtils.js
```diff
@@ -1287,167 +1287,142 @@ class RTCUtils extends Listenable {
...
             ...options.desktopSharingExtensionExternalInstallation,
             desktopSharingSources: options.desktopSharingSources
         };
 
-        return new Promise((resolve, reject) => {
-            const successCallback = function(stream) {
-                resolve(handleLocalStream(stream, options.resolution));
-            };
+        options.devices = options.devices || [ 'audio', 'video' ];
+        options.resolution = options.resolution || '720';
```
bbaldino: maybe work for a subsequent commit, but these defaults feel like they should be coming from config or something.  or at least constants?

modules/RTC/RTCUtils.js
```diff
@@ -1287,167 +1287,142 @@ class RTCUtils extends Listenable {
...
+                new Error('Desktop sharing is not supported!'));
+        }
+
+        let gumPromise;
+
+        if (browser.supportsMediaStreamConstructor()) {
+            options.devices = options.devices.filter(device =>
+                device !== 'desktop');
+
+            const promiseWrappedGum = (requestedDevices, gumOptions) =>
```
bbaldino: what about a version of `getUserMediaWithConstraints` that returns a promise and this logic being there?

modules/RTC/RTCUtils.js
```diff
@@ -1287,167 +1287,142 @@ class RTCUtils extends Listenable {
...
-                    this.getUserMediaWithConstraints(device, s, e, options);
-                };
+        if (requestingDesktop && !screenObtainer.isSupported()) {
+            return Promise.reject(
+                new Error('Desktop sharing is not supported!'));
+        }
+
+        let gumPromise;
+
+        if (browser.supportsMediaStreamConstructor()) {
```
bbaldino: i think this name is a bit confusing.  if it wasn't for the comment on the `else` block below i wouldn't really get what was being checked for here.  i think at least the explaining comment should be here at the top, but wonder too if we could come up with a name that conveys what differentiates the two paths better.  maybe the `supportsMediaStreamConstructor` stays in the browser API, but we have another helper layer on top of that, something like "shouldGetAudioAndVideoStreamsSeparately()" which did this check inside of it?

bbaldino: also, the if/else code is so long here...i wonder if it's worth moving the logic to functions to keep this one more readable.

bbaldino:
```
let gumPromise;
if (shouldGetAudioAndVideoStreamsSeparately()) {
  gumPromise = getAudioAndVideoStreamsSeparately()
} else {
  gumPromise = getAudioAndVideoStreams()
}
```
?

modules/RTC/RTCUtils.js
```diff
@@ -1287,167 +1287,142 @@ class RTCUtils extends Listenable {
...
+                            gumOptions);
+                    } else {
+                        resolve(null);
+                    }
+                });
+
+            gumPromise = promiseWrappedGum(options.devices, options)
+                .then(avStream => {
+                    const missingTracks
+                        = this._getMissingTracks(options.devices, avStream);
```
bbaldino: do we know that the scenario that required this error handling still exists?  i wonder if it's something we could get rid of.

modules/RTC/RTCUtils.js
paweldomas: hmm haven't we decided to not have any default options defined in the config, but have default handled by the code ?

modules/RTC/RTCUtils.js
paweldomas: The scenario is that we're trying to obtain audio and video stream with one GUM request, but only video will fail, because user granted mic, but not cam (and vice versa). So I'd say it's quite common and still needs to be handled.

modules/RTC/RTCUtils.js
bbaldino: ok so not config but a constant then

modules/RTC/RTCUtils.js
bbaldino: ok, i wasn't thinking so much that the case didn't exist, but that the need for our special error-handling (as opposed to the built in error) was no longer needed, but maybe that isn't the case.

virtuacoplenny:
I'm asking that this be blocked until I write new unit tests for this.

virtuacoplenny:
Force pushing to rebase and get some build related fixes.

PR #708, opened 2018-02-05
fix(connection-quality): set min percentage when exceeding a bitrate threshold

virtuacoplenny:
In some cases a high bitrate does not equate to a high quality
percentage, because the target itself might be quite high, such
as the case for 1080. In those cases, give credit for the bitrate
being high.

### DIFF
--- a/modules/connectivity/ConnectionQuality.js
+++ b/modules/connectivity/ConnectionQuality.js
@@ -58,6 +58,13 @@ const kSimulcastFormats = [
 ];
 
 /**
+ * The maximum bitrate to use as a measurement against the participant's current
+ * bitrate. This cap helps in the cases where the participant's bitrate is high
+ * but not enough to fulfill high targets, such as with 1080p.
+ */
+const MAX_TARGET_BITRATE = 2500;
+
+/**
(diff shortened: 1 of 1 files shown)


virtuacoplenny:
I went with the first recommendation for now to make sure the idea of adjusting the connection quality does not get lost and so others can chime in. (To be fair to the recommender, I'm not sure I implemented the recommendation correctly.)

There might be a couple issues with the GSM bars, but the target issue for this PR is the scenario with 1080, where the participant is sending up a high bitrate but the bitrate is not enough to meet the target bitrate for 1080, which is 4000kpbs in an enum in ConnectionQuality.js but is adjusted to 3600kps in ConnectionQuality#_calculateConnectionQuality.

virtuacoplenny:
After a quick chat yesterday with Boris I've tried another implementation that essentially sets a max bitrate to compare the current bitrate to. I've at least used a constant variable this time around.

virtuacoplenny:
Jenkins test this please.

virtuacoplenny:
rebased onto latest master to maybe fix the failing tests.

PR #709, opened 2018-02-06
ref(browser): Use js-utils.

hristoterezov:


modules/browser/BrowserCapabilities.js
```diff
@@ -1,7 +1,9 @@
-import BrowserDetection from './BrowserDetection';
-import capabilitiesDB from './capabilities.json';
+import { getLogger } from 'jitsi-meet-logger';
+import { BrowserDetection } from 'js-utils';
 
-// TODO: All methods should use capabilities.json when possible.
+const logger = getLogger(__filename);
+
+// TODO: Move this code to js-utils.
```
virtuacoplenny: What is "this code"? The entire file?

modules/browser/BrowserCapabilities.js
```diff
@@ -1,7 +1,9 @@
-import BrowserDetection from './BrowserDetection';
-import capabilitiesDB from './capabilities.json';
+import { getLogger } from 'jitsi-meet-logger';
+import { BrowserDetection } from 'js-utils';
```
virtuacoplenny: An aside: why was the package called js-utils? "Utils" is a fairly broad term. What "utils" are you expecting to go there and what should not go in there?

modules/browser/BrowserCapabilities.js
```diff
@@ -16,49 +18,11 @@ export default class BrowserCapabilities extends BrowserDetection {
      *
      * @param {boolean} [isUsingIFrame] - True if Jitsi Meet is loaded in iframe
```
virtuacoplenny: Update comments? Were these constructor arguments moved into js-utils BrowserDetection?

modules/browser/BrowserCapabilities.js
hristoterezov: Yes!

modules/browser/BrowserCapabilities.js
hristoterezov: I expect everything that need to be shared between our projects to go there. I don't mind changing it if you have better ideas, or even you can go and change it. But I don't think this is related to this PR.

modules/browser/BrowserCapabilities.js
hristoterezov: Yes! Will fix it!

modules/browser/BrowserCapabilities.js
virtuacoplenny: Why not do it now?

hristoterezov: By moving I mean moving also starting using the capabilities.json approach. It won't be 1h work and It is pretty serious change that should be well tested.

modules/browser/BrowserCapabilities.js
saghul: I don't want to open the can of worms, but I wholeheartedly agree with @virtuacoplenny here.

bgrozev: I would like to see some of the basic utilities from lib-jitsi-meet moved to the utils package in the future (some of them are already being used in e.g. jidesha by being copy/pasted). So I'm fine with the name being broad (although I'd personally go with "jitsi-meet-utils"). In any case, I don't think we should block this PR on the name of an existing package.

hristoterezov: @bgrozev @saghul I don't mind changing the name and I don't mind if somebody else change it. But I don't see how the name or the package is related to the change. We were using `BrowserDetection` from lib-jitsi-meet and now we are using the same class and code but from package.....

virtuacoplenny: I don't think anyone is saying to change the name now or as part of this change. I was asking as a side question why it has the name, hence why I used the word "aside" while asking. I've approved the changes. I'm hesitant to merge until the tests are passing again, which they're being worked on actively.

paweldomas: why was this name not consulted with anyone or was it ? then it suddenly because too much of a change and it will stay like that forever now

paweldomas: eventually is not so bad if not only "jitsi-meet" project will be using it

hristoterezov: @paweldomas I didn't know I'm supposed to discuss the names. It can be super easily changed(1h work maximum). I think there are 3 files only that are using it. I don't have any preferences for the name! So guys feel free to discuss it and to change it any time! 

Also @paweldomas let's move the discussion somewhere else if you have any other questions!

paweldomas: I am fine with that name, but it could have been discussed, before repo was created.

PR #711, opened 2018-02-13
fix(SS): Set min and max frame rate to 5

hristoterezov:


bgrozev:
Any reason not to export this to a config.js property (or separate properties for min and max) so we can experiment easily?

modules/RTC/RTCUtils.js
```diff
@@ -324,7 +334,8 @@ function getConstraints(um, options) {
                 chromeMediaSourceId: options.desktopStream,
                 maxWidth: window.screen.width,
                 maxHeight: window.screen.height,
-                maxFrameRate: 3
+                maxFrameRate: SS_FRAME_RATE,
+                minFrameRate: SS_FRAME_RATE
```
virtuacoplenny: There is currently an unfortunate world where there is new gum flow (adapterjs) and old gum flow (non adapterjs). Most browsers are in new gum flow but some still depend on old. Do your changes affect new gum flow for sure?

hristoterezov: Good question. Will check it.

hristoterezov:
> Any reason not to export this to a config.js property (or separate properties for min and max) so we can experiment easily?

Yes. I don't think we'll experiment too much with it. Are you expecting this to change per deployment? Do you expect to change it very frequently?

modules/RTC/RTCUtils.js
hristoterezov: Actually I tested it on Chrome and FF it appears to work. And now when I'm looking at the adapter we are using it as polyfill, so the new parameters are passed to it.

hristoterezov:
> > Yes.
> 
> Which is?

The next sentence - "I don't think we'll experiment too much with it." :)

> > I don't think we'll experiment too much with it.
> 
> So you know the magic number 5/5 to be optimal?

No I don't know if it is going to be optimal. But testing with different values can be done by changing the constant and building it. You probably won't just play with the values on production deployment anyway. IMHO we should add it to the config only if we expect different values for different deployments and I didn't think this is the case.

> > Are you expecting this to change per deployment?
> 
> Yes. I would personally like to experiment with 30/30 for sharing video, and I can imagine that some people might choose a higher value for their deployments. Knowing how screensharing behaves at different frame rates would be very useful. Ideally it should also go to the overrides whitelist.

Since you are sure that there we'll be deployments with different values for the frame rate constraint and also since if you are sure that we actually want to allow users to change it, I'll put it in the config.

modules/RTC/RTCUtils.js
hristoterezov: @virtuacoplenny I checked the code again. And as you suspected it seems that for SS we are using always the old flow!

I plan to change:
https://github.com/jitsi/lib-jitsi-meet/pull/711/files#diff-7d21489f2da42a0c08c868715b66dcddL454
Sorry I missed it :(

And then when we fix the bug it should just work :)

virtuacoplenny: Okay

modules/RTC/RTCUtils.js
```diff
@@ -1271,28 +1313,35 @@ class RTCUtils extends Listenable {
...
      * in RTCUtils#_newGetUserMediaWithConstraints.
      *
-     * @param {Object} desktopSharingExtensionExternalInstallation
-     * @param {string[]} desktopSharingSources
+     * @param {Object} options
+     * @param {Object} options.desktopSharingExtensionExternalInstallation
+     * @param {string[]} options.desktopSharingSources
+     * @param {Object} options.frameRate
+     * @param {Object} options.frameRate.min - Minimum fps
+     * @param {Object} options.frameRate.max - Maximum fps
```
bgrozev: I think that's options.gumOptions.frameRate.{min,max}, no?

hristoterezov: fixed!

bgrozev:
LGTM, but I haven't looked at all the details.

modules/RTC/RTCUtils.js
```diff
@@ -459,13 +490,21 @@ function newGetConstraints(um = [], options = {}) {
             constraints.video = {};
         }
 
+        const {
+            frameRate = {
+                min: SS_DEFAULT_FRAME_RATE,
+                max: SS_DEFAULT_FRAME_RATE
+            }
```
virtuacoplenny: What happens in the case where only max or min is not defined, but the frameRate option is defined?

hristoterezov: We won't set the undefined value. Do you think we should handle this? It is getting pretty complicated if we want to handle it. For example if the minimum is not set but the maximum is set to 3, we can't use the default value for the minimum because it is 5. I'm not sure adding such a complexity in the code is worth it and maybe the user don't want to set any constraint about the undefined value.

virtuacoplenny: I don't care much about handling it now but I'm wondering what the side effects are, if any.

hristoterezov: Oh. OK. Sorry I didn't understood from the first time :) I'll test to check.

modules/RTC/RTCUtils.js
```diff
@@ -460,14 +473,46 @@ function newGetConstraints(um = [], options = {}) {
...
+function getSSConstraints(options = {}) {
+    const { max, min }
+        = options.frameRate || {
+            min: SS_DEFAULT_FRAME_RATE,
+            max: SS_DEFAULT_FRAME_RATE
         };
+
+    const constraints = {
+        chromeMediaSource: options.source,
+        chromeMediaSourceId: options.desktopStream,
```
virtuacoplenny: If source is 'screen' then options.desktopStream will be undefined. I want to verify with you that chrome doesn't complain about it being undefined.

PR #715, opened 2018-02-27
Adjust name of the js library

antobinary:
After running `npm install` I end up with a minified library for lib-jitsi-meet. I had to adjust the name to get the example to run


antobinary:
@jitsi-jenkins Done

PR #717, opened 2018-03-07
Implements the promised based getStats. Enables them for Safari and FF.

damencho:
Adds stats and audio levels for Safari. Enables the new getStats API for Firefox, that will get rid of the following warning:
'non-maplike pc.getStats access is deprecated, and will be removed in the near future! See http://w3c.github.io/webrtc-pc/#getstats-example for usage.'

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -539,15 +539,17 @@ TraceablePeerConnection.prototype.getTrackBySSRC = function(ssrc) {
  */
 TraceablePeerConnection.prototype.getSsrcByTrackId = function(id) {
 
-    for (const localTrack of this.localTracks.values()) {
-        if (localTrack.getTrack().id === id) {
-            return this.getLocalSSRC(localTrack);
-        }
+    const findTrackById = track => track.getTrack().id === id;
+    const localTrack = this.getLocalTracks().find(findTrackById);
+
+    if (localTrack) {
+        return this.getLocalSSRC(localTrack);
     }
-    for (const remoteTrack of this.getRemoteTracks()) {
-        if (remoteTrack.getTrack().id === id) {
-            return remoteTrack.getSSRC();
-        }
+
+    const remoteTrack = this.getRemoteTracks().find(findTrackById);
+
+    if (remoteTrack) {
+        return remoteTrack.getSSRC();
     }
 
     return null;
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -532,6 +532,28 @@ TraceablePeerConnection.prototype.getTrackBySSRC = function(ssrc) {
 };
 
 /**
+ * Tries to find SSRC number for given {@link JitsiTrack} id. It will search
+ * both local and remote tracks bound to this instance.
+ * @param {string} id
+ * @return {number|null}
+ */
+TraceablePeerConnection.prototype.getSsrcByTrackId = function(id) {
+
+    for (const localTrack of this.localTracks.values()) {
+        if (localTrack.getTrack().id === id) {
+            return this.getLocalSSRC(localTrack);
+        }
+    }
+    for (const remoteTrack of this.getRemoteTracks()) {
+        if (remoteTrack.getTrack().id === id) {
+            return remoteTrack.getSSRC();
+        }
+    }
+
+    return null;
+};
+
+/**
  * Called when new remote MediaStream is added to the PeerConnection.
  * @param {MediaStream} stream the WebRTC MediaStream for remote participant
  */
@@ -2303,9 +2325,8 @@ TraceablePeerConnection.prototype.getStats = function(callback, errback) {
     // TODO (brian): After moving all browsers to adapter, check if adapter is
     // accounting for different getStats apis, making the browser-checking-if
     // unnecessary.
-    if (browser.isFirefox()
-            || browser.isTemasysPluginUsed()
-            || browser.isReactNative()) {
+    if (browser.isTemasysPluginUsed()
+        || browser.isReactNative()) {
         this.peerconnection.getStats(
             null,
             callback,
@@ -2314,10 +2335,15 @@ TraceablePeerConnection.prototype.getStats = function(callback, errback) {
                 // Making sure that getStats won't fail if error callback is
                 // not passed.
             }));
-    } else if (browser.isSafariWithWebrtc()) {
-        // FIXME: Safari's native stats implementation is not compatibile with
-        // existing stats processing logic. Skip implementing stats for now to
-        // at least get native webrtc Safari available for use.
+    } else if (browser.isSafariWithWebrtc() || browser.isFirefox()) {
+        // uses the new Promise based getStats
+        this.peerconnection.getStats()
+            .then(callback)
+            .catch(errback || (() => {
+
+                // Making sure that getStats won't fail if error callback is
+                // not passed.
+            }));
     } else {
         this.peerconnection.getStats(callback);
     }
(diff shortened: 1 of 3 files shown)


modules/statistics/RTPStatsCollector.js
```diff
@@ -940,3 +967,342 @@ StatsCollector.prototype.processAudioLevelReport = function() {
 };
 
 /* eslint-enable no-continue */
+
+/**
+ * New promised based getStats report processing.
+ * Tested with chrome, firefox and safari. Not switching it on for for chrome as
+ * frameRate stat is missing and calculating it using framesSent,
+ * gives values double the values seen in webrtc-internals.
```
gpolitis: My hat is off to you for catching this 🥇

modules/statistics/RTPStatsCollector.js
```diff
@@ -940,3 +967,342 @@ StatsCollector.prototype.processAudioLevelReport = function() {
...
+};
+
+/**
+ * Converts the value to a non-negative number.
+ * If the value is either invalid or negative then 0 will be returned.
+ * @param {*} v
+ * @return {number}
+ * @private
+ */
+StatsCollector.prototype.getNonNegativeValue = function(v) {
```
gpolitis: We use this method to get the ssrc, the bytes, the packets and the lost packets (negative because of retransmissions, maybe?). These look like unsigned positive ints, so I'm a bit confused why we need this (I'm not debating whether we need it or not, I'm just trying to understand why we need it). So, are we flipping the sign in an effort to work around some webrtc bug? If so, can we add a small comment documenting this? If there's a bug report on the webrtc side of things, we can include that too.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2314,10 +2335,15 @@ TraceablePeerConnection.prototype.getStats = function(callback, errback) {
...
-        // at least get native webrtc Safari available for use.
+    } else if (browser.isSafariWithWebrtc() || browser.isFirefox()) {
+        // uses the new Promise based getStats
+        this.peerconnection.getStats()
+            .then(callback)
+            .catch(errback || (() => {
+
+                // Making sure that getStats won't fail if error callback is
+                // not passed.
+            }));
```
virtuacoplenny: If this is promise based then maybe you should return this.peerconnection.getStats() so that in the future the callbacks can be removed. This would only be if nothing depends on TraceablePeerConnection#getStats not returning a promise.

modules/statistics/RTPStatsCollector.js
```diff
@@ -722,6 +742,16 @@ StatsCollector.prototype.processStatsReport = function() {
         }
     }
 
+    this.eventEmitter.emit(
```
virtuacoplenny: Why was this moved? It's close to where byteSentStats is done being calculated?

Honestly, I've always had a hard time understanding the 400 line block of multiple nested ifs that is processStatsReport. I'm glad you tried breaking this apart.

modules/statistics/RTPStatsCollector.js
```diff
@@ -940,3 +967,342 @@ StatsCollector.prototype.processAudioLevelReport = function() {
 };
 
 /* eslint-enable no-continue */
+
+/**
+ * New promised based getStats report processing.
+ * Tested with chrome, firefox and safari. Not switching it on for for chrome as
```
virtuacoplenny: *for for

modules/statistics/RTPStatsCollector.js
```diff
@@ -242,6 +250,7 @@ export default function StatsCollector(
      * @private
      */
     this._getStatValue = this._defineGetStatValueMethod(keys);
+    this._getNewStatValue = this._defineNewGetStatValueMethod(keys);
```
virtuacoplenny: Did you find it more clear to have separate instance variables for the get stats method instead of sharing one but conditionally selecting which _defineXXX got called?

modules/statistics/RTPStatsCollector.js
```diff
@@ -2,6 +2,7 @@ import browser from '../browser';
 import { browsers } from 'js-utils';
```
virtuacoplenny: Just had a thought. Tell me what you think: put the stats formatting into a separate module. That separates the collecting from the formatting. The separate module could then conditionally load the right formatter for rtpstatscollector to use (or the collector conditionally loads the right formatter).

modules/RTC/TraceablePeerConnection.js
damencho: Currently, RTPStatsCollector implements both the promise and non-promise getStats. So we always call it passing callbacks. 
We can avoid passing callbacks, with very little code duplications here:
https://github.com/jitsi/lib-jitsi-meet/blob/e3adbca60887eacd964b8336fc477202e270117f/modules/statistics/RTPStatsCollector.js#L314
Do you want me to change it? But then TraceablePeerConnection.getStats in one case will depend on the callbacks and in the other case will just return Promise:
 * @returns {Promise|void}, is this ok?

modules/statistics/RTPStatsCollector.js
damencho: Emitting BYTE_SENT_STATS depends on local variable and I left it in that method, while extracting: this._processAndEmitReport();, as it is used in two places. 
I can also move the emit BYTE_SENT_STATS inside _processAndEmitReport, by passing byteSentStats to the _processAndEmitReport method. Does this sound ok?

modules/statistics/RTPStatsCollector.js
damencho: will fix it.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -531,6 +531,28 @@ TraceablePeerConnection.prototype.getTrackBySSRC = function(ssrc) {
...
 
+/**
+ * Tries to find SSRC number for given {@link JitsiTrack} id. It will search
+ * both local and remote tracks bound to this instance.
+ * @param {string} id
+ * @return {number|null}
+ */
+TraceablePeerConnection.prototype.getSsrcByTrackId = function(id) {
+
+    for (const localTrack of this.localTracks.values()) {
```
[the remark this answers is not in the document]
damencho: Will change that.

modules/RTC/TraceablePeerConnection.js
virtuacoplenny: I guess we can wait if the promise isn't actually used yet. I was being too preemptive.

modules/statistics/RTPStatsCollector.js
damencho: Yep, makes sense. I was worried about spreading the if statements deciding which getStats implementation is used, but this thing can also be fixed. Will change that. Thanks.

modules/statistics/RTPStatsCollector.js
damencho: What do you mean by formatter? So currently what has been collected is been formatted in _processAndEmitReport which is common, no matter of the type of the stats...

modules/statistics/RTPStatsCollector.js
virtuacoplenny: No, it's okay where it is probably. I just couldn't see the intent of the move without some context.

modules/statistics/RTPStatsCollector.js
virtuacoplenny: I mean separating the collecting and the formatting logic. Conceptually they seem like two different concepts and the collecting logic doesn't necessarily need to know about formatting. That way the collector can use an arbitrary formatter, thereby hiding that there are two implementations for formatting.

virtuacoplenny: But I now get the feeling, since you mentioned instance variable somewhere else, that the collector and formatting may not have the clean separation already that could allow the collector to take and use an arbitrary formatter.

PR #718, opened 2018-03-08
Bumps uglifyjs-webpack-plugin and its dependencies.

damencho:


package.json
```diff
@@ -49,7 +49,7 @@
     "karma-webpack": "1.8.1",
     "precommit-hook": "3.0.0",
     "string-replace-loader": "1.3.0",
-    "uglifyjs-webpack-plugin": "1.1.2",
+    "uglifyjs-webpack-plugin": "1.2.2",
```
lyubomir: @damencho, we use uglifyjs-webpack-plugin in both lib-jitsi-meet and jitsi-meet and, because of the building of lib-jitsi-meet as part of `nom install` in jitsi-meet, I'd recommend keeping the common dependencies of the two projects including uglifyjs-webpack-plugin at the same version. Do you have a respective PR in jitsi-meet?

damencho: Just created one https://github.com/jitsi/jitsi-meet/pull/2586

PR #720, opened 2018-03-19
fix(edge): get p2p working up to ice failure

virtuacoplenny:
Edge does not support trickle ice. In an attempt to work around
such, and get p2p working, some other work was needed to even
get to the point of ice failure. This is that some other work.


modules/RTC/ortc/RTCPeerConnection.js
```diff
@@ -1937,15 +1972,43 @@ export default class ortcRTCPeerConnection extends yaeti.EventTarget {
         this._dtlsTransport = dtlsTransport;
     }
 
+    /**
+     * Pairs down STUN and TURN servers to only one. This works around an issue
```
bbaldino: `Pairs` -> `Pares`

PR #723, opened 2018-03-22
feat: Exposes sending JSON payload through XMPP.

bgrozev:
Includes only the payload when firing events for JSON transported
through XMPP.


JitsiConference.js
```diff
@@ -42,6 +42,7 @@ import VideoType from './service/RTC/VideoType';
 import VideoSIPGW from './modules/videosipgw/VideoSIPGW';
 import * as VideoSIPGWConstants from './modules/videosipgw/VideoSIPGWConstants';
 import * as XMPPEvents from './service/xmpp/XMPPEvents';
+import JITSI_MEET_MUC_TOPIC from './modules/xmpp/ChatRoom';
```
virtuacoplenny: Can you confirm for me that this imports the correct value? I would assume you needed to do `import { JITSI_MEET_MUC_TOPIC } from './modules/xmpp/ChatRoom'`.

bgrozev: Yes, fixed! I have no idea how this used to work...

JitsiConference.js
```diff
@@ -2054,20 +2055,38 @@ JitsiConference.prototype._fireIncompatibleVersionsEvent = function() {
...
+
+            message[JITSI_MEET_MUC_TOPIC] = '';
+
+            if (to === '') {
+                this.sendTextMessage(message);
+            } else {
+                this.sendPrivateTextMessage(to, message);
+            }
+        }
+    };
```
virtuacoplenny: I want your thoughts only because the change set still looks okay to me right now:
What about having new methods for sending a private message and a public message? My idea being that functions should try to do one thing only, and generally a boolean used to do different behaviors is a sign maybe multiple functions are warranted. With separate functions we know sendEndpointMessage always goes through the bridge data/websock channel.

bgrozev: Thanks for bringing this up. This was written for shipit, so I didn't really think about it at the time.

I am concerned about this interface getting too complex and inconsistent, but I don't know what the best way out is. There are 8 distinct cases, according to:
a. Whether the message is private or public
b. Whether we send it over the bridge channel or over the MUC
c. Whether we are passing a string or a JSON object (to be encoded in a special way)

IMO 8 or 4 separate methods is too much, so I'd like to bring it down to 2. Perhaps we can handle "c" according to the type of the argument. And between "a" and "b" I would choose to have different methods for private vs public, which leaves using an argument for "b".

One argument for using the same function to send over the bridge and over the MUC is that on the receive side we don't really care which channel the message came from, we just fire the same event in both cases. I think that's a good design, because it masks some of the complexity. Also, it's easy to chose a reasonable default (either one is reasonable, IMO).

So I would suggest:
```
sendMessage(message, sendThroughVideobridge = true)
sendPrivateMessage(message, sendThroughVideobridge = true)
```
Where `message` is either a string or an object and objects are encoded as needed.

We can then deprecate the rest of the methods. What do you think?

virtuacoplenny: What's the use case for setting `sendThroughVideobridge` to true versus false? I realized that I don't know for what reasons true or false would be set. Maybe that's why my desire was to make the branching point for functions on `sendThroughVideobridge`, and your explanation has revealed that ignorance to me. That said, splitting on public/private also makes logical sense to me, and maybe more sense from an api consumer standpoint, but I definitely don't have context to say anything more than "sounds reasonable."

bgrozev: Sending through the bridge will make the message available to the bridge but not jicofo (and not jigasi, today), sending through the MUC makes it available to jicofo and jigasi but not the bridge. It's very technical, which is why I'd like it to be somewhat hidden.

virtuacoplenny: Okay. I'm still not sure of why you need to send some something to the bridge but not jicofo/jigasi. Do you have an example, either real or hypothetical? Would there be a time when you need to send to all?

modules/xmpp/ChatRoom.js
```diff
@@ -102,13 +105,11 @@ function tryParseJSONAndVerify(jsonString) {
         // so we must check for that, too.
         // Thankfully, null is falsey, so this suffices:
         if (json && typeof json === 'object') {
-            const topic = json[VALID_FIELD_NAMES[0]];
-            const payload = json[VALID_FIELD_NAMES[1]];
+            const topic = json[JITSI_MEET_MUC_TOPIC];
+            const payload = json.payload;
```
virtuacoplenny: 👍

JitsiConference.js
```diff
@@ -2077,11 +2081,75 @@ JitsiConference.prototype.sendEndpointMessage = function(to, payload) {
...
+ * jitsi-videobridge (via the COLIBRI data channel or web socket), or through
+ * the XMPP MUC. Currently only objects can be sent through jitsi-videobridge.
+ */
+JitsiConference.prototype.sendMessage = function(
+        message,
+        to = '',
+        sendThroughVideobridge = false) {
+
+    if (sendThroughVideobridge) {
+        if (!(typeof message === 'object')) {
```
virtuacoplenny: I see this typeof check happening often. Maybe it'd be clearer to store the message type once.

JitsiConference.js
```diff
@@ -2077,11 +2081,75 @@ JitsiConference.prototype.sendEndpointMessage = function(to, payload) {
...
+                typeof message} through jitsi-videobridge.`);
+
+            return;
+        }
+        this.sendEndpointMessage(to, message);
+    } else {
+        // We only support strings (i.e. regular text messages) and objects (
+        // encapsulated in a specific JSON format)
+        if (typeof message !== 'string' && typeof message !== 'object') {
+            logger.error(`Can not send a message of type  ${typeof message}`);
```
virtuacoplenny: I wonder if it's worthwhile to have all type check error handling in one place, be it all at the top or as a default return at the bottom. Not sure how much of a clarity would it would be considering this function has split logic.

JitsiConference.js
```diff
@@ -2077,11 +2081,75 @@ JitsiConference.prototype.sendEndpointMessage = function(to, payload) {
...
+            try {
+                messageToSend = JSON.stringify(messageToSend);
+            } catch (e) {
+                logger.error('Can not send a message, stringify failed: ', e);
+
+                return;
+            }
+        }
+
+        if (to === '') {
```
virtuacoplenny: Only a musing: maybe it's better to do a truthy check for "to" and send the private message rather than checking for empty to. In my mind it's less complex to say "if it has a to" rather than "if to is empty". What do you think?

JitsiConference.js
```diff
@@ -2077,11 +2081,75 @@ JitsiConference.prototype.sendEndpointMessage = function(to, payload) {
...
+            // Broadcast
```
virtuacoplenny: This comment should probably be more explicit if it's needed at all. Saying "broadcast" my itself doesn't say much to someone like me because I could use the word (maybe incorrectly as) I broadcast/send my message to specific people.

JitsiConference.js
```diff
@@ -2077,11 +2081,75 @@ JitsiConference.prototype.sendEndpointMessage = function(to, payload) {
  * Sends a broadcast message via the data channel.
  * @param payload {object} the payload of the message.
  * @throws NetworkError or InvalidStateError or Error if the operation fails.
+ * @deprecated Use 'sendMessage' instead. TODO: this should be private.
  */
 JitsiConference.prototype.broadcastEndpointMessage = function(payload) {
     this.sendEndpointMessage('', payload);
```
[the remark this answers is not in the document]
bgrozev: No, they are actually the underlying implementation(s) being used by `sendMessage`. They just need to stop being exposed.

JitsiConference.js
```diff
@@ -2077,11 +2081,72 @@ JitsiConference.prototype.sendEndpointMessage = function(to, payload) {
...
+    if (sendThroughVideobridge) {
+        this.sendEndpointMessage(to, message);
+    } else {
+        let messageToSend = message;
+
+        if (messageType === 'object') {
+            // Encapsulate the object in the jitsi-meet format, and convert it
+            // to JSON.
+            messageToSend = { payload: message };
+            messageToSend[JITSI_MEET_MUC_TOPIC] = '';
```
virtuacoplenny: if you want to be fancy, you can do 
```
{
  payload: message,
  [JITSI_MEET_MUC_TOPIC]: ''
}
```

modules/xmpp/ChatRoom.js
```diff
@@ -102,13 +105,11 @@ function tryParseJSONAndVerify(jsonString) {
         // so we must check for that, too.
         // Thankfully, null is falsey, so this suffices:
         if (json && typeof json === 'object') {
-            const topic = json[VALID_FIELD_NAMES[0]];
-            const payload = json[VALID_FIELD_NAMES[1]];
+            const topic = json[JITSI_MEET_MUC_TOPIC];
```
bbaldino: why do we need this muc topic thing again?  i thought we were going to use a different element?

bgrozev: Eventually we will. But I'm not doing the change right now, because we need to send these from jigasi :(

bbaldino: ok...but why does sending them from jigasi affect it?

bgrozev: Because jigasi can only send a jitsi Message, and can not add an xml extension to it...

bbaldino: Ah, didn't realize it was sending it through the web.  Bummer.  So will there be a way we can get around that in the future?

bgrozev: As soon as we figure out (and implement) a way to make jigasi send the extension

PR #724, opened 2018-03-22
parse and emit identity information in MUC_MEMBER_JOINED event

bbaldino:


modules/xmpp/ChatRoom.js
```diff
@@ -430,6 +432,29 @@ export default class ChatRoom extends Listenable {
...
+                }
+            }
+            const groupInfo = node.children.find(c => c.tagName === 'group');
+
+            if (groupInfo) {
+                identity.group = groupInfo.value;
+            }
+
+            return identity;
+        };
```
virtuacoplenny: I'm worried that having this inline causes more confusion to an already fairly lengthy function. If this piece is re-usable, would it be possible to move this out on its own? Maybe some xml parsing utils somewhere?

modules/xmpp/ChatRoom.js
```diff
@@ -430,6 +432,29 @@ export default class ChatRoom extends Listenable {
...
+                for (const tag of [ 'id', 'name', 'avatar' ]) {
+                    const child
+                        = userInfo.children.find(c => c.tagName === tag);
+
+                    if (child) {
+                        identity.user[tag] = child.value;
+                    }
+                }
+            }
+            const groupInfo = node.children.find(c => c.tagName === 'group');
```
virtuacoplenny: If this is a node, would it be possible to use built in selectors to find the first child instead of iterating?

modules/xmpp/ChatRoom.spec.js
```diff
@@ -1,9 +1,13 @@
...
+import ChatRoom, { parser } from './ChatRoom';
 import { $pres } from 'strophe.js';
+import XMPPEvents from '../../service/xmpp/XMPPEvents';
 
 // This rule makes creating the xml elements take up way more
 // space than necessary.
 /* eslint-disable newline-per-chained-call */
+// These rules makes the xml strings harder to read
+/* eslint-disable operator-linebreak */
+/* eslint-disable max-len */
```
virtuacoplenny: nit: you can combine the ignored rules onto one line, `eslint-disable rule, rule, rule`.

bbaldino: fixed

modules/xmpp/ChatRoom.js
bbaldino: this isn't an actual `node`, `node` is just the term they used when transforming the xml to json.

PR #729, opened 2018-03-28
fix(chat): use querySelector over querySelectorAll

virtuacoplenny:
react-native does not support querySelectorAll. jitsi-meet's
react-native does have a polyfill for querySelector so use
that instead.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -418,7 +418,7 @@ export default class ChatRoom extends Listenable {
                 && this.options.hiddenDomain
                     === jid.substring(jid.indexOf('@') + 1, jid.indexOf('/'));
 
-        const xEl = pres.querySelectorAll('x')[0];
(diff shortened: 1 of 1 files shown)


modules/xmpp/ChatRoom.js
```diff
@@ -418,7 +418,7 @@ export default class ChatRoom extends Listenable {
                 && this.options.hiddenDomain
                     === jid.substring(jid.indexOf('@') + 1, jid.indexOf('/'));
 
-        const xEl = pres.querySelectorAll('x')[0];
+        const xEl = pres.querySelector('x');
```
bbaldino: so this just gets the first result?

virtuacoplenny: Yeah, it just gets the first result.

virtuacoplenny: Well one difference is what's returned. If nothing is found, querySelector will return null/falsy. querySelectorAll will return an empty array.

PR #730, opened 2018-03-30
feat: enableTcc, enableRemb, minParticipants.

gpolitis:


gpolitis:
@paweldomas could you PTAL?

modules/xmpp/moderator.js
```diff
@@ -188,6 +188,29 @@ Moderator.prototype.createConferenceIq = function() {
...
+                    name: 'enableRemb',
+                    value: Boolean(this.options.conference.enableRemb)
+                }).up();
+    }
+    if (this.options.conference.minParticipants !== undefined) {
+        elem.c(
+                'property', {
+                    name: 'minParticipants',
+                    value: this.options.conference.minParticipants
+                }).up();
```
virtuacoplenny: Are defaults set somewhere?

gpolitis: The defaults are currently set in Jicofo. I documented this in the jitsi-meet config.js file, as per Pawel's suggestion. Let me know if you believe they should live elsewhere.

PR #731, opened 2018-03-30
Switches camera id to mandatory when using old gum flow.

damencho:
When it fails we retry with different resolutions, and if that doesn't work we remove device id and let gum to decide which device to use.

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -38,8 +38,10 @@ const logger = Logger.getLogger(__filename);
 const USER_MEDIA_PERMISSION_PROMPT_TIMEOUT = 1000;
 
 /**
+ * Gets a lower resolution and if there is none left other than the requested
+ * returns null.
  *
- * @param resolution
+ * @param resolution the resolution to get.
  */
 function getLowerResolution(resolution) {
     if (!Resolutions[resolution]) {
@@ -57,6 +59,10 @@ function getLowerResolution(resolution) {
             res = value;
         }
     });
+
+    if (resName === resolution) {
+        resName = null;
+    }
 
     return resName;
 }
@@ -246,11 +252,14 @@ export default {
      *
      * @param {boolean} (firePermissionPromptIsShownEvent) - if event
      *      JitsiMediaDevicesEvents.PERMISSION_PROMPT_IS_SHOWN should be fired
+     * @param originalOptions - internal use only, to be able to store the
+     * originally requested options.
      * @returns {Promise.<{Array.<JitsiTrack>}, JitsiConferenceError>}
      *     A promise that returns an array of created JitsiTracks if resolved,
      *     or a JitsiConferenceError if rejected.
      */
-    createLocalTracks(options = {}, firePermissionPromptIsShownEvent) {
+    createLocalTracks(
+            options = {}, firePermissionPromptIsShownEvent, originalOptions) {
         let promiseFulfilled = false;
 
         if (firePermissionPromptIsShownEvent === true) {
@@ -336,7 +345,25 @@ export default {
                                 reason: 'unsupported resolution'
                             }));
 
-                        return this.createLocalTracks(options);
+                        return this.createLocalTracks(
+                            options,
+                            undefined,
+                            originalOptions || Object.assign({}, options));
+                    }
+
+                    // we tried everything, if there is a mandatory
+                    // device id, remove it and let gum find a device to
+                    // use
+                    if (originalOptions
+                        && error.gum.constraints
+                        && error.gum.constraints
+                        && error.gum.constraints.video
+                        && error.gum.constraints.video.mandatory
+                        && error.gum.constraints.video.mandatory.sourceId
+                            === options.cameraDeviceId) {
+                        originalOptions.cameraDeviceId = undefined;
+
+                        return this.createLocalTracks(originalOptions);
                     }
                 }
 
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -38,10 +38,12 @@ const logger = Logger.getLogger(__filename);
 const USER_MEDIA_PERMISSION_PROMPT_TIMEOUT = 1000;
 
 /**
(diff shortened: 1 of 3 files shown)


JitsiMeetJS.js
```diff
@@ -38,8 +38,10 @@ const logger = Logger.getLogger(__filename);
 const USER_MEDIA_PERMISSION_PROMPT_TIMEOUT = 1000;
 
 /**
+ * Gets a lower resolution and if there is none left other than the requested
+ * returns null.
  *
- * @param resolution
+ * @param resolution the resolution to get.
```
bbaldino: good idea to clean this up, maybe we can do:
```
Gets the next lowest desirable resolution to try for a camera.  If the given resolution 
is already the lowest acceptable resolution, returns null

@param resolution the current resolution
@return the next lowest resolution from the given one, or null if it is already 
the lowest acceptable resolution
```

JitsiMeetJS.js
```diff
@@ -336,7 +345,25 @@ export default {
...
+                            undefined,
+                            originalOptions || Object.assign({}, options));
+                    }
+
+                    // we tried everything, if there is a mandatory
+                    // device id, remove it and let gum find a device to
+                    // use
+                    if (originalOptions
+                        && error.gum.constraints
+                        && error.gum.constraints
```
bbaldino: duplicate check here

JitsiMeetJS.js
```diff
@@ -336,7 +345,25 @@ export default {
...
+                        && error.gum.constraints.video
+                        && error.gum.constraints.video.mandatory
+                        && error.gum.constraints.video.mandatory.sourceId
+                            === options.cameraDeviceId) {
```
bbaldino: do we actually need to check that it is equal to the one in options?  is there a scenario where it wouldn't be?

JitsiMeetJS.js
```diff
@@ -336,7 +345,25 @@ export default {
...
+                        originalOptions.cameraDeviceId = undefined;
+
+                        return this.createLocalTracks(originalOptions);
```
bbaldino: just double checking we don't need to re-pass `firePermissionPromptIsShownEvent`, right?

JitsiTrackError.js
```diff
@@ -99,7 +102,8 @@ function JitsiTrackError(error, options, devices) {
                         || constraintName === 'minHeight'
                         || constraintName === 'maxHeight'
                         || constraintName === 'width'
-                        || constraintName === 'height')) {
+                        || constraintName === 'height'
+                        || constraintName === 'deviceId')) {
```
bbaldino: will the error here actually have this as the constraintName?  won't it still be an error on width or height or something?

JitsiMeetJS.js
damencho: I think so, it should had shown something already.

JitsiTrackError.js
damencho: No, this is the case where you unplug the camera. And there is actually no such deviceId. 
I was explaining you offline, that in such case there is no way to detect whether this failure is because of audio or video.
This will result if anybody now unplugs a camera which is set as default in settings and reloads we will always retry all resolutions before fallback to another device if any.

JitsiMeetJS.js
```diff
@@ -246,11 +254,14 @@ export default {
...
      *      JitsiMediaDevicesEvents.PERMISSION_PROMPT_IS_SHOWN should be fired
+     * @param originalOptions - internal use only, to be able to store the
+     * originally requested options.
      * @returns {Promise.<{Array.<JitsiTrack>}, JitsiConferenceError>}
      *     A promise that returns an array of created JitsiTracks if resolved,
      *     or a JitsiConferenceError if rejected.
      */
-    createLocalTracks(options = {}, firePermissionPromptIsShownEvent) {
+    createLocalTracks(
+            options = {}, firePermissionPromptIsShownEvent, originalOptions) {
```
virtuacoplenny: Why is it necessary to have "originalOptions"? For example, are the mutations of options so deep that keeping a shallow copy of the original is insufficient?

modules/RTC/RTCUtils.js
```diff
@@ -243,9 +243,7 @@ function getConstraints(um, options = {}) {
             }
 
             // Old style.
```
virtuacoplenny: Does the "old style" comment still apply? Does new style also need a mandatory video?

JitsiMeetJS.js
damencho: Yep, `createLocalTracks ` is calling itself and so the original copy is lost.

modules/RTC/RTCUtils.js
damencho: I was not sure about the comment, what does it mean? 
I can remove it.

bbaldino: i do think this comment still applies, but it doesn't strike me as a very useful comment even if that is the case.

virtuacoplenny: I looked at the isNewStyleConstraintsSupported flow some more. In short, it may not need supporting for this change. This getConstriants function should be part of old gum flow. isNewStyleConstraintsSupported includes firefox, which shouldn't be using old gum (btw, camera switching doesn't work anymore on firefox...). isNewStyleConstraintsSupported includes edge which hasn't been able to device switch according to the ortc readme, and it includes temasys, which is pretty much going to be no longer supported. That leaves react native under isNewStyleConstraintsSupported, but I think it react native uses the option facingMode instead of cameraDeviceId.

damencho: Hum, switching camera for FF works for me, but it is sooo slow and you need to wait.

damencho: Yep performance overall, performance analyses show paint and graphics taking a lot of time.

virtuacoplenny: For my reference: Off Main Thread Paint (OMTP) was introduced to Mac FireFox with 59. I just checked ESR (52) and it's nowhere near as sluggish as 59.

PR #736, opened 2018-04-18
feat(recording): show the YouTube live stream URL

virtuacoplenny:
- Pass you_tube_broadcast_id to Jibri so it can create a
  YouTube link for the live stream.
- Emit the liveStreamViewURL (the YouTube link) when it
  is received from the participant doing the recording.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -425,8 +425,9 @@ export default class ChatRoom extends Listenable {
             const liveStreamViewURLItem
                 = pres.getElementsByTagName('live-stream-view-url')[0];
 
-            member.liveStreamViewURL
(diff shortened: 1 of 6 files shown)


bbaldino:
DON'T MERGE THIS YET

modules/xmpp/ChatRoom.js
```diff
@@ -812,6 +839,13 @@ export default class ChatRoom extends Listenable {
         const membersKeys = Object.keys(this.members);
 
         if (!isSelfPresence) {
+            if (this.members[from].liveStreamViewURL) {
```
virtuacoplenny: @bbaldino, do you think I should be defensive here instead of assuming the member exists?

PR #741, opened 2018-05-10
feat(recording): frontend logic can support live streaming and recording

virtuacoplenny:
Instead of either live streaming or recording, now both can live together.
The changes to facilitate such include the following:
- Remove the old recording.js module which allowed one recording at a time.
- Remove events for live stream url changes as the url is now part of a
  sesssion and not fired independently.
- Availability of sipgw and recording are gone. Instead sessions have a
  failure reason. For sipgw sessions, store that failure and emit it to
  listeners.
- Create a new recordingManager singleton that can start/stop sessions
  and handle updating known state of those sessions. Known state is
  emitted through one event.
- Create a JibriSession model to encapsulate state of a session.


virtuacoplenny:
Jenkins test this please.

modules/recording/JibriSession.js
```diff
@@ -0,0 +1,212 @@
...
+        this._liveStreamViewURL = url;
+    }
+
+    /**
+     * Sets the last known status for this recording session.
+     *
+     * @param {string} status - The new status to set.
+     * @returns {void}
+     */
+    setStatus(status = null) {
```
bbaldino: inconsistency between undefined and null usage?

PR #752, opened 2018-06-05
Leave jvb connection only with turns server.

damencho:


modules/xmpp/strophe.jingle.js
```diff
@@ -369,7 +369,8 @@ class JingleConnectionPlugin extends ConnectionPlugin {
                 const options = this.xmpp.options;
 
                 if (options.useStunTurn) {
-                    this.jvbIceConfig.iceServers = iceservers;
+                    this.jvbIceConfig.iceServers
```
bgrozev: Can we add a comment that explains why we're doing this? It's not obvious

damencho:
jenkins test this please

PR #754, opened 2018-06-06
Ff no media

hristoterezov:


saghul:
Why are we disabling P2P on Firefox? Can you at least describe why on the commit message?

gpolitis:
Hola Saul! this should have never been activated because sdp-interop does not support FF as the offerer https://github.com/jitsi/sdp-interop/issues/4. Hristo did some git archaeology and FF p2p was wrongfully enabled around a year ago. I agree we should document this.

hristoterezov:
@saghul See the description of the commit that disables it:

> Currently P2P won't succeed on FF because it's not supported by
> sdp-interop.

saghul:
@gpolitis But does it support it as the answerer?

@hristoterezov I think it needs some more background, like what @gpolitis said.

gpolitis:
>  But does it support it as the answerer?

Yeah, which is the normal use-case in the non-P2P scenario.

saghul:
> Yeah, which is the normal use-case in the non-P2P scenario.

Well, In P2P one has to be the offerer and one the answerer. Something I discussed with someone some time ago (maybe Emil?) was to support asymmetric P2P: that is, support P2P as long as it's the other party who sends the offer. Edge suffers from the same problem.

Good to know, at any rate!

PR #755, opened 2018-06-06
fix(migration): do not do double offer-answer for Chrome 63 and up

virtuacoplenny:
When switching bridges, removing the "data" section from the SDP
and doing a double offer-answer results in an error starting on
Chrome 63. So on 63 and up, do a single off-answer without
removing the "data" section.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -16,6 +16,7 @@ import SDPDiffer from './SDPDiffer';
 import SDPUtil from './SDPUtil';
 import SignalingLayerImpl from './SignalingLayerImpl';
 
+import browser from '../browser';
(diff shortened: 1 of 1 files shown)


damencho:
Fixes https://github.com/jitsi/jicofo/issues/282

paweldomas:
Have you verified that the data channels are operational after migration without removing the data section ? I'm asking because the description doesn't say "on Chrome 63 the workaround is no longer necessary".

virtuacoplenny:
I hadn't check; I didn't think to check. Testing again I see the data channel be closed when I take down the bridge but do not see another opened after getting switched to another bridge.

paweldomas:
Yes, that was the problem which was being worked around. New data channel was not being re-opened.

PR #756, opened 2018-06-07
Layer suspension

bbaldino:
Handle a updates from the bridge about whether or not the endpoint is selected by anyone else, so that it may suspend its higher layers if they are not in use.

Once this is ready, let me merge it so I can clean up the commit messages.

### DIFF
--- a/modules/RTC/BandwidthLimiter.js
+++ b/modules/RTC/BandwidthLimiter.js
@@ -1,0 +1,80 @@
+/* global __filename */
+
+import transform from 'sdp-transform';
+import { getLogger } from 'jitsi-meet-logger';
+
+const logger = getLogger(__filename);
+
+/**
+ * This will save a bandwidth limit (per mline) that should be used
+ * and then, when given an SDP, will enforce that limit.
+ * Note that this will affect *outgoing* bandwidth usage, but the SDP
+ * it must modify to implement that is the *remote* description
+ */
+export default class BandwidthLimiter {
+
+    /**
+     * Create a new BandwidthLimiter
+     */
+    constructor() {
+        /**
+         * Map of mline media type to a bandwidth
+         * limit (in kbps).  Any mlines present in this
+         * map will have the associated bandwidth limit
+         * (or 'null' for no limit) enforced, meaning that
+         * it will potentially overwrite a limit
+         * already set in the given sdp.  However, if an
+         * mline is NOT present in this map, any limit
+         * in the given sdp will not be touched.
+         */
+        this._bandwidthLimits = {};
+    }
+
+    /**
+     * Set a bandwidth limit for given mline.  If
+     * limitKbps is null, the limit will be removd
+     * @param mediaType the mline media type to set the
+     * bandwidth limit on
+     * @param limitKbps the bandwidth limit, in kbps
+     */
+    setBandwidthLimit(mediaType, limitKbps) {
+        this._bandwidthLimits[mediaType] = limitKbps;
+    }
+
+    /**
+     * Enforce any configured bandwidth limits (or
+     * lack thereof) in the given sdp
+     * @param sdp the session description
+     * @returns a potentially modified session description
+     * with any configured bandwidth limits set
+     */
+    enforceBandwithLimit(sdp) {
+        logger.debug('Enforcing any configured bandwidth limits');
+        const desc = transform.parse(sdp);
+
+        desc.media.forEach(mLine => {
+            const limitKbps = this._bandwidthLimits[mLine.type];
+
+            if (typeof limitKbps !== 'undefined') {
+                if (limitKbps === null) {
+                    logger.debug(
+                        `Removing bandwidth limit for mline ${mLine.type}`);
+                    delete mLine.bandwidth;
+                } else {
+                    logger.debug(`Enforcing limit ${limitKbps}kbps`
+                        + ` for mline ${mLine.type}`);
+                    mLine.bandwidth = [
+                        {
+                            type: 'AS',
+                            limit: limitKbps
+                        }
+                    ];
+                }
+            }
+        });
+
+        return transform.write(desc);
+    }
+
+
+}
--- a/modules/RTC/BandwidthLimiter.js
+++ b/modules/RTC/BandwidthLimiter.js
@@ -1,7 +1,7 @@
 /* global __filename */
 
-import transform from 'sdp-transform';
 import { getLogger } from 'jitsi-meet-logger';
+import transform from 'sdp-transform';
 
 const logger = getLogger(__filename);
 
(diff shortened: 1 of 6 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -205,13 +206,21 @@ export default function TraceablePeerConnection(
...
+
     /**
      * TracablePeerConnection uses RTC's eventEmitter
      * @type {EventEmitter}
      */
     this.eventEmitter = rtc.eventEmitter;
     this.rtxModifier = new RtxModifier();
 
+    /**
+     * Whether or not this endpoint has been selected
```
paweldomas: do you utilise full line length of 80 in this file ?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1852,6 +1873,21 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(
...
+    if (this.options.enableLayerSuspension) {
+        logger.debug('Layer suspension enabled,'
+            + `currently selected? ${this.isSelected}`);
+        const bitrateCapKbps = getSuspensionBitrateKbps(this.isSelected);
+
+        this.bandwidthLimiter.setBandwidthLimit('video', bitrateCapKbps);
+        logger.debug(`Layer suspension got bitrate cap of ${bitrateCapKbps}`);
+        description.sdp
+            = this.bandwidthLimiter.enforceBandwithLimit(description.sdp);
+        logger.warn(
```
paweldomas: logging on the warning level should happen only if something goes wrong

service/RTC/RTCEvents.js
```diff
@@ -14,6 +14,8 @@ const RTCEvents = {
     DOMINANT_SPEAKER_CHANGED: 'rtc.dominant_speaker_changed',
     LASTN_ENDPOINT_CHANGED: 'rtc.lastn_endpoint_changed',
 
+    SELECTED_UPDATE: 'rtc.selected_update',
```
paweldomas: majority of the RTC events use CHANGED word instead of UPDATED. Does it make sense to name it SELECTED_CHANGED ? or IS_SELECTED_CHANGED ?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1043,6 +1052,18 @@ function extractSSRCMap(desc) {
     return ssrcMap;
 }
 
+/**
+ * @param allSsrcs all the send ssrcs for this client.
```
paweldomas: 'allSsrcs' does not match argument name 'isSelected'

webintoit:
Awesome feature

modules/RTC/RTC.js
```diff
@@ -278,6 +278,23 @@ export default class RTC extends Listenable {
...
         this.addListener(RTCEvents.LASTN_ENDPOINT_CHANGED,
             this._lastNChangeListener);
+
+        this.addListener(
+            RTCEvents.SELECTED_UPDATE,
+            isSelected => {
+                const jvbSession = this.conference.jvbJingleSession;
+
+                if (jvbSession) {
+                    jvbSession.peerconnection.setIsSelected(isSelected);
```
paweldomas: I don't like how it uses private variables, but I guess you have figured out how to add the lest code.
Is it possible that SELECTED_UPDATE would arrive before jvbSession is ready and would we care (over WebSocket data channel for example ) ?

modules/RTC/TraceablePeerConnection.js
bbaldino: fixed

modules/RTC/TraceablePeerConnection.js
bbaldino: fixed

modules/RTC/TraceablePeerConnection.js
bbaldino: fixed

modules/RTC/RTC.js
bbaldino: well this isn't technically a private variable (or at least it's not indicated that it is, by being named with a leading `_`), but I did look into other ways (that would have been 'cleaner') but they would require a lot of work and justifying it for this (where this solution wasn't _that_ bad) didn't seem right.

bbaldino: as far as getting the update before the session is ready, i don't think we care too much since this can be viewed as a best effort thing anyway.  since we default `isSelected` to true, i don't think this scenario would cause the endpoint to send low quality when we need high, if anything it would continue to send high when we only needed low which isn't so bad.

service/RTC/RTCEvents.js
bbaldino: fixed

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1852,6 +1875,21 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(
         });
     }
 
+    if (this.options.enableLayerSuspension) {
```
paweldomas: do you mind describing this new option in TraceablePeerConnection constructor like few others are ?

modules/RTC/BandwidthLimiter.js
```diff
@@ -0,0 +1,80 @@
+/* global __filename */
+
+import transform from 'sdp-transform';
```
paweldomas: can we have the imports sorted alphabetically ? we try to do that in jitsi-meet

modules/RTC/BandwidthLimiter.js
```diff
@@ -0,0 +1,80 @@
...
+         * Map of mline media type to a bandwidth
+         * limit (in kbps).  Any mlines present in this
+         * map will have the associated bandwidth limit
+         * (or 'null' for no limit) enforced, meaning that
+         * it will potentially overwrite a limit
+         * already set in the given sdp.  However, if an
+         * mline is NOT present in this map, any limit
+         * in the given sdp will not be touched.
+         */
+        this._bandwidthLimits = {};
```
paweldomas: I think Saul would have asked here to use a Map and I would ask to describe the key/value types. If Map would be used it would be `@type {Map<string, number}`

modules/RTC/BandwidthLimiter.js
```diff
@@ -0,0 +1,80 @@
...
+        logger.debug('Enforcing any configured bandwidth limits');
+        const desc = transform.parse(sdp);
+
+        desc.media.forEach(mLine => {
+            const limitKbps = this._bandwidthLimits[mLine.type];
+
+            if (typeof limitKbps !== 'undefined') {
+                if (limitKbps === null) {
+                    logger.debug(
+                        `Removing bandwidth limit for mline ${mLine.type}`);
```
paweldomas: it doesn't seem like the line limit of 80 is used, can you go through all the new code and comments ? I don't want to point every single one. I haven't opened this in any editor, but this line seems much longer than all others and judging by that the comments seem to be broken too early.

modules/RTC/BandwidthLimiter.js
```diff
@@ -0,0 +1,80 @@
...
+         * mline is NOT present in this map, any limit
+         * in the given sdp will not be touched.
+         */
+        this._bandwidthLimits = {};
+    }
+
+    /**
+     * Set a bandwidth limit for given mline.  If
+     * limitKbps is null, the limit will be removd
+     * @param mediaType the mline media type to set the
```
paweldomas: could you please describe the types of each `@param` ? we try to do that in jitsi-meet. This applies to all code added in this PR

modules/RTC/TraceablePeerConnection.js
bbaldino: nice catch, fixed

modules/RTC/BandwidthLimiter.js
bbaldino: fixed

modules/RTC/BandwidthLimiter.js
bbaldino: fixed

modules/RTC/BandwidthLimiter.js
bbaldino: fixed...somewhat begrudgingly.

modules/RTC/BandwidthLimiter.js
bbaldino: fixed

modules/RTC/RTC.js
paweldomas: Actually I gave this part second thought and why can't we move it to JitsiConference ? I don't like that now RTC module has knowledge about JitsiConference, JingleSessionPC and is in fact dealing with kind of internal parts of JitsiConference. It should not be a big change.

paweldomas: In fact I wanted to get rid of this.conference field in RTC. It's used in two places both are kind of hacky.

paweldomas: Or even JingleSessionPC (the non p2p one) could be listening to that even directly - it has the RTC module instance.

bbaldino: > Or even JingleSessionPC (the non p2p one) could be listening to that even directly - it has the RTC module instance.

This would be nice...It didn't listen to any RTCEvents before so it didn't seem consistent to add that there, but I will try that out and update here if you think that'll be alright.

PR #757, opened 2018-06-07
fix(package.json): Use sdp-interop/simulcast from jitsi npm org.

hristoterezov:


package.json
```diff
@@ -21,8 +21,8 @@
     "jitsi-meet-logger": "github:jitsi/jitsi-meet-logger#6fff754a77a56ab52499f3559105a15886942a1e",
     "js-utils": "github:jitsi/js-utils#d4b78721b754a15c2588b7b0c621a9fff6fa1363",
     "react-native-callstats": "3.27.0",
-    "sdp-interop": "0.1.13",
-    "sdp-simulcast": "0.2.1",
+    "@jitsi/sdp-interop": "0.1.13",
+    "@jitsi/sdp-simulcast": "0.2.1",
```
saghul: These are incorrectly sorted now.

PR #758, opened 2018-06-08
M67

saghul:
With the update of react-native-webrtc to M67, the ability to stop the camera
when a track is disabled was introduced, so there is no longer a need for doing
a local O/A to remove the track for video muting.


saghul:
@jitsi/mobile This is now ready for review. Note that this can only be used in conjunction with the M67 branch of react-native-webrtc.

PR #760, opened 2018-06-11
fix(migration): reject the data channel to restart it

virtuacoplenny:
I've been looking at this with Brian (who has been doing most of the leg work...) and while this works we do not understand exactly why yet. The bridge is getting the sdp without the data channel but it is unknown where that is getting stripped.

Starting on chrome 63, an error occurs when renegotiating
with the new bridge on bridge migration when the SDP sections
do not match, which occurs when removing the data channel
section. Not doing the double renegotiate prevents the data
channel from being reopened. The way around this, while still
avoing the error, is to reject the data channel section instead
of removing it to trigger a new data channel.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -978,16 +978,17 @@ export default class JingleSessionPC extends JingleSession {
     replaceTransport(jingleOfferElem, success, failure) {
         this.room.eventEmitter.emit(XMPPEvents.ICE_RESTARTING, this);
 
-        // We need to first set an offer without the 'data' section to have the
-        // SCTP stack cleaned up. After that the original offer is set to have
-        // the SCTP connection established with the new bridge.
+        // We need to first reject the 'data' section to have the SCTP stack
+        // cleaned up to signal the known data channel is now invalid. After
+        // that the original offer is set to have the SCTP connection
+        // established with the new bridge.
         const originalOffer = jingleOfferElem.clone();
 
         jingleOfferElem
             .find('>content[name=\'data\']')
             .attr('senders', 'rejected');
 
-        // First set an offer without the 'data' section
+        // First set an offer with a rejected 'data' section
         this.setOfferAnswerCycle(
             jingleOfferElem,
             () => {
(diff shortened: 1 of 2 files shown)


bbaldino:
So I tried to look into *why* this fixes the issue, and was unable to come up with an answer in the amount of time I was willing to spend.  Some findings:

1) The colibri messages we send from Jicofo to the failover bridge (the one we are moving clients to) appear to be the same in both the good case (where this fix is present) and the bad case (where we don't reject the app line and then re-add it).  This would appear to point to an issue on the chrome side, but:

2) I forked the webrtc samples and tweaked a demo to do ice restart with data channels ([branch here](https://github.com/bbaldino/samples/tree/dc_ice_restart)).  This sets up 3 peer connections: one which simulates the client, one for the original bridge and one for the failover bridge.  Restarting ice between the client and the failover bridge works fine: data channel reconnects without any special treatment.

I don't see how it could be any sort of sctp setup issue, since this is a brand new bridge and, from the logs I added, it appeared to be doing all the same steps the original bridge did (after all, why wouldn't it).

I'm still not even sure where this problem lies (chrome or us): although the webrtc demo clearly works as expected, maybe in jitsi-meet there is something else we're triggering in chrome that needs to be worked around (i.e. this fix) or perhaps something about moving an ongoing call to a new bridge is breaking the sctp connection setup in some way (though ICE itself succeeds fine).

virtuacoplenny:
@paweldomas, do you know how/why removing "data" used to work in re-establishing the data channel and how/why setting the m-line to rejected might also work?

paweldomas:
> @paweldomas, do you know how/why removing "data" used to work in re-establishing the data channel and how/why setting the m-line to rejected might also work?

I guess that it's because the SCTP association must be destroyed, so that Chrome will try to re-establish a new one. The bridge acts as a server and only passively listens for new SCTP connection to be established, but I guess that Chrome is not doing that until the old one is destroyed.

It's possible that when testing with two Chromes both sides act as "client" and are actively trying to re-establish the SCTP connection or something like that. We could confirm that by watching .pcap capture of the traffic or maybe checking on the code.

virtuacoplenny:
Force pushed to resolve conflicts in JingleSessionPC imports.

PR #768, opened 2018-06-20
fix(presence): do not escape xml values

virtuacoplenny:
When receiving presence, the XML is converted to JSON. In
the process, Strophe.getText is used, which calls its
xmlescape utility function. This is not desired as the
raw value is desired, especially for display name.

Note: this issue only affects presence as it is the only
place that call the helper packet2JSON, which is the only
place that calls Strophe.getText.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -30,7 +30,10 @@ export const parser = {
             const text = Strophe.getText(child);
 
             if (text) {
-                node.value = text;
+                // Using Strophe.getText will do work for traversing all direct
+                // child text nodes but returns an escaped value, which is not
+                // desirable at this point.
+                node.value = Strophe.xmlunescape(text);
             }
             nodes.push(node);
             this.packet2JSON(child, node.children);


virtuacoplenny:
@bbaldino, an alternative I thought of was looking for text nodes myself in the XML, essentially implementing my own version of Strophe.getText without the escaping.

bbaldino:
@virtuacoplenny how bad is the custom implementation of `getText`? Bummer we have to do the unescape, but might be better than a chunk of custom code we have to maintain.

virtuacoplenny:
Really not bad. We might even be able to make it simpler but I was avoiding that case to be safe.

```javascript
    getText: function (elem) {
        if (!elem) { return null; }

        var str = "";
        if (elem.childNodes.length === 0 && elem.nodeType === Strophe.ElementType.TEXT) {
            str += elem.nodeValue;
        }

        for (var i = 0; i < elem.childNodes.length; i++) {
            if (elem.childNodes[i].nodeType === Strophe.ElementType.TEXT) {
                str += elem.childNodes[i].nodeValue;
            }
        }

        return Strophe.xmlescape(str);
    }
```

bbaldino:
oh, so it'd be exactly what's there but with `Strophe.xmlescape(str);` line removed?

virtuacoplenny:
Yeah, it'd be..."heavily inspired" by Strophe#getText without bringing over escaping.

bbaldino:
I think what you have here (calling it and then calling Strophe's own xmlunescape) is probably safer against future strophe changes, so unless you think there's some performance concern with doing the escape/unescape I think it's probably better this way.

PR #773, opened 2018-06-27
feat(sanity): axe IE and Temasys plugin support 🔥🔥🔥

saghul:


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2443,21 +2406,10 @@ TraceablePeerConnection.prototype.addIceCandidate = function(
  * @returns {void}
  */
 TraceablePeerConnection.prototype.getStats = function(callback, errback) {
-    // TODO: Is this the correct way to handle Opera, Temasys?
     // TODO (brian): After moving all browsers to adapter, check if adapter is
     // accounting for different getStats apis, making the browser-checking-if
     // unnecessary.
-    if (browser.isTemasysPluginUsed()
-        || browser.isReactNative()) {
```
virtuacoplenny: Maybe this line being removed is why the tests fail?

saghul: Oh, I missed that! Thanks for pointing out Lenny!

modules/RTC/JitsiTrack.js
```diff
@@ -272,31 +267,20 @@ export default class JitsiTrack extends EventEmitter {
...
      * @param container the HTML container which can be 'video' or 'audio'
-     * element. It can also be 'object' element if Temasys plugin is in use and
-     * this method has been called previously on video or audio HTML element.
+     * element.
      *
-     * @returns potentially new instance of container if it was replaced by the
-     *          library. That's the case when Temasys plugin is in use.
+     * @returns {void}
      */
     attach(container) {
```
paweldomas: note for me to check if usages on the return value have been updated

modules/RTC/RTCUIHelper.js
```diff
@@ -1,82 +0,0 @@
-/* global $ */
-
-import browser from '../browser';
-
-const logger = require('jitsi-meet-logger').getLogger(__filename);
-
-const RTCUIHelper = {
```
paweldomas: note to me to check if all usages have been updated

modules/RTC/RTCUtils.js
```diff
@@ -1144,11 +1038,34 @@ class RTCUtils extends Listenable {
...
+                            'devicechange',
+                            () => this.enumerateDevices(
+                                    onMediaDevicesListChanged));
+                    } else {
+                        pollForAvailableMediaDevices();
+                    }
+                });
             }
+
+            resolve();
```
paweldomas: have you looked if it's possible to change the init method to not be a promise  (can be done in separate commit/PR if you want to look into that) ?

saghul: I thought of it. While this would indeed simplify some stuff we do today, I worry we might need the process to be async in the future.

So I'd let it be as-is for a bit and revisit.

PR #776, opened 2018-06-29
fix(tracks): set remote tracks to default to videoType camera

virtuacoplenny:
This fix resolves a race condition exposed by 1bd6ee4b. Before 1bd6ee4b,
JitsiConference#replaceTrack would call JingleSessionPC#replaceTrack to update
remote participants with a new local track through jingle and then call
JitsiConference#sendCommand to update remote participants with the local track's
videoType through presence. However, replaceTrack would not be waited on for
success or failure, and instead sendCommand would end up getting fired first,
then replaceTrack completing. So remote participants would end up usually
getting the videoType first through the sendCommand call and then the track
update through the replaceTrack call.

With the promisification of JitsiConference#replaceTrack, the order of execution
is guaranteed to be JingleSessionPC#replaceTrack and then
JitsiConference#sendCommand, thereby making it more likely that the track update
comes in first and then the videoType update. (Although network conditions could
still modify the order of reception.) This causes a period of time when a remote
track update could be received without a known videoType, such as when the
remote participant joined as video muted and has un-video muted. For now default
to type CAMERA to avoid logic relying on videoType existence from blowing up.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -12,6 +12,7 @@ import RTC from './RTC';
 import RTCUtils from './RTCUtils';
 import browser from '../browser';
 import RTCEvents from '../../service/RTC/RTCEvents';
+import VideoType from '../../service/RTC/VideoType';
 import RtxModifier from '../xmpp/RtxModifier';
 
 // FIXME SDP tools should end up in some kind of util module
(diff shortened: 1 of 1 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -709,7 +710,11 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track) {
     }
 
     const muted = peerMediaInfo.muted;
-    const videoType = peerMediaInfo.videoType; // can be undefined
+
+    // peerMediaInfo.videoType can be undefined so set a default type in case
+    // there is logic depending on type existing and rely on a presence update
+    // to come in to update to the correct type.
+    const videoType = peerMediaInfo.videoType || VideoType.CAMERA;
```
bbaldino: so why default it to camera here instead of have logic that checks the type handle the case where it is `undefined`?

virtuacoplenny: Honestly, I couldn't decide which to do; either set a default when creating the remote jitsi track or handle undefined as needed. I couldn't tell for sure what places would break if videoType were undefined so this seemed like a good enough place to handle the undefined. Do you have a preference for handling errors due to undefined as they show up?

bbaldino: after we talked about it earlier, i think keeping this as undefined and making other code handle undefined is better in the long run.

if we find that to be too problematic, then we know this event needs to wait until we know what the videoType is (or those places need to listen for another event).

i don't like marking it as camera because we are explicitly guessing at information we don't actually have/know.

PR #778, opened 2018-07-04
DO NOT MERGE - core: refactor initialization not to return a Promise

saghul:
There is nothing asynchronous about the initialization process (anymore), thus
turn it into a synchronous method.

In addition, WebRTC support is absolute, it cannot change from not being
supported to being supported (as it plreviously could, thanks to Temasys) so get
rid of the ancillary logic to support that.

Last, introduce a way to check if WebRTC is supported in the current
environment: JitsiMeetJS.isWebRtcSupported().

### DIFF
--- a/JitsiMeetJS.js
+++ b/JitsiMeetJS.js
@@ -198,7 +198,7 @@ export default _mergeNamespaceAndModule({
             Statistics.sendLog(JSON.stringify(logObject));
         }
 
-        return RTC.init(options || {});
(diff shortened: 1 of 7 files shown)


saghul:
I need this as part of the routing refactor. @hristoterezov @lyubomir PTAL.

saghul:
Changed branch, please follow up on https://github.com/jitsi/lib-jitsi-meet/pull/779

PR #780, opened 2018-07-06
Respect enableRemb and enableTcc for P2P.

gpolitis:
Please do not merge this yet, I would also like to provide a knob for overriding the googCpuOveruseDetection option.

What would people think if in the meet config.js included a config.pcConfiguration (https://www.w3.org/TR/webrtc/#rtcconfiguration-dictionary) instead of the various options that we have now (config.stunServers, config.googIPv6). This would allow us to override/experiment any available peer-connection configuration parameter.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2196,6 +2196,15 @@ TraceablePeerConnection.prototype._createOfferOrAnswer = function(
             this.trace(
                 `create${logName}OnSuccess::preTransform`, dumpSDP(resultSdp));
 
+            if (isOffer) {
+                if (!this.options.enableRemb) {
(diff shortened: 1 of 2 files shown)


saghul:
> What would people think if in the meet config.js included a config.pcConfiguration (https://www.w3.org/TR/webrtc/#rtcconfiguration-dictionary) instead of the various options that we have now (config.stunServers, config.googIPv6). This would allow us to override/experiment any available peer-connection configuration parameter.

Me likey.

gpolitis:
I don't recall the original motivation behind this PR nor do I see a problem letting the browsers negotiate remb & tcc by themselves. The config.js modification that I suggested seems to be regarded positively, but it was merely that, a suggestion (and not the main object of this PR). So I'm closing this PR, for now at least, as invalid.

PR #782, opened 2018-07-17
feat(RTCUtils): default to 720p

saghul:


bbaldino:
what happened to glorious 1080?

saghul:
It's still configurable but the performance impact was too big so we changed it in config back to 720 months ago.

PR #784, opened 2018-07-19
feat(conference): expose method to select multiple participants

virtuacoplenny:


modules/RTC/RTC.js
```diff
@@ -172,10 +172,10 @@ export default class RTC extends Listenable {
         /**
          * The endpoint ID of currently selected participant or <tt>null</tt> if
          * no user is selected.
-         * @type {string|null}
+         * @type {Array|null}
          * @private
          */
-        this._selectedEndpoint = null;
+        this._selectedEndpoints = null;
```
virtuacoplenny: Instead of null I should use undefined so that sendSelectedEndpointsMessage will be called with undefined and default to an empty array.

virtuacoplenny:
Force pushed to resolve conflicts in RTC.js due to the new changes around storing of _maxFrameHeight.

PR #785, opened 2018-07-20
ref(video-quality): cache max frame height and send on channel open

virtuacoplenny:
Currently it is possible to try to change the max receive video
frame height before the data channel is open. In that case an error
will be thrown. This change makes it so that the desired frame height
is saved and sent on channel open, avoiding the thrown error the
max receive video frame height logic is exercised through the
JitsiConference api.

### DIFF
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -165,10 +165,10 @@ export default class RTC extends Listenable {
          * The number representing the maximum video height the local client
          * should receive from the bridge.
          *
-         * @type {number|null}
+         * @type {number|undefined}
          * @private
          */
-        this._maxFrameHeight = null;
+        this._maxFrameHeight = undefined;
 
         /**
          * The endpoint ID of currently pinned participant or <tt>null</tt> if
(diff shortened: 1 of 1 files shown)


modules/RTC/RTC.js
```diff
@@ -327,14 +336,17 @@ export default class RTC extends Listenable {
...
      *
      * @param {number} maxFrameHeightPixels the maximum frame height, in pixels,
      * this receiver is willing to receive.
      * @returns {void}
      */
     setReceiverVideoConstraint(maxFrameHeight) {
-        if (this._channel) {
+        this._maxFrameHeight = maxFrameHeight;
+
+        if (this._channel && this._channelOpen) {
```
bbaldino: shouldn't there be some logic somewhere that waits on the channel open event and, when it fires, sends the message we tried to send it before the channel opened?

virtuacoplenny: That's probably incredibly crucial.

modules/RTC/RTC.js
```diff
@@ -260,11 +260,17 @@ export default class RTC extends Listenable {
                     this._pinnedEndpoint);
                 this._channel.sendSelectedEndpointMessage(
                     this._selectedEndpoint);
+
+                if (typeof this._maxFrameHeight !== 'undefined') {
+                    this._channel.sendReceiverVideoConstraintMessage(
+                        this._maxFrameHeight);
```
bbaldino: should `this._maxFrameHeight` be cleared after it's been successfully sent? (either via the cache or normally?)

virtuacoplenny: I don't think it needs to be cleared, so in this case I don't think it should. The way the caching works is the last selected maxFrameHeight is always saved when the maxFrameHeight is set, so it's essentially a reference to the last selected maxFrameHeight but its usage happens to be for sending the selected frame height on channel open. If I cleared it here, it would get set again when the client sets a new maxFrameHeight, unless I want to conditionally cache.

PR #786, opened 2018-07-20
Change layer suspension to use parameters in RTPSender

bbaldino:
We no longer suspend unused simulcast layers via a bandwidth cap in the
SDP, instead we'll use the new parameters in RTPSender to enable and
disable streams explicitly.  The main advantage here is the RTPSender
method ramps up immediately when we re-enable the layers (as opposed to
the SDP bandwidth cap which took 30+ seconds).


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2527,12 +2503,44 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function(track) {
...
 
+const handleLayerSuspension = function(peerConnection, isSelected) {
+    const videoSender = peerConnection.getSenders()
+        .find(sender => sender.track.kind === 'video');
+    if (!videoSender) {
+        logger.warn("handleLayerSuspension unable to find video sender");
+        return;
+    }
+    if (!videoSender.getParameters) {
+        logger.debug("Browser doesn't support RTPSender parameters");
```
virtuacoplenny: Are these logs noisy?

virtuacoplenny: Does it make more sense to you to put the support checks in here versus near the same place as this.options.enableLayerSuspension? I can see both ways but moving them out keeps the function focused on always executing a certain path.

bbaldino: > Are these logs noisy?

I didn't worry too much since I put it on debug.  What do you think?

> Does it make more sense to you to put the support checks in here versus near the same place as this.options.enableLayerSuspension? I can see both ways but moving them out keeps the function focused on always executing a certain path.

My thinking was that since this check gets into the 'details' of the implementation, it felt better to have it here rather than the caller. So the split of responsibilities is: the first part checks if this is even enabled, then it defers everything else to the implementation.

bbaldino:
had to rebase and force push

PR #793, opened 2018-08-01
feat(RTC): refactor logic for polling for device changes

saghul:
- Don't use it on supported browsers
- Make sure no stray timers are left

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -107,44 +107,29 @@ const isAudioOutputDeviceChangeAvailable
     = typeof featureDetectionAudioEl.setSinkId !== 'undefined';
 
 let availableDevices;
+let availableDevicesPollTimer;
 
 /**
- * "rawEnumerateDevicesWithCallback" will be initialized only after WebRTC is
- * ready. Otherwise it is too early to assume that the devices listing is not
- * supported.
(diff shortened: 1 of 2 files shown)


modules/RTC/RTCUtils.js
```diff
@@ -1041,12 +985,32 @@ class RTCUtils extends Listenable {
...
+                            // the first time we call enumerateDevices().
+                            // This is the initial step.
+                            if (typeof availableDevices === 'undefined') {
+                                availableDevices = pds.slice(0);
+                            } else if (compareAvailableMediaDevices(pds)) {
+                                onMediaDevicesListChanged(pds);
+                            }
+
+                            availableDevicesPollTimer
+                                = window.setTimeout(
```
virtuacoplenny: Why prefer setTimeout over setInterval?

saghul: Nice catch! No reason, I was moving the code around. Come to think of it, it makes more sense. I'll update it, please don't merge it, or feel free to amend my commit or add one of your own with the change. 👍

virtuacoplenny: Okay, up to you. I noticed soon after writing the question that the original was using a setTimeout; I'm not expecting you to change it. The reason why I haven't merged yet is because there's a (flaky) test failure.

PR #794, opened 2018-08-01
feat: Adds a couple of permanent analytics properties.

gpolitis:


modules/xmpp/ChatRoom.js
```diff
@@ -613,6 +614,36 @@ export default class ChatRoom extends Listenable {
                     this.eventEmitter.emit(XMPPEvents.BRIDGE_DOWN);
                 }
                 break;
+            case 'conference-properties':
```
bgrozev: These should be accepted if they come from jicofo (`member.isFocus`), no?

gpolitis: Yes, you're right. I pushed a fix.

PR #795, opened 2018-08-03
feat: Measures end-to-end RTT with ping over the data channel.

bgrozev:


modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+ */
+const E2E_PING_RESPONSE = 'e2e-ping-response';
+
+/**
+ * Saves e2e ping related state for a single JitsiParticipant.
+ */
+class ParticipantState {
+    /**
+     * Creates a ParticipantState
+     * @param {JitsiParticipant} participant
```
bbaldino: more docs for these? (is this the local participant or a remote one?)

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+        // request. Start at 1 so we can consider only thruthy values valid.
+        this.lastRequestId = 1;
+
+        // If the data channel was already open (this is likely a participant
+        // joining an existing conference) send a request immediately.
+        if (e2eping.dataChannelOpened) {
+            this.sendRequest();
+        }
+
+        this.pingInterval = window.setInterval(
```
bbaldino: i was expecting that e2eping would actually be responsible for a lot of the stuff that's being done here.  maybe it's just due to naming, but i was expecting `ParticipantState` to basically just hold data, and `E2eping` to be doing all the management of pings (when they should be sent, the current sequence number, etc.) and would just fire some round trip event or something whenever a new one was calculated.  is there an E2eping instance for each remote participant?  (if so, i wonder if there is anything to the idea of a single broadcast ping that all remote participants would respond to rather than multiple individuals ones?  but maybe that would be harder to manage, i dunno)

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+        if (request) {
+            request.rtt = window.performance.now() - request.timeSent;
+            this.e2eping.conference.eventEmitter.emit(
+                E2ePingEvents.E2E_RTT_CHANGED,
+                this.participant,
+                request.rtt);
+        }
+
+        this.maybeSendAnalytics();
+    }.bind(this); // eslint-disable-line no-invalid-this
```
bbaldino: why define the functions this way instead of normal function definitions?

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+        // either, because our data channel might not have initialized.
+        // So once the data channel initializes, send requests to everyone.
+        // Wait an additional 200ms to give a chance to the remote side (if it
+        // also just connected as is the case for the first 2 participants in a
+        // conference) to open its data channel.
+        conference.on(
+            JitsiConferenceEvents.DATA_CHANNEL_OPENED,
+            () => {
+                this.dataChannelOpened = true;
+                for (const id in this.participants) {
```
bbaldino: i think using `for..of` will make this cleaner

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+
+    /**
+     * Handles a ping request coming from another participant.
+     * @param {JitsiParticipant} participant the participant who sent the
+     * request.
+     * @param {Object} request the request.
+     */
+    handleRequest(participant, request) {
```
bbaldino: if we only use the participant id here, what about just passing the id in directly instead of the participant?

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+        }
+    }
+
+    /**
+     * Handles a ping response coming from another participant
+     * @param {JitsiParticipant} participant the participant who sent the
+     * response.
+     * @param {Object} response the response.
+     */
+    handleResponse(participant, response) {
```
bbaldino: same thought about participant id

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+    /**
+     * Handles a participant joining the conference. Starts to send ping
+     * requests to the participant.
+     * @param {JitsiParticipant} participant the participant that joined.
+     */
+    participantJoined(participant) {
+        const id = participant.getId();
+
+        if (this.participants[id]) {
+            // State state left?
```
bbaldino: what's this comment?

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+
+        // We don't want to wait the whole interval before sending the first
+        // request, but we can't send it immediately after the participant joins
+        // either, because our data channel might not have initialized.
+        // So once the data channel initializes, send requests to everyone.
+        // Wait an additional 200ms to give a chance to the remote side (if it
+        // also just connected as is the case for the first 2 participants in a
+        // conference) to open its data channel.
+        conference.on(
+            JitsiConferenceEvents.DATA_CHANNEL_OPENED,
```
bbaldino: should this be relying on the more generic 'bridge channel' (which can be using websocket or data channel) instead of datachannel directly?

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,333 @@
...
+            }
+        }
+        logger.info(
+            `Initializing e2e ping; pingInterval=${
+                this.pingIntervalMs}, analyticsInterval=${
+                this.analyticsIntervalMs}.`);
+
+        // Only subscribe to user join/leave events if sending pings is enabled.
+        if (this.pingIntervalMs > 0) {
+            conference.on(
```
bbaldino: i think it'd be nice to remove the need to pass the conference object to this class...what would it look like to have the creator of this class do the wiring?  i.e. have it add a handler for the conference events and attach it to the handler methods here?  and we could pass the conference's event emitter directly.  this still leaves the sendMessage method but that could just be passed as a lambda which wraps the conference method call.

modules/e2eping/e2eping.js
bgrozev: I initially tried this:
```
handleRespone(response) {
// ...
}
```

It resulted in them not being properly bound :shrug:

modules/e2eping/e2eping.js
bgrozev: This results in more inline code in the constructor

modules/e2eping/e2eping.js
bgrozev: It's supposed to say "stale", but I guess it's redundant now that there's a log message. Removed

modules/e2eping/e2eping.js
bgrozev: It does, it's just that the event has the old data channel specific name.

modules/e2eping/e2eping.js
bbaldino: changing `this.handleRequest(participant, payload);` to `this.handleRequest(participant.getId(), payload);`?

modules/e2eping/e2eping.js
bgrozev: There is a single E2ePing instance. And one (what is now) ParticipantWrapper per remote partcipant.

bgrozev: I've added a TODO for using a broadcast, it's definitely a better solution

modules/e2eping/e2eping.js
bgrozev: `VM1279:1 Uncaught TypeError: XXX is not iterable`

modules/e2eping/e2eping.js
bgrozev: I switched it

modules/e2eping/e2eping.js
bgrozev: I did that, but I don't think the results are great. Let me know if you want me to keep or revert the commit.

modules/e2eping/e2eping.js
bbaldino: sorry, i didn't notice that was an object.  it would be:
`for (const key of Object.keys(participants)) {`

though in the past i think saul has pushed for using an explicit map when we want a map instead of an object

JitsiConference.js
```diff
@@ -245,7 +245,32 @@ JitsiConference.prototype._init = function(options = {}) {
 
     this.room.updateDeviceAvailability(RTC.getDeviceAvailability());
 
-    this.e2eping = new E2ePing(this, config);
+    this.e2eping = new E2ePing(
```
bbaldino: i think this looks better, thanks.  only tweak i'd say is move this code to a helper method.

bbaldino:
i think we should get to the bottom of the function definition thing.  i don't get why a normal definition doesn't work (like in [BridgeChannel](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/BridgeChannel.js)

bgrozev:
> i think we should get to the bottom of the function definition thing. i don't get why a normal definition doesn't work

I didn't exactly get to the bottom of it, but I did what Hristo suggested as consistent with the other code, which is use the cleaner function definition syntax and add a `bind` in the constructor. The funny thing is that I need the explicit bind in `ParticipantWrapper` (otherwise it is bound to `window`), but I don't need it in `E2ePing`.

modules/e2eping/e2eping.js
```diff
@@ -0,0 +1,334 @@
...
+        // either, because our data channel might not have initialized.
+        // So once the data channel initializes, send requests to everyone.
+        // Wait an additional 200ms to give a chance to the remote side (if it
+        // also just connected as is the case for the first 2 participants in a
+        // conference) to open its data channel.
+        for (const id in this.participants) {
+            if (this.participants.hasOwnProperty(id)) {
+                const participantWrapper = this.participants[id];
+
+                window.setTimeout(participantWrapper.sendRequest, 200);
```
bbaldino: i think this is the source of your bind issues.  this is just passing the member function so the `this` won't be bound to participantWrapper, it'll be bound to where it's called from (which in this case will be `window`).  Doing it this way will fix it:

`window.setTimeout(() => participantWrapper.sendRequest(), 200);`

bgrozev: I see. I think I prefer the explicit `bind` in the constructor so that you don't have to worry about it when you call the function.

bbaldino: So then why bind all the functions if this one is the only being used in a way where it's invoked without `this` being properly bound?  I don't think it is common to proactively rebind `this` to member functions in case they will be passed as a standalone function.  If we're going to use the bind method, I think it should be done at the call-site when it is needed, as in:

`window.setTimeout(participantWrapper.sendRequest.bind(participantWrapper), 200);`

The issue only occurs when the function is passed as a standalone function (like as a callback here in `window.setTimeout`) and refers to `this`, so normal function calls (other places these functions are called via `participantWrapper.XXX`) work as expected without the bind.

paweldomas: Generally an arrow function is the preferred way by the team, so what Brian suggested at the top would be the way to go.

paweldomas: But I see this has been merged and not a big deal, so feel free to ignore my comment

hristoterezov: I'm with the impression that bind is the proffered way by the team. In jitsi-meet for example bind is used in most of those cases. Why do you think the arrow function is the preferred way?

paweldomas: Bind was the "preferred way" before anyone in the team knew about the arrow function. If you search for usages in jitsi-meet (which is never code than lib-jitsi-meet) 90% of the cases are with the arrow function. I'm not going to argue about this as I don't care

hristoterezov: Hmm... Just to be clear my intention is not to start argument, my only intention is to know which is the "preferred way" in order to follow it.

Probably I'm missing something because in every React component we are doing binds. So I'm not really sure what you are talking about.

paweldomas: We're doing binds in every components only to avoid binding on every single render

PR #796, opened 2018-08-03
JingleSessionPC cleanup

paweldomas:


modules/xmpp/JingleSession.js
```diff
@@ -23,21 +23,35 @@ export default class JingleSession {
...
         this.iceConfig = iceConfig;
 
+        /**
+         * Indicates whether this instance is an initiator or an answerer of
+         * the Jingle session.
+         * @type {boolean}
+         */
+        this.isInitiator = isInitiator;
+
+        this.initiator = this.isInitiator ? this.localJid : this.remoteJid;
```
bbaldino: what about naming these `initiatorJid`/`respondJid`?

modules/xmpp/JingleSession.js
```diff
@@ -23,21 +23,35 @@ export default class JingleSession {
...
+
+        this.responder = this.isInitiator ? this.remoteJid : this.localJid;
```
saghul: why not make these getter properties?

modules/xmpp/JingleSessionPC.js
```diff
@@ -242,48 +243,54 @@ export default class JingleSessionPC extends JingleSession {
...
+                = options.disableSimulcast
+                    || (options.preferH264 && !options.disableH264);
+            pcOptions.preferH264 = options.preferH264;
             pcOptions.enableFirefoxSimulcast
-                = this.room.options.testing
-                    && this.room.options.testing.enableFirefoxSimulcast;
-            pcOptions.enableLayerSuspension
-                = this.room.options.enableLayerSuspension;
+                = options.testing && options.testing.enableFirefoxSimulcast;
+            pcOptions.enableLayerSuspension = options.enableLayerSuspension;
```
saghul: Not sure if it has any impact, but this is not going to work if there is no simulcast, is it? Maybe check for simulcast too then?

paweldomas: This commit refactors the options, but I don't want to do any changes to the logic behind it.

paweldomas:
jenkins test this please

PR #802, opened 2018-08-14
Analytics event for Jingle session-initiate timeout

paweldomas:


JitsiConference.js
```diff
@@ -361,7 +369,7 @@ JitsiConference.prototype._init = function(options = {}) {
  */
 JitsiConference.prototype.join = function(password) {
     if (this.room) {
-        this.room.join(password);
+        this.room.join(password).then(() => this._maybeSetSITimeout());
```
saghul: drive-by-comment: would it make sense to return the promise here?

paweldomas: Jitsi-meet (if I remember correctly) uses lib-jitsi-meet in a more event based way, so the app waits for the events rather than tracking a Promise. Maybe having it here would add confusion on what should be used. Also there's no failure case in room.join() (to add it another an event listener would have to be used).

PR #805, opened 2018-08-30
fix(device-list): workaround for devicechange being fired twice

virtuacoplenny:
Chrome fires devicechange twice. This causes the DEVICE_LIST_CHANGED
event to fire twice. The listener in jitsi-meet has async logic
to handle the event, including creation of new tracks. So it
can happen that two devicechange events fire, the jitsi-meet
listener fires twice, creates duplicate tracks, and uses both
tracks. Diffing the device list for an actual change is a workaround
against extra devicechange events.

Note: this does not solve the issue in jitsi-meet of it not
handling quick successive DEVICE_LIST_CHANGED events.

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -879,25 +879,28 @@ class RTCUtils extends Listenable {
                     RTCEvents.DEVICE_LIST_AVAILABLE,
                     availableDevices);
 
+
+                // Use a shared callback to handle both the devicechange event
+                // and the polling implementations. This prevents duplication
+                // and works around a chrome bug (verified to occur on 68) where
+                // devicechange fires twice in a row, which can cause async post
+                // devicechange processing to collide.
+                const updateKnownDevices = () => this.enumerateDevices(pds => {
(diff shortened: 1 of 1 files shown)


modules/RTC/RTCUtils.js
```diff
@@ -879,25 +879,28 @@ class RTCUtils extends Listenable {
...
+                        updateKnownDevices);
                 } else {
                     // Periodically poll enumerateDevices() method to check if
                     // list of media devices has changed.
-                    availableDevicesPollTimer = window.setInterval(() => {
-                        this.enumerateDevices(pds => {
-                            // We don't fire RTCEvents.DEVICE_LIST_CHANGED for
-                            // the first time we call enumerateDevices().
-                            // This is the initial step.
-                            if (typeof availableDevices === 'undefined') {
```
saghul: Where is this logic now?

virtuacoplenny: availableDevices gets set about 10-15 lines above and so isn't undefined by the time this fires...I hope.

PR #813, opened 2018-09-14
Exports additional statistics through ConnectionQuality

bgrozev:


JitsiConference.js
```diff
@@ -2530,6 +2553,57 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
...
+        const analyticsKeys = [
+
+            // The number of jitsi-videobridge instances currently used for the
+            // conference.
+            'bridge-count',
+
+            // The conference creation time (set by jicofo).
+            'created-ms',
+            'octo-enabled'
+        ];
```
virtuacoplenny: You don't want to just log the whole properties object?

JitsiConference.js
```diff
@@ -2530,6 +2553,57 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
...
+
+        analyticsKeys.forEach(key => {
+            if (properties[key] && properties[key].value !== undefined) {
```
virtuacoplenny: Can a property go from defined to undefined and therefore be interesting to log?

JitsiConference.js
```diff
@@ -2530,6 +2553,57 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
...
 
+/**
+ * Called when the chat room reads a new list of properties from jicofo's
+ * presence. The properties may have changed, but they don't have to.
+ *
+ * @param {Object} properties - The properties keyed by the property name
+ * ('key').
+ * @private
+ */
+JitsiConference.prototype._propertiesChanged = function(properties) {
```
virtuacoplenny: What do you think of a name like updateProperties? It's more specific to what this does, but that might also be a problem depending on your intention.

JitsiConference.js
```diff
@@ -2530,6 +2553,57 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
...
+    const changed = isEqual(properties, this.properties);
+
+    this.properties = properties;
+    if (changed) {
+        this.eventEmitter.emit(JitsiConferenceEvents.PROPERTIES_CHANGED);
```
virtuacoplenny: Timing wise, does it make a difference that this event is fired before analytics updates itself?

JitsiConference.js
```diff
@@ -2530,6 +2553,57 @@ JitsiConference.prototype._onIceConnectionEstablished = function(
...
+};
+
+/**
+ * Gets a conference property with a given key.
+ *
+ * @param {string} key - The key.
+ * @returns {*} The value
+ */
+JitsiConference.prototype.getProperty = function(key) {
+    if (!this.properties) {
```
virtuacoplenny: Is there not always a this.properties defined? You're adding this for safety?

JitsiConference.js
virtuacoplenny: Also, I noticed for JitsiConferenceEvents.SERVER_REGION_CHANGE I noticed you also emit out the new region but here you don't emit the properties. Is there a reason for the difference?

modules/xmpp/ChatRoom.js
```diff
@@ -625,32 +624,18 @@ export default class ChatRoom extends Listenable {
...
-                        switch (key) {
-
-                        // The number of jitsi-videobridge instances currently
-                        // used for the conference.
-                        case 'bridge-count':
-
-                        // The conference creation time (set by jicofo).
-                        case 'created-ms':
-                        case 'octo-enabled':
-                            Statistics.analytics.addPermanentProperties({
```
virtuacoplenny: Nice. Getting extra responsibilities out.

JitsiConference.js
bgrozev: I'd rather not, I don't know what we'll end up putting there (right now we use only the 3 properties mentioned here)

JitsiConference.js
bgrozev: That's a good point. I can't think of why any of the existing ones would go to undefined, and I'd have to add the code to remove a "permanent property". So I'll keep it like this (which is the same as the current behaviour)

JitsiConference.js
bgrozev: It didn't seem good to emit the properties because they have some sort of structure that JitsiConference maintains. But now that I think about it, it's just a map, so I'll emit it.

JitsiConference.js
bgrozev: It's true that it is always defined with today's code, but it depends on code in another file. That is, if ChatRoom.js decides to emit an event with undefined, this.properties will be undefined. So I'd rather keep the check for safety.

modules/connectivity/ConnectionQuality.js
```diff
@@ -287,9 +287,9 @@ export default class ConnectionQuality {
 
         conference.on(
             ConferenceEvents.PROPERTIES_CHANGED,
-            () => {
+            properties => {
                 this._localStats.bridgeCount
-                    = conference.getProperty('bridge-count');
+                    = Number((properties || {})['bridge-count']);
```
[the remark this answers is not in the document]
bgrozev: Jitsi-meet handles the NaN case correctly (it doesn't display the line).

JitsiConference.js
```diff
@@ -14,13 +14,14 @@ import {
     createP2PEvent
 } from './service/statistics/AnalyticsEvents';
 import AvgRTPStatsReporter from './modules/statistics/AvgRTPStatsReporter';
-import ComponentsVersions from './modules/version/ComponentsVersions';
+import authenticateAndUpgradeRole from './authenticateAndUpgradeRole';
```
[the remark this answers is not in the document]
bgrozev: I tried to order them, but I don't know all the rules (I don't understand the file depth, and whether upper case matters).

PR #817, opened 2018-09-19
Share code amongst "Chrome-like" browsers

saghul:


saghul:

modules/RTC/RTCUtils.js
```diff
@@ -814,10 +814,7 @@ class RTCUtils extends Listenable {
 
             this.getStreamID = ({ id }) => id;
             this.getTrackID = ({ id }) => id;
-        } else if (browser.isChrome() // this is chrome < 61
-                || browser.isOpera()
-                || browser.isNWJS()
-                || browser.isElectron()
+        } else if (browser.isChromeLike() // this is chrome < 61
```
bbaldino: is this comment still accurate?

modules/browser/BrowserCapabilities.js
```diff
@@ -20,6 +20,7 @@ export default class BrowserCapabilities extends BrowserDetection {
         super();
         logger.info(
             `This appears to be ${this.getName()}, ver: ${this.getVersion()}`);
+        this.usesNewGumFlow() && logger.debug('Using the new GUM flow');
```
bbaldino: seems like an odd place to log this, do we not have a log for this when we go down the new gum path?  if not i'd rather have it there than inside this util class

modules/browser/BrowserCapabilities.js
```diff
@@ -47,6 +48,17 @@ export default class BrowserCapabilities extends BrowserDetection {
         return !this.isEdge() && !this.isFirefox();
     }
 
+    /**
+     * Checks if the current browser is "Chrome like", that is, it's either
+     * Chrome / Chromium or uses it as its engine.
+     */
+    isChromeLike() {
```
bbaldino: we have both this and https://github.com/jitsi/js-utils/pull/6 ?

modules/browser/BrowserCapabilities.js
```diff
@@ -241,11 +241,32 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean}
      */
     usesNewGumFlow() {
-        return (this.isChrome()
-                && !this.isVersionLessThan('61'))
-            || this.isFirefox()
-            || this.isSafariWithWebrtc();
+        const REQUIRED_VERSION = 61;
```
bbaldino: `REQUIRED_CHROME_VERSION`?

modules/browser/BrowserCapabilities.js
```diff
@@ -241,11 +241,32 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+            // eslint-disable-next-line no-undef
+            const version = Number.parseInt(process.versions.chromium, 10);
+
+            return version >= REQUIRED_VERSION;
+        }
+
+        if (this.isChromeLike()) {
+            const ua = navigator.userAgent;
+
+            if (ua.match(/Chrome/)) {
```
bbaldino: this also seems like overalp with https://github.com/jitsi/js-utils/pull/6 ?

modules/RTC/RTCUtils.js
saghul: Yes: we use that code for anything which uses the Chrome engine and its version is < 61.

modules/browser/BrowserCapabilities.js
saghul: Here is where we log the browser name, so I thought it was an appropriate place. I'm ok with dropping it altogether though, it was helpful when debugging.

bbaldino: ok, my vote would be to log it in the flow that decides whether or not to use the new gum flow, but won't hold up the PR on the change.  just seems a bit weird to call out this one behavior of the browser capabilities in a log message here when there are lots of different flows we make based on the version.

modules/browser/BrowserCapabilities.js
bbaldino: I guess what I don't understand is if we already detect chrome-like things as chrome, that means this check is effectively the same as:

```
 isChromeLike() {
    return this.isChrome();
}
```
no?  electron/js/opera will already be detected as chrome-like and therefore chrome, no? so what is this adding?  or maybe i'm missing something.

modules/browser/BrowserCapabilities.js
bbaldino: so we detect chrome-like things as chrome, but don't parse the chrome version but instead the actual product (electron, etc.) version?  this seems like it would get confusing/misleading, as we represent things as chrome but parse a version that isn't actually the chrome version?

saghul: No, that's not what happens. Sort of. When `isChrome` returns true, that browser is detected as Chrome, so the version is the Chrome version. This is the case for Chrome, Chromum, Brave and after my js-utils patch, Vivaldi.

Then we have Electron, NWJS and Opera for which `isChrome` returns false, and their version is the product version. These are "Chrome like" and we want to extract the Chrome version by hand. I'll add a comment which hopefully clarifies this.

Thanks for being so thorough!

saghul:
@bbaldino Updated, PTAL. I clarified this with comments as much as I could. Hopefully it's clearer now!

modules/browser/BrowserCapabilities.js
```diff
@@ -241,11 +248,40 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+
+        if (this.isChrome()) {
+            return !this.isVersionLessThan(REQUIRED_CHROME_VERSION);
+        }
+
+        if (this.isFirefox() || this.isSafariWithWebrtc()) {
+            return true;
+        }
+
+        if (this.isNWJS()) {
```
bbaldino: why is this needed as a separate case from isChromeLike?  Is it because NWJS represents the chrome version differently than others?  If so, could that be handled somewhere else so it remains more uniform with the rest of the isChromeLike tests?

bbaldino: if it is just the version thing, i wonder if it might be clearer to still have this as a case within the `if (this.isChromeLike()) {` block below?

saghul:
>  and i wonder if something like `isChromiumBased` would be clearer?

I think you're right. Thanks for the suggestion! Will fix.

modules/browser/BrowserCapabilities.js
saghul: >  Is it because NWJS represents the chrome version differently than others?

Yep, it's not in `navigator.userAgent` but in `process`.

saghul: I don't think the extra indentation level adds much IMHO. That's why I left it separate. I'll give it a shot when I rename the function and see how it feels like.

saghul:
@bbaldino Updated, PTAL!

saghul:

saghul:
Note: merged with a failing iOS build. (0 tests failed).

PR #818, opened 2018-09-20
Refactor permissions

saghul:


JitsiConferenceEventManager.js
```diff
@@ -391,52 +391,6 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
...
-            }
-            if (config.tagName === 'video') {
-                isVideoAvailable = config.value === 'true';
-            }
-        });
-
-        let availableDevices;
-
-        if (conference.myUserId() === from) {
-            availableDevices = conference.availableDevices;
```
virtuacoplenny: What a strange practice of modifying the object in place like this. It's possible `conference.availableDevices` might not be referenced anymore. I'm also unsure exactly how it differs from RTCUtils's available devices.

JitsiMediaDevices.js
```diff
@@ -54,26 +110,58 @@ const JitsiMediaDevices = {
...
-            return permissions.audio === true;
-        default:
-            return permissions.video === true && permissions.audio === true;
-        }
-    },
+        return new Promise(resolve => {
+            // Shortcut: first check if we already know the permission was
+            // granted.
+            if (type in this._grantedPermissions) {
+                resolve(this._grantedPermissions[type] === true);
```
virtuacoplenny: What other values can it be besides a boolean?

JitsiMediaDevices.js
```diff
@@ -54,26 +110,58 @@ const JitsiMediaDevices = {
...
+                default:
+                    promises.push(
+                        navigator.permissions.query({ name: 'camera' }));
+                    promises.push(
+                        navigator.permissions.query({ name: 'microphone' }));
+                }
+
+                Promise.all(promises).then(
+                    r => resolve(r.every(Boolean)),
+                    () => resolve(false)
```
virtuacoplenny: You think it's worthwhile to update the known permissions at this point? Maybe a premature optimization.

doc/API.md
```diff
@@ -88,7 +88,7 @@ JitsiMeetJS.setLogLevel(JitsiMeetJS.logLevels.ERROR);
         - groupId - group identifier, two devices have the same group identifier if they belong to the same physical device; for example a monitor with both a built-in camera and microphone
     - ```setAudioOutputDevice(deviceId)``` - sets current audio output device. ```deviceId``` - id of 'audiooutput' device from ```JitsiMeetJS.enumerateDevices()```, '' is for default device.
     - ```getAudioOutputDevice()``` - returns currently used audio output device id, '' stands for default device.
-    - ```isDevicePermissionGranted(type)``` - returns true if user granted permission to media devices. ```type``` - 'audio', 'video' or ```undefined```. In case of ```undefined``` will check if both audio and video permissions were granted.
+    - ```isDevicePermissionGranted(type)``` - returns a Promise with resolves to true if user granted permission to media devices. ```type``` - 'audio', 'video' or ```undefined```. In case of ```undefined``` will check if both audio and video permissions were granted.
```
virtuacoplenny: Maybe "which resolves to true"?

saghul: :+1: will fix.

JitsiMediaDevices.js
saghul: They will be updated as soon as streams are created, so I think having a single place where that info comes from makes the code less involved. WDYT?

JitsiMediaDevices.js
saghul: `undefined`, when we don't know yet. I'll replace this with `Boolean(XXX)` it reads better.

JitsiConferenceEventManager.js
saghul: Good catch! I'll chase it down and remove it if we no longer need it.

saghul:
@virtuacoplenny Updated, PTAL!

JitsiMediaDevices.js
virtuacoplenny: Sounds fair and right.

saghul:

saghul:
Note: merged with iOS build failure (0 failed tests).

PR #820, opened 2018-09-26
Updates API docs to include connectvityEvents

lsiv568:
As a result of a [Jitsi community discussion](https://community.jitsi.org/t/enable-local-statistics/15070/3) it was determined that `connectivityEvents` had not made their way into the docs.  

This PR updates the API documentation to include `connectivityEvents` and marks `CONNECTION_STATS` as deprecated.


lsiv568:
@jitsi-jenkins I have signed the individual icla.

PR #821, opened 2018-09-27
feat(TPC): fix handling layer suspension when RTPSender is not supported

saghul:
Just skip it in such environments, such as React Native.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2428,6 +2428,12 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function(track) {
 };
 
 const handleLayerSuspension = function(peerConnection, isSelected) {
+    if (!peerConnection.getSenders) {
+        logger.debug('Browser doesn\'t support RTPSender');
+
+        return;
+    }
+
     const videoSender = peerConnection.getSenders()
         .find(sender => sender.track.kind === 'video');
 


saghul:
@bbaldino PTAL.

I also thought of updating https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/browser/BrowserCapabilities.js#L188 but ultimately I think that check should go away and just check for the existence of `RTPSender`, or `getSenders()` in `RTCPeerConnection`'s prototype.

Right now `supportsRtpSender` is only used here: https://github.com/jitsi/lib-jitsi-meet/blob/e6648504e153c29ee350d1c5910025381ab55590/modules/RTC/TraceablePeerConnection.js#L2101 which I *think* only works on Firefox. In that case, maybe we should replace the check with `isFirefox()` now that Chrome also supports `RTPSender`?

I'm happy to make the above suggested changes if you agree.

bbaldino:
I think the reason I didn't use `supportsRtpSender` is that chrome actually had rtp sender support but didn't have the support for setting the constraints like this to suspend layers, so checking for rtp sender support alone wasn't enough.  Looking now, though, I'd be all for a `supportsLayerSuspension` check that validated the proper calls were there (though it seems the level of implementation for these sorts of things can vary quite a bit, so we may still be stuck on doing the check inside `supportsLayerSuspension` based on browser type/version rather than capabilities).  I'm also fine with this change as-is for now.

saghul:
Thanks Brian. I landed it as-is, will do more improvements in another round.

PR #822, opened 2018-10-01
Adds server-side speaker stats handling.

damencho:
When the main domain advertises component with type 'speakerstats' every participant will send message to the component with the name of the room where this happen.

### DIFF
--- a/modules/statistics/SpeakerStatsCollector.js
+++ b/modules/statistics/SpeakerStatsCollector.js
@@ -52,10 +52,10 @@ export default class SpeakerStatsCollector {
     _onDominantSpeaker(dominantSpeakerId) {
         const oldDominantSpeaker
             = this.stats.users[this.stats.dominantSpeakerId];
-        const newDominantSpaker = this.stats.users[dominantSpeakerId];
+        const newDominantSpeaker = this.stats.users[dominantSpeakerId];
 
         oldDominantSpeaker && oldDominantSpeaker.setIsDominantSpeaker(false);
-        newDominantSpaker && newDominantSpaker.setIsDominantSpeaker(true);
+        newDominantSpeaker && newDominantSpeaker.setIsDominantSpeaker(true);
         this.stats.dominantSpeakerId = dominantSpeakerId;
     }
 
--- a/modules/statistics/SpeakerStatsCollector.js
+++ b/modules/statistics/SpeakerStatsCollector.js
@@ -39,9 +39,11 @@ export default class SpeakerStatsCollector {
         conference.addEventListener(
             JitsiConferenceEvents.DISPLAY_NAME_CHANGED,
             this._onDisplayNameChange.bind(this));
-        conference.xmpp.addListener(
-            XMPPEvents.SPEAKER_STATS_RECEIVED,
-            this._updateStats.bind(this));
+        if (conference.xmpp) {
+            conference.xmpp.addListener(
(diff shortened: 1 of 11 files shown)


modules/xmpp/xmpp.js
```diff
@@ -154,21 +155,34 @@ export default class XMPP extends Listenable {
...
 
-            this.connection.ping.hasPingSupport(
-                pingJid,
-                hasPing => {
-                    if (hasPing) {
+            this.caps.getFeaturesAndIdentities(pingJid)
+                .then(({ features, identities }) => {
+                    if (features.has(Strophe.NS.PING)) {
                         this.connection.ping.startInterval(pingJid);
-                    } else {
```
hristoterezov: Shouldn't we keep the warn here instead of moving it to the catch?

damencho:
@hristoterezov Can I have a new review, I've added an update of the stats to the joining participants.

modules/xmpp/xmpp.js
```diff
@@ -553,4 +580,91 @@ export default class XMPP extends Listenable {
...
+        // no speaker stats component advertised
+        if (!this.speakerStatsComponentAddress || !roomJid) {
+            return;
+        }
+
+        const msg = $msg({ to: this.speakerStatsComponentAddress });
+
+        msg.c('speakerstats', {
+            xmlns: 'http://jitsi.org/jitmeet',
+            room: `${roomJid}` })
```
hristoterezov: Unnecessary template - you can use just `roomJid` directly.

modules/xmpp/xmpp.js
```diff
@@ -553,4 +580,91 @@ export default class XMPP extends Listenable {
...
     }
+
+    /**
+     * Notifies speaker stats component if available that we are the new
+     * dominant speaker in the conference.
+     * @param {String} roomJid - The room jid where the speaker event occurred.
+     */
+    sendDominantSpeakerEvent(roomJid) {
+        // no speaker stats component advertised
+        if (!this.speakerStatsComponentAddress || !roomJid) {
```
hristoterezov: I'm wondering if this is room specific function why it's not defined in ChatRoom.js?

damencho: Because of the speakerStatsComponentAddress which is retrieved here in xmpp.js and this is send from statistics.js where there is no room ... it can be handled in JitsiConferenceEventManager and can be made per room and in xmpp there can be a getter for speakerStatsComponentAddress if you think that is better.

modules/statistics/SpeakerStatsCollector.js
```diff
@@ -38,6 +39,9 @@ export default class SpeakerStatsCollector {
         conference.addEventListener(
             JitsiConferenceEvents.DISPLAY_NAME_CHANGED,
             this._onDisplayNameChange.bind(this));
+        conference.xmpp.addListener(
+            XMPPEvents.SPEAKER_STATS_RECEIVED,
```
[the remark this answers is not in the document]
damencho: What do you mean in conflict?

modules/xmpp/xmpp.js
damencho: Fixed.

modules/statistics/SpeakerStatsCollector.js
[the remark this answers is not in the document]
hristoterezov: I got confused. You can ignore the comment!

PR #826, opened 2018-10-04
Update to Babel 7 (needed for RN 0.57 update)

saghul:


saghul:
I disabled Android testing because it gives an OOM. That's the only test failure at the moment.

PR #829, opened 2018-10-17
cleanup: Removes the user-agent string from presence (unused).

bgrozev:


modules/xmpp/ChatRoom.spec.js
```diff
@@ -47,17 +47,17 @@ describe('ChatRoom', () => {
                 to: 'tojid',
                 from: 'fromjid'
             })
-            .c('user-agent').t('user-agent-text').up();
+            .c('element-name').t('element-name-text').up();
```
paweldomas: the 'element-name' is just an arbitrary name for the purpose of this unit test ? not something we're using, right ?

bgrozev: Yes, I renamed it to make it clear that it is generic

PR #831, opened 2018-10-23
ignore duplicate jibri status updates

bbaldino:


modules/recording/RecordingManager.js
```diff
@@ -198,6 +198,18 @@ class RecordingManager {
             logger.warn(
                 'Ignoring recording presence update',
                 'Received a new session with status off.');
+        }
```
virtuacoplenny: Does this need the early return still?

bbaldino: fixed

bbaldino:
@virtuacoplenny hold off on merging for now

PR #833, opened 2018-11-05
Removes quotes around token from the docs.

damencho:


damencho:
Wow, @aaronkvanmeerten, that was a quick one:
```
created: update-tokens-doc 24 seconds ago
aaronkvanmeerten approved these changes 19 seconds ago
```
What did you do in those 5 seconds :P

PR #835, opened 2018-11-19
feat(BworserCapabilities): Allow newer Safari to use video

mishurov:


webintoit:
It only  works with `unified plan` option disabled.
image
Nice present to NY.

mishurov:
@webintoit, here, on the screenshot, is a qemu VM with clean Mojave installation and Safari Technology Preview 70 with default settings. Well, except for Mock Capture Devices.

screenshot_2018-11-21_09-44-32

webintoit:
There is fatal jingle error when trying to use as `unified plan`
```
OperationError: Failed to set remote offer sdp: Media section has more than one track specified with a=ssrc lines which is not supported with Unified Plan.
(anonymous function)
```

mishurov:
When in BrowserCapabilities I return `true` in `usesUnifiedPlan()`, `supportsSdpSemantics()` and return `false` in `supportsDataChannels()` for Safari, it is able to create rooms and communicate with other participants but throws an error when it connects to existing rooms.

I presume that the fault is on their side, Safari doesn't add `a=rtcp-mux` to `m=application 9 DTLS/SCTP 5000` for unified plan, without `rtcp-mux` sdp-interop can't convert unified plan to plan b. I probably try to analyse WebKit's sources to be sure.

mishurov:
Unified Plan works more or less on both sides. The same laptop runs Jitsi and Jibri servers in Docker, a window manager and a VM with macOS, everything is incredibly slow. It's hard to debug subtleties.

screenshot_2018-11-22_23-55-34

saghul:
Folks, please let's got get entangled here. VP8 support is not related to Unified Plan support, that's a totally different story.

saghul:
@mishurov Please can you keep this PR about VP8 only and move the unified plan related commits to another PR?

mishurov:
That's the problem, it's interconnected. Safari now sets Unified Plan by default, it can be switched to Plan B only via the normally unexposed developer menu, which isn't nice for a regular user.

The current state of the diff of the PR is a semi-working Safari in the Unified Plan mode. It's usable but there're issues with audio levels and errors into JS console about SDP and m-lines. As I can understand at the moment, it's related to sdp-interop and its caching. Also according to Safari's SDPs, it supports rtx in ssrc groups, it's turned off in BrowserCapabilities due to, again, problems with plan-b to unified plan conversion.

saghul:
I see. First, let me thank you for the work here. This is non-trivial stuff (as you have noticed). ❤️ 

Now, it wouldn't be the first time Safari TP has something on by default, but later ships with it off. So, given it's still in TP, we can do VP8 first, and once that's out of the way, work on the remaining issues. RTX, for example, is not a hard requirement, we can do unified, then RTX and so on. Breaking these changes into smaller chunks will really help us get through them quicker.

mishurov:
You're welcome, there're actually people using Jibri who demanded working Safari. Well, I'll make it work with plan-b and remove intermediate commits by rebasing.

Since both Chrome and Safari now work in two modes, I think that BrowserCapabilities should infer semantics as well, by parsing generated SDP because Safari doesn't tell it through connection configuration. It'll give the ability to reject incompatible modes or configure connection accordingly.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1155,6 +1155,33 @@ const enforceSendRecv = function(localDescription) {
     return localDescription;
 };
 
+/**
+ * Neither Safari nor Chromium adds rtcp-mux to an application channel
+ * The function makes sure that it is set to pass sdp-interop validation
+ * @param {Object} localDescription the SDP object as defined by WebRTC.
+ */
+const injectRtcpMux = function(localDescription) {
```
bbaldino: i think this would be better in `SdpTransformUtil.js`

bbaldino: also, did this code already exist somewhere?  if not, how were things working?  electron uses chromium, so i would think it would've had this issue.  or is this only in the unified case? (in which case, i guess firefox does this correctly on its own?)

mishurov: Things are working because Chromium uses plan-b for now and it doesn't need sdp-interloop conversion, it's sdp-interloop that doesn't validate an SDP. There was [an issue for Firefox about the data channel and rtcp-mux](https://github.com/jitsi/sdp-interop/issues/13 ) on sdp-interop's GitHub, Firefox seems not to set rtcp-mux in some cases as well.

I anticipate that sdp-interloop is going to endure a lot of modifications as the question inevitably comes to Chrome's unified-plan, they are to drop plan-b next year or may be bit later.

Doesn't matter anyway. This piece of code will be removed because we decided to make Safari work in plan-b mode first, I'll update the diff soon.

mishurov:
I limited the Safari support to plan-b and squashed everything in the PR into one commit.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1881,7 +1882,7 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(description) {
 
     // Safari WebRTC errors when no supported video codec is found in the offer.
     // To prevent the error, inject H264 into the video mLine.
-    if (browser.isSafariWithWebrtc()) {
+    if (browser.isSafariWithWebrtc() && !browser.isSafariWithVP8()) {
```
bbaldino: but doesn't the new safari still have 264?

bbaldino: ok, re-read the comment again and i misunderstood.  but maybe this should still be a bit different to make sure we insert something that the current safari supports?

mishurov: Chrome codecs
```
a=rtpmap:96 VP8/90000
a=rtpmap:97 rtx/90000
a=rtpmap:98 red/90000
```

Firefox codecs
```
a=rtpmap:120 VP8/90000
a=rtpmap:121 VP9/90000
a=rtpmap:126 H264/90000
```

Safari codecs
```
a=rtpmap:99 rtx/90000
a=rtpmap:96 H264/90000
a=rtpmap:100 VP8/90000
```

They at least will be using VP8

mishurov:
One more update after testing triple connection: Chromium, Safari and Firefox. I also renamed remove stream and replace methods as they're more related to RTPSender support rather than particularly Firefox.

modules/browser/BrowserCapabilities.js
```diff
@@ -109,7 +121,7 @@ export default class BrowserCapabilities extends BrowserDetection {
         // FIXME bandwidth stats are currently not implemented for FF on our
         // side, but not sure if not possible ?
         return !this.isFirefox() && !this.isEdge()
-            && !this.isSafariWithWebrtc();
+            && !(this.isSafariWithWebrtc() && !this.isSafariWithVP8());
```
saghul: Did safari 12.1 also implement proper BW stats? If so, I don't think the VP8 check here is correct. They may both rely on the version being 12.1, but the semantics are different. Can you please clarify this?

modules/browser/BrowserCapabilities.js
```diff
@@ -241,6 +255,39 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+     * Checks if the browser uses unified plan by default.
+     *
+     * @returns {boolean}
+     */
+    isUnifiedPlanDefault() {
+        if (this.isFirefox()) {
+            return true;
+        }
+
+        if (this.supportsSdpSemantics()) {
```
saghul: This branch is not currently run, and the rought plan moving forward is to set SDP semantics to Plan B until we fully migrate to Unified, so I don't think we'll ever need this check for Chrome.

modules/browser/BrowserCapabilities.js
```diff
@@ -241,6 +255,39 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+            // semantics is set explicitly in RTC.js
+            const pc = new RTCPeerConnection();
+            const semantics = pc.getConfiguration().sdpSemantics;
+
+            pc.close();
+
+            return semantics === 'unified-plan';
+        }
+
+        if (this.isSafariWithVP8()) {
```
saghul: Same as above, semantically using VP8 support as the condition here is very counter intuitive. Also, This check should be part of the `usesUnifiedPlan` function.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1392,8 +1392,8 @@ TraceablePeerConnection.prototype._addStream = function(mediaStream) {
  * @param {MediaStream} mediaStream
  */
 TraceablePeerConnection.prototype._removeStream = function(mediaStream) {
-    if (browser.isFirefox()) {
-        this._handleFirefoxRemoveStream(mediaStream);
+    if (browser.supportsRtpSender()) {
```
bbaldino: thanks for renaming these, much better

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2069,7 +2069,7 @@ TraceablePeerConnection.prototype.close = function() {
  * @private
  */
 const _fixAnswerRFC4145Setup = function(offer, answer) {
-    if (!browser.isChrome()) {
+    if (!browser.isChrome() && !browser.isSafariWithVP8()) {
```
bbaldino: tests like this make me wonder if we should name this check something else.  it isn't so much that safari has vp8 here, it's that it's a safari version with more modern support.

mishurov: I agree but I don't know how to name it.

modules/RTC/TraceablePeerConnection.js
mishurov: There's also a function which is more related to sdp-interloop caching for unified plan than to Firefox, I'll rename that too.

modules/browser/BrowserCapabilities.js
mishurov: I've made some fixes for sdp-interop, Safari works in both modes with those changes.
https://github.com/jitsi/sdp-interop/pull/28
I think it's more reasonable to use `usesUnifiedPlan`.

modules/browser/BrowserCapabilities.js
mishurov: I'll remove that. I've just wanted to make a generic function.

modules/browser/BrowserCapabilities.js
mishurov: I believe it does. They've made a hell of a lot of changes in WebKit's included webrtc library
https://trac.webkit.org/changeset/237075/webkit

saghul: Cool. Then please check if the browser is safari and if the version is 12.1, since BW stats are really not related to VP8 support, they just happened to both be released in 12.1

mishurov:
Lads, I tend to agree that changes to sdp interop are a half measure. I hate that solution as well to be honest. Let's make Safari users happy with Plan B and make more fundamental solutions later.

mishurov:
https://webkit.org/blog/8672/on-the-road-to-webrtc-1-0-including-vp8/

>Unified Plan is an experimental feature that is currently turned on by default in Safari Technology Preview and turned off by default in Safari in iOS 12.2 and macOS 10.14.4 betas.

modules/browser/BrowserCapabilities.js
webintoit: Implements, but no proper, it needs to be adapted. Lets do it in other PR.

modules/browser/BrowserCapabilities.js
saghul: Let's drop this now, since Safari won't implement it and it's Chrome only but we already pass the plan-b constraint.

mishurov:
I removed Safari from the bandwidth stats, removed the default plan method, rebased and merged the latest master.

I signed the CLA long ago, GitHub required to do that the moment I created the PR.

saghul:
Jenkins ok to test.

webintoit:
@mishurov, great, thanks! 
Should we enable simulcast for safari ? When intercept safari user-agent to `google-chrome`, simulcast  works fine (but stats throws incompatibility errors). Maybe  it is better to do it in other iteration.

saghul:
I was under the impression that simulcast only worked when Safari is in Unified plan. I'll give it a quick test. That would be awesome of course!

mishurov:
@webintoit, you seem right, according to https://webkit.org/blog/8672/on-the-road-to-webrtc-1-0-including-vp8/

> simulcast is now supported for both H.264 and VP8

I thought it was only for h.264.

And further:
> Each individual encoding can be controlled using RTCRtpSender.setParameters on the sender side. Simulcast is currently activated using SDP munging.

As I understand, that is exactly what the code in TraceablePeerConnection.js does. So, in theory, it would be sufficient just to turn the supportsSimulcast() switch on for Safari 12.1+ in BrowserCapabilities.js

saghul:
@mishurov I'm afgrad I need to ask you to sign the CLA again. When you signed it were in the process of moving from Atlassian to 8x8, so the CLA you signed is not valid. Here is the new CLA: https://jitsi.org/icla (same link, redirects elsewhere now). Sorry for the trouble.

mishurov:
@saghul, no problem. What about the simulcast, should I include it in the commit?

saghul:
> no problem. What about the simulcast, should I include it in the commit?

Thank you! Yes please, add simulcast support, I just tested it and it works fine.

webintoit:
image
Unfortunately. Simulcast needs to be adapted too.

saghul:
@webintoit were you running on plan B or unified? I had no problems with plan B.

webintoit:
> @webintoit were you running on plan B or unified? I had no problems with plan B.

Plan B.

saghul:
I'm running the same. Safari TP or 12.1 release?

webintoit:
> I'm running the same. Safari TP or 12.1 release?

Release one. Trying to clear caches. Will report

webintoit:
image
I think it is because of adapter.js, will play with it

saghul:
I got that last error when enabling unified plan. The sdp-interop PR is supposed to fix that.

webintoit:
image

mishurov:
I can't reproduce the bug in my Mojave VM with the latest updates and Safari 12.1 release. Plan B, patched sdp-interop.

webintoit:
> I can't reproduce the bug in my Mojave VM with the latest updates and Safari 12.1 release. Plan B, patched sdp-interop.

Oh, i did not patched sdp-interop yet. 
This line causes the error and jingle fails for me. Everything is Ok when i intercept user-agent to `chrome`. 
https://github.com/jitsi/lib-jitsi-meet/blob/efa099a2890f9b090422f734ab1509ffbb328285/modules/RTC/TraceablePeerConnection.js#L2166

My environment is same. 
image

Will try without rtpSender.

webintoit:
When we are creating the tracks after join, simulcast causes the crashes, but
It works in both cases: 
  - safe check for `sender.track` property when `browser.supportsRTPSender()` enabled. 
   - disable `browser.supportsRTPSender`

image

saghul:
@webintoit Thanks for digging into this. I think we need to add the checks, since pretty much every browser supports RTPSender now.

mishurov:
Yet Webkit's blog explicitly stated that encoding could be controlled via RTCRtpSender.setParameters which is being done in prototype.createAnswer. So the problem may be related to adapter.js as @webintoit suggested earlier.

webintoit:
>  since pretty much every browser supports RTPSender now.

Chrome with enabled RTPSender is also working  fine, except
image

saghul:
I think there are 2 issues at play here. Part of your changes use RTPSender also on Safari, but when using plan B that's not needed IIRC, since simulcast is enabled by munging.

mishurov:
I've traced getSenders() in Safari, it returns senders whose tracks may be null. I've added an additional check and rebased again.

saghul:
@mishurov Excellent, thanks! I'm updating to macOS 10.14.4 so I'll give it another round of testing with Safari 12.1 and Safari 12.2 (TP) with and without the unified plan flag enabled.

Just to confirm, the sdp-interop patch is only needed for unified, right?

mishurov:
@saghul, quite right, I removed several bugs during conversion from unified plan to plan b on Safari and Chrome yet, as I remember, there still were several troubles, I inferred that it wasn't related to sdp-interop.

At the moment we were deploying Jitsi and some people badly wanted it to work with Safari and it wasn't clear whether Apple would make Plan-B default in the release.

I haven't tested but I suppose, the current sdp-unterop shall work with Plan-B without errors simply because the conversion is being performed only in Firefox in that setting.

saghul:
The sdp-interop PR landed, thanks! Can you please update the depenency as part of this PR?

mishurov:
Done

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2163,7 +2163,7 @@ TraceablePeerConnection.prototype.createAnswer = function(constraints) {
     if (browser.supportsRtpSender() && this.isSimulcastOn()) {
         const videoSender
             = this.peerconnection.getSenders().find(sender =>
-                sender.track.kind === 'video');
+                sender.track !== null && sender.track.kind === 'video');
```
saghul: I'm still getting this error when using unified:

`TypeError: undefined is not an object (evaluating 'this.peerconnection.getSenders().find(function(e){return null!==e.track&&"video"===e.track.kind}).setParameters'`

I think we need to check for falseyness, not null.

saghul: Hum, I tried a couple of things locally and I never get video when in unified plan.

mishurov: I'm trying to reproduce the error right now. I disabled video support in BrowserCapabilites for unified plan in Safari because even with the new sdp-interop Safari doesn't work flawlessly.

webintoit: We need additional check here. 
```js
// videosender is absent for `startmuted` here
const videosender = this.peerconnection.getSenders().find(sender => {
     return sender.track && sender.track.kind === 'video';
})

//... undefined here
videosender && videosender.setParameters({
    ... simulcastParams,
    transactionId: "" //have no idea what value should be here, but safari requires it
})
```

saghul: Yeah, sorry for the noise. Let's ignore video on unified for the purposes of this PR.

This is the proper fix for that error:

```
diff --git a/modules/RTC/TraceablePeerConnection.js b/modules/RTC/TraceablePeerConnection.js
index e6bef13..3c6c87e 100644
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2160,10 +2160,12 @@ const _fixAnswerRFC4145Setup = function(offer, answer) {
 };
 
 TraceablePeerConnection.prototype.createAnswer = function(constraints) {
-    if (browser.supportsRtpSender() && this.isSimulcastOn()) {
+    // TODO: Apply this on Firerox only since it's the only one which only does
+    // simulcast with RID.
+    if (browser.isFirefox() && this.isSimulcastOn()) {
         const videoSender
             = this.peerconnection.getSenders().find(sender =>
-                sender.track !== null && sender.track.kind === 'video');
+                sender.track.kind === 'video');
         const simParams = {
             encodings: [
                 {

```

As you can see I went back to explicitly checking for Firefox. The problem is that while other browsers support RTPSender, that particular funcion applies only if RID simulcast is used, which we only support on Firefox, and in an experimental capacity.

mishurov: I updated the diff with this fix.

saghul:
Thank you so much for your work @mishurov ! ❤️

PR #838, opened 2018-11-30
fix(RTC): Pass options allowing GUM to work with config

guusdk:
With this change, an Android device will adhere to the value defined in the 'resolution' value of config.js, something it currently does not.

The only functional change is in the first commit. The other commits refactor the solution a bit, to clean things up.


guusdk:
I think I've found the one repo where I didn't contribute to before? :)

saghul:
Change looks good, thanks @guusdk! I'll give it a test and let you know.

PR #839, opened 2018-12-03
deps: update to webpack 4

saghul:


virtuacoplenny:
The changes here look fine, and lib-jitsi-meet builds fine, but jitsi-meet has an error while building the native app. Is it a conflict with webpack in jitsi-meet?

```Unable to resolve module `lib-jitsi-meet/lib-jitsi-meet.min` from `/home/jenkins/jitsi-meet/react/features/base/lib-jitsi-meet/_.native.js`: Module `lib-jitsi-meet/lib-jitsi-meet.min` does not exist in the Haste module map```

**Comment by saghul** [comment]
Thanks, I'll take a look. I think I have an idea of what's wrong, note the change I had to make in the Makefile...

**Comment by virtuacoplenny** [comment]
Jenkins test this please.

File: webpack.config.js [hunk 1]
```diff
@@ -1,31 +1,14 @@
...
-        cache: true,
-        extractComments: true,
-        parallel: true,
-        sourceMap: true
-    }));
-}
 
 const config = {
     devtool: 'source-map',
+    mode: minimize ? 'development' : 'production',
```
hristoterezov: @saghul @virtuacoplenny Shouldn't this be exactly the opposite? I think at the moment when you pass -p we will be in development mode.

saghul: Looks like I am, indeed and idiot. You're totally right!

PR #840, opened 2018-12-04
When user leaves conference, and there are no other members, timeout …

golyo:
When user leaves conference, and there are no other members, timeout error occures.


sapkra:
Already seems to be fixed in master:
https://github.com/jitsi/lib-jitsi-meet/blob/80ba679eca27b8f50bcefee004f21d746ecf8554/modules/xmpp/ChatRoom.js#L976-L996

PR #770, opened 2018-06-21
Uses json-message extension in Chatroom.js

pvgupta24:
Relies on 
* [jitsi/jitsi#495](https://github.com/jitsi/jitsi/pull/495)
* [jitsi/jigasi#111](https://github.com/jitsi/jigasi/pull/111)

PR #842, opened 2018-12-08
Enable sending JSON messages via MUC

mmoanis:
To allow clients sending JSON messages -for examples in polls- the `sendMessage` should only add the `JITSI_MEET_MUC_TYPE` if it was not present in the message. Otherwise it override the original message type.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2292,10 +2292,15 @@ JitsiConference.prototype.sendMessage = function(
         let elementName = 'body';
 
         if (messageType === 'object') {
-            try {
+            elementName = 'json-message';
+
+            // Mark as valid JSON message if not already
+            if (!messageToSend.hasOwnProperty(JITSI_MEET_MUC_TYPE)) {
                 messageToSend[JITSI_MEET_MUC_TYPE] = '';
+            }
+
+            try {
                 messageToSend = JSON.stringify(messageToSend);
-                elementName = 'json-message';
             } catch (e) {
                 logger.error('Can not send a message, stringify failed: ', e);
 


virtuacoplenny:
Could you provide more context on why you need JITSI_MEET_MUC_TYPE to be a different value?

mmoanis:
The purpose is allow `sendMessage` to be used from libjitsi to send a json message. A scenario would be to send json messages to other participants not through the data channels.
For example if I want to send a json like 
```
{
    "type": "POLL_STARTED",
    "id": "uniqueid",
    "question": "how are you"
}
``` 

So the current implementation override the type field. Am I missing something on how to send a json message?

virtuacoplenny:
I think the confusion for me is that it already does have support for sending json. For example, If I just do `APP.conference._room.sendMessage({ test: 1 }, '2f7f8fa7')` I can send a private json message to someone. But you are right in that the type field gets overridden, for purposes that I can't even begin to remember, so I don't know if it's an oversight or intentional. @bgrozev might remember.

virtuacoplenny:
Jenkins test this please.

virtuacoplenny:
Looking at #770 more I'm pretty sure this is just a bug. Seems like initially the implementation set an empty type field outside of the passed in message, but 770 modifies the passed in message with an empty type.

mmoanis:
Yeah exactly. I can send JSON messages now but the type is overridden. I want to send some actions between clients and therefore I am using the field type to be consistent with the rest of actions.

mmoanis:
I am not sure that #770 caused this. I think it has been always the same case since #723 
What I can guess about why it was made this way, since `JITSI_MEET_MUC_TYPE` is used to know if message is a JSON message or a chat message, there was no use case of clients sending messages in JSON

virtuacoplenny:
JITSI_MEET_MUC_TYPE is used by jigasi and its value is set. Its value is parsed in subtitles/middleware in jitsi-meet. It was probably set to an empty string by default. Initially it had a separate key/value for json message and message type, but 770 started writing over the message.

virtuacoplenny:
Testing locally the change seems fine. I'll just merge it.

bgrozev:
If I remember correctly the original use of this was to differentiate between json objects being sent and text sent from users that also happens to parse as JSON (e.g. we want people to be able to send "{a: 1}" as a message in the chat). With the move to a separate XML format for JSON messages (#770) this isn't necessary anymore. Forcing the type to an empty string doesn't make sense to me (the parsing code has to handle the case of a missing type anyway). This solution seems fine to me, but we should also be able to remove the code altogether (provided parsing does actually work as expected).

PR #845, opened 2018-12-10
fix(remote-description): remove the default id of "-"

virtuacoplenny:
Starting in Chrome 71, tracks without a stream are
given the msid "-", which in plan b incorrectly
triggers an onaddstream even with a default local
stream.

Do not assume I know what I'm doing; I assume I don't know what I'm doing. This is the chromium change that led me to my proposed change: https://bugs.chromium.org/p/webrtc/issues/detail?id=9880. Also, I did not gate the changes behind a "chrome 71" check but maybe I should.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1106,7 +1106,7 @@ const normalizePlanB = function(desc) {
                     }
                 }
 
-                mLine.ssrcs = newSsrcLines;
+                mLine.ssrcs = replaceDefaultUnifiedPlanMsid(newSsrcLines);
             }
         });
     }
@@ -1119,6 +1119,27 @@ const normalizePlanB = function(desc) {
         sdp: resStr
     });
 };
+
+/**
+ * Unified plan differentiates a remote track not associated with a stream using
+ * the msid "-". The msid "-" can incorrectly trigger an onaddstream event on
+ * the peer connection, passing in a default local track. To prevent onaddstream
+ * from firing, revert back to old behavior of using " " for msid.
+ *
+ * @param {Array<Object>} ssrcLines - The ssrc lines from a remote description.
+ * @private
+ * @returns {Array<Object>} The ssrcsLines with any "-" msids replaced with " ".
+ */
+function replaceDefaultUnifiedPlanMsid(ssrcLines = []) {
+    return ssrcLines.map(ssrc => {
+        if (ssrc.value && ssrc.value.startsWith('-')) {
+            ssrc.value
+                = ssrc.value.substring(1, ssrc.value.length);
+        }
+
+        return ssrc;
+    });
+}
 
 /**
  * Makes sure that both audio and video directions are configured as 'sendrecv'.


virtuacoplenny:
I have the changes locally. In chrome, when converting firefox's remote description to plan-b, I'm searching for an ssrc line with the "-" msid. If I find it, I drop all of the lines for that ssrc, change the cname line to be something like `a=ssrc:3154498619 cname:recv-only-1`, add add the cname line back.

It does resolve the issue where chrome gets an onaddstream event with its own default track when video muting itself. However, I noticed that it does not really matter what I set as the cname. For example, I set it as "asdfrecv-only-1" and the issue is still gone. Any thoughts on that?

Testing on older firefox has proven to be trickier than expected. Firefox 50 works but 45 will not without local changes due to adapterjs being brought in over a year ago.

saghul:
@gpolitis Don't we still use reconly SSRCs when the local user has no video from the start? (#config.startWithAudioMuted=true)

@virtuacoplenny no need to look FF < 52, we don't support it: https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/browser/capabilities.json#L44

bbaldino:
> Unfortunately we don't have test cases for any of this, and all of this is pretty complex, largely undocumented and pretty fragile, but, as I mentioned in our offline discussion, a safer approach may be to replace "-" lines that look like this
> 
> a=ssrc:3154498619 msid:- 0I/xJ6Le
> a=ssrc:3154498619 mslabel:-
> a=ssrc:3154498619 label:0I/xJ6Le
> 
> To something that looks like this
> 
> a=ssrc:3154498619 cname:recv-only-1
> 
> Now, the question is, what should be the cname.. This functionality existed in older Firefox versions, so we should probably find an old version and check what we were doing there.

Hey @gpolitis, if these are recvonly streams the far side doesn't need to know about, I'm wondering: should Jicofo be the entity dropping these before building the jingle offer for the far side?

gpolitis:
It would be nice if we were you able to check what the cnames of recv-only ssrcs look like in older FFs, just to make sure we preserve the existing behavior and avoid breaking something. I remember they were not all the same fixed string, they were somehow dynamic.

gpolitis:
> Hey @gpolitis, if these are recvonly streams the far side doesn't need to know about

@bbaldino afaik --it may have changed recently, so I may be wrong-- the far side needs to know about the recv-only ssrcs in order to accept RTCP feedback from the recv-only participant. We need this logic at least in the p2p case; we can get away with it in routed calls because of RTCP termination.

gpolitis:
> @gpolitis Don't we still use reconly SSRCs when the local user has no video from the start? (#config.startWithAudioMuted=true)

Yes, you're right about that Saul, we do.

gpolitis:
jenkins test this please

gpolitis:
jenkins test this please

virtuacoplenny:
@gpolitis, tests won't pass until https://github.com/jitsi/jitsi-meet-torture/pull/219 (or https://github.com/jitsi/jitsi-meet-torture/pull/218 and https://github.com/jitsi/jitsi-meet-torture/pull/217), and the tcp test failures are looked at.

damencho:
@gpolitis and @virtuacoplenny the tests pass now.

damencho:
But firefox tests are disabled :( Do you want this PR tested with chrome <-> firefox before merging?

virtuacoplenny:
Yes, I would prefer this PR be tested with Firefox/Chrome before merging.

damencho:
jenkins test this please

PR #851, opened 2019-01-02
feat(ScreenObtainer): fix support for old getDisplayMedia

saghul:
Chrome 71 does not support inline installs, but has navigator.getDisplayMedia
behind a flag. Support this case.

### DIFF
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -282,8 +282,8 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean} {@code true} if the browser supposrts getDisplayMedia.
      */
     supportsGetDisplayMedia() {
-        return navigator.mediaDevices
-            && navigator.mediaDevices.getDisplayMedia !== undefined;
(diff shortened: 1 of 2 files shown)


modules/browser/BrowserCapabilities.js
```diff
@@ -282,8 +282,8 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean} {@code true} if the browser supposrts getDisplayMedia.
      */
     supportsGetDisplayMedia() {
-        return navigator.mediaDevices
-            && navigator.mediaDevices.getDisplayMedia !== undefined;
+        return typeof navigator.getDisplayMedia !== 'undefined'
+            || typeof navigator.mediaDevices.getDisplayMedia !== 'undefined';
     }
```
damencho: @saghul Should we check here whether navigator.mediaDevices is undefined? We are checking it in supportsDeviceChangeEvent.

saghul: I don't this it's needed at this point, since all browsers we support have it for sure.

PR #853, opened 2019-01-07
ref: open WebRTC DataChannel

paweldomas:
DEPENDS ON: https://github.com/jitsi/jitsi-videobridge/pull/723

Changes the behavior to actively open new WebRTC Data Channel instead of waiting for the JVB to do it.
Adds ICE_RESTART_SUCCESS event used to re-initialize the data channel in case of ICE restart where the conference could have been moved to another bridge.

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -74,6 +74,17 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
         // else: there are no DataChannels in P2P session (at least for now)
     });
 
+    chatRoom.addListener(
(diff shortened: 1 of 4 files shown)


bbaldino:
will this always be the case (client always opens datachannel instead of the bridge)? or will it be a runtime thing that will depend on some value (like `isInitiator` or something?) .  i just did a bunch of work here in the bridge refactor and was assuming we always opened on the bridge side, though i have the code for the client opening it, just not using it right now.

paweldomas:
In this Pr it's always the case. The `isInitiator` affects other things like ICE controlling side and DTLS sever/client role, so I don't want to touch it.

PR #857, opened 2019-01-18
feat(screenshare): support remote wireless screenshare

virtuacoplenny:
- ProxyConnectionService is exposed and meant to be the
  facade for creating and updating a peer connection with
  another peer, outside of the muc.
- ProxyConnectionPC wraps JingleSessionPC so the peer
  connection handling can be reused.

### DIFF
--- a/modules/proxyconnection/ProxyConnectionPC.js
+++ b/modules/proxyconnection/ProxyConnectionPC.js
@@ -5,6 +5,7 @@ import RTCEvents from '../../service/RTC/RTCEvents';
 import XMPPEvents from '../../service/xmpp/XMPPEvents';
 
 import JingleSessionPC from '../xmpp/JingleSessionPC';
+import { DEFAULT_STUN_SERVERS } from '../xmpp/xmpp';
 
 import { ACTIONS } from './constants';
 
@@ -22,7 +23,9 @@ export default class ProxyConnectionPC {
      * Initializes a new {@code ProxyConnectionPC} instance.
      *
      * @param {Object} options - Values to initialize the instance with.
-     * @param {boolean} options.isInitiator - If true, the local client should
+     * @param {Object} [options.iceConfig] - The {@code RTCConfiguration} to use
+     * for the peer connection.
+     * @param {boolean} [options.isInitiator] - If true, the local client should
      * send offers. If false, the local client should send answers. Defaults to
      * false.
      * @param {Function} options.onRemoteStream - Callback to invoke when a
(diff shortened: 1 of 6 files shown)


modules/proxyconnection/ProxyConnectionPC.js
```diff
@@ -0,0 +1,392 @@
...
+            this._onRemoteStream
+        );
+
+        const peerConnection = new JingleSessionPC(
+            undefined, // sid
+            undefined, // localJid
+            this._options.peerJid, // remoteJid
+            connectionStub, // connection
+            {
+                offerToReceiveAudio: false,
```
paweldomas: I know this is tailored to our specific needs, but how about adding `receiveAudio` as an option as well ? I was thinking it could be more usable for something else this way.

modules/proxyconnection/ProxyConnectionService.js
```diff
@@ -0,0 +1,301 @@
...
+        // Grab the webrtc media stream and pipe it through the same processing
+        // that would occur for a locally obtained media stream.
+        const mediaStream = jitsiRemoteTrack.getOriginalStream();
+        const jitsiLocalTracks = RTC.newCreateLocalTracks(
+            [
+                {
+                    deviceId:
+                        `proxy:${this._peerConnection.getPeerJid()}`,
+                    mediaType: 'video',
+                    videoType: 'desktop',
```
paweldomas: So this looks very specific to our case given that the video type is hardcoded as 'desktop'. Then you can probably disregard my other comment about `receiveAudio`.

modules/proxyconnection/ProxyConnectionPC.js
```diff
@@ -0,0 +1,392 @@
...
+         * However, {@code ProxyConnectionPC} always assume a peer-to-peer
+         * connection so the ice configuration is hard-coded with defaults.
+         *
+         * @type {Object}
+         */
+        const iceConfigStub = {
+            p2p: {
+                iceServers: [
+                    {
+                        urls: 'stun:stun.l.google.com:19302'
```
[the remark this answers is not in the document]
virtuacoplenny: I made an exported constant to reduce the duplication. Let me know if that's acceptable. spot doesn't load the jitsi-meet config file so the exported constant was my compromise. I did let the config be settable through the constructor though.

modules/proxyconnection/ProxyConnectionService.js
virtuacoplenny: This proxy connection is definitely tailored for jitsi spot use case only. I'd be surprised, and a little afraid, if people really started using this proxy. That said, I tried making these fields more configurable.

PR #858, opened 2019-01-18
Fix: Old JitsiConference instance remains listening on incoming calls…

takarogabor:
… after leave - rejoin to same room

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -110,6 +110,7 @@ export default function JitsiConference(options) {
         throw new Error(errmsg);
     }
     this.eventEmitter = new EventEmitter();
+    this.incomingListener = this.onIncomingCall.bind(this);
     this.options = options;
     this.eventManager = new JitsiConferenceEventManager(this);
     this.participants = {};
@@ -462,6 +463,8 @@ JitsiConference.prototype.leave = function() {
         room.removeListener(
             XMPPEvents.CONFERENCE_PROPERTIES_CHANGED,
             this._updateProperties);
+
+        this.eventManager.removeXMPPListeners();
 
         this.room = null;
 
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -530,6 +530,17 @@ JitsiConferenceEventManager.prototype.setupRTCListeners = function() {
 };
 
 /**
+ * Removes event listeners related to conference.xmpp
+ */
+JitsiConferenceEventManager.prototype.removeXMPPListeners = function() {
+    const conference = this.conference;
+
+    conference.xmpp.removeListener(
+        XMPPEvents.CALL_INCOMING,
+        conference.incomingListener);
+};
+
+/**
  * Setups event listeners related to conference.xmpp
  */
 JitsiConferenceEventManager.prototype.setupXMPPListeners = function() {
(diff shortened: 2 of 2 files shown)


JitsiConference.js
```diff
@@ -463,6 +464,8 @@ JitsiConference.prototype.leave = function() {
             XMPPEvents.CONFERENCE_PROPERTIES_CHANGED,
             this._updateProperties);
 
+        this.eventManager.removeXMPPListeners();
```
saghul: `setupXMPPListeners` adds many more listeners than this function removes. I'd say it should undo everything `setupXMPPListeners` did.

saghul:
Jenkinsplease test this please.

saghul:

saghul:
@hristoterezov can you PTAL? @takarogabor This looks pretty good, thanks!

saghul:
Maybe @virtuacoplenny ?

JitsiConferenceEventManager.js
```diff
@@ -529,38 +530,64 @@ JitsiConferenceEventManager.prototype.setupRTCListeners = function() {
         });
 };
 
+/**
+ * Removes event listeners related to conference.xmpp
+ */
+JitsiConferenceEventManager.prototype.removeXMPPListeners = function() {
+    const conference = this.conference;
```
virtuacoplenny: Why do you need to use this intermediary variable? It's for convenience?

JitsiConferenceEventManager.js
```diff
@@ -586,6 +613,15 @@ JitsiConferenceEventManager.prototype.setupXMPPListeners = function() {
         });
 };
 
+/**
+ * Add XMPP listener and save its reference for remove on leave conference.
+ */
+JitsiConferenceEventManager.prototype._addConferenceXMPPListener = function(
+        eventName, listener) {
+    this.xmppListeners[eventName] = listener;
+    this.conference.xmpp.addListener(eventName, listener);
```
virtuacoplenny: Have you considered or looked into leveraging eventEmitter's removeAllListeners instead of maintaining an object of listeners? Not sure how safe that is though. The "Listenable" class is essentially a wrapped EventEmitter and adding a hook into removeAllListeners seems desirable. I'm sure there are a bunch more leaks in lib-jitsi-meet, as jitsi-meet relies on page reload after a meeting ends, and maybe having that convenience method would be nice, but I haven't looked into the edge cases.

JitsiConferenceEventManager.js
takarogabor: Yes, just for convenience. Basically I've copied from setupXMPPListeners.

JitsiConferenceEventManager.js
takarogabor: I was thinking to use removeAllListeners but I have to remove listeners by conference not by event type. I plan to join multiple rooms/conferences same time (only one with audio/video streams other rooms just for sending messages) and removeAllListeners would remove all conferences listener.

virtuacoplenny: I see. So each conference initializes its own JitsiConferenceEventManager and its own eventEmitter, but listeners get set on the xmppConnection itself, and the xmppConnection gets shared by different conferences.

PR #860, opened 2019-01-23
Polls/Voting Feature in lib-jitsi-meet

mmoanis:
Related to [#3827](https://github.com/jitsi/jitsi-meet/pull/3827)

Implement API in lib-jitsi-meet for managing a poll inside the conference room. API is not documented for now.

The API export 3 events to conference that represent:
1. Poll started
2. Someone voted in a poll
3. Poll ended

Additionally, the poll module keeps track of the current poll by saving its state to a prosody module and listens to messages from the module if for example a user just joined a conference room to get notified about the poll.

### DIFF
--- a/modules/polls/polls.js
+++ b/modules/polls/polls.js
@@ -53,6 +53,8 @@ export default class Polls {
      */
     startPoll(roomjid, poll, choices, question) {
         if (!poll || !choices || !question) {
+            logger.error('startPoll called with invalid arguments.');
+
             return;
         }
 
@@ -62,8 +64,6 @@ export default class Polls {
 
             return;
         }
-
-        logger.log(`Poll Start: ${poll} ${choices} ${question}`);
 
         this._doStartPoll(choices, poll, question);
 
@@ -88,14 +88,14 @@ export default class Polls {
      */
     voteInPoll(roomjid, choiceID) {
         if (!choiceID) {
+            logger.error('voteInPoll called with invalid choiceID');
+
             return;
         }
 
-        logger.log(`Voted for ${choiceID}`);
-
         const myid = this.conference.myUserId();
 
-        this._doVote(choiceID, myid);
+        this._toggleVote(choiceID, myid);
 
         // inform others about my vote.
         this.conference.sendMessage({
@@ -133,8 +133,6 @@ export default class Polls {
      * Emit end poll event.
      */
     _doEndPoll() {
-        logger.log('Poll has ended');
-
         this.choices = null;
         this.poll = null;
         this.question = null;
@@ -165,11 +163,11 @@ export default class Polls {
     }
 
     /**
-     * Emit user voted event.
+     * Toggle vote of a user to a given choice. Emits user voted event.
      * @param {string} choiceID - Choice ID.
      * @param {string} userID - ID of user who voted.
      */
-    _doVote(choiceID, userID) {
+    _toggleVote(choiceID, userID) {
         if (!this.choices.hasOwnProperty(choiceID)) {
             logger.warn(`Ignoring an invalid choice ID ${choiceID}`);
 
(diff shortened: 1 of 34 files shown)


mmoanis:
CLA signed for 8x8

modules/polls/polls.js
```diff
@@ -0,0 +1,251 @@
...
+const MESSAGE_POLL_STARTED = 'poll-started';
+const MESSAGE_POLL_VOTE = 'poll-vote';
+
+
+/**
+ * A module for managing poll session related functionality.
+ * This module handles communication with all participants regarding the poll
+ * session management. It also keep the record of what poll is going
+ * on using the server side prosody-module if it is present.
+ */
```
virtuacoplenny: I really appreciate how you tried to encapsulate polls.

modules/polls/polls.js
```diff
@@ -0,0 +1,251 @@
...
+     * @param {string} roomJid - Room JID
+     * @param {Object} poll - Object containing the information about the poll.
+     * @param {Object} choices - Object containing poll choices by their ID.
+     * @param {Object} question - Object with information about
+     * the question asked.
+     * @returns {void}
+     */
+    startPoll(roomjid, poll, choices, question) {
+        if (!poll || !choices || !question) {
+            return;
```
virtuacoplenny: Should an error be logged here?

modules/polls/polls.js
```diff
@@ -0,0 +1,251 @@
...
+        }
+
+        if (this.poll && this.choices && this.question) {
+            logger.warn('Attempt to create a poll '
+                + 'while one is currently active. Please call endPoll first');
+
+            return;
+        }
+
+        logger.log(`Poll Start: ${poll} ${choices} ${question}`);
```
virtuacoplenny: I'm uncertain how I feel about logging potentially private details.

modules/polls/polls.js
```diff
@@ -0,0 +1,251 @@
...
+            question
+        );
+    }
+
+    /**
+     * Emit user voted event.
+     * @param {string} choiceID - Choice ID.
+     * @param {string} userID - ID of user who voted.
+     */
+    _doVote(choiceID, userID) {
```
virtuacoplenny: Maybe this should be a toggleVote? I don't feel strongly about it though.

mmoanis: Yeah, its better

modules/polls/polls.js
mmoanis: fixed

modules/polls/polls.js
mmoanis: speaking of such, currently assigning IDs for example happens actually in the UI. I think it should be moved down here, right? Is there a way to generate unique IDs in lib-jitsi-meet. I didn't want to add a new dependency package.

mmoanis:
Fixed review about logging and function names.

I would like to ask about the possibility of generating unique IDs inside lib-jitsi-meet because I think building the data structure that represent the poll should be here not in jitsi-meet.

modules/polls/polls.js
virtuacoplenny: What about prosody? Could that store the source of truth for all polls in the muc? It has uuid capabilities. Boris will also be reviewing your PRs and he's more knowledge about that side of the world.

virtuacoplenny:
Jenkins test this please.

modules/polls/polls.js
mmoanis: Yeah currently prosody module store that but it only stores it and tell people when they join about it.

kneza23:
Any progress on this?

please-wait:
I believe the change is tied to this in Jitsi-meet: https://github.com/jitsi/jitsi-meet/pull/3827

It would be a very cool feature to have.

Aaryan-kapur:
Was this integrated @mmoanis

saghul:
https://github.com/jitsi/jitsi-meet/pull/3827#issuecomment-698333629

PR #861, opened 2019-01-27
feat(AnalyticsAdapter): Add reset functionality.

hristoterezov:


modules/statistics/AnalyticsAdapter.js
```diff
@@ -101,6 +101,24 @@ class AnalyticsAdapter {
         });
     }
 
+    /**
+     * Reset the state to the initial one.
+     *
+     * @returns {void}
+     */
+    reset() {
```
saghul: The constructor now duplicates this. How about we just call reset in the constructor then?

PR #863, opened 2019-01-30
fix(presence): emit display name change when cleared

virtuacoplenny:
Currently when a participant removes the local display name, other clients are not updating their UIs because displayName in the nickname presence update is undefined.


damencho:
Oh .. there was something about clearing the displayname ... but I do not rember what was it ... I was purposing the same change ... but it was two years ago ... maybe we had problems in the UI with clearing it ...

PR #865, opened 2019-01-31
fix(ie11): lazy-set the permission check promise

virtuacoplenny:
JitsiMediaDevices is a class that gets initialized
on export. The constructor creates a Promise, but
Promise is not supported on older browsers (IE11).
To avoid breaking IE11, initialize and cache the
promise later.

This PR brings up the question again of how to handle IE11 breaking due to the lib using newer syntax that IE11 cannot support. The lib gets loaded first through a script tag and then jitsi-meet loads with its polyfills. Each past breakage has kicked the can down the road, but the can stopped being kicked once it was assumed any breakages in IE11 did not matter but I guess that's not the case because the unsupported browser warning still has to display.

### DIFF
--- a/JitsiMediaDevices.js
+++ b/JitsiMediaDevices.js
@@ -19,6 +19,7 @@ class JitsiMediaDevices {
     constructor() {
         this._eventEmitter = new EventEmitter();
         this._grantedPermissions = {};
+        this._permissionsApiSupported = null;
 
         RTC.addListener(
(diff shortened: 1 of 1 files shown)


virtuacoplenny:
Trying out a polyfills bundle instead. https://github.com/jitsi/jitsi-meet/pull/3853. I really don't want to discourage using es6 in singleton initialization.

PR #867, opened 2019-02-06
Fixed exporting the DTMF manager

NNesh:
Fixed: "Uncaught TypeError: y.a is not a constructor" (JitsiDTMFManager is undefined)


saghul:
@NNesh has you signed the CLA?

NNesh:
@saghul No, I haven't! I don't want to sign the CLA due to one line fix. You may to fix this if you want. I just didn't know about signing the CLA. :smile: Thanks for your attention :relaxed:

saghul:
Sorry, we cannot take your patch then.

PR #876, opened 2019-02-22
feat(ChatRoom): add 'disableFocus' option

paweldomas:
Adds a flag which can be used to disable talking to Jicofo. It's intended to be used only in jitsi-meet-spot which joins a special room.

### DIFF
--- a/modules/xmpp/moderator.js
+++ b/modules/xmpp/moderator.js
@@ -363,22 +363,26 @@ Moderator.prototype.parseConfigOptions = function(resultIq) {
  *
  * @param {Function} callback - the function to be called back upon the
  * successful allocation of the conference focus
+ * @returns {Promise} - Resolved when Jicofo allows to join the room. It's never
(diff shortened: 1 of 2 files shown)


modules/xmpp/moderator.js
```diff
@@ -363,22 +363,26 @@ Moderator.prototype.parseConfigOptions = function(resultIq) {
  *
  * @param {Function} callback - the function to be called back upon the
  * successful allocation of the conference focus
+ * @returns {Promise} - Resolved when Jicofo allows to join the room. It's never
+ * rejected and it'll keep on pinging Jicofo forever.
  */
-Moderator.prototype.allocateConferenceFocus = function(callback) {
-    // Try to use focus user JID from the config
-    this.setFocusUserJid(this.options.connection.focusUserJid);
+Moderator.prototype.allocateConferenceFocus = function() {
```
virtuacoplenny: No need to support both promises and callback, right?

PR #877, opened 2019-02-22
fix(screenshare): do not limit resolution for fake screenshare

virtuacoplenny:
A camera source can be used as a screenshare source.
Normally for screenshare resolution is not capped and
the same will be true for the fake screenshare source.


virtuacoplenny:
@bbaldino

PR #878, opened 2019-03-06
fix(screenshare): specify source type for fake and proxy

virtuacoplenny:
Spot is the consumer of screensharing using a camera
input and through a proxy connection. To differentiate
which one is active, declare a source type on the
created "desktop" stream.


virtuacoplenny:
More specifically, this PR could allow the jitsi-meet external/iframe api to say screensharing is enabled and pass in the source type of the stream. Then Spot can set its presence to say screensharing is on with type of device/proxy/whatever-var-names and show different UI based on that.

PR #880, opened 2019-03-13
feat(ProxyConnectionService): use JitsiConnection to fetch TURN credentials

paweldomas:


modules/proxyconnection/ProxyConnectionService.js
```diff
@@ -34,12 +37,21 @@ export default class ProxyConnectionService {
...
          * Holds a reference to the collection of all callbacks.
          *
          * @type {Object}
          */
-        this._options = options;
+        this._options = {
+            iceConfig: jitsiConnection
+                && jitsiConnection.xmpp.connection.jingle.p2pIceConfig,
+            ...otherOptions
+        };
```
virtuacoplenny: We'll handle refreshing some other day, yeah?

virtuacoplenny: Also, are we guaranteed to have p2pIceConfig defined?

paweldomas: Yes, I was thinking to do that in the xmpp module. In spot they will be fetched each time new screensharing connection is created ?

paweldomas: we don't guarantee, but then it should give `iceConfig` `undefined` and then it should default to `DEFAULT_STUN_SERVERS` as far as I understand

PR #881, opened 2019-03-19
ref(electron): allow passing in desktop source info for capturing

virtuacoplenny:


modules/RTC/ScreenObtainer.js
```diff
@@ -178,36 +178,50 @@ const ScreenObtainer = {
      * @param {Array<string>} [options.desktopSharingSources] - Array with the
      * sources that have to be displayed in the desktop picker window ('screen',
      * 'window', etc.).
+     * @param {Object} [options.screenshareSource] - Information about a desktop
```
saghul: all the options here are desktopSharingXXX, maybe this one should follow?

virtuacoplenny:
Decided to keep using the existing global variable implementation for electron desktop picker so this PR has been closed.

PR #883, opened 2019-03-24
Fix typos around permissions

virtuacoplenny:


JitsiMediaDevices.js
```diff
@@ -156,7 +156,19 @@ class JitsiMediaDevices {
                 }
 
                 Promise.all(promises).then(
-                    r => resolve(r.every(Boolean)),
+                    r => {
+                        const grantedPermissions = r.every(permissionStatus => {
```
webintoit: `isPermissionsGranted`

JitsiMediaDevices.js
```diff
@@ -156,7 +156,19 @@ class JitsiMediaDevices {
...
+                            // The status attribute is deprecated, and state
+                            // should be used instead, but check both for now
+                            // for backwards compatibility.
+                            const grantStatus = permissionStatus.state
+                                || permissionStatus.status;
+
+                            return grantStatus === 'granted';
```
webintoit: `granted` Should be moved to upper level constant

JitsiMediaDevices.js
```diff
@@ -156,7 +156,19 @@ class JitsiMediaDevices {
                 }
 
                 Promise.all(promises).then(
-                    r => resolve(r.every(Boolean)),
+                    r => {
```
webintoit: what is `r` ?  is `response ` better here ?

saghul: response shortened as r is a generally used convention

PR #888, opened 2019-04-12
fix(JitsiLocalTrack): realDeviceId.

hristoterezov:


modules/RTC/JitsiLocalTrack.js
```diff
@@ -136,6 +136,8 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // created (until getConstraints() support), however we can associate
         // tracks with real devices obtained from enumerateDevices() call as
         // soon as it's called.
+        // NOTE: this.deviceId corresponds to the device id specified in GUM constraints and this._realDeviceId seems to
+        // correspond to the id of a matching device from the available device list.
```
virtuacoplenny: Did you know why there are two ids? Is it because exact constraints aren't used so another device could be obtained?

modules/RTC/RTCUtils.js
```diff
@@ -592,6 +592,9 @@ function onMediaDevicesListChanged(devicesReceived) {
 
     sendDeviceListToAnalytics(availableDevices);
 
+    // Used by tracks to update the real device id before the consumer of lib-jitsi-meet receives the new device list.
+    eventEmitter.emit(RTCEvents.DEVICE_LIST_WILL_CHANGE, devicesReceived);
```
virtuacoplenny: This event name seems misleading, because the device list has already changed and there is nothing between will change and has changed. But I'm assuming your added it to make sure certain listeners get processed first. Is that correct? Were there any alternatives you were thinking about to intercept DEVICE_LIST_CHANGED to make sure tracks get updated first?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -265,8 +265,14 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
-        const device = devices.find(
-            d => d.kind === `${track.kind}input` && d.label === track.label);
+        const kind = `${track.kind}input`;
+        let device = devices.find(d => d.kind === kind && d.label === track.label);
+
+        if (!device && this._realDeviceId === 'default') { // the default device has been changed.
+            const label = (track.label || '').replace('Default - ', '');
+
+            device = devices.find(d => d.kind === kind && d.label === label);
+        }
```
virtuacoplenny: So what was the bug causing this issue? Maybe a comment here on why this replacing is needed would be good.

modules/RTC/JitsiLocalTrack.js
hristoterezov: The device id passed in the constructor is the device id passed to GUM. Currently it serves as fallback if we can't obtain the device id. The use case that you mentioned sounds like one possible reason. Maybe there are other...

modules/RTC/RTCUtils.js
hristoterezov: Correct! I agree the name may be a little misleading, although you can argue that the list is changed when DEVICE_LIST_CHANGED is fired. Unfortunately I can't think of better name. Any ideas?

modules/RTC/JitsiLocalTrack.js
hristoterezov: If the default device was 'A' and for some reason the default device is changed to 'B' the label for the track will remain 'Default - A' but the label for the device in the device list will be 'A'.

I've put '// the default device has been changed.' comment. Do you want me to add the above explanation?

virtuacoplenny: Yes.

modules/RTC/JitsiLocalTrack.js
virtuacoplenny: Tangential question: would relying on label over device id have fixed this issue as well then?

modules/RTC/RTCUtils.js
virtuacoplenny: RECEIVED? That sounds kind of dumb too. At least change the DEVICE_LIST_WILL_CHANGE callback from _onDeviceListChanged to something not clashing with DEVICE_LIST_CHANGED.

modules/RTC/JitsiLocalTrack.js
hristoterezov: What is the issue? That we have 2 variables for device id or that the device id was wrong when the default device is changed?

If you are talking about the default device issue, I'd say no because the issue is that the track uses the old label...

if you are talking about the 2 variables for device id, I'm not sure. I guess we can fix it even if we use the device ids but it will need further investigation...

modules/RTC/RTCUtils.js
hristoterezov: Changed the name of the method.

modules/RTC/JitsiLocalTrack.js
hristoterezov: Done!

PR #893, opened 2019-04-23
Removes Edge support.

damencho:


modules/RTC/ortc/RTCPeerConnection.js
```diff
@@ -1,2146 +0,0 @@
-/* global __filename, RTCIceGatherer, RTCIceTransport, RTCDtlsTransport,
-RTCRtpSender, RTCRtpReceiver */
-
-import { getLogger } from 'jitsi-meet-logger';
-import yaeti from 'yaeti';
```
saghul: you can also remove this dependency from package.json

damencho: Removed it.

dennishansen:
Why did we remove edge support? If we add it back, should we have any issues?

damencho:
This removes edge support for the non chromium version, which anyway had its problems and soon will no longer exist.

PR #899, opened 2019-05-02
Add video background blur

theunafraid:
TODO 

- [x]  Fix no data source error
- [x] Add comments

Improvements:
- [ ] Remove scaling by adding second MediaStream captured at lower resolution.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -399,11 +399,11 @@ export default class JitsiLocalTrack extends JitsiTrack {
             return Promise.resolve();
         }
 
+        this._effectEnabled = enableFlag;
+
         if (this.isMuted()) {
             return Promise.resolve();
         }
-
-        this._effectEnabled = enableFlag;
 
         return this._switchStream(enableFlag ? this._startStreamEffect(effect) : this._stopStreamEffect())
                     .catch(error => {
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -342,7 +342,6 @@ export default class JitsiLocalTrack extends JitsiTrack {
      */
     _stopStreamEffect() {
         this._streamEffect.stopEffect();
-        this._streamEffect = null;
 
         return this._originalStream;
     }
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -32,7 +32,6 @@ const logger = getLogger(__filename);
  * One <tt>JitsiLocalTrack</tt> corresponds to one WebRTC MediaStreamTrack.
  */
 export default class JitsiLocalTrack extends JitsiTrack {
-
     /**
      * Constructs new JitsiLocalTrack instance.
      *
@@ -73,14 +72,11 @@ export default class JitsiLocalTrack extends JitsiTrack {
             mediaType,
             videoType);
 
+        this._setEffectInProgress = false;
         const effect = effects.find(e => e.isEnabled(this));
 
-        this._effectEnabled = false;
-        this._streamChangeInProgress = false;
-
         if (effect) {
-            this._setStream(this._startStreamEffect(effect));
-            this._effectEnabled = true;
+            this._startStreamEffect(effect);
         }
 
         /**
@@ -324,96 +320,99 @@ export default class JitsiLocalTrack extends JitsiTrack {
      *
      * @private
      * @param {*} effect - Represents effect instance
-     * @returns {MediaStream}
+     * @returns {void}
      */
     _startStreamEffect(effect) {
         this._streamEffect = effect;
         this._originalStream = this.stream;
-        this._streamEffect.startEffect(this._originalStream);
-
-        return this._streamEffect.getStreamWithEffect();
+        this._setStream(this._streamEffect.startEffect(this._originalStream));
     }
 
     /**
      * Stops the effect process and returns the original stream.
      *
      * @private
-     * @returns {MediaStream}
+     * @returns {void}
      */
     _stopStreamEffect() {
-        this._streamEffect.stopEffect();
-
-        return this._originalStream;
+        if (this._streamEffect) {
+            this._streamEffect.stopEffect();
+            this._setStream(this._originalStream);
+            this._originalStream = undefined;
+        }
     }
 
     /**
-     * Switches the video stream associated with this track and updates the containers.
+     * Stops the currently used effect (if there is one) and starts the passed effect (if there is one).
      *
-     * @private
-     * @param {MediaStream} newStream - The MediaStream to replace
-     * @returns {Promise}
+     * @param {Object|undefined} effect - The new effect to be set.
      */
-    _switchStream(newStream: MediaStream) {
-        if (this._streamChangeInProgress === true) {
-            return Promise.reject(new Error('Stream change already in progress!'));
+    _switchStreamEffect(effect) {
+        if (this._streamEffect) {
+            this._stopStreamEffect();
+            this._streamEffect = undefined;
         }
-
-        this._streamChangeInProgress = true;
-
-        const conference = this.conference;
-
-        if (!conference) {
-            return Promise.resolve();
+        if (effect) {
+            this._startStreamEffect(effect);
         }
-
-        return conference.removeTrack(this)
-            .then(() => {
-                this._setStream(newStream);
-
-                if (!this.isAudioTrack()) {
-                    this.containers.forEach(
-                        cont => RTCUtils.attachMediaStream(cont, this.stream));
-                }
-
-                return conference.addTrack(this);
-            })
-            .then(() => {
-                this._streamChangeInProgress = false;
-            })
-            .catch(error => {
-                this._streamChangeInProgress = false;
-                logger.error('Failed to switch to new stream!', error);
-                throw error;
-            });
     }
 
     /**
      * Sets the effect and switches between the modified stream and original one.
      *
-     * @param {Boolean} enableFlag - Flag to start or stop the effect processing
-     * @param {*} effect - Represents the effect instance to use
+     * @param {Object} effect - Represents the effect instance to be used.
      * @returns {Promise}
      */
-    enableEffect(enableFlag: Boolean, effect) {
-        if (this._effectEnabled === enableFlag) {
+    setEffect(effect) {
+        if (typeof this._streamEffect === 'undefined' && typeof effect === 'undefined') {
             return Promise.resolve();
         }
 
-        this._effectEnabled = enableFlag;
+        if (typeof effect !== 'undefined' && !effect.isEnabled(this)) {
+            return Promise.reject(new Error('Incompatible effect instance!'));
+        }
+
+        if (this._setEffectInProgress === true) {
+            return Promise.reject(new Error('setEffect already in progress!'));
+        }
 
         if (this.isMuted()) {
+            this._streamEffect = effect;
+
             return Promise.resolve();
         }
 
-        return this._switchStream(enableFlag ? this._startStreamEffect(effect) : this._stopStreamEffect())
-                    .catch(error => {
-                        if (enableFlag) {
-                            this._stopStreamEffect();
-                        }
-                        this._effectEnabled = false;
-                        logger.error('Failed to switch to new stream!', error);
-                        throw error;
-                    });
+        const conference = this.conference;
+
+        if (!conference) {
+            this._switchStreamEffect(effect);
+
+            return Promise.resolve();
+        }
+
+        this._setEffectInProgress = true;
+
+        // TODO: Create new JingleSessionPC method for replacing a stream in JitsiLocalTrack without offer answer.
+        return conference.removeTrack(this)
+            .then(() => {
+                this._switchStreamEffect(effect);
+                if (this.isVideoTrack()) {
+                    this.containers.forEach(cont => RTCUtils.attachMediaStream(cont, this.stream));
+                }
+
+                return conference.addTrack(this);
+            })
+            .then(() => {
+                this._setEffectInProgress = false;
+            })
+            .catch(error => {
+                // Any error will be not recovarable and will trigger CONFERENCE_FAILED event. But let's try to cleanup
+                // everyhting related to the effect functionality.
+                this._setEffectInProgress = false;
+                this._switchStreamEffect();
+                logger.error('Failed to switch to the new stream!', error);
+                throw error;
+            });
     }
 
     /**
@@ -481,11 +480,14 @@ export default class JitsiLocalTrack extends JitsiTrack {
                 this.track.enabled = !muted;
             }
         } else if (muted) {
-
             promise = new Promise((resolve, reject) => {
                 logMuteInfo();
                 this._removeStreamFromConferenceAsMute(
                     () => {
+                        if (this._streamEffect) {
+                            this._stopStreamEffect();
+                        }
+
                         // FIXME: Maybe here we should set the SRC for the
                         // containers to something
                         // We don't want any events to be fired on this stream
@@ -495,12 +497,6 @@ export default class JitsiLocalTrack extends JitsiTrack {
                         resolve();
                     },
                     reject);
-            });
-
-            promise.then(() => {
-                if (this._effectEnabled) {
-                    this._stopStreamEffect();
-                }
             });
         } else {
             logMuteInfo();
@@ -553,8 +549,8 @@ export default class JitsiLocalTrack extends JitsiTrack {
                     throw new JitsiTrackError(TRACK_NO_STREAM_FOUND);
                 }
 
-                if (this._effectEnabled) {
-                    this._setStream(this._startStreamEffect(this._streamEffect));
+                if (this._streamEffect) {
+                    this._startStreamEffect(this._streamEffect);
                 }
 
                 this.containers.map(
@@ -644,9 +640,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * @returns {Promise}
      */
     dispose() {
-        if (this._effectEnabled) {
-            this._setStream(this._stopStreamEffect());
-        }
+        this._switchStreamEffect();
 
         let promise = Promise.resolve();
 
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -32,6 +32,7 @@ const logger = getLogger(__filename);
  * One <tt>JitsiLocalTrack</tt> corresponds to one WebRTC MediaStreamTrack.
  */
 export default class JitsiLocalTrack extends JitsiTrack {
+
     /**
      * Constructs new JitsiLocalTrack instance.
      *
@@ -43,6 +44,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * JitsiRemoteTrack
      * @param trackInfo.mediaType the MediaType of the JitsiRemoteTrack
      * @param trackInfo.videoType the VideoType of the JitsiRemoteTrack
+     * @param trackInfo.effects the effects array contains the effect instance to use
      * @param trackInfo.resolution the video resolution if it's a video track
      * @param trackInfo.deviceId the ID of the local device for this track
      * @param trackInfo.facingMode the camera facing mode used in getUserMedia
@@ -60,7 +62,8 @@ export default class JitsiLocalTrack extends JitsiTrack {
         sourceType,
         stream,
         track,
-        videoType
+        videoType,
+        effects = []
     }) {
         super(
             /* conference */ null,
@@ -69,6 +72,16 @@ export default class JitsiLocalTrack extends JitsiTrack {
             /* streamInactiveHandler */ () => this.emit(LOCAL_TRACK_STOPPED),
             mediaType,
             videoType);
+
+        const effect = effects.find(e => e.isEnabled(this));
+
+        this._effectEnabled = false;
+        this._streamChangeInProgress = false;
+
+        if (effect) {
+            this._setStream(this._startStreamEffect(effect));
+            this._effectEnabled = true;
+        }
 
         /**
          * The ID assigned by the RTC module on instance creation.
@@ -307,6 +320,104 @@ export default class JitsiLocalTrack extends JitsiTrack {
     }
 
     /**
+     * Starts the effect process and returns the modified stream.
+     *
+     * @private
+     * @param {*} effect - Represents effect instance
+     * @returns {MediaStream}
+     */
+    _startStreamEffect(effect) {
+        this._streamEffect = effect;
+        this._originalStream = this.stream;
+        this._streamEffect.startEffect(this._originalStream);
+
+        return this._streamEffect.getStreamWithEffect();
+    }
+
+    /**
+     * Stops the effect process and returns the original stream.
+     *
+     * @private
+     * @returns {MediaStream}
+     */
+    _stopStreamEffect() {
+        this._streamEffect.stopEffect();
+        this._streamEffect = null;
+
+        return this._originalStream;
+    }
+
+    /**
+     * Switches the video stream associated with this track and updates the containers.
+     *
+     * @private
+     * @param {MediaStream} newStream - The MediaStream to replace
+     * @returns {Promise}
+     */
+    _switchStream(newStream: MediaStream) {
+        if (this._streamChangeInProgress === true) {
+            return Promise.reject(new Error('Stream change already in progress!'));
+        }
+
+        this._streamChangeInProgress = true;
+
+        const conference = this.conference;
+
+        if (!conference) {
+            return Promise.resolve();
+        }
+
+        return conference.removeTrack(this)
+            .then(() => {
+                this._setStream(newStream);
+
+                if (!this.isAudioTrack()) {
+                    this.containers.forEach(
+                        cont => RTCUtils.attachMediaStream(cont, this.stream));
+                }
+
+                return conference.addTrack(this);
+            })
+            .then(() => {
+                this._streamChangeInProgress = false;
+            })
+            .catch(error => {
+                this._streamChangeInProgress = false;
+                logger.error('Failed to switch to new stream!', error);
+                throw error;
+            });
+    }
+
+    /**
+     * Sets the effect and switches between the modified stream and original one.
+     *
+     * @param {Boolean} enableFlag - Flag to start or stop the effect processing
+     * @param {*} effect - Represents the effect instance to use
+     * @returns {Promise}
+     */
+    enableEffect(enableFlag: Boolean, effect) {
+        if (this._effectEnabled === enableFlag) {
+            return Promise.resolve();
+        }
+
+        if (this.isMuted()) {
+            return Promise.resolve();
+        }
+
+        this._effectEnabled = enableFlag;
+
+        return this._switchStream(enableFlag ? this._startStreamEffect(effect) : this._stopStreamEffect())
+                    .catch(error => {
+                        if (enableFlag) {
+                            this._stopStreamEffect();
+                        }
+                        this._effectEnabled = false;
+                        logger.error('Failed to switch to new stream!', error);
+                        throw error;
+                    });
+    }
+
+    /**
      * Asynchronously mutes this track.
      *
      * @returns {Promise}
@@ -371,6 +482,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
                 this.track.enabled = !muted;
             }
         } else if (muted) {
+
             promise = new Promise((resolve, reject) => {
                 logMuteInfo();
                 this._removeStreamFromConferenceAsMute(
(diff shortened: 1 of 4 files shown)


theunafraid:
> Do we need to store and pass the video blur as media type or video type? Why?
> 
> Can't we store and pass this information separately? Maybe another flag?

Yes, will pass as option as we discussed offline

modules/RTC/JitsiLocalTrack.js
```diff
@@ -190,7 +213,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * issues.
      */
     _initNoDataFromSourceHandlers() {
-        if (this.isVideoTrack() && this.videoType === VideoType.CAMERA) {
+        if (this.isVideoTrack() && (this.videoType === VideoType.CAMERA)) {
```
hristoterezov: Do we need this change?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -301,6 +324,60 @@ export default class JitsiLocalTrack extends JitsiTrack {
         }
     }
 
+    /**
+     * Starts the video blur effect
```
hristoterezov: Add returns line

modules/RTC/JitsiLocalTrack.js
```diff
@@ -301,6 +324,60 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            cont => RTCUtils.attachMediaStream(cont, this.stream));
+        this._addStreamToConferenceAsUnmute();
+        this.effects.push('blur');
+        Statistics.sendAnalytics(TRACK_VIDEOBLUR, { value: 'enabled' });
+
+        return Promise.resolve();
+    }
+
+    /**
+     * Stops the video blur effect
```
hristoterezov: Add returns line

modules/RTC/JitsiStreamBlurEffect.js
```diff
@@ -0,0 +1,178 @@
+import JitsiStreamEffect from './JitsiStreamEffect';
```
hristoterezov: Add JSDoc for this file

modules/RTC/JitsiStreamBlurEffect.js
```diff
@@ -0,0 +1,178 @@
...
+                        for (let j = 0; j < this._width; ++j) {
+
+                            x = Math.floor(j * this._scaleFactor);
+                            y = Math.floor(i * this._scaleFactor);
+
+                            segData.data[(i * this._width) + j] = value.data[(y * value.width) + x];
+                        }
+                    }
+
+                    // set
```
hristoterezov: I don't really see why we have this comment. We should remove it or update it.

modules/RTC/JitsiStreamEffect.js
```diff
@@ -0,0 +1,35 @@
+/**
+ *
```
hristoterezov: Add JSDoc.

modules/RTC/RTCUtils.js
```diff
@@ -1386,7 +1386,7 @@ class RTCUtils extends Listenable {
             const umDevices = options.devices || [ 'audio', 'video' ];
             const requestedCaptureDevices = umDevices.filter(device =>
                 device === 'audio'
-                || (device === 'video' && browser.supportsVideo()));
+                || ((device === 'video') && browser.supportsVideo()));
```
hristoterezov: Do we need this change?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -424,11 +507,16 @@ export default class JitsiLocalTrack extends JitsiTrack {
                         logger.warn(
                             `${this}: video type has changed after unmute!`,
                             this.videoType, streamInfo.videoType);
-                        this.videoType = streamInfo.videoType;
```
hristoterezov: Why did you remove this line?

modules/RTC/RTC.js
```diff
@@ -102,7 +102,8 @@ function _newCreateLocalTracks(mediaStreamMetaData = []) {
             sourceType,
             stream,
             track,
-            videoType: videoType || null
+            videoType,
```
hristoterezov: Why did you remove the `null`? Aren't we checking for `null` somewhere?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -366,6 +443,11 @@ export default class JitsiLocalTrack extends JitsiTrack {
                 this.track.enabled = !muted;
             }
         } else if (muted) {
+
```
hristoterezov: Unnecessary empty line.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -69,6 +74,19 @@ export default class JitsiLocalTrack extends JitsiTrack {
             mediaType,
             videoType);
 
+        this.effects = effects;
+
+        this._videoBlurEnabled = this.effects.indexOf('blur') !== -1
+            && mediaType === MediaType.VIDEO;
+
+        this._jitsiStreamBlurEffect = null;
```
hristoterezov: You can leave this set to `undefined`, isn't it? Isn't that going to be better? WDYT?

modules/RTC/RTCUtils.js
theunafraid: No, sorry forgot to revert that change

modules/RTC/JitsiLocalTrack.js
theunafraid: By mistake, thought it was my change before adding the effects option

modules/RTC/RTC.js
theunafraid: By mistake, corrected

modules/RTC/JitsiLocalTrack.js
theunafraid: Not needed

modules/RTC/JitsiLocalTrack.js
theunafraid: Agree, undefined is better.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -20,9 +20,11 @@ import RTCEvents from '../../service/RTC/RTCEvents';
 import VideoType from '../../service/RTC/VideoType';
 import {
     TRACK_UNMUTED,
-    createNoDataFromSourceEvent
+    createNoDataFromSourceEvent,
+    TRACK_VIDEOBLUR
 } from '../../service/statistics/AnalyticsEvents';
 import Statistics from '../statistics/statistics';
+import JitsiStreamBlurEffect from './JitsiStreamBlurEffect';
```
hristoterezov: Could you please leave an empty line? This way it will be consistent with jitsi-meet imports...

modules/RTC/JitsiLocalTrack.js
```diff
@@ -174,6 +191,11 @@ export default class JitsiLocalTrack extends JitsiTrack {
         RTCUtils.addListener(RTCEvents.DEVICE_LIST_WILL_CHANGE, this._onDeviceListWillChange);
 
         this._initNoDataFromSourceHandlers();
+
+        if (this._videoBlurEnabled === true && this._jitsiStreamBlurEffect) {
```
hristoterezov: I'm asking out of curiosity - Why don't you include this logic into the `if` that initialises  `this. _jitsiStreamBlurEffect`?

service/statistics/AnalyticsEvents.js
```diff
@@ -216,6 +216,14 @@ export const ICE_STATE_CHANGED = 'ice.state.changed';
  */
 export const TRACK_UNMUTED = 'track.unmuted';
 
+/**
+ * Indicates that a track has video blur enabled
+ *
+ * Properties:
+ *      value: 'enabled' or 'disabled'
+ */
+export const TRACK_VIDEOBLUR = 'track.videoblur';
```
hristoterezov: Why don't you create an event creator function?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -301,6 +323,64 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
     }
 
+    /**
+     * Starts the video blur effect
+     *
+     * @returns {Promise}
+     */
+    startVideoBlurEffect() {
+        if (this.getType() !== MediaType.VIDEO) {
+            return Promise.reject();
```
hristoterezov: Add `new Error(...)` as an argument

modules/RTC/RTC.js
```diff
@@ -76,7 +76,7 @@ function createLocalTracks(tracksInfo, options) {
  *     sourceType: The desktopsharing source type
  * }}
  */
-function _newCreateLocalTracks(mediaStreamMetaData = []) {
+function _newCreateLocalTracks(mediaStreamMetaData = [], options = {}) {
```
hristoterezov: Add `options` to the JSDoc.

modules/RTC/JitsiStreamBlurEffect.js
```diff
@@ -0,0 +1,183 @@
...
+     */
+    constructor(stream) {
+
+        super(stream);
+
+        let n = 1;
+
+        // Original video stream settings
+        const firstVideoTrack = this._stream.getVideoTracks().shift();
+        const { height, frameRate, width } = firstVideoTrack.getSettings();
```
hristoterezov: What happens if the browser doesn't support `getSettings`?

modules/RTC/RTC.js
```diff
@@ -102,7 +102,8 @@ function _newCreateLocalTracks(mediaStreamMetaData = []) {
             sourceType,
             stream,
             track,
-            videoType: videoType || null
+            videoType: videoType || null,
+            effects: options.effects
```
hristoterezov: Isn't that going to add the effect property even on audio tracks? 

Can't we make `obtainAudioAndVideoPermissions` to include it in `tracksInfo` array? This way we wouldn't need this parameter and the effect property will be assigned only to the correct `JitsiLocalTrack`.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -777,6 +871,14 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // we aren't receiving any data from the source. We want to notify
         // the users for error if the stream is muted or ended on it's
         // creation.
+
+        // For video blur enabled use the original video stream
+        if (this._videoBlurEnabled === true) {
```
hristoterezov: How about the following implementation:
```
   const stream =  this._videoBlurEnabled ? this._originalVideoStream : this.stream;

        return stream.getTracks().some(track =>
            (!('readyState' in track) || track.readyState === 'live')
                && (!('muted' in track) || track.muted !== true));
```

hristoterezov: Also in this case I guess you have to check if this._originalVideoStream exists...

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +342,64 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+    }
+
+    /**
+     * Stops the video blur effect
+     *
+     * @returns {Promise}
+     */
+    stopVideoBlurEffect() {
+        if (this.getType() !== MediaType.VIDEO) {
+            return Promise.reject();
```
hristoterezov: Add new Error(...) as an argument

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +342,64 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            return Promise.resolve();
+        }
+
+        this._videoBlurEnabled = false;
+        if (this._jitsiStreamBlurEffect) {
+            this._jitsiStreamBlurEffect.stopEffect();
+        }
+
+        this._setStream(this._originalVideoStream);
+        this.containers.map(
```
hristoterezov: why `map` and not `forEach`?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +342,64 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            cont => RTCUtils.attachMediaStream(cont, this.stream));
+        this._addStreamToConferenceAsUnmute();
+        this.effects = this.effects.filter(flt => flt !== 'blur');
```
hristoterezov: why do we need this?

hristoterezov:
@theunafraid It seems if you change the blur setting during video mute your video is not muted anymore and the video mute stops working.

hristoterezov:
@theunafraid Another bug switching the state doesn't seem to work properly. I've joined with 2 participant - A and B. 

Bug#1:
Steps to reproduce:
1) A joins the call
2) B joins the call
3) A starts the video blur

Expected result:
B receives the blurred video.

Actual result:
B receives not blurred video.

Bug#2:
1) A joins the call
2) A enables the video blur
3) B joins the call and receives the blurred video
4) A disables the blurred video

Expected result:
B receives the original video.

Actual result:
B doesn't receive video.

hristoterezov:
@theunafraid I've noticed that with your PR the size of lib-jitsi-meet bundle is 1.7 MB and before that it was ~600KB. This is huge increase. Any idea if we can fix this?

theunafraid:
@hristoterezov I think this is because of tensorflow model...Not sure how to fix at the moment

hristoterezov:
@theunafraid Increasing the size of the bundle 3 times for one feature is a very serious issue. We should work on fixing this somehow.

modules/RTC/JitsiLocalTrack.js
theunafraid: Renamed _originalVideoStream to _originalStream for general use so that no check is needed

modules/RTC/RTC.js
theunafraid: I thought that maybe in the future we need an effect/processing of audio stream also. And i can give and example if we discuss offline. We can move it there for audio/video media but then i think i have to adapt JitsiLocalTrack also.

service/statistics/AnalyticsEvents.js
theunafraid: I used Statistics.sendAnalytics as I saw it being used without a creator function - in RTCUtils.js

modules/RTC/JitsiStreamBlurEffect.js
theunafraid: I will use getContraints as alternative but saw in RTC.js _newCreateLocalTracks the following call :         const { deviceId, facingMode } = track.getSettings();

modules/RTC/RTC.js
theunafraid: Also the effects option is used from outside lib-jitsi-meet using JitsiMeetJS.js createLocalTracks otherwise it is empty array

theunafraid:
@hristoterezov Fixed the 3 found bugs

modules/RTC/RTC.js
theunafraid: From JSDoc of JitsiMeetJS.js createLocalTracks 
* @param options Object with properties / settings specifying the tracks
* which should be created. should be created or some additional
* configurations about resolution for example.

modules/RTC/RTC.js
theunafraid: Also added JSDoc for JitsiMeetJS.js createLocalTracks

modules/RTC/JitsiLocalTrack.js
theunafraid: Moved code to initialization if branch

modules/RTC/RTC.js
theunafraid: Moved effects to trackInfo

saghul:
FTR, I'd like to show my opposition to the direction this is taking. I find the size increate to be unaccetable for just this one feature.

This could be implemented as a separate bundle which implements the blur effect itself. What lib-jitsi-meet would need is a way to register transformations / filters / whatever for the local track.

hristoterezov:
@saghul The last versions of the PR are addressing the issue you are referring to. We have separate bundle build in jitsi-meet and the actual filter implementation is loaded dynamically when we need it and passed to lib-jitsi-meet trough the API.

Are you sure you are looking into the latest version of the code?

saghul:
@hristoterezov That's awesome to hear! Looks like I hadn't clicked on the "refresh" banner thing 🤦‍♂  sorry for the noise!

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +335,121 @@ export default class JitsiLocalTrack extends JitsiTrack {
         }
     }
 
+    /**
+     * Sets the effect type that is used to modify the original stream
+     * @param {*} effectType
+     */
+    setEffect(effectType) {
+        this.EffectType = effectType;
```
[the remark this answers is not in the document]
theunafraid: Done, renamed this.EffectType to this._effectType

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +335,121 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+        this._originalStream = this.stream;
+    }
+
+    /**
+     * Starts the effect
+     *
+     * @returns {Promise}
+     */
+    startEffect() {
```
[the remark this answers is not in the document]
theunafraid: Removed setEffect and added a parameter to startEffect

modules/RTC/JitsiLocalTrack.js
```diff
@@ -449,6 +607,14 @@ export default class JitsiLocalTrack extends JitsiTrack {
                     throw new JitsiTrackError(TRACK_NO_STREAM_FOUND);
                 }
 
+                this._originalStream = this.stream;
+
+                if (this.effectEnabled) {
+                    this._streamEffect = new this._effect(this._originalStream);
+                    this._setStream(this._streamEffect.getStreamWithEffect());
+                    this._streamEffect.startEffect();
+                }
```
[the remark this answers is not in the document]
theunafraid: Created new function _streamEffectEnable which is used in many places in this file.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -385,6 +536,12 @@ export default class JitsiLocalTrack extends JitsiTrack {
                 this.track.enabled = !muted;
             }
         } else if (muted) {
+
+            if (this.effectEnabled) {
```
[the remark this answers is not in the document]
theunafraid: Used new function _streamEffectEnable.

modules/RTC/RTCUtils.js
```diff
@@ -1409,7 +1410,8 @@ class RTCUtils extends Listenable {
                 mediaStreamsMetaData.push({
                     stream: videoStream,
                     track: videoStream.getVideoTracks()[0],
-                    videoType: VideoType.CAMERA
+                    videoType: VideoType.CAMERA,
+                    effectTypes: options.effects
```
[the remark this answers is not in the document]
theunafraid: Renamed to just effects and used property shorthand

modules/RTC/RTC.js
```diff
@@ -83,7 +85,8 @@ function _newCreateLocalTracks(mediaStreamMetaData = []) {
             sourceType,
             stream,
             track,
-            videoType
+            videoType,
+            effectTypes
```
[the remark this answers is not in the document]
theunafraid: Renamed to effects

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +337,140 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+        return startEffectPromise;
+    }
+
+    /**
+     * Stops the effect
+     *
+     * @returns {Promise}
+     */
+    stopEffect() {
+        if (this.effectEnabled === false) {
```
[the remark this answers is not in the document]
theunafraid: Created a new function _switchStream which contains most of the common code and replaced startEffect and stopEffect with enableEffect function.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +337,140 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            const modifiedStream = this._streamEffect.getStreamWithEffect();
+
+            this._setStream(modifiedStream);
+
+            if (!this.isAudioTrack()) {
+                this.containers.forEach(
+                    cont => RTCUtils.attachMediaStream(cont, this.stream));
+            }
+
+            promises.push(this._addStreamToConferenceAsUnmute());
```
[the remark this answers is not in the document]
theunafraid: Reworked the code and moved it to new function _switchStream.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +337,140 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+        });
+
+        const startEffectPromise = Promise.all(promises);
```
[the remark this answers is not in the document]
theunafraid: Code from this function is not in _switchStream.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +337,140 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+
+        startEffectPromise.then(() => {
```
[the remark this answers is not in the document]
theunafraid: Code is now in _switchStream.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +337,140 @@ export default class JitsiLocalTrack extends JitsiTrack {
         }
     }
 
+    /**
+     * Starts the effect and sets the effect type that is used to modify the original stream
+     * @param {*} effect
```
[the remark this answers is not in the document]
theunafraid: Added description for effect parameter

modules/RTC/JitsiLocalTrack.js
```diff
@@ -69,6 +72,20 @@ export default class JitsiLocalTrack extends JitsiTrack {
             mediaType,
             videoType);
 
+        const effect = effects.find(e => e.isEnabled(this));
+
+        this.effectEnabled = false;
+        this.streamChangeInProgress = false;
```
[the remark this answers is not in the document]
theunafraid: Added _ to both

modules/RTC/JitsiLocalTrack.js
```diff
@@ -42,6 +43,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * JitsiRemoteTrack
      * @param trackInfo.mediaType the MediaType of the JitsiRemoteTrack
      * @param trackInfo.videoType the VideoType of the JitsiRemoteTrack
+     * @param trackInfo.effects the effect applied to the original stream
```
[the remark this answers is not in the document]
theunafraid: Updated the comment

modules/RTC/JitsiLocalTrack.js
```diff
@@ -69,6 +72,20 @@ export default class JitsiLocalTrack extends JitsiTrack {
             mediaType,
             videoType);
 
+        const effect = effects.find(e => e.isEnabled(this));
+
+        this.effectEnabled = false;
```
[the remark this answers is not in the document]
theunafraid: Added _

modules/RTC/JitsiLocalTrack.js
theunafraid: Removed empty line

modules/RTC/JitsiLocalTrack.js
```diff
@@ -59,7 +61,8 @@ export default class JitsiLocalTrack extends JitsiTrack {
         sourceType,
         stream,
         track,
-        videoType
+        videoType,
+        effectTypes = []
```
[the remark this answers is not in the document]
theunafraid: Renamed to effects

modules/RTC/JitsiLocalTrack.js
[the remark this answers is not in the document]
theunafraid: Replaced startEffect and stopEffect with enableEffect function

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +335,99 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+
+        return removeStreamPromise.then(() => {
+            this._setStream(newStream);
+
+            if (!this.isAudioTrack()) {
+                this.containers.forEach(
+                    cont => RTCUtils.attachMediaStream(cont, this.stream));
+            }
+
+            return this._addStreamToConferenceAsUnmute();
```
hristoterezov: `_removeLocalTrackAsMute` and `_addStreamToConferenceAsUnmute` are specifically designed for the video mute use case. Maybe we can create a new method that will replace the webrtc stream for a JitsiLocalTrack. WDYT? Probably we can do this even without Offer/Answer cycle.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +335,99 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+    }
+
+    /**
+     * Switches the video stream associated with this track and updates the containers.
+     *
+     * @private
+     * @param {MediaStream} newStream - The MediaStream to replace
+     * @returns {Promise}
+     */
+    _switchStream(newStream: MediaStream) {
```
hristoterezov: Since we use `_switchStream` only in `enableEffect`, why don't we include it's code in `enableEffect`?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -320,6 +335,99 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
 
+    /**
+     * Starts or stops the effect process and returns the corresponding stream.
+     *
+     * @private
+     * @param {Boolean} enable - Used to start or stop the effect processing
+     * @param {*} effect - Optional parameter represents effect class to create instance
+     * @returns {MediaStream}
+     */
+    _streamEffectEnable(enable: Boolean, effect) {
```
hristoterezov: Since the code is totally different for `enable===true` and `enable===false` I think it makes sense to have to methods. Maybe `_startStreamEffect` and `_stopStreamEffect`. WDYT?

modules/RTC/JitsiLocalTrack.js
theunafraid: I was thinking that since it is a generic implementation maybe it can be used in the future

modules/RTC/JitsiLocalTrack.js
```diff
@@ -449,6 +562,12 @@ export default class JitsiLocalTrack extends JitsiTrack {
                     throw new JitsiTrackError(TRACK_NO_STREAM_FOUND);
                 }
 
+                if (this._effectEnabled) {
+                    const modifiedStream = this._streamEffectEnable(true, this._effect);
+
+                    this._setStream(modifiedStream);
```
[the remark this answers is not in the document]
theunafraid: I agree, but i thought that we need the code to be more readable

modules/RTC/JitsiLocalTrack.js
theunafraid: I could also use removeTrack and addTrack from JitsiConference to get the same result

modules/RTC/JitsiLocalTrack.js
theunafraid: Create the two methods _startStreamEffect and _stopStreamEffect

modules/RTC/JitsiLocalTrack.js
```diff
@@ -536,6 +655,12 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * @returns {Promise}
      */
     dispose() {
+        if (this.effectEnabled) {
```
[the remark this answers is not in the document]
theunafraid: Fixed

modules/RTC/RTC.js
```diff
@@ -53,7 +53,8 @@ function createLocalTracks(tracksInfo, options) {
             ...trackInfo,
             deviceId,
             facingMode: options.facingMode,
-            rtcId: rtcTrackIdCounter
+            rtcId: rtcTrackIdCounter,
+            effect: options.effects
```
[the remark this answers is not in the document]
theunafraid: Fixed

modules/RTC/JitsiLocalTrack.js
hristoterezov: removeTrack and addTrack is fine for now. As further optimisation we can create JingleSession method that replaces a stream without doing Offer Answer. But let's address this trough another PR when we have more time.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -306,6 +319,103 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            return Promise.reject(new Error('Stream change already in progress!'));
+        }
+
+        this._streamChangeInProgress = true;
+
+        if (!this.conference) {
+            return Promise.resolve();
+        }
+
+        const conference = this.conference;
```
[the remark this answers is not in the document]
theunafraid: But this way the assignment will not be executed because of the earlier return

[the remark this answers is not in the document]
hristoterezov: Yes but we'll have 2 times the operation to search for the property... I don't think there's really a big performance tradeoff. I just like the solution for the early assignment of the `conference` more. If you don't like it you can keep the code as it is.

PR #900, opened 2019-05-09
Summary: Fix the issue that local tracks won't be added to session

tianjyan:
when we invoke JitsiConference.addTrack, we will add the track to session first and then cache the track to JitsiConference.rtc. This scenario could not cover all the cases and will cause remote device could not get local device‘s tracks. JitsiConference.p2pJingleSession may initialize after JitsiConference._doReplaceTrack executes and before JitsiConference._setupNewTrack executes. In this case, p2pJingleSession will initialize with empty localtracks in JitsiConference.rtc.

In this fix, I cache the track to JitsiConference.rtc before JitsiConference._doReplaceTrack, so that p2pJingleSession could initialize with the localtracks.

Please have a look when you are available. Thanks~

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1145,20 +1145,21 @@ JitsiConference.prototype.replaceTrack = function(oldTrack, newTrack) {
             return Promise.reject(
                 new JitsiTrackError(JitsiTrackErrors.TRACK_IS_DISPOSED));
         }
+        this.onLocalTrackRemoved(oldTrack);
     }
     if (newTrack) {
         if (newTrack.disposed) {
             return Promise.reject(
                 new JitsiTrackError(JitsiTrackErrors.TRACK_IS_DISPOSED));
         }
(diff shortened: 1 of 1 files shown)


tianjyan:
CLA signed for 8x8

ashitikov:
I can confirm that this problem still exists and reproduces stable. 
@jallamsetty1 @bbaldino is there any chance to review this pull request again?

sapkra:
Does this problem still exist? Can someone try to reproduce it?

tianjyan:
@sapkra I afraid the answer is Yes. PR updated.

saghul:
This fix feels incorrect on the surface. You are calling `onLocalTrackRemoved` before the track is actually removed, same goes for setting up the new track.

Maybe it's just semantics, but it reads weird.

Can you provide a short snipped that we can run, which exhibits the issue?

PR #901, opened 2019-05-12
Call Analytics' handler dispose() if exists.

markmadlangbayan:
If a custom handler has a dispose() function implemented, it will be called if the AnalyticsAdapter's dispose is also called.


markmadlangbayan:
I signed the CLA.

PR #906, opened 2019-05-31
feat(JitsiConnection): disconnect->return Promise

hristoterezov:


modules/xmpp/xmpp.js
```diff
@@ -92,7 +92,11 @@ export default class XMPP extends Listenable {
         // they wanted to utilize the connected connection in an unload handler
         // of their own. However, it should be fairly easy for them to do that
         // by registering their unload handler before us.
-        $(window).on('beforeunload unload', this.disconnect.bind(this));
+        $(window).on('beforeunload unload', ev => {
+            this.disconnect(ev).catch(() => {
+                // ignore errors in order to not brake the unload.
+            });
```
virtuacoplenny: To confirm, uncaught promise rejections somehow prevent other unload logic from executing?

modules/xmpp/xmpp.js
```diff
@@ -490,50 +496,62 @@ export default class XMPP extends Listenable {
...
+     * @param {Object} ev - Optionally, the event which triggered the necessity to
      * disconnect from the XMPP server (e.g. beforeunload, unload).
+     * @returns {Promise} - Resoves when the disconnect process is finished or rejects with an error.
      */
     disconnect(ev) {
         if (this.disconnectInProgress || !this.connection) {
             this.eventEmitter.emit(JitsiConnectionEvents.WRONG_STATE);
 
-            return;
+            return Promise.reject(new Error('Wrong connection state!'));
```
virtuacoplenny: Would it be useful to add information about there being a disconnect in progress or there isn't a connection?

JitsiConnection.js
```diff
@@ -74,13 +74,14 @@ JitsiConnection.prototype.attach = function(options) {
 
 /**
  * Disconnect the client from the server.
+ * @returns {Promise} - Resoves when the disconnect process is finished or rejects with an error.
```
virtuacoplenny: resoves -> resolves

modules/xmpp/xmpp.js
```diff
@@ -490,50 +496,62 @@ export default class XMPP extends Listenable {
...
+                    // Consequently, it may be fine to have the source code and
+                    // comment it in or out depending on whether we want to run with
+                    // it for some time.
+                    this.connection.options.sync = true;
+                }
             }
-        }
 
-        this.connection.disconnect();
+            this.connection.disconnect();
```
virtuacoplenny: To confirm, this disconnect call doesn't return a promise that can be re-used? The way the

modules/xmpp/xmpp.js
hristoterezov: Yep. Noticed this on Electron! very strange...

modules/xmpp/xmpp.js
hristoterezov: hmm... in jitsi-meet we don't care...

Actually I don't see benefit because I imagine in both cases you would just ignore this.

modules/xmpp/xmpp.js
hristoterezov: According to strophe's docs and their current source code there isn't a promise. At the beginning I was also hoping there is :) but unfortunately ...

PR #911, opened 2019-06-10
Add error description

AlexRAV:
For now, a developer as a user of the library doesn't see any errors when using .lock() method without moderator role except promise rejection without any error. I've added error throwing with description to clarify the issue for users.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1086,7 +1086,7 @@ JitsiConference.prototype.isModerator = function() {
  */
 JitsiConference.prototype.lock = function(password) {
     if (!this.isModerator()) {
-        throw new Error('You are not moderator.');
+        return Promise.reject('You are not moderator.');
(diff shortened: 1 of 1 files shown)


JitsiConference.js
```diff
@@ -1086,7 +1086,7 @@ JitsiConference.prototype.isModerator = function() {
  */
 JitsiConference.prototype.lock = function(password) {
     if (!this.isModerator()) {
-        return Promise.reject();
+        throw new Error('You are not moderator.');
```
damencho: This changes the behavior of the API which is not desired. Just pass the error when rejecting the promise ...

AlexRAV:
@damencho hello. I've changed throwing an error for a promise rejecting with an error. Should I do something else?

damencho:
Jenkins test this please.

PR #915, opened 2019-06-13
Notify for kick and mute

damencho:


JitsiConference.js
```diff
@@ -135,6 +135,9 @@ export default function JitsiConference(options) {
     };
     this.isMutedByFocus = false;
 
+    // when muted by focus we receive and the jid of the initator of the mute
```
hristoterezov: maybe `and` is redundant?

modules/xmpp/ChatRoom.js
```diff
@@ -804,6 +804,40 @@ export default class ChatRoom extends Listenable {
...
+                = $(pres)
+                .find('>x[xmlns="http://jabber.org/protocol/muc#user"]>item>actor');
+
+            let actorNick;
+
+            if (actorSelect.length) {
+                actorNick = actorSelect.attr('nick');
+            }
+
+            membersKeys.forEach(jid => {
```
hristoterezov: Why not:
```
actorJid  = membersKeys.find(jid => Strophe.getResourceFromJid(jid) === actorNick);
```

JitsiConference.js
```diff
@@ -1411,6 +1423,35 @@ JitsiConference.prototype.onMemberLeft = function(jid) {
...
+ * Designates an event indicating that we were kicked from the XMPP MUC.
+ * @param {boolean} isSelfPresence - whether it is for local participant
+ * or another participant.
+ * @param {string} actorJid - the jid of the participant who was initiator
+ * of the kick.
+ * @param {string?} participantJid - when it is not a kick for local participant,
+ * this is the jid of the participant which was kicked.
+ */
+JitsiConference.prototype.onMemberKicked = function(isSelfPresence, actorJid, participantJid) {
+    const actorId = Strophe.getResourceFromJid(actorJid);
```
hristoterezov: It seems that we are anyway using the resource only, why then we are passing the full jid with the event from ChatRoom.js?

damencho: Yep, will fix it.

PR #917, opened 2019-06-17
fix(caps): validate sha before caching features

virtuacoplenny:
Attemps to fix this race condition:
1. Web client joins a conference
2. Electron client joins the same conference.
3. Web client requests capabilities of Electron
   using version 1111.
4. Electron adds remote control capability,
   settings its version to 2222.
5. Electron responds to web client's requests
   but with features of 2222.
6. Web client caches request under 1111.

### DIFF
--- a/modules/xmpp/Caps.js
+++ b/modules/xmpp/Caps.js
@@ -154,7 +154,7 @@ export default class Caps extends Listenable {
                         const receivedNode = `${user.node}#${sha}`;
 
                         if (receivedNode !== node) {
-                            logger.error(`Expected node ${node} but received ${
+                            logger.warn(`Expected node ${node} but received ${
                                 receivedNode}`);
                         }
 
--- a/modules/xmpp/Caps.js
+++ b/modules/xmpp/Caps.js
@@ -5,6 +5,8 @@ import { b64_sha1, Strophe } from 'strophe.js'; // eslint-disable-line camelcase
 import XMPPEvents from '../../service/xmpp/XMPPEvents';
 import Listenable from '../util/Listenable';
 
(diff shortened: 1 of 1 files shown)


modules/xmpp/Caps.js
```diff
@@ -120,10 +145,20 @@ export default class Caps extends Listenable {
             const node = user ? `${user.node}#${user.version}` : null;
 
             return this._getDiscoInfo(jid, node, timeout)
-                .then(({ features }) => {
+                .then(({ features, identities }) => {
                     if (user) {
-                        // TODO: Maybe use the version + node + hash as keys?
-                        this.versionToCapabilities[user.version] = features;
+                        const sha = generateSha(
```
paweldomas: When I'm looking at XEP[1] it seems that the node should be included in the response. I'm wondering if at the time when it happens the user has already advertised the new version and would comparing the version received to the current one would be possible ? I mention that, because it feels would be more safe to depend on the client's version calculation instead of trying to do it here.

[1]: https://xmpp.org/extensions/xep-0115.html

virtuacoplenny: It could also be that the remote participant advertised an updated version, but the local participant hasn't received that update yet so the version is outdated locally. I initially was trying to fix the issue by checking the version but it wasn't being updated in time, and sometimes it isn't updated at all which looks like another bug.

paweldomas: I see

virtuacoplenny: I kind of do like the idea, though, of including the version in the info query. I'd much rather have that than recalculate the sha. However, this would require patching strohpe.disco.js; right now strohpe.disco.js always has an array of the latest features and for version it uses whatever version is specified in the info query.

PR #919, opened 2019-06-17
fix(conference): fire failed event on o/a failure

hristoterezov:


JitsiConferenceEventManager.js
```diff
@@ -180,15 +184,6 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
         JitsiConferenceEvents.CONFERENCE_FAILED,
         JitsiConferenceErrors.GRACEFUL_SHUTDOWN);
 
-    chatRoom.addListener(XMPPEvents.JINGLE_FATAL_ERROR,
-        (session, error) => {
-            if (!session.isP2P) {
```
paweldomas: The old code would not fail the conference if p2p O/A fails. P2P is kind of optional and it should not break the conference when fails.

hristoterezov: Oh. So are you saying that we don't need a reload because we'll anyway fall back to the jvb peer connection anyway?

saghul: Yep, that's it.

hristoterezov: Yep. Just tested and it starts to use the jvb one. 

It's fixed now to emit the events only for non p2p peerconnection.

PR #920, opened 2019-06-17
fix(presence): send latest presence on conference join

virtuacoplenny:
Attempts to fix the following issue:
1. Electron user starts the conference join process.
2. Electron user adds remote control capability and
   attempts to send presence.
3. Electron user presence send is blocked because the
   conference has not been joined.
4. Electron user joins the conference.
5. The remote control capability is not emitted until
   some other presence change triggers another
   presence send. Until then, other uses will not see
   remote control option available for the electron
   user.

### DIFF
--- a/modules/xmpp/Caps.js
+++ b/modules/xmpp/Caps.js
@@ -5,6 +5,8 @@ import { b64_sha1, Strophe } from 'strophe.js'; // eslint-disable-line camelcase
 import XMPPEvents from '../../service/xmpp/XMPPEvents';
 import Listenable from '../util/Listenable';
 
+const logger = require('jitsi-meet-logger').getLogger(__filename);
+
 /**
  * The property
  */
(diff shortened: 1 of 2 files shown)


damencho:
I see in caps there is a submit parameter to the addFeature method, is this to solve a problem where addFeature was called without allowing it to submit(send presence)?
Maybe we can store the version on joining and when we joined the version is different we can then send the presence? WDYT?

virtuacoplenny:
> I see in caps there is a submit parameter to the addFeature method, is this to solve a problem where addFeature was called without allowing it to submit(send presence)?

I think it was added to be able to call addFeature a bunch of times in a row without triggering a presence update for each. Maybe a better solution would have been to create a helper that could take in a list of features.

> Maybe we can store the version on joining and when we joined the version is different we can then send the presence? WDYT?
So add a conference join listener to caps and also cache the previous version? That would solve this issue but could this issue happen elsewhere too?

virtuacoplenny:
I still made the changes. Hope they're what you were talking about.

paweldomas:
There are many things that are added to presence and may have not been sent, before the room is joined. Many of them must have worked by coincidence that they were not needed, so early. I think the first version is correct to address all the cases.

virtuacoplenny:
@damencho, what do you not like about sending presence again on join?

damencho:
Yep, it is fine.

PR #921, opened 2019-06-19
Always uses 8 hex digits for the MUC nickname/endpoint ID.

bgrozev:
With Octo we need to have all colibri endpoint IDs be 8 hex digits. This PR ensures this by:
1. Disabling the 'useNicks' option which allowed a value from config.js to be prepended to the nickname
2. Disabling prepending the node (part before @) for authenticated users.

In practice this means that jibri will get a new random ID instead of "recorder-xxxxxx" (the x's here being random), and the 'useNicks' option will cease to work. As far as I am aware the 'useNicks' option is not used anywhere, apart from debugging, and nothing depends on the nickname for jibri starting with "recorder" or anything similar for other authenticated users.

If there is a need, we can continue to support the 'useNicks' option, but not together with Octo (we could make that configurable in jicofo).

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -401,26 +401,30 @@ export default class XMPP extends Listenable {
      * @param options
      */
     createRoom(roomName, options) {
-        // By default MUC nickname is the resource part of the JID
-        let mucNickname = Strophe.getNodeFromJid(this.connection.jid);
         let roomjid = `${roomName}@${this.options.hosts.muc}/`;
-        const cfgNickname
-            = options.useNicks && options.nick ? options.nick : null;
 
-        if (cfgNickname) {
-            // Use nick if it's defined
-            mucNickname = options.nick;
-        } else if (!this.authenticatedUser) {
-            // node of the anonymous JID is very long - here we trim it a bit
-            mucNickname = mucNickname.substr(0, 8);
+        let mucNickname;
+
+        // We use the room nickname (the resourcepart of the occupant JID, see XEP-0045)
+        // as the endpoint ID in colibri. We require endpoint IDs to be 8 hex digits because
+        // in some cases they get serialized into a 32bit field.
+        if (this.authenticatedUser) {
+            // Authenticated users h
+            mucNickname = RandomUtil.randomHexString(8).toLowerCase();
+        }
+        else (!this.authenticatedUser) {
+            // We try to use the first part of the node (which for anonymous users on prosody is a UUID) to match
+            // the previous behavior (and maybe make debugging easier).
+            mucNickname = Strophe.getNodeFromJid(this.connection.jid).substr(0, 8).toLowerCase();
+
+            // But if this doesn't have the required format we just generate a new random nickname.
+            let re = /[0-9a-f]{8}/g;
+            if (!re.test(mucNuckname)) {
+                mucNickname = RandomUtil.randomHexString(8).toLowerCase();
+            }
         }
 
-        // Constant JIDs need some random part to be appended in order to be
-        // able to join the MUC more than once.
-        if (this.authenticatedUser || cfgNickname !== null) {
-            mucNickname += `-${RandomUtil.randomHexString(6)}`;
-        }
-
+        logger.info(`Using MUC nickname $mucNickname`);
         roomjid += mucNickname;
 
         return this.connection.emuc.createRoom(roomjid, null, options);
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -409,7 +409,7 @@ export default class XMPP extends Listenable {
         // as the endpoint ID in colibri. We require endpoint IDs to be 8 hex digits because
         // in some cases they get serialized into a 32bit field.
         if (this.authenticatedUser) {
-            // Authenticated users h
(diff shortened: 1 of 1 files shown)


paweldomas:
The PR LGTM, but just out of curiosity: where does the 8 digit limitation come from ?

modules/xmpp/xmpp.js
```diff
@@ -401,26 +401,30 @@ export default class XMPP extends Listenable {
...
-        // able to join the MUC more than once.
-        if (this.authenticatedUser || cfgNickname !== null) {
-            mucNickname += `-${RandomUtil.randomHexString(6)}`;
+        let mucNickname;
+
+        // We use the room nickname (the resourcepart of the occupant JID, see XEP-0045)
+        // as the endpoint ID in colibri. We require endpoint IDs to be 8 hex digits because
+        // in some cases they get serialized into a 32bit field.
+        if (this.authenticatedUser) {
+            // Authenticated users h
```
bbaldino: looks like there was more to this comment?

modules/xmpp/xmpp.js
```diff
@@ -401,26 +401,30 @@ export default class XMPP extends Listenable {
...
+            mucNickname = RandomUtil.randomHexString(8).toLowerCase();
+        }
+        else (!this.authenticatedUser) {
+            // We try to use the first part of the node (which for anonymous users on prosody is a UUID) to match
+            // the previous behavior (and maybe make debugging easier).
+            mucNickname = Strophe.getNodeFromJid(this.connection.jid).substr(0, 8).toLowerCase();
+
+            // But if this doesn't have the required format we just generate a new random nickname.
+            let re = /[0-9a-f]{8}/g;
+            if (!re.test(mucNuckname)) {
```
bbaldino: if we didn't handle this case, the whole block would just become `let mucNickname = RandomUtil.randomHexString(8).toLowerCase();`, right? honestly testing that maybe the first 8 characters were hex only seems slim enough that it doesn't even feel like it's worth handling when we could just replace this all with one line.

bbaldino: and, i'd say, arguably the behavior of _sometimes_ the first bit of the UUID being used and sometimes it not is more confusing than it always being random.

damencho: It is easier to debug, as the connection jids are those visible in prosody logs, as I remember, when bosh connects and then you can correlate them with muc nicks when looking at browser logs or callstats.

bbaldino: but this will only happen in the case that the first 8 digits happen to all be hexadecimal characters.  if we want to make debugging easier i think we should log somewhere an association between the original connection jids and the generated muc usernames.

paweldomas: It is the case for all our current deployments that users have those first 8 digits hexadecimal except for jibris and other authenticated components, so it is worth keeping it. Looking up "some" association may not be so convenient. Also when I look at this PR first thing I would investigate is to see why is there such limitation on the backend instead of trying to kill now our old debugging habits :P

bbaldino: That may be the case, but why rely on behavior we're not enforcing and could easily berak?  I'm saying let's just change the method to:
```
createRoom(roomName, options) {
    let roomjid = `${roomName}@${this.options.hosts.muc}/`;
    let mucNickname = RandomUtil.randomHexString(8).toLowerCase();
    logger.info(`JID ${this.connection.jid} using MUC nickname $mucNickname`);
    roomjid += mucNickname;
    return this.connection.emuc.createRoom(roomjid, null, options);
}
```
This:
1) is consistent for all endpoints (no "if it's X, Y or Z endpoint it's random, otherwise it's not")
2) gives a mapping in the endpoint's logs to the ep id that will be used in the backend (something i think would be really nice, it's then easy to go from a callstats session to looking up what's going on with a certain endpoint in the backend)
3) simplifies the code in this method.

bgrozev: We're not relying on this behavior. We're just using it because it happens to make things easier in practice. If people are already using it I prefer to keep it available at the cost of the extra lines of code. Adding a log is nice (I will do that) but I don't think it helps much (I'd have to go look for the message and then go and amend my grep, and then keep checking the extra mapping if I'm dealing with more than one endpoint).

bgrozev:
@paweldomas I agree that this requirement from the backend is sort of arbitrary and generally not a good idea. We have it because we want to transport the source endpoint ID in Octo packets which have a 32bit fields for this. Otherwise we have to introduce another identifier for endpoints (known by jicofo and the bridges), and it would be a bigger effort.

paweldomas:
@bgrozev thanks for the explanation ! my curiosity is not satisfied :)

modules/xmpp/xmpp.js
paweldomas: @bbaldino now that you have offer an actual solution, I think I kind of like it. Also maybe it's worth to add `logger.info(`JID ${this.connection.jid} using MUC nickname $mucNickname`);` for the current version even without making it random for all yet.

paweldomas: @bbaldino even though I like your solution I would not go with it yet, because it will add a mandatory extra step of figuring out the association where right now it would be only required for jibris.

PR #922, opened 2019-06-25
fix(JingleSessionPC): do not execute peerconnection tasks after close

paweldomas:
Put the close call onto the modification queue and also prevent from executing any pending tasks once the close method has been called.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -196,7 +196,7 @@ export default class JingleSessionPC extends JingleSession {
         this._gatheringReported = false;
 
         this.lasticecandidate = false;
-        this.closed = false;
+        this._closed = false;
 
         /**
          * Indicates whether or not this <tt>JingleSessionPC</tt> is used in
@@ -239,13 +239,13 @@ export default class JingleSessionPC extends JingleSession {
      * cancelled.
      * @param {string} actionName
      * @return {boolean} <tt>true</tt> if this {@link JingleSessionPC} has
-     * entered {@link JingleSessionState.ENDED} or <tt>false</tt> otherwise.
+     * either entered {@link JingleSessionState.ENDED} or if the {@link close}
+     * method has been called.
      * @private
      */
     _assertNotEnded(actionName) {
-        if (this.state === JingleSessionState.ENDED) {
-            logger.log(
-                `The session has ended - cancelling action: ${actionName}`);
+        if (this.state === JingleSessionState.ENDED || this._closed) {
+            logger.log(`The session has ended - cancelling action: ${actionName}`);
 
             return false;
         }
@@ -324,6 +324,10 @@ export default class JingleSessionPC extends JingleSession {
                 // the check complete.
                 return;
             }
+            if (!this._assertNotEnded('onicecandidate')) {
+
+                return;
+            }
 
             // XXX this is broken, candidate is not parsed.
             const candidate = ev.candidate;
@@ -372,15 +376,14 @@ export default class JingleSessionPC extends JingleSession {
         // "closed" instead.
         // I suppose at some point this will be moved to onconnectionstatechange
         this.peerconnection.onsignalingstatechange = () => {
-            if (!this.peerconnection) {
+            if (!this.peerconnection
+                || !this._assertNotEnded('onsignalingstatechange')) {
                 return;
             }
             if (this.peerconnection.signalingState === 'stable') {
                 this.wasstable = true;
-            } else if (
-                (this.peerconnection.signalingState === 'closed'
-                || this.peerconnection.connectionState === 'closed')
-                && !this.closed) {
+            } else if (this.peerconnection.signalingState === 'closed'
+                || this.peerconnection.connectionState === 'closed') {
                 this.room.eventEmitter.emit(XMPPEvents.SUSPEND_DETECTED, this);
             }
         };
@@ -473,9 +476,6 @@ export default class JingleSessionPC extends JingleSession {
                 this.isReconnect = false;
                 break;
             case 'disconnected':
-                if (this.closed) {
-                    break;
-                }
                 this.isReconnect = true;
 
                 // Informs interested parties that the connection has been
@@ -507,6 +507,12 @@ export default class JingleSessionPC extends JingleSession {
                     logger.info('Doing local O/A due to '
                         + 'IS_SELECTED_CHANGED event');
                     this.modificationQueue.push(finishedCallback => {
+                        if (!this._assertNotEnded('setIsSelected O/A')) {
+                            finishedCallback();
+
+                            return;
+                        }
+
                         this._renegotiate()
                             .then(finishedCallback)
                             .catch(finishedCallback);
@@ -727,6 +733,12 @@ export default class JingleSessionPC extends JingleSession {
         // the assumption that candidates are spawned after the offer/answer
         // and XMPP preserves order).
         const workFunction = finishedCallback => {
+            if (!this._assertNotEnded('addIceCandidate')) {
+                finishedCallback();
+
+                return;
+            }
+
             for (const iceCandidate of iceCandidates) {
                 this.peerconnection.addIceCandidate(iceCandidate)
                     .then(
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -196,7 +196,7 @@ export default class JingleSessionPC extends JingleSession {
         this._gatheringReported = false;
 
         this.lasticecandidate = false;
-        this.closed = false;
+        this._closed = false;
```
bbaldino: is there a reason why we have this `closed` member when we already have `this.state`?

bbaldino: by that i mean, why doesn't whatever sets `closed` to true just change the state instead? why 2 things?

paweldomas: Yeah, so the thing is that `state` is updated after things go through a chain of callbacks here and there and when finally `onTerminated` will be called. I supposed it only even gets called when jingle 'session-terminate' is received. So this `state` is more like a Jingle state, but maybe I could use it instead of `_closed`. Currently the `_closed` flag guards against calling close more than once. I was just a bit scared to touch it to be honest :P

paweldomas:
I also plan to try and improve to have `this._assertNotEnded('setOfferAnswer')` called for every worker function in one place

bbaldino:
so i think we should be able to add this check in a single place (in `JingleSessionPC#_processQueueTasks`)

modules/xmpp/JingleSessionPC.js
paweldomas: >by that i mean, why doesn't whatever sets closed to true just change the state instead? why 2 things?

Right, I was afraid to not break some callbacks, but I'll give it a try and we'll see.

bbaldino: gotcha, i see.  i know it was already there, but since we're now relying on it more i was wondering because it seems more complex to have 2 things modeling this state.

modules/xmpp/JingleSessionPC.js
```diff
@@ -503,25 +503,48 @@ export default class JingleSessionPC extends JingleSession {
...
-                        + 'IS_SELECTED_CHANGED event');
-                    this.modificationQueue.push(finishedCallback => {
-                        if (!this._assertNotEnded('setIsSelected O/A')) {
-                            finishedCallback();
-
-                            return;
-                        }
+                    this._executeOnTheQueue('setIsSelected', finishedCallback => {
+                        logger.info('Doing local O/A due to IS_SELECTED_CHANGED event');
+                        this.peerconnection.setIsSelected(isSelected);
```
bbaldino: this used to not be done as part of the queue task but now it is, is that intentional?

modules/xmpp/JingleSessionPC.js
```diff
@@ -503,25 +503,48 @@ export default class JingleSessionPC extends JingleSession {
...
 
                         this._renegotiate()
-                            .then(finishedCallback)
-                            .catch(finishedCallback);
+                            .then(finishedCallback, finishedCallback);
```
bbaldino: personally i prefer the old syntax (explicitly setting `then` and `catch`) because it's clearer but maybe that's just me.

modules/xmpp/JingleSessionPC.js
```diff
@@ -503,25 +503,48 @@ export default class JingleSessionPC extends JingleSession {
...
+        return new Promise((resolve, reject) => {
+            this.modificationQueue.push(
+                finishCallback => {
+                    if (!this._assertNotEnded(actionName)) {
+                        finishCallback();
+
+                        return;
+                    }
+
+                    workFunction(finishCallback);
```
bbaldino: shouldn't there be a `resolve` somewhere in this block?

bbaldino: EDIT: ok, i remember now why we don't see this. was talking with lenny and we thought maybe this would be clearer:

```
_executeOnTheQueue(actionName, workFunction) {
    return new Promise((resolve, reject) => {
        const wrappedWorkFunction = finishedCallback => {
            if (!this.assertNotEnded(actionName)) {
                finishedCallback();
                return;
            }
            workFunction(finishedCallback);
         }
         const handler = error => {
             if (error) { 
                 reject(error);
             } else {
                  resolve();
             }
          this.modificationQueue.push(wrappedWorkFunction, handler);
       });
}
```

i was hoping we could not wrap everything in the promise by doing it that way as well, but think we're stuck there.

modules/xmpp/JingleSessionPC.js
```diff
@@ -733,25 +756,21 @@ export default class JingleSessionPC extends JingleSession {
...
-                        err => logger.error('addIceCandidate failed!', err));
+                allPromises.push(
+                    this.peerconnection.addIceCandidate(iceCandidate)
+                        .then(
+                            () => logger.debug('addIceCandidate ok!'),
+                            err => logger.error('addIceCandidate failed!', err)));
             }
 
-            finishedCallback();
+            Promise.all(allPromises).then(finishedCallback);
```
bbaldino: this is another change in behavior (maybe a bug fix?).  did you find an issue with this part? if not i'm wondering if we should just leave it for now.

modules/xmpp/JingleSessionPC.js
```diff
@@ -886,15 +899,14 @@ export default class JingleSessionPC extends JingleSession {
...
 
-        this.modificationQueue.push(
-            workFunction,
-            error => {
-                if (error) {
-                    logger.error('invite error', error);
-                } else {
+        this._executeOnTheQueue('invite', workFunction)
+            .then(
+                () => {
```
bbaldino: i think linter doesn't want braces for single-line lambdas

modules/xmpp/JingleSessionPC.js
paweldomas: So `then` and `catch` are technically two different things, doing `.catch` on `.then` will also catch any rejections that would happen in the `.then` chain. And usually it is not the intention of the person who writes it from my experience.

modules/xmpp/JingleSessionPC.js
paweldomas: I changed it thinking all peerconnection operations should be done on the queue.
Doing `setIsSelected` can be put into mid-execution of another tasks and I think it's bad.

modules/xmpp/JingleSessionPC.js
paweldomas: It is obviously not the case in a scenario simple like this one

modules/xmpp/JingleSessionPC.js
paweldomas: well yeah I can leave it, but then it will never be fixed and I think it is correct to wait for all promises to happen

paweldomas: I can put into separate commit if that would be better, I can also remove the change if you insist

modules/xmpp/JingleSessionPC.js
bbaldino: that may be true, just wondering if we want to change it as part of this change since it's already a bit complex (potentially).

modules/xmpp/JingleSessionPC.js
bbaldino: if someone wrote catch, i assume they mean to catch.  here, if an exception is throw in the then it'll be unhandled which doesn't seem good?

bbaldino: actually i guess we'd throw twice here (and execute the same method twice), so ignore what i said

modules/xmpp/JingleSessionPC.js
bbaldino: just nervous about all the changes in this commit since the main one is potentially pretty big, thinking it would be nice to isolate.  these bugs (and i do agree, btw, that they are likely bugs) are either 1) not TOO drastic, since we haven't run into them or 2) done this way because of other weird lib-jitsi-meet behavior

paweldomas:
https://github.com/jitsi/lib-jitsi-meet/pull/923

PR #923, opened 2019-06-25
fix(JingleSessionPC): prevent from queueing tasks after close

paweldomas:
Also do not execute any peerconnection callbacks like 'oniceconnectionstatechange', 'onicecandidate' etc. after close.

### DIFF
--- a/modules/util/AsyncQueue.js
+++ b/modules/util/AsyncQueue.js
@@ -1,0 +1,55 @@
+import async from 'async';
+
+/**
+ * A queue for async task execution.
+ */
+export default class AsyncQueue {
+    /**
+     * Creates new instance.
+     */
+    constructor() {
+        this.modificationQueue = async.queue(this._processQueueTasks.bind(this), 1);
+        this.stopped = false;
+    }
+
+    /**
+     * Internal task processing implementation which makes things work.
+     */
+    _processQueueTasks(task, finishedCallback) {
+        task(finishedCallback);
+    }
+
+    /**
+     * The 'workFunction' function will be given a callback it MUST call with either:
+     *  1) No arguments if it was successful or
+     *  2) An error argument if there was an error
+     * If the task wants to process the success or failure of the task, it
+     * should pass the {@code callback} to the push function, e.g.:
+     * queue.push(task, (err) => {
+     *     if (err) {
+     *         // error handling
+     *     } else {
+     *         // success handling
+     *     }
+     * });
+     *
+     * @param {function} workFunction - The task to be execute. See the description above.
+     * @param {function} [callback] - Optional callback to be called after the task has been executed.
+     */
+    push(workFunction, callback) {
+        if (this.stopped) {
+            callback && callback('The queue has been stopped');
+
+            return;
+        }
+        this.modificationQueue.push(workFunction, callback);
+    }
+
+    /**
+     * Shutdowns the queue. All already queued tasks will execute, but no future tasks can be added. If a task is added
+     * after the queue has been shutdown then the callback will be called with an error.
+     */
+    shutdown() {
+        this.stopped = true;
+    }
+}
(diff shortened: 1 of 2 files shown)


modules/util/AsyncQueue.js
```diff
@@ -0,0 +1,55 @@
...
+        task(finishedCallback);
+    }
+
+    /**
+     * The 'workFunction' function will be given a callback it MUST call with either:
+     *  1) No arguments if it was successful or
+     *  2) An error argument if there was an error
+     * If the task wants to process the success or failure of the task, it
+     * should pass the {@code callback} to the push function, e.g.:
+     * queue.push(task, (err) => {
```
bbaldino: `task` -> `workFunction`

(though i like task better here)

modules/xmpp/JingleSessionPC.js
```diff
@@ -2251,18 +2205,18 @@ export default class JingleSessionPC extends JingleSession {
...
 
-        // do not try to close if already closed.
-        this.peerconnection
-            && ((this.peerconnection.signalingState
-                    && this.peerconnection.signalingState !== 'closed')
-                || (this.peerconnection.connectionState
-                    && this.peerconnection.connectionState !== 'closed'))
-            && this.peerconnection.close();
+            // do not try to close if already closed.
+            this.peerconnection && this.peerconnection.close();
```
bbaldino: we could remove the need for a lot of the checks in the peer connection handlers by just assigning them to null before calling close

paweldomas: right

modules/xmpp/JingleSessionPC.js
```diff
@@ -2251,18 +2205,18 @@ export default class JingleSessionPC extends JingleSession {
      * Closes the peerconnection.
      */
     close() {
-        this.closed = true;
+        this.state = JingleSessionState.ENDED;
+        this.establishmentDuration = undefined;
 
-        // The signaling layer will remove it's listeners
-        this.signalingLayer.setChatRoom(null);
+        this.modificationQueue.push(finishCallback => {
```
hristoterezov: Have you considered removing all tasks from queue before adding the close there and shutting it down? Do you think something would break if we do it?

modules/util/AsyncQueue.js
```diff
@@ -0,0 +1,55 @@
...
+     *         // success handling
+     *     }
+     * });
+     *
+     * @param {function} workFunction - The task to be execute. See the description above.
+     * @param {function} [callback] - Optional callback to be called after the task has been executed.
+     */
+    push(workFunction, callback) {
+        if (this.stopped) {
+            callback && callback('The queue has been stopped');
```
saghul: `new Error` ?

modules/util/AsyncQueue.js
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * A queue for async task execution.
+ */
+export default class AsyncQueue {
+    /**
+     * Creates new instance.
+     */
+    constructor() {
+        this.modificationQueue = async.queue(this._processQueueTasks.bind(this), 1);
```
saghul: `_queue` ? AFAIS this can be used for anything.

modules/xmpp/JingleSessionPC.js
bbaldino: the queue has a method `kill` which does this.  it also clears the `drain` handler, but we don't currently rely on that.  i wish it had a better name, though.  it feels a little weird to call `kill` and then add another task.

paweldomas: I will try to see if I can kill the queue and then have the end task to execute

paweldomas: yes, it worked

paweldomas:
ups sorry I forgot to squash

PR #926, opened 2019-06-27
Fix listener issues in emuc Strophe plugin and JitsiConference

pip25:
* If a presence message was received from a room which the Emuc plugin did not know about, it would return undefined, causing Strophe to deregister the listener. Since this is by no means a fatal error, I modified the listeners to return true in such cases as well.
* the "removeCommandListener" method in JitsiConference did not have a handler argument, even though the underlying implementation in ChatRoom required one, so the call actually did not remove anything.


pip25:
Signed.

damencho:
Jenkins, test this, please.

PR #928, opened 2019-07-04
Detects whether a participant is jigasi and sets a property.

damencho:


JitsiConference.js
```diff
@@ -1351,12 +1351,26 @@ JitsiConference.prototype.onMemberJoined = function(
         JitsiConferenceEvents.USER_JOINED,
         id,
         participant);
+
+    const updateFeatures = features => {
```
hristoterezov: Can't we avoid having nested function?

JitsiConference.js
```diff
@@ -1351,12 +1351,26 @@ JitsiConference.prototype.onMemberJoined = function(
...
+    };
+
     this.xmpp.caps.getFeatures(jid)
-        .then(features => {
-            participant._supportsDTMF = features.has('urn:xmpp:jingle:dtmf:0');
-            this.updateDTMFSupport();
-        },
-        error => logger.warn(`Failed to discover features of ${jid}`, error));
+        .then(updateFeatures)
+        .catch(_ => { // eslint-disable-line no-unused-vars
```
hristoterezov: Why the unused var? maybe () => {...} will help fixing the eslint errors, isn't it?

JitsiConference.js
```diff
@@ -1351,12 +1351,26 @@ JitsiConference.prototype.onMemberJoined = function(
...
+            // there was probably a mismatch, lets try one more time and givup
```
hristoterezov: Why do we do the retry?

JitsiConference.js
damencho: Oh, yes.

JitsiConference.js
damencho: Cause we always hit the validation problem that Lenny recently fixed: https://github.com/jitsi/lib-jitsi-meet/commit/a89d67388f49cc18aae080cb388877380a7664b1

hristoterezov: But how we are hitting it since it is fixed?

PR #931, opened 2019-07-10
fix(BridgeChannel): don't log WS onerror events

saghul:
They contain no information about the failure (this is available in the onclose
event) and the event references the WS object itself, which causes hangs on
mobile.

### DIFF
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -276,8 +276,12 @@ export default class BridgeChannel {
             emitter.emit(RTCEvents.DATA_CHANNEL_OPEN);
         };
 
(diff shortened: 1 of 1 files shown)


modules/RTC/BridgeChannel.js
```diff
@@ -276,8 +276,10 @@ export default class BridgeChannel {
             emitter.emit(RTCEvents.DATA_CHANNEL_OPEN);
         };
 
-        channel.onerror = error => {
-            logger.error('Channel error:', error);
+        channel.onerror = event => {
+            if (this._mode !== 'websocket') {
```
hristoterezov: Maybe add a comment why. I guess what you have in the description of the PR will be enough.

PR #934, opened 2019-07-11
Adds confID to JitsiConference option and enables callstats siteID.

damencho:


bruderbarna:
I have a bit of an issue with this pull request.

After this change, you must add a `hosts` property with at least a single `domain` property to `JitsiMeetJS.initJitsiConference` function's second parameter, `options`.

Which would be totally fine, but
1) this optional -> required parameter neccessity change is undocumented (the api description doesn't even mention a `hosts` object in `initJitsiConference`'s options' argument's description)
2) If I'm not mistaken, this parameter is only used for callStats. This means it should not be required if callStatsID and callStatsSecret is null or not passed. AFAIK that means that statistics are disabled. So why create a Statistics object at all?

I'm quite the greenhorn when it comes to this library, so excuse me if I'm making incorrect assumptions.

damencho:
Hum, I will check that. But I do not see how something can work if you do not have hosts.domain in your config ...

damencho:
Hum, yeah hosts config is described in JitsiConnection doc and is not for initJitsiConference, but for jitsi-meet we pass same object values in both places. I will fix that. Thanks for looking into it and reporting.

damencho:
@bruderbarna Here it is https://github.com/jitsi/lib-jitsi-meet/pull/936, wdyt?

PR #935, opened 2019-07-11
feat(browsers): add check for env requiring interaction to unmute

virtuacoplenny:


modules/browser/BrowserCapabilities.js
```diff
@@ -100,6 +100,19 @@ export default class BrowserCapabilities extends BrowserDetection {
...
 
+    /**
+     * Returns whether or not the current environment needs a user interaction
+     * with the page before any unmute can occur.
+     *
+     * @returns {boolean}
+     */
+    isUserInteractionRequiredForUnmute() {
+        return !this.isReactNative()
+            && !this.isChrome()
```
saghul: Rather than being implicit, shouldn't we be explicit here and say `isFirefox` If not `isChrome` is unnecessary, since `isChromiumBased` covers it.

virtuacoplenny: It's hard to say. I haven't tested all browsers (how many are there anyway), and Safari also has the behavior. That's why at the time it was easier for me to do specify the browsers we know to be okay rather than the browsers we know to be bad. I'm okay with going explicitly with listing browsers if you prefer, and we can backfill as needed. (And I'll make the change later today unless you stop me)

saghul: No strong opinion here, just a thought :-)

PR #936, opened 2019-07-15
Fixes obtaining hosts.domain from the correct options, from JitsiConn.

damencho:


bruderbarna:
In reply to https://github.com/jitsi/lib-jitsi-meet/pull/934#issuecomment-511365964:

Thanks for the quick reply, and especially for the quick fix!

Seems more logical to me than before.
But I'm not sure this is 100% perfect. This change makes connection property of the options object absolutely required (instead of hosts), or this code will throw a can not read property options of undefined. Which is not documented yet again, unless `this.connection` is set somewhere I didn't notice.
At first thought I see no problem with having to pass the connection as a property in options, but then the fact that it's absolutely required should be documented.

damencho:
@bruderbarna 

> This change makes connection property of the options object absolutely required (instead of hosts), or this code will throw a can not read property options of undefined

The connection in the options is always passed from the code, and this is internal. https://github.com/jitsi/lib-jitsi-meet/blob/985a0bbfab72f88db668b64c25d5a4224452dfd6/JitsiConnection.js#L116

And `this.connection` is always set, few lines above: https://github.com/jitsi/lib-jitsi-meet/blob/e010df4573e918afde5ca0ab497629c3ef962896/JitsiConference.js#L234

> At first glance I see no problem with having to pass the connection as a property in options, but then the fact that it's absolutely required should be documented.

When this is internal I don't see why it should be documented. You should create JitsiConference using the initJitsiConference method which pass the connection and this is documented.

bruderbarna:
I see now! You're exactly right. Thanks for clarifying.

PR #940, opened 2019-07-17
feat(xmpp): allow setting resource on room jid

virtuacoplenny:
This functionality is being brought back for apps
that use ljm for its prosody logic but not for
video conferencing, ie Jitsi-Meet Spot. The desired
use is for Spot-TVs to be identifiable by their
jid alone and for the resource to be the part
that identifies them.


virtuacoplenny:
If the latest changes are going in the right path I'll need help testing on a deployment with authentication.

virtuacoplenny:
Damencho set up a deployment with auth I could test against and login and failed login still work.

PR #944, opened 2019-07-23
Remove isModerator logic from P2P call start 

daimoc:
Removing the moderator role check in P2P start to apply the "all moderators" logic everytime.
This fix avoid Duplicated P2P “session-initiate” when the 2 pears start the P2P call
and allow P2P call without moderator.


damencho:
Jenkins, test this please.

PR #946, opened 2019-07-24
feat(RTC): drop support for platforms without the MediaStream constru…

saghul:
…ctor

React Native also supports it now.


saghul:
DO NOT merge until https://github.com/jitsi/react-native-webrtc/pull/55 lands please.

PR #948, opened 2019-07-25
feat(JitsiRemoteTrack): Add logs.

hristoterezov:


modules/RTC/JitsiRemoteTrack.js
```diff
@@ -64,26 +73,35 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.muted = muted;
         this.isP2P = isP2P;
 
+        logger.debug(`New remote track added: ${this}`);
```
saghul: please don't log full objects like this, mobile suffers a lot

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -9,6 +9,15 @@ const RTCEvents = require('../../service/RTC/RTCEvents');
 let ttfmTrackerAudioAttached = false;
 let ttfmTrackerVideoAttached = false;
 
+/**
+ * List of container events that we are going to process. _onContainerEventHandler will be added as listener to the
+ * container for every event in the list.
+ */
+const containerEvents = [
+    'play', 'pause', 'waiting', 'suspend', 'stalled', 'ratechange', 'playing', 'loadstart',
```
saghul: sort alphabetically

modules/RTC/JitsiRemoteTrack.js
hristoterezov: Hmm... Why? We have `toString`, isn't it used?

saghul: Oh, we do? Then yeah, no worries!

PR #956, opened 2019-08-06
Make interface compatible with implementation

webintoit:


webintoit:
I have. Few times

modules/xmpp/JingleSession.js
```diff
@@ -190,6 +190,28 @@ export default class JingleSession {
...
+     * which should contain new ICE transport details.
+     * @param jingleOfferElem an element Jingle IQ that contains new offer and
+     *        transport info.
+     * @param success callback called when we succeed to accept new offer.
+     * @param failure function(error) called when we fail to accept new offer.
+     */
+     replaceTransport(jingleOfferElem, success, failure){}
+     
+     /**
+     * Will put and execute on the queue a session modify task. Currently it
```
hristoterezov: Let's modify this comment to be generic. The references for the implementation should be removed. Maybe you can leave it:
```
/**
 * Will put and execute on the queue a session modify task.
 * @param {jQuery} jingleContents - jQuery selector pointing to the jingle element of the session modify IQ. 
 */
```
WDYT?

PR #957, opened 2019-08-07
Refactors the "connection status duration" analytics event.

gpolitis:
This PR refactors 592f72a63c3fd57fc9473a16fa4deaeeeb4f7928. It adds
the video type in the event and it also does a better job at managing
and properly cleaning up the state map.

### DIFF
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -202,6 +202,13 @@ export default class JitsiRemoteTrack extends JitsiTrack {
     }
 
     /**
+     * Gets the content type of this video track.
+     */
+    getVideoType() {
+        return this.videoType;
+    }
+
+    /**
      * Handles track play events.
      */
     _playCallback() {
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -202,13 +202,6 @@ export default class JitsiRemoteTrack extends JitsiTrack {
     }
 
     /**
-     * @returns {VideoType} the video type of this video track.
-     */
-    getVideoType() {
-        return this.videoType;
-    }
-
-    /**
      * Handles track play events.
      */
     _playCallback() {
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -202,7 +202,7 @@ export default class JitsiRemoteTrack extends JitsiTrack {
     }
 
     /**
-     * Gets the content type of this video track.
+     * @returns {VideoType} the video type of this video track.
      */
     getVideoType() {
         return this.videoType;
--- a/modules/connectivity/ParticipantConnectionStatus.js
+++ b/modules/connectivity/ParticipantConnectionStatus.js
@@ -648,12 +648,11 @@ export default class ParticipantConnectionStatusHandler {
                 isInLastN} currentStatus => newStatus: ${
                 participant.getConnectionStatus()} => ${newState}`);
 
-        const oldConnectionStatus = this.connectionStatusMap[id];
+        const oldConnectionStatus = this.connectionStatusMap[id] || {};
 
         // Send an analytics event (guard on either the p2p flag or the connection status has changed
         // since the last time this code block run).
-        if (!oldConnectionStatus
-            || !('p2p' in oldConnectionStatus)
+        if (!('p2p' in oldConnectionStatus)
             || !('connectionStatus' in oldConnectionStatus)
             || oldConnectionStatus.p2p !== inP2PMode
             || oldConnectionStatus.connectionStatus !== newState) {
@@ -662,7 +661,7 @@ export default class ParticipantConnectionStatusHandler {
 
             this.maybeSendParticipantConnectionStatusEvent(id, nowMs);
 
-            this.connectionStatusMap[id] = Object.create(oldConnectionStatus || {}, {
+            this.connectionStatusMap[id] = Object.assign(oldConnectionStatus, {
                 connectionStatus: newState,
                 p2p: inP2PMode,
                 startedMs: nowMs
@@ -885,7 +884,7 @@ export default class ParticipantConnectionStatusHandler {
 
         this.maybeSendParticipantConnectionStatusEvent(id, nowMs);
 
-        this.connectionStatusMap[id] = Object.create(this.connectionStatusMap[id] || {}, {
+        this.connectionStatusMap[id] = Object.assign(this.connectionStatusMap[id] || {}, {
             videoType: type,
             startedMs: nowMs
         });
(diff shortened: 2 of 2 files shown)


modules/RTC/JitsiRemoteTrack.js
```diff
@@ -201,6 +201,13 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.emit(JitsiTrackEvents.TRACK_VIDEOTYPE_CHANGED, type);
     }
 
+    /**
+     * Gets the content type of this video track.
+     */
+    getVideoType() {
```
hristoterezov: Let's define this in JitsiTrack. It could be used for local tracks too.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -624,40 +648,56 @@ export default class ParticipantConnectionStatusHandler {
...
+        if (!('p2p' in oldConnectionStatus)
+            || !('connectionStatus' in oldConnectionStatus)
+            || oldConnectionStatus.p2p !== inP2PMode
+            || oldConnectionStatus.connectionStatus !== newState) {
+
+            const nowMs = Date.now();
+
+            this.maybeSendParticipantConnectionStatusEvent(id, nowMs);
 
+            this.connectionStatusMap[id] = Object.assign(oldConnectionStatus, {
```
hristoterezov: you can use:
```
this.connectionStatusMap[id] = {
    ...oldConnectionStatus,
     connectionStatus: newState,
     p2p: inP2PMode,
     startedMs: nowMs
};
```

Seems a little bit more readable to me :) WDYT?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -624,40 +648,56 @@ export default class ParticipantConnectionStatusHandler {
...
+                connectionStatus: newState,
+                p2p: inP2PMode,
+                startedMs: nowMs
+            });
+
+            // sometimes (always?) we're late to hook the TRACK_VIDEOTYPE_CHANGED event and the
+            // video type is not in oldConnectionStatus.
+            if (!('videoType' in this.connectionStatusMap[id])) {
+                this.connectionStatusMap[id].videoType = participant.getTracks().some(
```
hristoterezov: I'd select the first video track and return its videotype. Maybe you can use `JitsiParticipant.getTracksByMediaType`. I think it will be a little bit more clear what we try to do. WDYT?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -739,6 +779,29 @@ export default class ParticipantConnectionStatusHandler {
...
+    /**
+     * Flushes out any remaining participant connection status events as a result of the user
+     * leaving the conference.
+     */
+    onConferenceLeft() {
+        const nowMs = Date.now();
+
+        // eslint-disable-next-line guard-for-in
+        for (const id in this.connectionStatusMap) {
+            this.maybeSendParticipantConnectionStatusEvent(id, nowMs);
```
hristoterezov: You can use `this.onUserLeft` here, isn't it?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -739,6 +779,29 @@ export default class ParticipantConnectionStatusHandler {
...
     }
 
+    /**
+     * Flushes out any remaining participant connection status events as a result of the user
+     * leaving the conference.
+     */
+    onConferenceLeft() {
+        const nowMs = Date.now();
+
+        // eslint-disable-next-line guard-for-in
```
hristoterezov: Can't we somehow remove the eslint disable?

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -624,40 +648,56 @@ export default class ParticipantConnectionStatusHandler {
...
+                        && track.getVideoType() === VideoType.DESKTOP) ? VideoType.DESKTOP : VideoType.CAMERA;
+            }
+        }
         this._changeConnectionStatus(participant, newState);
     }
 
+    /**
+     * Computes the duration of the current connection status for the participant with the specified id (i.e. 15 seconds
+     * in the INTERRUPTED state) and sends a participant connection status event.
+     * @param id the jid of the participant
```
hristoterezov: It will be good if you add the type of the param like this:
```
 *  @param {string} id - The jid of the participant.
```

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -624,40 +648,56 @@ export default class ParticipantConnectionStatusHandler {
...
+     * @param nowMs the current time (in millis)
+     */
```
hristoterezov: Add `@returns {void}`.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -739,6 +779,29 @@ export default class ParticipantConnectionStatusHandler {
         return false;
     }
 
+    /**
+     * Flushes out any remaining participant connection status events as a result of the user
+     * leaving the conference.
```
hristoterezov: Same as above

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -739,6 +779,29 @@ export default class ParticipantConnectionStatusHandler {
...
+        // eslint-disable-next-line guard-for-in
+        for (const id in this.connectionStatusMap) {
+            this.maybeSendParticipantConnectionStatusEvent(id, nowMs);
+            delete this.connectionStatusMap[id];
+        }
+    }
+
+    /**
+     * Sends a last/final participant connection status event for the participant that left the conference.
+     * @param id the id of the participant that left the conference.
```
hristoterezov: Add type and returns please.

modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -808,4 +871,22 @@ export default class ParticipantConnectionStatusHandler {
 
         this.figureOutConnectionStatus(participantId);
     }
+
+    /**
+     * Sends a participant connection status event as a result of the video type
+     * changing.
+     * @param track the track
+     * @param type the video type
```
hristoterezov: Same as above.

modules/RTC/JitsiRemoteTrack.js
gpolitis: unfortunately we don't have access to the videoType field in JitsiTrack but I can add a corresponding getter in JitsiLocalTrack?

modules/connectivity/ParticipantConnectionStatus.js
gpolitis: sure, just to make sure i understand, you mean not use `Object.assign` but copy the fields one by one instead?

modules/RTC/JitsiRemoteTrack.js
saghul: FTR, there is lots of code already accessing videoType directly.

hristoterezov: @gpolitis , Actually @saghul is correct!

Could you please remove the method and just use it directly! I don't think this is good practice but let's be consistent :) WDYT

modules/connectivity/ParticipantConnectionStatus.js
hristoterezov: not one by one but using the spread operator(https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax). you can just copy the code from above :)

But this is optional, so if you prefer the `Object.assign` please keep it!

gpolitis: i feel like an old-schooler but TIL :)

PR #959, opened 2019-08-13
fix(connection-status): pass expected args to type change listener

virtuacoplenny:
The event TRACK_VIDEOTYPE_CHANGED passes only the video type
along, not the track.

This does remove a listener removal call. Alternatives thought of
- keep a cache of attached listeners to call off from
- change TRACK_VIDEOTYPE_CHANGED to pass "this" or the track ID as a second argument (added second to preserve backwards compatibiltiy)

### DIFF
--- a/modules/connectivity/ParticipantConnectionStatus.js
+++ b/modules/connectivity/ParticipantConnectionStatus.js
@@ -506,7 +506,7 @@ export default class ParticipantConnectionStatusHandler {
                 this._onSignallingMuteChanged);
             remoteTrack.on(
                 JitsiTrackEvents.TRACK_VIDEOTYPE_CHANGED,
-                this._onTrackVideoTypeChanged);
+                videoType => this._onTrackVideoTypeChanged(remoteTrack, videoType));
         }
     }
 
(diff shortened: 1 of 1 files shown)


modules/connectivity/ParticipantConnectionStatus.js
```diff
@@ -528,9 +528,6 @@ export default class ParticipantConnectionStatusHandler {
             remoteTrack.off(
                 JitsiTrackEvents.TRACK_MUTE_CHANGED,
                 this._onSignallingMuteChanged);
-            remoteTrack.off(
```
hristoterezov: We don't need to remove the listener?

virtuacoplenny: To encapsulate, yes. Practically and specifically for jitsi-meet, no, because jitsi-meet calls to remove all listeners. This is why I offered alternatives, and this was the "easiest" alternative.

hristoterezov: I agree this is the easiest way but shouldn't we try to remove the listener?

gpolitis: if we want to remove the listener I find the 2nd alternative most appealing because it avoids introducing extra state variables; I'm also happy to implement the 1st one or leave this PR as is. @hristoterezov do you have a preference?

hristoterezov: No

PR #961, opened 2019-08-16
ref(xmpp): remove references to moderate module

virtuacoplenny:
I don't see the "moderate" functionality being set somewhere and it is undefined when I try to call  xmpp#setMute and xmpp#eject directly. Not sure if these should be left in for some reason.


virtuacoplenny:
Jenkins test this please.

PR #963, opened 2019-08-20
deps: prepare for react-native 0.60

saghul:


saghul:
Merge together with https://github.com/jitsi/jitsi-meet/pull/4550

PR #968, opened 2019-09-03
fix(JingleSessionPC): no media flowing after ICE restart

paweldomas:
If DTLS fingerprint remains unchanged after ice-restart there will be no media flow even thought the ICE connection will be re-established.

The workaround is to set an all zeros fingerprint on the hacky offer/answer cycle which is currently done in order to reset the data channels and the sequence number counters on the RTP receivers.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -1040,6 +1040,15 @@ export default class JingleSessionPC extends JingleSession {
             .find('>content>description>ssrc-group')
             .remove();
 
+        // On the JVB it's not a real ICE restart and all layers are re-initialized from scratch as Jicofo does
+        // the restart by re-allocating new channels. Chrome (or WebRTC stack) needs to have the DTLS transport layer
+        // reset to start a new handshake with fresh DTLS transport on the bridge. Make it think that the DTLS
+        // fingerprint has changed by setting an all zeros key.
+        const newFingerprint = jingleOfferElem.find('>content>transport>fingerprint');
+
+        newFingerprint.attr('hash', 'sha-1');
+        newFingerprint.text('00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00:00');
+
         // First set an offer with a rejected 'data' section
         this.setOfferAnswerCycle(
             jingleOfferElem,


bbaldino:
did you try using this? https://developer.mozilla.org/en-US/docs/Web/API/RTCOfferOptions/iceRestart

paweldomas:
> did you try using this? https://developer.mozilla.org/en-US/docs/Web/API/RTCOfferOptions/iceRestart

This is used to create an ICE restart offer, but it does not apply in our case where it's Jicofo that creates it.

bbaldino:
So how does a proper fingerprint end up getting generated?

bbaldino:
> > did you try using this? https://developer.mozilla.org/en-US/docs/Web/API/RTCOfferOptions/iceRestart
> 
> This is used to create an ICE restart offer, but it does not apply in our case where it's Jicofo that creates it.

could we have Jicofo do the ice restart btw?

paweldomas:
> So how does a proper fingerprint end up getting generated?

You'd have to check the entire method. What happens there is when a 'transport-replace' offer arrives it contains the correct fingerprint generated by the JVB. That offer is modified to remove all remote SSRCs, disable data transport to re-initialize SCTP and now I'm adding a change which will put a dummy fingerprint to reset the DTLS layer. The modified offer is set first and full sRD/sLD cycle occurs after which the original 'transport-replace' offer is applied which contains all the valid data.

paweldomas:
> could we have Jicofo do the ice restart btw?

I'm not sure what do you mean exactly. It currently does the restart by reallocating channels on the bridge when the client sends a notification. That results in new ufrag/passwd which is the trigger for ICE restart on the client.

bbaldino:
> > could we have Jicofo do the ice restart btw?
> 
> I'm not sure what do you mean exactly. It currently does the restart by reallocating channels on the bridge when the client sends a notification. That results in new ufrag/passwd which is the trigger for ICE restart on the client.

Ok, guess I'm misunderstanding.  The incoming offer has a new ufrag and password but not a new fingerprint?

paweldomas:
> Ok, guess I'm misunderstanding. The incoming offer has a new ufrag and password but not a new fingerprint?

The incoming offer has everything what's needed.
LJM modifies this offer in order to reset things and only after that munged offer is executed it will set the original one received from Jicofo.

bbaldino:
> > Ok, guess I'm misunderstanding. The incoming offer has a new ufrag and password but not a new fingerprint?
> 
> The incoming offer has everything what's needed.
> LJM modifies this offer in order to reset things and only after that munged offer is executed it will set the original one received from Jicofo.

I get that, but the reason we need to wipe out the fingerprint in that incoming offer is because it's the same, right?

paweldomas:
> I get that, but the reason we need to wipe out the fingerprint in that incoming offer is because it's the same, right?

Yes. Currently JVB will change it only after 24 hours. The operation of generating new certificate is rather expensive, so in discussions with Boris and Emil we thought it's better to cheat the client.

bbaldino:
> > I get that, but the reason we need to wipe out the fingerprint in that incoming offer is because it's the same, right?
> 
> Yes. Currently JVB will change it only after 24 hours. The operation of generating new certificate is rather expensive, so in discussions with Boris and Emil we thought it's better to cheat the client.

I was talking with jonathan about this...are we sure chrome needs to do an entire DTLS restart and not just an ICE restart?  it doesn't seem like the fingerprint would need to change to redo ICE

paweldomas:
> I was talking with jonathan about this...are we sure chrome needs to do an entire DTLS restart and not just an ICE restart? it doesn't seem like the fingerprint would need to change to redo ICE

So it's not about Chrome, but about the bridge which will have all layers reset and not just ICE. I've updated the comment to describe it as JVB's limitation.

webintoit:
> > I was talking with jonathan about this...are we sure chrome needs to do an entire DTLS restart and not just an ICE restart? it doesn't seem like the fingerprint would need to change to redo ICE
> 
> So it's not about Chrome, but about the bridge which will have all layers reset and not just ICE. I've updated the comment to describe it as JVB's limitation.

Could you create new RTCPeerconnection here ? No tricks will be needed also for datachannel

paweldomas:
> Could you create new RTCPeerconnection here ? No tricks will be needed also for datachannel

I could, but that would make the code much more complex than that.

PR #971, opened 2019-09-06
Avoid bundle bloat

saghul:


saghul:
Green CI: http://ci.jitsi.org/view/pr-testing/job/jitsi-meet-PR-test/6658/testReport/

PR #978, opened 2019-09-26
Uses sendBeacon to send presence unavailable on unload.

damencho:


damencho:
Yep, we already are doing it for some other stuff, anyway. I was in a talk with the Strophe project lead about it, but implementing it there will take time, which we do not have. Chrome 78 will be released in less than 4 weeks and we need it :)

paweldomas:
Does it make sense to skip Strophe disconnect in case beacon is supported instead of doing both?

damencho:
It clears its internals and schedules a request to be sent as an XHR which got ignored by chrome ... So in FF we may end up sending to unavailable messages ... Need to check what exactly is the internal cleanup ... and whether we can just skip the the XHR send ...

paweldomas:
So when a beacon is sent the browser quits doing any other stuff or any other XHRs?

damencho:
Not actually. 
```
The Beacon interface schedules an asynchronous and non-blocking request to a web server.

Beacon requests use HTTP POST and do not require a response.
Beacon requests are guaranteed to be initiated before the page unloads.
```

Chrome just disable sync XHR on unload: https://bugs.chromium.org/p/chromium/issues/detail?id=827324

It continues to execute on unload but does not send sync XHR scheduled during this time, a thing we were using to be sure presence unavailable will be delivered.

PR #983, opened 2019-10-02
ref(JitsiConference): move sendTones impl to TPC

paweldomas:
Moves sendTones implementation to TraceablePeerConnection. It will be responsible for creating/storing a DTMFSender. JitsiConference will use the currently active TraceablePeerConnection to send the tones.

This fixes a bug where DTMF tones can not be sent after JingleSessionPC/TraceablePeerConnection instance changes. The JitsiDTMFManager instance was referencing invalid PeerConnection.


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2053,6 +2061,40 @@ TraceablePeerConnection.prototype.setVideoTransferActive = function(active) {
...
+ * @param {string} tones - The DTMF tones string as defined by {@code RTCDTMFSender.insertDTMF}, 'tones' argument.
+ * @param {number} duration - The amount of time in milliseconds that each DTMF should last. It's 200ms by default.
+ * @param {number} interToneGap - The length of time in miliseconds to wait between tones. It's 200ms by default.
+ *
+ * @returns {void}
+ */
+TraceablePeerConnection.prototype.sendTones = function(tones, duration = 200, interToneGap = 200) {
+    if (!this._dtmfSender) {
+        if (this.peerconnection.getSenders) {
+            this._dtmfSender = this.peerconnection.getSenders().map(s => s.dtmf)[0];
```
virtuacoplenny: Why map over all senders to get the first one?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2053,6 +2061,40 @@ TraceablePeerConnection.prototype.setVideoTransferActive = function(active) {
...
+            this._dtmfSender && logger.info(`${this} initialized DTMFSender using getSenders`);
+        }
+
+        if (!this._dtmfSender) {
+            const localAudioTrack = Array.from(this.localTracks.values()).find(t => t.isAudioTrack());
+
+            this._dtmfSender
+                = this.peerconnection.createDTMFSender
+                && localAudioTrack
```
virtuacoplenny: _dtmfSender could become a boolean. Not sure if that matters.

modules/RTC/TraceablePeerConnection.js
paweldomas: This is the only way I figured to do it in one line. Another way I was trying was to do `find(s => s.dtmf)`, but then you have to do `this._dtmfSender = result.dtmf`. It would be something like this:

```
const rtpSender = this.peerconnection.getSenders().find(s => s.dtmf);
this._dtmfSender = rtpSender.dtmf;
```
Do you prefer it this way or have another idea?

modules/RTC/TraceablePeerConnection.js
paweldomas: I was thinking it doesn't, because still falsy, but not 100% accurate.

modules/RTC/TraceablePeerConnection.js
paweldomas: Maybe it's more readable this way? WDYT?

virtuacoplenny: There's a pretty big difference between find and .map in that case though. With map, if index 0 does not have .dtmf the [0] will still take the undefined, whereas with the find would skip index 0. So that becomes the question, are you looking for the first dtmf in the senders array? Is every sender guaranteed to have dtmf?

paweldomas: hmm my observation was that doing map to undefined would kind of remove the item, but maybe I'm wrong will try that again

paweldomas: No, you're right, it would map to undefined

PR #992, opened 2019-10-16
Feature/set maxbitrate desktop

jallamsetty1:
Changes for applying max bitrate on the desktop stream
- Simulcast is disabled on desktop share for now
- max bit rate is set to 500Kbps for desktop share for JVB peerconnections
- cached Primary SSRCs are cleared when we switch between different video sources prompting the client to send out source-remove/source-add to the JVB
- contentHint on dekstop track is set to "detail" when a desktop share is started


JitsiMeetJS.js
```diff
@@ -386,6 +386,15 @@ export default _mergeNamespaceAndModule({
                     }
                 }
 
+                // set the contentHint to "detail" for desktop tracks
+                // eslint-disable-next-line prefer-const
+                for (let track of tracks) {
```
saghul: you can use const here no problem

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1883,6 +1885,45 @@ TraceablePeerConnection.prototype._insertUnifiedPlanSimulcastReceive
...
+            return;
+        }
+        const videoType = localTrack.videoType;
+        const trackId = localTrack.track.id;
+
+        this.peerconnection.getSenders()
+            .filter(s => s.track && s.track.id === trackId)
+            .forEach(sender => {
+                // eslint-disable-next-line prefer-const
+                let parameters = sender.getParameters();
```
saghul: you don't re-assign it, so you can use const here

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1883,6 +1885,45 @@ TraceablePeerConnection.prototype._insertUnifiedPlanSimulcastReceive
...
+
+                if (parameters.encodings && parameters.encodings.length) {
+                    logger.info('Setting max bitrate on video stream');
+                    parameters.encodings.forEach((encoding, idx) => {
```
saghul: can you use for..in here?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1883,6 +1885,45 @@ TraceablePeerConnection.prototype._insertUnifiedPlanSimulcastReceive
...
+                        encoding.maxBitrate = videoType === 'desktop'
```
saghul: does it make sense to check if `maxBitrate` is supported?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1669,7 +1669,8 @@ export default class JingleSessionPC extends JingleSession {
             // NOTE the code below assumes that no more than 1 video track
             // can be added to the peer connection.
             // Transition from no video to video (possibly screen sharing)
-            if (!oldTrack && newTrack && newTrack.isVideoTrack()) {
+            // or transition from camera to share
```
saghul: please finish the sentence with a period.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -26,6 +26,8 @@ const SIM_LAYER_1_RID = '1';
 const SIM_LAYER_2_RID = '2';
 const SIM_LAYER_3_RID = '3';
 const SIM_LAYER_RIDS = [ SIM_LAYER_1_RID, SIM_LAYER_2_RID, SIM_LAYER_3_RID ];
+const SIM_LAYER_BITRATES = [ 500000, 700000, 2000000 ];
```
gpolitis: Do we care to set a max for normal video encodings? If so maybe we should at least try to match the values in the client side engine https://cs.chromium.org/chromium/src/third_party/webrtc/media/engine/simulcast.cc?l=74&rcl=2f6e525099a0d5186e6b5b130aa26b78158d5aa7

gpolitis:
> cached Primary SSRCs are cleared when we switch between different video sources prompting the client to send out source-remove/source-add to the JVB

I don't see this in the RP, maybe it's not pushed yet? Regardless, I think re-using ssrcs was explicitly implemented in order to avoid having to re-negotiate across all clients. I could be wrong though, my memory isn't as good as it used to be and unfortunately we're not good at documenting technical choices /cc @paweldomas or @bbaldino or @bgrozev.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Sure

modules/RTC/TraceablePeerConnection.js
jallamsetty1: ok

JitsiMeetJS.js
jallamsetty1: ok sure

modules/RTC/TraceablePeerConnection.js
jallamsetty1: maxBitrate property in the encoding object is not usually set the first time when the encoder is created so I can't check for that. Maybe I can check if it is actually set after it is applied, will that suffice ? maxBitrate is supported from Chrome 69 onwards.

modules/xmpp/JingleSessionPC.js
jallamsetty1: sure

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Thanks George, will take the max bitrate values from here.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -26,6 +26,8 @@ const SIM_LAYER_1_RID = '1';
 const SIM_LAYER_2_RID = '2';
 const SIM_LAYER_3_RID = '3';
 const SIM_LAYER_RIDS = [ SIM_LAYER_1_RID, SIM_LAYER_2_RID, SIM_LAYER_3_RID ];
+const SIM_LAYER_BITRATES = [ 200000, 700000, 2500000 ];
```
bbaldino: can you add a unit here? `SIM_LAYER_BITRATES_BPS`?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1883,6 +1885,50 @@ TraceablePeerConnection.prototype._insertUnifiedPlanSimulcastReceive
...
+ * bitrate of the enocder doesn't exceed the configured value.
+ * This is needed for the desktop share until spec-complaint
+ * simulcast is implemented.
+ * @param {JitsiLocalTrack} localTrack - the local track whose
+ * max bitrate is to be configured.
+ */
+TraceablePeerConnection.prototype.setMaxBitRate = function(localTrack) {
+    const mediaType = localTrack.type;
+
+    if (mediaType !== MediaType.AUDIO) {
```
bbaldino: nit: can we do:
```
if (mediaType === MediaType.AUDIO) {
  return;
}
```
to avoid the extra layer of indentation for everything below it?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2500,7 +2549,12 @@ TraceablePeerConnection.prototype.generateNewStreamSSRCInfo = function(track) {
     if (ssrcInfo) {
         logger.error(`Will overwrite local SSRCs for track ID: ${rtcId}`);
     }
-    if (this.isSimulcastOn()) {
+
+    // configure simulcast only for camera tracks
+    const hasCameraTrack = this.getLocalTracks()
```
bbaldino: we use this twice, add a helper?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1668,8 +1668,10 @@ export default class JingleSessionPC extends JingleSession {
 
             // NOTE the code below assumes that no more than 1 video track
             // can be added to the peer connection.
-            // Transition from no video to video (possibly screen sharing)
-            if (!oldTrack && newTrack && newTrack.isVideoTrack()) {
+            // Transition from no video to video (unmute),
+            // transition from camera to desktop share
+            // or transition from one camera source to another.
+            if (newTrack && newTrack.isVideoTrack()) {
```
bbaldino: can you remind me what `isVideoTrack` denotes?

paweldomas: it tells whether it's an audio or a video track

modules/RTC/TraceablePeerConnection.js
jallamsetty1: sure

modules/RTC/TraceablePeerConnection.js
jallamsetty1: ok sure

modules/xmpp/JingleSessionPC.js
bbaldino: in that case @jallamsetty1 will this trigger on a camera change?

paweldomas: yes, it will

modules/RTC/TraceablePeerConnection.js
jallamsetty1: ok

gpolitis:
I was wondering how we can validate that this has had an impact (hopefully positive) on the user experience. Would it make sense to make this code configurable, do an A/B test and observe the impact on the freeze durations dashboard that we keep in Amplitude?

This approach would involve extending config.js with two new knobs: one for setting the max bitrate and another for enabling screen-share simulcast and in the events that we send in amplitude, we include whether any of the two knobs was activated.

```
testing: {
        octo: {
            probability: 1
        },
        screensharingSimulcast: .5,
        setMaxBitrates: .5
    }
```

jallamsetty1:
> I was wondering how we can validate that this has had an impact (hopefully positive) on the user experience. Would it make sense to make this code configurable, do an A/B test and observe the impact on the freeze durations dashboard that we keep in Amplitude?
> 
> This approach would involve extending config.js with two new knobs: one for setting the max bitrate and another for enabling screen-share simulcast and in the events that we send in amplitude, we include whether any of the two knobs was activated.
> 
> ```
> testing: {
>         octo: {
>             probability: 1
>         },
>         screensharingSimulcast: .5,
>         setMaxBitrates: .5
>     }
> ```

jallamsetty1:
Closed the pull request accidently, reopening it again :)

jallamsetty1:
> or even something more specific to also suggest that it involves solving a bitrate problem in lieu of simulcast? i just worry it's a bit vague as-is

@bbaldino, how about `enableNonSimulcastScreenshare` ? Are you talking about just the flag in config.js or everywhere else ?

bbaldino:
> > or even something more specific to also suggest that it involves solving a bitrate problem in lieu of simulcast? i just worry it's a bit vague as-is
> 
> @bbaldino, how about `enableNonSimulcastScreenshare` ? Are you talking about just the flag in config.js or everywhere else ?

I was trying to think of something that makes it clear we're 1) addressing a bitrate problem and 2) we have 2 solutions: simulcast and limiting the max bitrate.  so maybe it's `capScreenshareBitrate` or something?  that's not so clear that it's an either-or between this or simulcast, but i think we could make that clear via comments.

jallamsetty1:
> > > or even something more specific to also suggest that it involves solving a bitrate problem in lieu of simulcast? i just worry it's a bit vague as-is
> > 
> > 
> > @bbaldino, how about `enableNonSimulcastScreenshare` ? Are you talking about just the flag in config.js or everywhere else ?
> 
> I was trying to think of something that makes it clear we're 1) addressing a bitrate problem and 2) we have 2 solutions: simulcast and limiting the max bitrate. so maybe it's `capScreenshareBitrate` or something? that's not so clear that it's an either-or between this or simulcast, but i think we could make that clear via comments.

I think we can go with `capScreenshareBitrate` for now and make it clear via comments. I thought this change was going to be temporary until we implement spec-compliant simulcast for screenshare so this should do for now.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: done

PR #993, opened 2019-10-16
fix: npm run test will fail when karma is not installed globally

paweldomas:


paweldomas:
I don't really know what I'm doing here... @saghul @virtuacoplenny do you know why 'npm run test' would fail with karma not found? It sounds like it expect it to be installed globally, or maybe "npm install' is missing?

paweldomas:
Okay I think it's missing `npm install` in the PR script

PR #996, opened 2019-10-23
Add a version check to force user interaction to unmute for older Firefox only

daimoc:
Since version 68 , Firefox considers non-secure origins for getUserMedia request :
https://wiki.mozilla.org/Media/WebRTC/ReleaseNotes/68 : bug 1335740 Disable getUserMedia on non-secure origins.
So lib-jitsi-meet don't need to require a user interaction to unmute media for Firefox 68 and later.


damencho:
Jenkins, test this please.

PR #1008, opened 2019-12-03
fix(presenter mode): Resolve promise instead of throwing error when presenter track is unmuted

jallamsetty1:
When the presenter track is unmuted, we do not need to add the track to the conference.
Currently, the lib tries to lookup for the presenter stream by comparing the MediaStreamTrack kind to that of the JitsiLocalTrack media type which is different from "video". When the stream is not found, it throws a TRACK_NO_STREAM_FOUND error. 
We now resolve the promise immediately and not proceed with adding tracks to the conference if the track media type is set to "presenter".

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -525,6 +525,10 @@ export default class JitsiLocalTrack extends JitsiTrack {
             }
 
             promise.then(streamsInfo => {
+                // Do not add the presenter track to the conference
+                if (this.getType() === 'presenter') {
+                    return Promise.resolve();
+                }
                 const mediaType = this.getType();
                 const streamInfo
                     = browser.usesNewGumFlow()
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -525,11 +525,8 @@ export default class JitsiLocalTrack extends JitsiTrack {
             }
 
             promise.then(streamsInfo => {
-                // Do not add the presenter track to the conference
-                if (this.getType() === 'presenter') {
-                    return Promise.resolve();
-                }
-                const mediaType = this.getType();
+                // The track kind for presenter track is video as well.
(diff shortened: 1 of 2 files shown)


saghul:
Can you add `presenter` here: https://github.com/jitsi/lib-jitsi-meet/blob/master/service/RTC/MediaType.js and then use the constant for the comparison?

saghul:
Also, please try to keep the commit message < 80 chars.

paweldomas:
> Can you add `presenter` here: https://github.com/jitsi/lib-jitsi-meet/blob/master/service/RTC/MediaType.js and then use the constant for the comparison?

I've been wondering if it makes sense to have 1-1 Jitsi track type to media type mapping. `presenter` track is still a `video` media type.

paweldomas:
Also I've noticed that in jitsi-meet muting presenter track is used in place of disposing/removing the track which should be changed as well.

jallamsetty1:
@paweldomas I agree, I am working on re-using the JitsiLocalTrack on presenter un-mute. This change would not be necessary when that is done. I will just have to just fix the way we look up presenter track, i.e. change mediaType to 'video' when it is 'presenter' and not resolve the promise early.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -525,7 +525,8 @@ export default class JitsiLocalTrack extends JitsiTrack {
             }
 
             promise.then(streamsInfo => {
-                const mediaType = this.getType();
+                // The track kind for presenter track is video as well.
+                const mediaType = this.getType() === MediaType.PRESENTER ? MediaType.VIDEO : this.getType();
```
paweldomas: I can't find `MediaType.PRESENTER` defined anywhere. Also I just noticed that in jitsi-meet the code overrides the track type: https://github.com/jitsi/jitsi-meet/blob/master/react/features/base/tracks/functions.js#L49 This type should live inside LJM. `track.type` is not a public var to be modified from jitsi-meet

jallamsetty1: Oops, I forgot to commit the changes to service/RTC/MediaType.js.

PR #1009, opened 2019-12-09
fix: Use RTCRtpSender.replaceTrack() for applying presenter effect on FF

jallamsetty1:


modules/RTC/JitsiLocalTrack.js
```diff
@@ -338,7 +338,6 @@ export default class JitsiLocalTrack extends JitsiTrack {
         if (this._streamEffect) {
             this._streamEffect.stopEffect();
             this._setStream(this._originalStream);
-            this._originalStream = undefined;
```
saghul: what is this change about?

JitsiConference.js
```diff
@@ -989,6 +989,32 @@ JitsiConference.prototype.replaceTrack = function(oldTrack, newTrack) {
...
 };
 
+/**
+ * Replaces the track at the lower level by going through the Jingle session
+ * and WebRTC peer connection. The track is replaced without the need for an
+ * offer/answer cycle.
+ * @param {JitsiLocalTrack} localTrack - the local track whose media stream has
+ * been updated.
+ */
+JitsiConference.prototype.replaceTrackWithoutOfferAnswer = function(localTrack) {
```
saghul: Not a big fan of this new API. Why not make this an internal function which gets called by `JitsiConference.prototype.replaceTrack` if supported?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -392,6 +391,22 @@ export default class JitsiLocalTrack extends JitsiTrack {
 
         this._setEffectInProgress = true;
 
+        // For firefox/safari, replace the stream without doing a offer answer with the remote peer.
```
saghul: Please don't mention browser names since that's bound to change.

jallamsetty1: ok

modules/RTC/JitsiLocalTrack.js
jallamsetty1: I will have to keep the original stream around so that I can remove it from the _addedStreams array in TraceablePeerconnection after the media stream is replaced.

PR #1014, opened 2019-12-19
Add conference duration timer

theunafraid:


modules/xmpp/xmpp.js
```diff
@@ -713,16 +720,16 @@ export default class XMPP extends Listenable {
...
      *
      * @param {string} msg - The message.
      */
     _onPrivateMessage(msg) {
         const from = msg.getAttribute('from');
 
-        if (!this.speakerStatsComponentAddress
-            || from !== this.speakerStatsComponentAddress) {
+        if (!((from === this.speakerStatsComponentAddress)
+            || (from === this.conferenceDurationComponentAddress))) {
```
damencho: Doesn't lint complain about the extra brackets?

damencho: Or just simplify it a bit ...

theunafraid: Removed the extra brackets

Issue #464, opened 2017-04-28
Multiple Video Tracks 


bilhamil:
Currently, lib-jitsi-meet only allows an endpoint to add one video track to a conference. I am wondering if it is at all planned to eventually support adding multiple video tracks? Is there a suggested work around? The most obvious use case would be having a webcam plus screenshare. I will note that, in my particular application, I would like to be able to support an arbitrary number of video tracks. My understanding is that the videobridge supports endpoints sending multiple video tracks to an extent, but that lib-jitsi-meet does not. 

I implemented a semi-functioning version of this in a clone of lib-jitsi-meet about a year ago, but things seem to have changed quite a bit since then. I am not sure, but it seems like the more difficult parts of the problem are managing multiple simulcast streams and bitrate adaptivity. I assume there are probably other complications. I guess I am just wondering if this is something on the team's radar?

Kidav:
I'm also interested in supporting at least two video tracks.

Kidav:
I think it would be nice to distinguish between  ss video track and camera video track , so SDPDiffer.toJingle have to mark video tracks by num or like so on

afrokick:
Any news?

saghul:
No news, we are still only supporting a single video track per participant at the moment. I'm not aware of any plans to change this.

saghul:
@afrokick I don't think that's currently possible with Jitsi. Maybe by hacking around and closing streams for pupil-pupil communication... but that would be very brittle, and we didn't design Jitsi Meet with that in mind.

Sorry about that.

Sylchauf:
You can use a new connection to share more than one video track.
I use this workaround to send camera track and share screen track. Be careful, the second connection must not send audio track.

saghul:
We have now added the ability to set "effects" on video tracks. Jitsi Meet now uses this feature to mix camera and screensharing into a single stream by using a canvas.

tiofabby:
Hi @saghul , how can we use such effect feature within lib-jitsi-meet? Is it something that could be documented in https://github.com/jitsi/lib-jitsi-meet/blob/master/doc/API.md#installation ?
Thank you!

saghul:
@jallamsetty1 can you please help?

jallamsetty1:
[...]

PR #1015, opened 2019-12-19
doc: add documentation for setEffect method on JitsiTrack

jallamsetty1:
- add documentation for setEffect as per customer request.
https://github.com/jitsi/lib-jitsi-meet/issues/464

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -444,6 +444,22 @@ We have the following methods for controling the tracks:
 
 12. isEnded() - returns true if track is ended
 
+13. setEffect(effect) - Applies the effect by swapping out the existing MediaStream on the JitsiTrack with the new 
+
+    MediaStream which has the desired effect. "undefined" is passed to this function for removing the effect and for
+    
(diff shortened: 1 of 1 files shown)


tiofabby:
Hi @jallamsetty1 
To use the presenter effect using lib-jitsi-meet, I need then to use setEffect(effect).
Looking at code in jitis-meet, effet is being initialized with:
`effect = await this._createPresenterStreamEffect(height, defaultCamera);`
Unfortunatly, I dont see such function '_createPresenterStreamEffect', calling 'createLocalPresenterTrack' being present in lib-jitis-meet.

Could you please then confirm that to use such Presenter Effect and teh setEffect() function, I would need to mimic code being in 'createLocalPresenterTrack' in jitis-meet in my own code right?

Thank you
Fabrice

jallamsetty1:
Hi Fabrice, yes you would have to mimic the code in ”createPresenterTrack” to create a new video track with lower resolution and then apply the effect on the existing desktop stream to create the presenter effect.

PR #1017, opened 2020-01-07
Add Chromium based Edge (Chredge) to the list of webrtc-capable browsers

jallamsetty1:
The older version of Bowser (<2.0) reported Edge beta as chrome but now in Bowser 2.7, it comes back as Edge which is not in the list of supported browsers.
We now check if the Edge version > 72 to determine if it is Chromium based Edge (Chredge).
Chredge is added to the list of webrtc-capable browsers but we still display the warning that it is not one of the fully supported browsers.

### DIFF
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -61,6 +61,7 @@ export default class BrowserCapabilities extends BrowserDetection {
      */
     isChromiumBased() {
         return this.isChrome()
+            || this.isChromiumBasedEdge()
             || this.isElectron()
             || this.isNWJS()
             || this.isOpera();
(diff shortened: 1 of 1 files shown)


modules/browser/BrowserCapabilities.js
```diff
@@ -66,6 +66,15 @@ export default class BrowserCapabilities extends BrowserDetection {
             || this.isOpera();
     }
 
+    /**
+     * Checks if the current browser is Edge that is Chromium based.
+     */
+    isChromiumBasedEdge() {
```
saghul: Do we really need to be this explicit? If we are not going to do anything specific to Edge I'd rather see this embedded in `isChromiumBased`. WDYT?

jallamsetty1:
> Nice! Please squash when merging.

Sure will do. Thanks Saul.

PR #1018, opened 2020-01-07
Remove code for old gUM flow and use newGumFlow on all clients.

jallamsetty1:
- Remove the code related to old GUM flow and force all clients to use the new GUM flow.
- Force all the clients except React Native to use adapter.js.
- Update the unit test for GUM to use the new gum API.
- Take options.resolution into consideration when constructing the constraints for the GUM call. We were ignoring options.resolution in the new GUM flow APIs.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -46,7 +46,6 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * @param trackInfo.mediaType the MediaType of the JitsiRemoteTrack
      * @param trackInfo.videoType the VideoType of the JitsiRemoteTrack
      * @param trackInfo.effects the effects array contains the effect instance to use
-     * @param trackInfo.resolution the video resolution if it's a video track
      * @param trackInfo.deviceId the ID of the local device for this track
      * @param trackInfo.facingMode the camera facing mode used in getUserMedia
      * call
@@ -89,33 +88,20 @@ export default class JitsiLocalTrack extends JitsiTrack {
         this.rtcId = rtcId;
         this.sourceId = sourceId;
         this.sourceType = sourceType;
-
-        if (browser.usesNewGumFlow()) {
-            // Get the resolution from the track itself because it cannot be
-            // certain which resolution webrtc has fallen back to using.
-            this.resolution = track.getSettings().height;
-            this.maxEnabledResolution = resolution;
-
-            // Cache the constraints of the track in case of any this track
-            // model needs to call getUserMedia again, such as when unmuting.
-            this._constraints = track.getConstraints();
-
-            // Safari returns an empty constraints object, construct the constraints using getSettings.
-            if (!Object.keys(this._constraints).length && videoType === VideoType.CAMERA) {
-                this._constraints = {
-                    height: track.getSettings().height,
-                    width: track.getSettings().width
-                };
-            }
-        } else {
-            // FIXME Currently, Firefox is ignoring our constraints about
-            // resolutions so we do not store it, to avoid wrong reporting of
-            // local track resolution.
-            this.resolution = browser.isFirefox() ? null : resolution;
-            this.maxEnabledResolution = this.resolution;
-        }
-
         this.deviceId = deviceId;
+        this.maxEnabledResolution = resolution;
+
+        // Cache the constraints of the track in case of any this track
+        // model needs to call getUserMedia again, such as when unmuting.
+        this._constraints = track.getConstraints();
+
+        // Safari returns an empty constraints object, construct the constraints using getSettings.
+        if (!Object.keys(this._constraints).length && videoType === VideoType.CAMERA) {
+            this._constraints = {
+                height: track.getSettings().height,
+                width: track.getSettings().width
+            };
+        }
 
         /**
          * The <tt>Promise</tt> which represents the progress of a previously
@@ -538,35 +524,17 @@ export default class JitsiLocalTrack extends JitsiTrack {
             // This path is only for camera.
             const streamOptions = {
                 cameraDeviceId: this.getDeviceId(),
+                constraints: { video: this._constraints },
                 devices: [ MediaType.VIDEO ],
                 effects: this._streamEffect ? [ this._streamEffect ] : [],
                 facingMode: this.getCameraFacingMode()
             };
 
-            if (browser.usesNewGumFlow()) {
-                promise
-                    = RTCUtils.newObtainAudioAndVideoPermissions(Object.assign(
-                        {},
-                        streamOptions,
-                        { constraints: { video: this._constraints } }));
-            } else {
-                if (this.resolution) {
-                    streamOptions.resolution = this.resolution;
-                }
-
-                promise
-                    = RTCUtils.obtainAudioAndVideoPermissions(streamOptions);
-            }
-
+            promise = RTCUtils.obtainAudioAndVideoPermissions(streamOptions);
             promise = promise.then(streamsInfo => {
                 // The track kind for presenter track is video as well.
                 const mediaType = this.getType() === MediaType.PRESENTER ? MediaType.VIDEO : this.getType();
-                const streamInfo
-                    = browser.usesNewGumFlow()
-                        ? streamsInfo.find(
-                            info => info.track.kind === mediaType)
-                        : streamsInfo.find(
-                            info => info.mediaType === mediaType);
+                const streamInfo = streamsInfo.find(info => info.track.kind === mediaType);
 
                 if (streamInfo) {
                     this._setStream(streamInfo.stream);
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -5,7 +5,6 @@ import { getLogger } from 'jitsi-meet-logger';
 import * as JitsiConferenceEvents from '../../JitsiConferenceEvents';
 import * as MediaType from '../../service/RTC/MediaType';
 import RTCEvents from '../../service/RTC/RTCEvents';
-import VideoType from '../../service/RTC/VideoType';
 import browser from '../browser';
 import Statistics from '../statistics/statistics';
 import GlobalOnErrorHandler from '../util/GlobalOnErrorHandler';
@@ -16,7 +15,6 @@ import BridgeChannel from './BridgeChannel';
 import JitsiLocalTrack from './JitsiLocalTrack';
 import RTCUtils from './RTCUtils';
 import TraceablePeerConnection from './TraceablePeerConnection';
-
 
 const logger = getLogger(__filename);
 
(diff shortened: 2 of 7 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -524,30 +514,19 @@ export default class JitsiLocalTrack extends JitsiTrack {
                 facingMode: this.getCameraFacingMode()
             };
 
-            if (browser.usesNewGumFlow()) {
-                promise
-                    = RTCUtils.newObtainAudioAndVideoPermissions(Object.assign(
+            if (this.resolution) {
+                streamOptions.resolution = this.resolution;
+            }
+            promise = RTCUtils.newObtainAudioAndVideoPermissions(Object.assign(
```
saghul: maybe drop the `new` from the name?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -88,21 +86,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
-            this._constraints = track.getConstraints();
-        } else {
-            // FIXME Currently, Firefox is ignoring our constraints about
-            // resolutions so we do not store it, to avoid wrong reporting of
-            // local track resolution.
-            this.resolution = browser.isFirefox() ? null : resolution;
-        }
+        // Get the resolution from the track itself because it cannot be
+        // certain which resolution webrtc has fallen back to using.
+        this.resolution = track.getSettings().height;
```
saghul: we don't have this on mobile, it throws an error because it's not implemented

modules/RTC/JitsiLocalTrack.js
```diff
@@ -88,21 +86,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
 
+        // Cache the constraints of the track in case of any this track
+        // model needs to call getUserMedia again, such as when unmuting.
+        this._constraints = track.getConstraints();
```
saghul: ditto

modules/RTC/JitsiLocalTrack.js
```diff
@@ -524,30 +514,19 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
                 // The track kind for presenter track is video as well.
                 const mediaType = this.getType() === MediaType.PRESENTER ? MediaType.VIDEO : this.getType();
-                const streamInfo
-                    = browser.usesNewGumFlow()
-                        ? streamsInfo.find(
-                            info => info.track.kind === mediaType)
-                        : streamsInfo.find(
-                            info => info.mediaType === mediaType);
+                const streamInfo = streamsInfo.find(
+                    info => info.track.kind === mediaType);
```
saghul: doesn't this fit in a single line now?

modules/RTC/RTC.js
```diff
@@ -260,18 +229,8 @@ export default class RTC extends Listenable {
...
-            : RTCUtils.obtainAudioAndVideoPermissions(options);
-
-        return obtainMediaPromise.then(tracksInfo => {
-            if (usesNewGumFlow) {
-                return _newCreateLocalTracks(tracksInfo);
-            }
-
-            return createLocalTracks(tracksInfo, options);
-        });
+        return RTCUtils.newObtainAudioAndVideoPermissions(options)
```
saghul: let's drop the "new" prefix from the APIs now

modules/RTC/RTCUtils.js
```diff
@@ -13,12 +13,10 @@ import { getLogger } from 'jitsi-meet-logger';
 import GlobalOnErrorHandler from '../util/GlobalOnErrorHandler';
 import JitsiTrackError from '../../JitsiTrackError';
 import Listenable from '../util/Listenable';
-import * as MediaType from '../../service/RTC/MediaType';
 import Resolutions from '../../service/RTC/Resolutions';
 import browser from '../browser';
 import RTCEvents from '../../service/RTC/RTCEvents';
 import screenObtainer from './ScreenObtainer';
-import SDPUtil from '../xmpp/SDPUtil';
```
saghul: sweet!

modules/RTC/RTCUtils.js
```diff
@@ -935,24 +539,11 @@ class RTCUtils extends Listenable {
     * on failure.
     **/
     getUserMediaWithConstraints(um, options = {}) {
-        const constraints = getConstraints(um, options);
+        const constraints = newGetConstraints(um, options);
```
saghul: drop new

modules/RTC/RTCUtils.js
```diff
@@ -1338,8 +794,11 @@ class RTCUtils extends Listenable {
             const audioTracks = avStream.getAudioTracks();
 
             if (audioTracks.length) {
-                const audioStream = new MediaStream(audioTracks);
+                const audioStream = new MediaStream();
 
+                for (let i = 0; i < audioTracks.length; i++) {
```
saghul: why is this change needed? is that media stream constructor not supported on some browsers? if so, please leave a comment.

modules/RTC/RTCUtils.js
```diff
@@ -1350,8 +809,11 @@ class RTCUtils extends Listenable {
             const videoTracks = avStream.getVideoTracks();
 
             if (videoTracks.length) {
-                const videoStream = new MediaStream(videoTracks);
+                const videoStream = new MediaStream();
```
saghul: ditto

modules/browser/BrowserCapabilities.js
```diff
@@ -104,10 +104,21 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean} true if the browser is supported, false otherwise.
      */
     isSupported() {
-        return this.isChromiumBased()
-            || this.isFirefox()
-            || this.isReactNative()
-            || this.isSafariWithWebrtc();
+        const REQUIRED_CHROME_VERSION = 69;
+
+        if (this.isChrome()) {
```
saghul: why do we check for Chrome excplicitly, if it's already part of the  this.isChromiumBased() check?

modules/browser/BrowserCapabilities.js
```diff
@@ -276,40 +287,14 @@ export default class BrowserCapabilities extends BrowserDetection {
...
-
     /**
      * Checks if the browser uses webrtc-adapter. All browsers using the new
      * getUserMedia flow and Edge.
      *
      * @returns {boolean}
      */
     usesAdapter() {
-        return this.usesNewGumFlow() || this.isEdge();
+        return this.isSupported() || this.isEdge();
```
saghul: This is not correct. The adapter doesn't really support RN, so we shouldn't use it there. Also, I think we can now drop the Edge check, since it's not supported. Edgium is, but that's already covered by the previous check. I think this should read `return !this.isReactNative();`

modules/RTC/RTCUtils.js
```diff
@@ -112,228 +97,6 @@ function initEnumerateDevicesWithCallback() {
...
-                }
-
-                // Old style.
-                constraints.audio.optional.push({
-                    sourceId: options.micDeviceId
-                });
-            }
-
-            // if it is good enough for hangouts...
-            constraints.audio.optional.push(
```
saghul: where are these set now?

modules/RTC/RTCUtils.js
```diff
@@ -112,228 +97,6 @@ function initEnumerateDevicesWithCallback() {
...
-        }
-
-        setResolutionConstraints(
-            constraints, isNewStyleConstraintsSupported, options.resolution);
-    }
-    if (um.indexOf('audio') >= 0) {
-        if (browser.isReactNative()) {
-            // The react-native-webrtc project that we're currently using
-            // expects the audio constraint to be a boolean.
-            constraints.audio = true;
```
saghul: note to self, list this limitation.

modules/RTC/RTCUtils.js
```diff
@@ -353,6 +116,7 @@ function getConstraints(um, options = {}) {
  * @param {Object} options.frameRate - used only for dekstop sharing.
  * @param {Object} options.frameRate.min - Minimum fps
  * @param {Object} options.frameRate.max - Maximum fps
+ * @param {string} options.resolution - resolution constraints.
```
saghul: Why is this here now? `resolution` was part of the old flow, the new flow uses `constraints`, which is already a parameter, with the video constrainsts.

modules/RTC/JitsiLocalTrack.js
jallamsetty1: Sure, I will drop the "new" prefix from all the APIs.

modules/RTC/RTCUtils.js
jallamsetty1: Maybe I am confused, I was told that some client apps still use options.resolution to override the default constraints. If that is not the case anymore, I will remove resolution from the options.

modules/browser/BrowserCapabilities.js
jallamsetty1: Yep we don't need the chrome check.

modules/RTC/RTCUtils.js
saghul: Right now, the resolution option in config.js is only used in Firefox, well and anything not using the new gUM flow. I always disliked that the new `constraints` option ignored the legacy `resolution` option, so there is that.

I think what we need here is a decission, more than anything.

modules/RTC/RTCUtils.js
jallamsetty1: These are set in newGetConstraints() now. I didn't change the audio constraints like we discussed before, thought it will be better to make those changes in a separate PR after thorough testing on all browsers.

modules/RTC/RTCUtils.js
jallamsetty1: Ok, I will bring it up with tin the standup to see what we want to do here.

modules/browser/BrowserCapabilities.js
jallamsetty1: Ok, didn't know that adapter doesn't support RN, also I wasn't sure why Edge was here when it is not supported.

modules/RTC/RTCUtils.js
jallamsetty1: The media stream constructor doesn't seem to work when we use MediaStreamTrackMock(unit tests) instead of MediaStreamTrack.

modules/RTC/JitsiLocalTrack.js
jallamsetty1: For now on RN I went back to using the resolution passed to the constructor of JitsiLocalTrack and I am passing `{ video: true }` as constraints when unmuting video.

saghul: That's a step backwards from the previous behavior, the resolution was respeccted. Let me look into implementing those 2 to see if we can avoid leaving this for later.

modules/RTC/RTCUtils.js
saghul: Aha. Maybe leave a comment? I wouldn't have guessed :-)

modules/RTC/JitsiLocalTrack.js
jallamsetty1: Ok sure, will wait for you to implement it. We have another option as well - we could construct the constraints from the options and then use those constraints while unmuting video (just saying).

modules/RTC/RTCUtils.js
jallamsetty1: Sure, will do.

saghul:
can you PTAL? https://github.com/react-native-webrtc/react-native-webrtc/pull/739 it's currently only getConstraints, I'm working on getSettings now.

PR #1020, opened 2020-01-20
Remove isEdge checks

jallamsetty1:


modules/RTC/RTCUtils.js
```diff
@@ -1397,8 +1396,7 @@ class RTCUtils extends Listenable {
     isDeviceChangeAvailable(deviceType) {
         return deviceType === 'output' || deviceType === 'audiooutput'
             ? isAudioOutputDeviceChangeAvailable
-            : browser.isChromiumBased()
-                || browser.isFirefox() || browser.isEdge();
+            : browser.isChromiumBased() || browser.isFirefox();
```
saghul: Doesn't this check cover all browsers now? If Safari is the exception we should be explicit about it IMHO.

modules/browser/BrowserCapabilities.js
```diff
@@ -32,7 +32,6 @@ export default class BrowserCapabilities extends BrowserDetection {
     doesVideoMuteByStreamRemove() {
         return !(
             this.isFirefox()
-            || this.isEdge()
```
saghul: Since we are here, maybe just say `this.isChromiumBased` instead of negating every other browser.

modules/browser/BrowserCapabilities.js
```diff
@@ -139,17 +128,15 @@ export default class BrowserCapabilities extends BrowserDetection {
...
     }
 
     /**
      * Checks if the current browser supports WebRTC datachannels.
      * @return {boolean}
      */
     supportsDataChannels() {
-        // NOTE: Edge does not yet implement DataChannel.
-        return !this.isEdge();
+        return true;
```
saghul: Maybe let's dump this function altogether?

modules/browser/BrowserCapabilities.js
```diff
@@ -170,7 +157,6 @@ export default class BrowserCapabilities extends BrowserDetection {
      */
     supportsRtpStatistics() {
         return this.isChromiumBased()
-            || this.isEdge()
```
saghul: Is there a browser where this is not supported now? If not, maybe let's drop the function completely.

modules/statistics/RTPStatsCollector.js
```diff
@@ -573,7 +548,6 @@ StatsCollector.prototype.processStatsReport = function() {
             });
         }
 
-        // NOTE: Edge's proprietary stats via RTCIceTransport.msGetStats().
```
saghul: I think you need to drop the code just following the comment, `msType` was the Microsoft proprietary stuff the comment mentions.

modules/browser/BrowserCapabilities.js
jallamsetty1: sure

jallamsetty1: sure

modules/browser/BrowserCapabilities.js
jallamsetty1: sure

modules/statistics/RTPStatsCollector.js
jallamsetty1: I thought so but there was no isEdge check around it so I let it be, thanks for confirming.

modules/RTC/RTCUtils.js
jallamsetty1: How about RN ? Is input device (both cam and mic) changing supported there ?

modules/browser/BrowserCapabilities.js
jallamsetty1: sure, we could drop it.

modules/RTC/RTCUtils.js
saghul: No, but we don't support device preview anyway, it's switched as the user selects, so I don't think it's relevant.

modules/browser/BrowserCapabilities.js
saghul: 👍

modules/RTC/RTCUtils.js
jallamsetty1: ok makes sense then, thanks for the explanation.

jallamsetty1: I believe the reason Safari is an exception is because we have disabled video on Safari after it switched to Unified plan by default. Once we add unified plan support, we could remove that exception as well.

saghul: Ah, gotcha! Good point! Let's keep it for now then.

saghul:

saghul:

saghul:

PR #1021, opened 2020-01-24
fix: use consistent log format for TIME prints

paweldomas:
...it make it easier to process the logs in automated way.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -196,7 +196,7 @@ class JingleConnectionPlugin extends ConnectionPlugin {
             break;
         }
         case 'transport-replace':
-            logger.info('(TIME) Start transport replace', now);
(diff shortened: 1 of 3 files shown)


modules/xmpp/strophe.jingle.js
```diff
@@ -207,7 +207,7 @@ class JingleConnectionPlugin extends ConnectionPlugin {
             sess.replaceTransport($(iq).find('>jingle'), () => {
                 const successTime = window.performance.now();
 
-                logger.info('(TIME) Transport replace success!', successTime);
+                logger.info('(TIME) Transport replace success:\n', successTime);
```
damencho: Is this intentional or it was supposed to be `\t`?

paweldomas: 🤦‍♂ yes should be \t

PR #1022, opened 2020-01-24
ref: deprecate and rename 'bosh' option to 'serviceUrl'

paweldomas:
It can either be a websocket or bosh URL, so the current naming is confusing.

A change to jitsi-meet can come in separately, because 'bosh' option is still handled.


saghul:
❤️❤️❤️

doc/API.md
```diff
@@ -219,7 +219,7 @@ This objects represents the server connection. You can create new ```JitsiConnec
     - appID - identification for the provider of Jitsi Meet video conferencing services. **NOTE: not implemented yet. You can safely pass ```null```**
     - token - secret generated by the provider of Jitsi Meet video conferencing services. The token will be send to the provider from the Jitsi Meet server deployment for authorization of the current client.
     - options - JS object with configuration options for the server connection. You can change the following properties there:
-        1. bosh -
+        1. serviceUrl - XMPP service URL. For  example 'wss://server.com/xmpp-websocket' for Websocket or '//server.com/http-bind' for BOSH.
```
saghul: Maybe document the deprecated one (saying it’s deprecated) to avoid confusion?

paweldomas: done!

PR #1023, opened 2020-01-27
Updates speaker stats to ignore hidden participants.

damencho:


modules/statistics/SpeakerStatsCollector.js
```diff
@@ -26,6 +26,7 @@ export default class SpeakerStatsCollector {
         const userId = conference.myUserId();
 
         this.stats.users[userId] = new SpeakerStats(userId, null, true);
+        this.getParticipantById = conference.getParticipantById.bind(conference);
```
hristoterezov: I'd add `conference` as prop instead of adding a method.

Issue #484, opened 2017-05-16
Optimize for npm install


joshfarrant:
Apologies if this has been discussed already, but I can't find any other conversations on this topic.

Is there a reason that this package isn't really optimized for being installed through npm? I'm trying to integrate this into a React app and I'm having a bit of a hard time getting everything up and running.

It would be nice if we could simply `npm install lib-jitsi-meet` within a project and then just `import JitsiMeetJS from 'lib-jitsi-meet';`, however this currently doesn't seem to be possible.

Obviously the package isn't currently published to npm, but that's not an issue as we can install directly from this repo. Doing that, however, seems to do three things. It...

1. Clones the repo
2. Installs production dependencies only, not devDependencies
3. Runs `postinstall` npm script (`webpack -p`) and fails because of missing devDependencies.

Navigating to `<my-react-project>/node_modules/lib-jitsi-meet` and running `npm install` manually does install these devDependencies and run the `postinstall` script successfully, creating `lib-jitsi-meet.min.js` and `lib-jitsi-meet.min.map`.

At this point, trying to `import JitsiMeetJS from 'lib-jitsi-meet';` creates errors as the `"main"` property of the `package.json` references the `./index.js` file, rather than the built (and transpiled) `lib-jitsi-meet.min.js` file created by webpack.

If we update the `"main"` property of the `package.json` to be `./lib-jitsi-meet.min.js` then `import JitsiMeetJS from 'lib-jitsi-meet';` does in fact import `JitsiMeetJS` as expected. Now, however, we get a `ReferenceError: Strophe is not defined`.

Adding `import strophe;` does resolve this, however we then get `Missing strophe-plugins (disco and caps plugins are required)!`. At this point it's clear that the strophe and strophe-plugins modules aren't included in the `lib-jitsi-meet.min.js` file that webpack builds, even though they are listed as project dependencies. Is there a reason that the choice was made to require these modules globally, rather than just importing them whenever they're needed? From looking through the codebase it looks like the same goes for jQuery.

This all can be solved by the user including `<script>` tags in their html, as you recommend, or possibly by using webpack's `ProvidePlugin`, but it seems like it would be far nicer to try to do as much as possible when packaging the module in the first place.

Interested to hear your thoughts on this.

damencho:
[...]

PR #1026, opened 2020-01-30
Prepare repository to publish to npm

sapkra:
Related: #484 

I thought about versioning but I'm still not happy with this solution.

Some ideas:
- `1.[jitsi_meet_tag].[numeric_tag]`
- `1.[jitsi_meet_tag].0` and `1.[jitsi_meet_tag].0-preview.[numeric_tag]`
- replace current versioning with semver

But I'm not sure how your tagging is working and defined. Is there a documentation about that e.g. for the numeric one like `819`?

### DIFF
--- a/.npmignore
+++ b/.npmignore
@@ -1,0 +1,2 @@
+*
+!lib-jitsi-meet.*
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,6 +1,6 @@
 {
-  "name": "lib-jitsi-meet",
-  "version": "0.0.0",
+  "name": "@jitsi/lib-jitsi-meet",
+  "version": "2.0.5276",
   "lockfileVersion": 1,
   "requires": true,
   "dependencies": {
@@ -2652,14 +2652,6 @@
         "is-plain-object": "^2.0.4",
         "kind-of": "^6.0.2",
         "shallow-clone": "^3.0.0"
-      },
-      "dependencies": {
-        "kind-of": {
-          "version": "6.0.3",
-          "resolved": "https://registry.npmjs.org/kind-of/-/kind-of-6.0.3.tgz",
-          "integrity": "sha512-dcS1ul+9tmeD95T+x28/ehLgd9mENa3LsvDTtzm3vyBEO7RPptvAD+t44WVXaUjTBRcrpFeFlC8WCruUR456hw==",
-          "dev": true
-        }
       }
     },
     "collection-visit": {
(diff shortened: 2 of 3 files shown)


damencho:
819 is jenkins build id. 
My proposal will be to be consistent with the rest of the projects, like https://github.com/jitsi/jitsi-videobridge/blob/4b50618445a607298dadb1319d5a197397654692/resources/build_deb_package.sh#L31

saghul:
Thanks for getting this started @sapkra! We had some internal discussions about this in the past, and currently the only blocking issue (that I'm aware of) is automation. We need to make sure that every merge to this repo results in a new release so we can then update Jitsi Meet.

As for versioning, I think we should start at 2.0 given that we are changing the package name, that's kinda a big deal.

Then what I proposed is to have the automation raise the patch level by 1 on every release, unless the previous one doesn't exist. This would be the case if you make a minor (or major change) as you'd bump the minor (or major, depending on the case) as part of the PR you send.

sapkra:
Ok. I just had a lot of problems again using yarn. I think they are caching packages based on the version in the package.json which results in installing of old dependencies because the version is always 0.0.0.

To fix that I had to delete my yarn.lock file everytime I've ran yarn install or publish my own package. Because I couldn't see any progress on the official side I just published my own package and wanted to share my changes to push this forward.

I think it's s big deal to don't have an official package because someone else could publish the library to npm with adding some malicious code.

saghul:
We have control over “lib-jitsi-meet “ and I believe nobody can publish under @jitsi/ without being part of the Jitsi org. Isn’t this right?

sapkra:
Ok I see, but I think if we want to publish it regularly to npm it should be in the same organization like all the other packages published by jitsi. This would be the cleanest way in my opinion. The old package could be marked as deprecated.

saghul:
Correct, we plan on publishing on @jitsi/ and deprecating the old one. The publication will be done by a CI.

everflux:
I can't find the published npm artifacts, is this still ongoing?

saghul:
Yes, the current world events have pushed this back, alas.

sapkra:
@everflux You can use my package `@lyno/lib-jitsi-meet` if you want. I'm updating it regularly (usually for every stable release) until there is an official package.

jordisala1991:
@sapkra are you still mainaining the `@lyno/lib-jitsi-meet`? It would be nice to have the latest releases of this package.

sapkra:
@jordisala1991 I haven't had the time to update my cluster to the new version yet, so I haven't released the new version yet. But since you asked for it, I just did it now. I released the new version a few minutes ago, including the updated TypeScript type definitions which are automatically generated, so some are unfortunately missing.

jordisala1991:
Thank you so much!

sapkra:
Closing it for now. All my changes to it can be found here: https://github.com/lynoapp/lib-jitsi-meet

PR #1027, opened 2020-01-31
WIP: Generate TypeScript definitions

### LINKED ISSUE #1025: Add TypeScript definitions
The common way to use types in JavaScript became TypeScript. The most projects dropped the flow support and migrated to TypeScript. I would really appreciate it if this project could either.

The best way would be to completely migrate to TypeScript but it would also be ok if a TypeScript definition file would exist to provide the typings.

sapkra:
This is the starting point for generating typescript definitions for lib-jitsi-meet using the jsdoc types already used in this project. I also had to add some missing ones and also fix a few of them.

There are still some problems which can be found in the `jsdoc.log`. I was not able to solve most of them.

TODO:
- [ ] Fix errors and add missing types
- [ ] Generate types using [webpack plugin](https://github.com/Jaid/jsdoc-tsd-webpack-plugin)
- [ ] Add types.d.ts to package

Closes #1025

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -74,6 +74,7 @@ const JINGLE_SI_TIMEOUT = 5000;
  * Creates a JitsiConference object with the given name and properties.
  * Note: this constructor is not a part of the public API (objects should be
  * created using JitsiConnection.createConference).
+ * @constructor
  * @param options.config properties / settings related to the conference that
  * will be created.
  * @param options.name the name of the conference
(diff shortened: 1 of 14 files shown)


sapkra:
I'm closing this PR because I think it would be easier to just port the repository to TypeScript step-by-step.

To do this you can use `allowJs: true` inside the tsconfig.json and remove it again after the port is completely done.

PR #1039, opened 2020-03-12
Add system audio sharing support / Implement AudioMixer

andrei-gavrilescu:
PR adds support for returning the system audio stream along with the screen sharing audio stream. 
On mac only tab audio can be shared, where as on windows one can share desktop audio as well.
The PR also features an AudioMixer used by jitsi-meet to merge together the local audio track with  system audio, resulting in screen sharing with audio functionality.


andrei-gavrilescu:
I've added a new function to the lib-jitsi-meet interface. I'll update de documentation accordingly if we decide that it's ok.

modules/webaudio/AudioMixer.js
```diff
@@ -0,0 +1,94 @@
...
+        this._started = true;
+
+        // Create ChannelMergerNode and connect all MediaStreams to it.
+        this._channelMerger = this._audioContext.createChannelMerger(this._streamsToMix.length);
+
+        for (const stream of this._streamsToMix) {
+            const streamMSS = this._audioContext.createMediaStreamSource(stream);
+
+            streamMSS.connect(this._channelMerger);
+
```
saghul: please drop the empty line

modules/webaudio/AudioMixer.js
```diff
@@ -0,0 +1,94 @@
...
+        this._started = false;
+        this._streamsToMix = [];
+    }
+
+    /**
+     * Adds the MediaStream from a JitsiLocalTrack to the list of streams to be mixed.
+     *
+     * @param {JitsiLocalTrack} jitsiLocalTrack - JitsiLocalTrack containing audio stream.
+     */
+    addJitsiLocalTrack(jitsiLocalTrack) {
```
saghul: I mentioneed it in the other PR, but for completelenes:  I think we should drop this function and just deal with streams in this entity

modules/webaudio/AudioMixer.js
```diff
@@ -0,0 +1,94 @@
...
+        if (!this._streamsToMix.length) {
+            logger.warn('No MediaStream\'s added to AudioMixer, nothing will happen.');
+
+            return null;
+        }
+
+        this._started = true;
+
+        // Create ChannelMergerNode and connect all MediaStreams to it.
+        this._channelMerger = this._audioContext.createChannelMerger(this._streamsToMix.length);
```
saghul: this is not initialized in the constructor

modules/webaudio/AudioMixer.js
```diff
@@ -0,0 +1,94 @@
...
+
+        return this._mixedMSD.stream;
+    }
+
+    /**
+     * Disconnect the ChannelMergerNode stopping the audio mix process.References to MediaStreams are also cleared.
+     *
+     * @returns {void}
+     */
+    reset() {
```
saghul: this is not resilient to calling it before start, channelMerger would be undefined

andrei-gavrilescu: woopsy, fixed.

modules/webaudio/AudioMixer.js
andrei-gavrilescu: Please see my comment about this in the associated jitsi-meet PR.

modules/webaudio/AudioMixer.js
andrei-gavrilescu: dropped.

modules/webaudio/AudioMixer.js
andrei-gavrilescu: It's not, we need the number of tracks that will be mixed in when we create it.

modules/RTC/ScreenObtainer.js
```diff
@@ -312,15 +312,20 @@ const ScreenObtainer = {
             getDisplayMedia = navigator.mediaDevices.getDisplayMedia.bind(navigator.mediaDevices);
         }
 
-        getDisplayMedia({ video: true })
+        getDisplayMedia({ video: true,
+            audio: true })
```
saghul: nit: can you put this on the same line as getDisplayMedia ?

chrisguttandin:
Hi @andrei-gavrilescu, I hope it's okay to comment on a closed PR.

It noticed that the `ChannelMergerNode` is probably used in an unintended way. The code currently connects every `MediaStreamAudioSourceNode` to the first input of the `ChannelMergerNode` and leaves the other inputs unconnected. If that is intended and the goal is to just sum up all signals the `ChannelMergerNode` could be replaced with a `GainNode`.

andrei-gavrilescu:
Hello @chrisguttandin , no problem  whatsoever. You're right, the `ChannelMergeNode` doesn't need to be used there.
At the time of the implementation I was under the impression that it was needed for `MediaStream` mixing. Now, because all the stream are connected to the same input it does the job, however this could have easily been achieved by simply connecting the streams to the destination, I don't think even a `GainNode` is needed.
Thanks a bunch for taking the time to look over this, I'll change the implementation accordingly in https://github.com/jitsi/lib-jitsi-meet/pull/1061.

chrisguttandin:
Yes, you're right. The `GainNode` isn't really necessary as well. In case you think about adding a limiter to avoid clipping, we recently created one for a similar use case.

https://github.com/chrisguttandin/limiter-audio-worklet

Since it has zero documentation so far please let me know if you're interested. I'm happy to help.

PR #1040, opened 2020-03-12
Disco info fixes

paweldomas:


damencho:
So this in practice will do discovery only once? 
Not sure whether we are not doing discovery twice in case of remote control and jigasi ...

damencho:
Even without doing discovery twice, which probably is not right... Wouldn't this change getFeatures return features only once ... the first time it is executed? What if we got feature update later on (which is what is probably happening with jigasi and remote desktop), we will never return those.

damencho:
Aaaa no, no I realized what this change does, haha, better late than never.

paweldomas:
This will not cause it to be called only once. It only picks up any request that is already in progress. Also any participant has this process independent from each other.
About update of features, I don't know(yet) how this logic is implemented in jitsi-meet. I don't know the part which checks for  feature update, but I would  expect it to be called again and this would be way after the first results are returned.

PR #1046, opened 2020-03-15
feat(Amplitude): Add isP2P attribute to the transport.stats analytics event

jallamsetty1:


hristoterezov:
Don't we already have this as user prop?

jallamsetty1:
@hristoterezov, how do we set the user props ? Are these set using Statistics.analytics.addPermanentProperties() ? 
```
Statistics.analytics.addPermanentProperties({
            p2pFailed: false,
            forceJvb121: false
 });
```
Is that what you were referring to ?

paweldomas:
> Don't we already have this as user prop?

No, we don't have it yet. Even if we would it might not be reliable to for the transport stats if those would be emitted while p2p is not yet established or if jvb transport is emitted after p2p has been established.

modules/statistics/AvgRTPStatsReporter.js
```diff
@@ -941,6 +941,7 @@ export default class AvgRTPStatsReporter {
             return;
         }
         const transportStats = {
+            isP2P: data.transport[0].p2p,
```
paweldomas: can we keep the the p2p flag named `p2p` like other parts of the stats  report?

paweldomas: LGTM otherwise

jallamsetty1: ok sure

hristoterezov:
> > Don't we already have this as user prop?
> 
> No, we don't have it yet. Even if we would it might not be reliable to for the transport stats if those would be emitted while p2p is not yet established or if jvb transport is emitted after p2p has been established.

@paweldomas Make sense!
@jallamsetty1 Please ignore my previous comment!

PR #1053, opened 2020-03-19
ref(xmpp): adjust last success time tracker for bosh and websocket

paweldomas:


modules/xmpp/StropheLastSuccess.js
```diff
@@ -0,0 +1,36 @@
...
+    /**
+     * Starts tracking requests on the given connection.
+     *
+     * @param {Object} stropheConnection - Strophe connection instance.
+     */
+    startTracking(stropheConnection) {
+        const originalRawInput = stropheConnection.rawInput;
+
+        stropheConnection.rawInput = function(...args) {
+            this._lastSuccess = new Date().getTime();
```
saghul: use performane.now() ?

paweldomas: do you know if that's available on both web and mobile RN?

saghul: We polyfill it with Date.now() IIRC because of callstats, but please double check

paweldomas: well  I dont see point in doing that to be honest

paweldomas: having it in the "since epoch" format makes more sense in this context for me

paweldomas:
jenkins test this please

PR #1059, opened 2020-03-27
API.md - Important Mention of Add Local Track to Conference

WesUnwin:
In most use cases, the developer will want to add the newly acquired local track to the room.

As this is extremely common and required to actually share the track with others I would like to recommend that step 5 quickly mentions this.

This furthermore makes it clear that createLocalTracks() does not automatically do anything like this.

With this added, these steps now produce a visible result (excellent for a "hello world" demonstration)

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -511,7 +511,9 @@ room.on(JitsiMeetJS.events.conference.CONFERENCE_JOINED, onConferenceJoined);
 
 5. You also may want to get your local tracks from the camera and microphone:
 ```javascript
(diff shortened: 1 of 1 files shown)


WesUnwin:
Hi why did this get closed? After remaining untouched since Mar 2020, its just closed with no explanation. I already accepted the contribution agreement

WesUnwin:
This is not a question, this is a proposed addition to API.md which I is super helpful to beginners.

jallamsetty1:
The docs have been moved to the handbook repo and this PR is not valid anymore.

PR #1060, opened 2020-03-27
Add restart service instruction in tokens doc

c100k:


sapkra:

PR #1062, opened 2020-03-31
Unified plan changes - 1/2

jallamsetty1:
These are the first set of changes to LJM. The next PR will be take will have changes that will take into account the new sdp-interop module.

1. Reject m-lines with port 0 only when a=bundle-only is not present
2. Use unique session id per TPC.
3. Translate a=rid to Jingle only when simulcast through RIDs, i.e., ssrc-less simulcast is supported.
4. Trigger renegotation only when negotiationneeded event is fired on the browser.
5. Do not disable simulcast for screensharing in unified plan.
6. Fix the order of the simulcast streams for Firefox.
7. Use replaceTrack for applying stream effects in unified plan.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -393,11 +393,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
 
         this._setEffectInProgress = true;
 
-        // For firefox/safari, replace the stream without doing a offer answer with the remote peer.
-        if (browser.supportsRtpSender()) {
+        if (browser.usesUnifiedPlan()) {
             this._switchStreamEffect(effect);
+            if (this.isVideoTrack()) {
+                this.containers.forEach(cont => RTCUtils.attachMediaStream(cont, this.stream));
+            }
 
-            return conference.replaceTrackWithoutOfferAnswer(this)
+            return conference.replaceTrack(this, this)
                 .then(() => {
                     this._setEffectInProgress = false;
                 })
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -50,18 +50,6 @@ export class TPCUtils {
     }
 
     /**
-     * Obtains local tracks for given {@link MediaType}.
-     * @param {MediaType} mediaType - audio or video.
-     * @return {Array<JitsiLocalTrack>} - array containing the local tracks
-     * attached to the peerconnection of the given media type.
-     */
-    _getLocalTracks(mediaType) {
-        const tracks = Array.from(this.pc.localTracks.values());
-
-        return tracks.filter(track => track.getType() === mediaType);
-    }
-
-    /**
      * Obtains stream encodings that need to be configured on the given track.
      * @param {JitsiLocalTrack} localTrack
      */
@@ -301,16 +289,23 @@ export class TPCUtils {
     }
 
     /**
-     *
-     * @param {boolean} active
-     */
+    * Enables/disables audio transmission on the peer connection. When
+    * disabled the audio transceiver direction will be set to 'inactive'
+    * which means that no data will be sent nor accepted, but
+    * the connection should be kept alive.
+    * @param {boolean} active - true to enable audio media transmission or
+    * false to disable.
+    * @returns {false} - returns false always so that renegotiation is not automatically
+    * triggered after this operation.
+    */
     setAudioTransferActive(active) {
         return this.setMediaTransferActive('audio', active);
     }
 
     /**
      * Set the simulcast stream encoding properties on the RTCRtpSender.
-     * @param {*} track - the current track in use for which the encodings are to be set.
+     * @param {JitsiLocalTrack} track - the current track in use for which
+     * the encodings are to be set.
      */
     setEncodings(track) {
         const transceiver = this.pc.peerconnection.getTransceivers()
@@ -322,17 +317,23 @@ export class TPCUtils {
     }
 
     /**
-     *
-     * @param {*} mediaType
-     * @param {boolean} active
+     * Enables/disables media transmission on the peerconnection by changing the direction
+     * on the transceiver for the specified media type.
+     * @param {String} mediaType - 'audio' or 'video'
+     * @param {boolean} active - true to enable media transmission or false
+     * to disable.
+     * @returns {false} - returns false always so that renegotiation is not automatically
+     * triggered after this operation
      */
     setMediaTransferActive(mediaType, active) {
         const transceivers = this.pc.peerconnection.getTransceivers()
             .filter(t => t.receiver && t.receiver.track && t.receiver.track.kind === mediaType);
+        const localTracks = Array.from(this.pc.localTracks.values())
+            .filter(track => track.getType() === mediaType);
 
         if (active) {
             transceivers.forEach(transceiver => {
-                if (this._getLocalTracks(mediaType).length > 0) {
+                if (localTracks.length) {
                     transceiver.direction = 'sendrecv';
                     const parameters = transceiver.sender.getParameters();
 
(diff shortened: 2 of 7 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -0,0 +1,365 @@
...
+    }
+
+    /**
+     * Obtains local tracks for given {@link MediaType}.
+     * @param {MediaType} mediaType - audio or video.
+     * @return {Array<JitsiLocalTrack>} - array containing the local tracks
+     * attached to the peerconnection of the given media type.
+     */
+    _getLocalTracks(mediaType) {
+        const tracks = Array.from(this.pc.localTracks.values());
```
saghul: why do we need to create thiss extra array?

modules/browser/BrowserCapabilities.js
```diff
@@ -270,7 +279,15 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean}
      */
     usesAdapter() {
-        return this.usesNewGumFlow();
+        return !this.isFirefox() && !this.isReactNative();
```
saghul: why not on Firefox?

modules/browser/BrowserCapabilities.js
```diff
@@ -270,7 +279,15 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+    }
+
+    /**
+     * Checks if the browser uses RIDs/MIDs for siganling the simulcast streams
+     * to the bridge instead of the ssrcs.
+     */
+    usesRidsForSimulcast() {
```
saghul: Is this in preparation for part 2?

modules/xmpp/JingleSessionPC.js
```diff
@@ -510,8 +511,44 @@ export default class JingleSessionPC extends JingleSession {
...
+         * The negotiationneeded event is fired whenever we shake the media on the
+         * RTCPeerConnection object.
+         */
         this.peerconnection.onnegotiationneeded = () => {
+            const state = this.peerconnection.signalingState;
+            const remoteDescription = this.peerconnection.remoteDescription;
+
             this.room.eventEmitter.emit(XMPPEvents.PEERCONNECTION_READY, this);
+            if (browser.usesUnifiedPlan() && state.toString() === 'stable'
+                && remoteDescription && typeof remoteDescription.sdp === 'string') {
```
saghul: style: this needs an extra indent

modules/xmpp/JingleSessionPC.js
```diff
@@ -510,8 +511,44 @@ export default class JingleSessionPC extends JingleSession {
...
+        /**
+         * The negotiationneeded event is fired whenever we shake the media on the
+         * RTCPeerConnection object.
+         */
         this.peerconnection.onnegotiationneeded = () => {
+            const state = this.peerconnection.signalingState;
+            const remoteDescription = this.peerconnection.remoteDescription;
+
             this.room.eventEmitter.emit(XMPPEvents.PEERCONNECTION_READY, this);
+            if (browser.usesUnifiedPlan() && state.toString() === 'stable'
```
saghul: is the toString() necesssary?

modules/xmpp/JingleSessionPC.js
```diff
@@ -510,8 +511,44 @@ export default class JingleSessionPC extends JingleSession {
...
+                    const oldSdp = new SDP(this.peerconnection.localDescription.sdp);
+
+                    this._renegotiate()
+                        .then(() => {
+                            const newSdp = new SDP(this.peerconnection.localDescription.sdp);
+
+                            this.notifyMySSRCUpdate(oldSdp, newSdp);
+                            finishedCallback();
+                        })
+                        .catch(() => {
```
saghul: no error logging here?

modules/xmpp/JingleSessionPC.js
```diff
@@ -543,7 +580,7 @@ export default class JingleSessionPC extends JingleSession {
     sendIceCandidate(candidate) {
         const localSDP = new SDP(this.peerconnection.localDescription.sdp);
 
-        if (candidate && !this.lasticecandidate) {
+        if (candidate && candidate.candidate !== '' && !this.lasticecandidate) {
```
saghul: checkc the length > 0 maybe? I this what breaks FF74? Can we have it as a separate PR?

modules/xmpp/SDP.js
```diff
@@ -566,9 +570,11 @@ SDP.prototype.rtcpFbFromJingle = function(elem, payloadtype) { // XEP-0293
 // construct an SDP from a jingle stanza
 SDP.prototype.fromJingle = function(jingle) {
     const self = this;
+    const sessionId = (new Date()).getTime();
```
saghul:
```suggestion
    const sessionId = Date.now();
```

modules/RTC/TPCUtils.js
```diff
@@ -0,0 +1,365 @@
...
+                    transceiver.direction = 'recvonly';
+                }
+            });
+        } else {
+            transceivers.forEach(transceiver => {
+                transceiver.direction = 'inactive';
+            });
+        }
+
+        return true;
```
saghul: this function always returns true

modules/RTC/TPCUtils.js
```diff
@@ -0,0 +1,365 @@
...
+        const transceiver = this.pc.peerconnection.getTransceivers()
+            .find(t => t.sender && t.sender.track && t.sender.track.kind === track.getType());
+        const parameters = transceiver.sender.getParameters();
+
+        parameters.encodings = this._getStreamEncodings(track);
+        transceiver.sender.setParameters(parameters);
+    }
+
+    /**
+     *
```
saghul: can you please document this?

modules/RTC/TPCUtils.js
jallamsetty1: pc.localTracks is a map and therefore we need to construct an array from the interable object. I moved this to setMediaTransferActive() since there is not other function using it. I think I had one more before which I no longer needed.

modules/xmpp/JingleSessionPC.js
jallamsetty1: done

modules/xmpp/JingleSessionPC.js
jallamsetty1: fixed it

modules/xmpp/JingleSessionPC.js
jallamsetty1: I changed it but we are still not sure what the issue is with FF74 so I am leaving it here for now.

modules/RTC/TPCUtils.js
jallamsetty1: thanks for pointing it out, this should be returning false always so that renegotiation is not triggered.

modules/RTC/TPCUtils.js
jallamsetty1: Sure

modules/browser/BrowserCapabilities.js
jallamsetty1: Ok , will revert this for now. I will bump up the adapter to 7.5.0 after I test it on all browsers.

modules/browser/BrowserCapabilities.js
jallamsetty1: I need this here  - https://github.com/jitsi/lib-jitsi-meet/pull/1062/files#diff-902544c7807a8d275d7f45764155d76bR322
so that we don't create ssrc-groups for rids in the jingle message.

modules/xmpp/JingleSessionPC.js
jallamsetty1: I can't recall why I added that check but looks like its not necessary.

modules/xmpp/JingleSessionPC.js
jallamsetty1: eslint is complaining about the extra indent so I had to remove it.

jallamsetty1:
@saghul can you please a look again ? I rebased with master and force-pushed again. Once this is merged I have the other change with sdp-interop.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -393,11 +393,13 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
-        // For firefox/safari, replace the stream without doing a offer answer with the remote peer.
-        if (browser.supportsRtpSender()) {
+        if (browser.usesUnifiedPlan()) {
             this._switchStreamEffect(effect);
+            if (this.isVideoTrack()) {
+                this.containers.forEach(cont => RTCUtils.attachMediaStream(cont, this.stream));
+            }
 
-            return conference.replaceTrackWithoutOfferAnswer(this)
+            return conference.replaceTrack(this, this)
```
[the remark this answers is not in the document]
jallamsetty1: Yes, though the track doesn't need to be removed (just replaced) at the peerconnection level, it still needs to be removed and added back at the conference level for the remote participant to update its view.

PR #1064, opened 2020-04-03
fix(xmpp): prevent from calling disconnect twice

paweldomas:
Prevents from calling disconnect on the Strophe connection if it already has been called or not connected.

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -288,10 +288,9 @@ export default class XMPP extends Listenable {
         } else if (status === Strophe.Status.DISCONNECTED) {
             // Stop ping interval
             this.connection.ping.stopInterval();
-            const wasIntentionalDisconnect = this.disconnectInProgress;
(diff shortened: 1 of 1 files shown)


modules/xmpp/xmpp.js
```diff
@@ -534,15 +543,13 @@ export default class XMPP extends Listenable {
      * @returns {Promise} - Resolves when the disconnect process is finished or rejects with an error.
      */
     disconnect(ev) {
-        if (this.disconnectInProgress || !this.connection) {
-            this.eventEmitter.emit(JitsiConnectionEvents.WRONG_STATE);
-
-            return Promise.reject(new Error('Wrong connection state!'));
+        if (this.disconnectInProgress) {
+            return this.disconnectInProgress;
```
damencho, line 547: Isn't this supposed to return a promise not a boolean or undefined?

damencho: aaa it will be the promise few lines below.

Issue #512, opened 2017-06-17
Making a room password protected


guusdk:
XEP-0045 defines two `room_config` fields that relate to the password protection of a room:

    <field
         var='muc#roomconfig_passwordprotectedroom'
         type='boolean'
         label='Whether a Password is Required to Enter'/>
    <field
         var='muc#roomconfig_roomsecret'
         type='text-single'
         label='The Room Password'/>

Prosody appears to work with only the latter, while Openfire requires you to set both, and will not make the room password protected when setting only the `roomsecret` field. 

Arguably, setting the room password implies that a user wants the the room to be password protected (which is raised as an [Openfire issue here](https://issues.igniterealtime.org/browse/OF-1350), but the specification is somewhat confusing at this point (as [discussed here](https://mail.jabber.org/pipermail/standards/2017-June/032990.html))

At some point, the original author of the code wondered about this behavior, given the 'fixme' comment left at https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/ChatRoom.js#L857

To be safe, the client should always set both values - that will always have the desired effect, no matter what interpretation is implemented server-sided.

debalance:
Today I set up Jitsi Meet with ejabberd, and I have exactly this same problem:
The password gets set, but "password_protected" is still false.

weiss:
Same here, also with ejabberd.  FWIW, @guusdk's fix (#513) still applies to the current master branch without conflicts and resolves the issue for me.

saghul:
They have been updated already.

PR #513, opened 2017-06-17
Modifying room password protection

guusdk:
When modifying the room protection (setting or unsetting a password), the `muc#roomconfig_passwordprotectedroom` field should always be defined. This fixes issue #512.

PR #1068, opened 2020-04-06
Modifying room password protection

paweldomas:
Rebased version of https://github.com/jitsi/lib-jitsi-meet/pull/513

When modifying the room protection (setting or unsetting a password), the muc#roomconfig_passwordprotectedroom should always be defined. This fixes issue #512.


paweldomas:
I had to rebase the original PR in order run the PR tests

PR #1070, opened 2020-04-06
fix(strophe.jingle.js): exception on ICE config modification

paweldomas:
If ICE config object gets modified while there's any native
PeerConnection instance holding a reference, it will cause
an exception about final object being modified.

It happens when trying to reconnect XMPP Websocket on RN and
is caused by the fact that new TURN config is fetched after
the connection is re-established.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -149,6 +149,9 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
             logger.info(
                 `Marking session from ${fromJid
                 } as ${isP2P ? '' : '*not*'} P2P`);
+
+            const iceConfig = isP2P ? this.p2pIceConfig : this.jvbIceConfig;
+
             sess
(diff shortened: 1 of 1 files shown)


modules/xmpp/strophe.jingle.js
```diff
@@ -149,14 +149,20 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
...
                     $(iq).find('jingle').attr('sid'),
                     $(iq).attr('to'),
                     fromJid,
                     this.connection,
                     this.mediaConstraints,
-                    isP2P ? this.p2pIceConfig : this.jvbIceConfig,
+
+                    // Makes a copy in order to prevent exception thrown on RN when either this.p2pIceConfig or
+                    // this.jvbIceConfig is modified and there's a PeerConnection instance holding a reference
+                    { ...iceConfig },
```
saghul, line 165: This only does a shallow clone. In order to be extra safe I'd do `JSON.parse(JSON.stringify(data));`

PR #1074, opened 2020-04-07
Update to improve readability

JamesNewton:
The options object /must/ contain a muc of 'conference.meet.jit.si' but line 131 can be changed to any room name.

### DIFF
--- a/doc/example/example.js
+++ b/doc/example/example.js
@@ -1,14 +1,15 @@
 /* global $, JitsiMeetJS */
+room_name = 'massmind';
 
 const options = {
     hosts: {
-        domain: 'jitsi-meet.example.com',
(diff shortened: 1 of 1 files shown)


JamesNewton:
K, done.
James Newton
MassMind.org <http://www.massmind.org/>
1-970-462-7764

doc/example/example.js
```diff
@@ -1,14 +1,15 @@
 /* global $, JitsiMeetJS */
+room_name = 'massmind';
```
sapkra, line 2:
```suggestion
const room_name = 'massmind';
```
But I don't think that it will improve the readability that much if the related code is that far apart.

sapkra:
Closing because PR has not been updated.

PR #1077, opened 2020-04-08
fix(stats): Use the standard props for icecandidate stats on Firefox

jallamsetty1:


modules/statistics/RTPStatsCollector.js
```diff
@@ -21,9 +21,9 @@ KEYS_BY_BROWSER_TYPE[browsers.FIREFOX] = {
     'bytesReceived': 'bytesReceived',
     'bytesSent': 'bytesSent',
     'framerateMean': 'framerateMean',
-    'ip': 'ipAddress',
```
fippo, line 24: actually the standard thing is "address" because it may contain a mdns hostname. Chrome (and therefore Safari) don't do this yet though, see [here](https://bugs.chromium.org/p/chromium/issues/detail?id=968203)
If you need to map anyway...

jallamsetty1: Yes, you are right. We'll get rid of the map when we switch to spec compliant getStats on chrome as well.

PR #1085, opened 2020-04-13
Fix protocol replace for ws-ka

ashitikov:
This code fixes a problem we get, when our url of service contains 'ws' symbols inside domain name. Our service url looks like 'wss://vs.myws.com' - In this case keep-alive will fail because of incorrect replace: 'https://vs.myhttp.com'. This pull request should fix this problem.


ashitikov:
Signed.

ashitikov:
Closed due to wrong branch management in favor #1102

PR #1088, opened 2020-04-13
Add current audio/video codec to RTP statistics blob

Syonyk:
In preparation for some development work on VP9 (and to better allow understanding the state of the system when switching between P2P and JVB communications), I've added "current codec reporting" to the stats object when on a browser that supports it (Chrome-based).  The current stream audio and video codecs are added to the stats object with resolution, framerate, etc.

I've plumbed this through to the "user stats overlay" in jitsi-meet, and it works properly, showing the changes in codec when transitioning between P2P and JVB connections on a server configured properly.

Passes lint, and I've signed the code contributor document under this email address.


saghul:
Ssorry for the delay, and thanks for getting this effort started! @jallamsetty1 can you PTAL? The one thing that stands out to me is the use of the Google custom stat, isn't there a standardd way to get the codec name?

JonathanLennox:
Jenkins, add to whitelist.

jallamsetty1:
@Syonyk , thank you for the PR ! It LGTM. Will you be able to add the codec stats to the promise based stats implementation (which is used by FF and Safari) as well ?
@saghul, we are using the non-standard callback stats on chrome so we would have to use the "goog" stats there.

Syonyk:
I looked into the codec stats for FF/Safari, and didn't see an obvious parameter name for it.  Practically speaking, if it only works on Chrome, that's still fine for development purposes.  You can see the codecs of everyone in the meeting from a single Chrome instance joined in.

fippo:
the spec way is to traverse the stats from the outbound-rtp stats via the codecId to the codec. I think only chrome has implemented that way however.

PR #1093, opened 2020-04-15
Support inviting guests, with links with JWT tokens: allow tokendomain attribute in configuration

ramon-garcia:
This pull request adds an attribute tokendomain to the configuration object inside 'hosts', to specify a specific domain for login a user that has a JWT token.

The use case is an organization that has a domain for its own users, but wants to invite guests for some meetings with JWT tokens.

Best regards.


ramon-garcia:
Please consider this patch.

How can one, otherwise, support local users authenticated by LDAP, and guests authenticated with jwt tokens?

PR #1095, opened 2020-04-15
e2ee: generate silence in case of audio decryption errors

fippo:
This handles decryption errors for audio differently than the
current version. Instead of forwarding horrible noise to the decoder,
replace the bytes with magic opus bytes for silence:
```
0xd8fffe
```

Those bytes were captured on a modified version of
  https://webrtc.github.io/samples/src/content/peerconnection/endtoend-encryption/
with more dumping and after disabling the track.

### DIFF
--- a/modules/e2ee/E2EEContext.js
+++ b/modules/e2ee/E2EEContext.js
@@ -314,6 +314,13 @@ export default class E2EEcontext {
                 return controller.enqueue(encodedFrame);
             }, e => {
                 logger.error(e);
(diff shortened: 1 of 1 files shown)


saghul:
Maybe wait until this is merged to test? https://github.com/jitsi/lib-jitsi-meet/pull/1096

fippo:
Note: this only puts silence on decrypt errors and decryption only happens when we think that there is a key set.
And yes, lets test first :-)

fippo:
confirmed garbled audio which was not nice :-)

PR #1096, opened 2020-04-15
hips: fix enabling E2EE for tracks added later to the conference

saghul:


modules/e2ee/E2EEContext.js
```diff
@@ -31,6 +31,9 @@ const unencryptedBytes = {
     undefined: 1 // frame.type is not set on audio
 };
 
+// Flag to set on senders / receivers to avoid setting up the encryption transform
+// more than once.
+const kJitsiE2EE = Symbol('kJitsiE2EE');
```
fippo, line 36: oh I didn't know the symbol trick!

JitsiConference.js
```diff
@@ -3327,3 +3313,43 @@ JitsiConference.prototype.setE2EEKey = function(key) {
...
+/**
+ * Setup E2EE for the sending side, if supported.
+ * Note that this is only done for the JVB Peer Connecction.
+ *
+ * @returns {void}
+ */
+JitsiConference.prototype._setupSenderE2EEForTrack = function(track) {
+    const jvbPc = this.jvbJingleSession ? this.jvbJingleSession.peerconnection : null;
+
+    if (!this.isP2PActive() && jvbPc && this._e2eeCtx) {
```
paweldomas, line 3326: do you really  need the `isP2PActive()` check?

paweldomas: I suppose you don't because you're working with the JVB peerconnection anyway

saghul: Right, I could simplify!

saghul:
Updated, PTAL @fippo @paweldomas

PR #1098, opened 2020-04-16
e2ee: decode static black frame for decryption errors

fippo:
similar to #1095 use a black keyframe instead of displaying a potentially
unencrypted stream. This was again captured with a modified version of
the webrtc sample, additional logging and a resolution of 320x180


fippo:
I think we need a different approach but this is an ok stop-gap.

The better approach would be to emit a signal about the decryption status and set it to false in case of errors. I would prefer to do that after workers though.

fippo:
@saghul also assuming we are now tagging commits with e2ee instead of hips since we are no longer hiding :-)

saghul:
Jenkins, add to whitelist.

saghul:
Jenkins, please test this please.

PR #1099, opened 2020-04-16
e2ee: remove encodedFrameType workaround

fippo:
no longer required as of Chrome 84.0.4108.0 and has been merged into M83 as well.


fippo:
[omahaproxy](https://omahaproxy.appspot.com/) says linux dev is still 83.0.4103.14 but that probably contains a bunch of other bugs and should get a new release early next week.

PR #1100, opened 2020-04-16
fix(firefox): Fix simulcast on FF ESR

jallamsetty1:
Make sure we don't have a=simulcast lines on the sendonly m-lines in the remote description

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -122,7 +122,15 @@ export class TPCUtils {
                 if (mline.type === 'video' && i !== idx) {
                     sdp.media[i].rids = undefined;
                     sdp.media[i].simulcast = undefined;
+
+                    // eslint-disable-next-line camelcase
+                    sdp.media[i].simulcast_03 = undefined;
                 }
+            });
+
+            return new RTCSessionDescription({
+                type: desc.type,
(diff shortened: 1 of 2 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -122,8 +122,19 @@ export class TPCUtils {
                 if (mline.type === 'video' && i !== idx) {
                     sdp.media[i].rids = undefined;
                     sdp.media[i].simulcast = undefined;
+
+                    // eslint-disable-next-line camelcase
+                    sdp.media[i].simulcast_03 = undefined;
                 }
             });
+
+            // sdp-transform seems to be inserting a= simulcast line for all the video m-lines even if we make
```
paweldomas, line 131: are we not in control of sdp-transform? is it not a better place for a fix? also what does `simulcast_03` stand for?

jallamsetty1: nope we are not in control of sdp-transform. `simulcast_03` is the old simulcast draft 03 which is implemented in FF older versions.

jallamsetty1: returning early doesn't modify the behavior, if we detect that the a=simulcast line isalready  present in the first video m-line, then we don't have to try to set it again.

PR #1104, opened 2020-04-17
fix(safari): construct track constraints from getSettings

jallamsetty1:
MediaStreamTrack getConstraints on Safari returns an empty object.
Enable device selection on Safari.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -96,6 +96,14 @@ export default class JitsiLocalTrack extends JitsiTrack {
             // Cache the constraints of the track in case of any this track
             // model needs to call getUserMedia again, such as when unmuting.
             this._constraints = track.getConstraints();
+
+            // Safari returns an empty constraints object, construct the constraints using getSettings.
+            if (!Object.keys(this._constraints).length && videoType === VideoType.CAMERA) {
+                this._constraints = {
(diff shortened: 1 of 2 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -96,6 +96,14 @@ export default class JitsiLocalTrack extends JitsiTrack {
             // Cache the constraints of the track in case of any this track
             // model needs to call getUserMedia again, such as when unmuting.
             this._constraints = track.getConstraints();
+
+            // Safari returns an empty constraints object, construct the constraints using getSettings.
+            if (!Object.keys(this._constraints).length && browser.isSafari() && videoType === VideoType.CAMERA) {
```
saghul, line 101: Why the browser check?

jallamsetty1: I wasn't sure if RN implemented getConstraints(), didn't want to mess it up when we start using the new gUM flow on RN as well. If it is implemented there, I can remove the browser check.

saghul: Removing the old gUM flow is not a small change, IMHO that is better handled when that change happens.

jallamsetty1: Sure, makes sense. I pushed the change with the browser check removed. Thanks for reviewing !

PR #1107, opened 2020-04-20
e2ee: also enable on p2p connections

fippo:
also enables e2ee on peer-to-peer connections. While a bit unnecessary
it avoids the confusing UX of things working until a third person joins

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1113,8 +1113,6 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
     newTrack._setConference(this);
 
     this.eventEmitter.emit(JitsiConferenceEvents.TRACK_ADDED, newTrack);
-
-    this._setupSenderE2EEForTrack(newTrack);
 };
 
 /**
@@ -1694,6 +1692,14 @@ JitsiConference.prototype.onRemoteTrackAdded = function(track) {
 JitsiConference.prototype.onCallAccepted = function(session, answer) {
     if (this.p2pJingleSession === session) {
         logger.info('P2P setAnswer');
+
+        // Setup E2EE.
+        const localTracks = this.getLocalTracks();
+
+        for (const track of localTracks) {
+            this._setupSenderE2EEForTrack(session, track);
+        }
+
         this.p2pJingleSession.setAnswer(answer);
     }
 };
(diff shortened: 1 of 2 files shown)


fippo:
@saghul I don't think that is quite enough and needs additional handling in _setupSenderE2EEForTrack (and the receiver), right?

fippo:
(also is this more complicated? what about transitions...)

saghul:
I'll check, I think some more stuff is needed.

fippo:
I *think* I found them all. API change: _setupSenderE2EEForTrack gets the jvb/p2p session as first argument

JitsiConference.js
```diff
@@ -3341,10 +3352,14 @@ JitsiConference.prototype._setupSenderE2EEForTrack = function(track) {
  * @returns {void}
  */
 JitsiConference.prototype._setupReceiverE2EEForTrack = function(track) {
-    const jvbPc = this.jvbJingleSession ? this.jvbJingleSession.peerconnection : null;
+    if (!this._e2eeCtx) {
+        return;
+    }
+    const session = track.isP2P ? this.p2pJingleSession : this.jvbJingleSession;
+    const pc = session && session.peerconnection;
```
saghul, line 3359: use `getActiveTPC` instead

JitsiConference.js
```diff
@@ -3320,17 +3331,17 @@ JitsiConference.prototype.setE2EEKey = function(key) {
...
-JitsiConference.prototype._setupSenderE2EEForTrack = function(track) {
-    const jvbPc = this.jvbJingleSession ? this.jvbJingleSession.peerconnection : null;
-
-    if (jvbPc && this._e2eeCtx) {
-        const sender = jvbPc.findSenderForTrack(track.track);
+JitsiConference.prototype._setupSenderE2EEForTrack = function(session, track) {
+    if (!this._e2eeCtx) {
+        return;
+    }
+    const pc = session.peerconnection;
```
saghul, line 3338: We ccould use geetActiveTPC` here as well instead of passsing the session around.

fippo: but I want the peerconnection the track is received on, no matter what the current establishment status of the p2p ice connection is?

saghul: That is taken care of by caller function already, isn't it?

fippo: no, this is called from JitsiConference.prototype.onCallAccepted (line 1701) for p2p and JitsiConference.prototype._acceptJvbIncomingCall for the sfu call. In both cases we know exactly where this is originating

saghul: Ah, gotcha.

PR #1110, opened 2020-04-21
sdp: remove remove(Session|Media)Lines

fippo:
removes two unused methods for mangling SDP.

### DIFF
--- a/modules/xmpp/SDP.js
+++ b/modules/xmpp/SDP.js
@@ -147,34 +147,6 @@ SDP.prototype.mangle = function() {
     this.raw = this.session + this.media.join('');
 };
 
-// remove lines matching prefix from session section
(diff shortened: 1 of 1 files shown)


fippo:
they've been unused [even here](https://github.com/ESTOS/strophe.jingle/blob/master/strophe.jingle.sdp.js#L44) so are probably an artefact of earlier SDP munging things ;-)

fippo:
ah... [used here](https://github.com/ESTOS/colibri.js/blob/master/colibri.js#L379). But that focus module is gone for a while.

fippo:
@paweldomas i hear you pulled the shorter one ;-)

hristoterezov:
Waiting for @paweldomas to approve since you requested a review from him. Otherwise I would merge :(

PR #1111, opened 2020-04-22
e2ee: use per-participant crypto contexts

fippo:
currently unused, will allow moving to per-participant keys

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -244,7 +244,10 @@ export default function JitsiConference(options) {
     this.maxFrameHeight = null;
 
     if (browser.supportsInsertableStreams()) {
-        this._e2eeCtx = new E2EEContext({ salt: this.options.name });
+        this._e2eeCtx = new E2EEContext({
+            myUserId: this.myUserId(),
+            salt: this.options.name
+        });
     }
 }
 
@@ -3448,7 +3451,7 @@ JitsiConference.prototype._setupReceiverE2EEForTrack = function(track) {
         const receiver = pc.findReceiverForTrack(track.track);
 
         if (receiver) {
-            this._e2eeCtx.handleReceiver(receiver, track.getType());
+            this._e2eeCtx.handleReceiver(receiver, track.getType(), track.id());
         } else {
             logger.warn(`Could not handle E2EE for remote ${track.getType()} track: receiver not found`);
         }
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -244,10 +244,7 @@ export default function JitsiConference(options) {
     this.maxFrameHeight = null;
 
     if (browser.supportsInsertableStreams()) {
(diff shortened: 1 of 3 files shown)


fippo:
timing-wise there is no way this can be unset, correct?
The alternative i considered was passing the local id to handleSender which would be nice an symmetrical to what needs to be done for handleReceiver

saghul:
Let's punt on this for now, I generally don't like to addd thingss until they are used / useful.

fippo:
lets keep it open then, I can do the other bits of moving to per-sender keys here as well

fippo:
moved to a data structure which allows per-participant keys without doing that yet.
Still looking for feedback on the api symmetry.

modules/e2ee/Worker.js
```diff
@@ -45,246 +35,290 @@ const code = `
...
+
+            contexts.forEach(async context => {
+                let key;
+
+                if (keyBytes) {
+                    key = await context.deriveKey(keyBytes);
+                } else {
+                    key = false;
+                }
+                context._currentKeyIndex++;
```
fippo, line 315: accessing private variables suggests this should be moved into a method

fippo:
per-participant keys should be a one-liner (including the local id in the deriveKey salt) after this

JitsiConference.js
```diff
@@ -236,7 +236,10 @@ export default function JitsiConference(options) {
     this._conferenceJoinAnalyticsEventSent = false;
 
     if (browser.supportsInsertableStreams()) {
-        this._e2eeCtx = new E2EEContext({ salt: this.options.name });
+        this._e2eeCtx = new E2EEContext({
+            myUserId: this.myUserId(),
```
fippo, line 240: the alternative would be to pass myUserId() in handleSender

saghul: Yeah, I like having that symmetry in the API :-)

modules/e2ee/E2EEContext.js
```diff
@@ -58,8 +59,9 @@ export default class E2EEcontext {
      *
      * @param {RTCRtpReceiver} receiver - The receiver which will get the decoding function injected.
      * @param {string} kind - The kind of track this receiver belongs to.
+     * @param {string} participantId - The (muc participant) id that this receiver belongs to.
```
saghul, line 62: Let's note mention the fact that this is MUC related, since that might change at some point. For all intents and purposes this is just an identifier for the participant, and that's that.

fippo:
meh, this breaks encryption even. If setKey is called *before* encode/decode the key is set, the contexts setKey method won't be called. So we need to store the raw keyBytes in setKey and then derive the key in encode/decode

fippo:
i'll have the next PR up for that shortly. But first we need to fix a regression

PR #1112, opened 2020-04-22
e2ee: move to a worker

fippo:
Moveѕ e2ee operations to a worker that is included as text/blob for now
to simplify deployment.


fippo:
setting `name: 'E2EE Worker'` is actually what enables the cool inspectability in the `sources` tab of Chrome devtools :+1:

PR #1116, opened 2020-04-22
e2ee: don't pass plain frames to decoder if we are encrypting

fippo:
If we are encrypting and the key index from a (potentially unencrypted)
frame exceeds our key ring size (1 currently) drop the frame.

This is a heuristic. We currently don't have the signaling for whether a
remote end does encrypt its frames or not.

### DIFF
--- a/modules/e2ee/E2EEContext.js
+++ b/modules/e2ee/E2EEContext.js
@@ -341,6 +341,13 @@ export default class E2EEcontext {
                     controller.enqueue(encodedFrame);
                 }
             });
+        } else if (keyIndex >= this._cryptoKeyRing.length
+                && this._cryptoKeyRing[this._currentKeyIndex % this._cryptoKeyRing.length]) {
+            // If we are encrypting but don't have a key for the remote drop the frame.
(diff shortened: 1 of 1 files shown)


fippo:
@saghul does this make sense? Also this will be a lot easier if we do signal whether a key is used

modules/e2ee/E2EEContext.js
```diff
@@ -341,6 +341,13 @@ export default class E2EEcontext {
                     controller.enqueue(encodedFrame);
                 }
             });
+        } else if (keyIndex >= this._cryptoKeyRing.length
+                && this._cryptoKeyRing[this._currentKeyIndex % this._cryptoKeyRing.length]) {
+            // If we are encrypting but don't have a key for the remote drop the frame.
```
saghul, line 346: decrypting?

fippo: no, this checks whether *we* have a key set. We don't know what the remote does

saghul: 🤦

PR #1119, opened 2020-04-24
feat(video-quality): control the sender resolution based on video quality settings

jallamsetty1:
For simulcast case, enable the stream encodings that have a resolution <= user setting.
For non-simulcast case, use applyConstraints to change the resolution of the stream.

This currently doesn't work on FF in the simulcast case 
https://bugzilla.mozilla.org/show_bug.cgi?id=1401592 is implementation bug for bringing RTCRtpParameters/setParameters/getParameters up to spec. I will open a new bug for this specific case and update it here.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -235,6 +235,12 @@ export default function JitsiConference(options) {
     this.recordingManager = new RecordingManager(this.room);
     this._conferenceJoinAnalyticsEventSent = false;
 
+    /**
+     * Max frame height that the user prefers to send to the remote participants.
+     * @type {number}
+     */
+    this.maxFrameHeight = null;
+
     if (browser.supportsInsertableStreams()) {
         this._e2eeCtx = new E2EEContext({ salt: this.options.name });
     }
@@ -1879,6 +1885,13 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
                 // to be turned off here.
                 if (this.isP2PActive() && this.jvbJingleSession) {
                     this._suspendMediaTransferForJvbConnection();
+                } else if (this.jvbJingleSession && this.maxFrameHeight) {
+                    // Apply user preferred max frame height if it was called before this
+                    // jingle session was created.
+                    this.jvbJingleSession.setSenderVideoConstraint(this.maxFrameHeight)
+                        .catch(err => {
+                            logger.error(`Sender video constraints failed on jvb session - ${err}`);
+                        });
                 }
 
                 // Setup E2EE.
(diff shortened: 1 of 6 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -521,6 +522,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
                         this._unregisterHandlers();
                         this.stopStream();
                         this._setStream(null);
+                        this.emit(TRACK_MUTE_CHANGED, this);
```
paweldomas, line 525: um moving `this.emit(TRACK_MUTE_CHANGED, this);` from the line 592 doesn't change much only diverges the flow and scatters the update to different places. What were you trying to achieve?

JitsiConference.js
```diff
@@ -3261,6 +3261,26 @@ JitsiConference.prototype.setReceiverVideoConstraint = function(
...
+ * @returns {Promise} promise that will be resolved when the operation is
+ * successful and rejected otherwise.
+ */
+JitsiConference.prototype.setSenderVideoConstraint = function(maxFrameHeight) {
+    const promises = [];
+
+    if (this.jvbJingleSession) {
+        promises.push(this.jvbJingleSession.setSenderVideoConstraint(maxFrameHeight));
+    }
+    if (this.p2pJingleSession) {
```
paweldomas, line 3277: So I don't think this is going to work out well. P2P connection comes and goes, so you will have the instance here only by luck.

I think you also need to store the `maxFrameHeight` value in the `JitsiConference` or in a dedicated submodule and then whenever JVB or P2P session is created apply the setting in case it didn't exist at the time this method was called.

modules/RTC/JitsiLocalTrack.js
jallamsetty1: I had to move this because we need to wait for the track to be removed/added from the peerconnection before emitting the `TRACK_MUTE_CHANGED` event. Otherwise, the video sender would not be present when I am trying to set the parameters on the sender immediately after the event is fired.

JitsiConference.js
jallamsetty1: Thanks for pointing it out, I see how it can fail in some cases.

modules/RTC/JitsiLocalTrack.js
paweldomas: do you mean `this._addStreamToConferenceAsUnmute`?

paweldomas: I think that's what the old code was doing, it was chaining `this._addStreamToConferenceAsUnmute` on to `promise` and then the events were emitted on line https://github.com/jitsi/lib-jitsi-meet/pull/1119/files/74aff9367bf7343722f0dd4104e966bc3a9529e6#diff-922e8edf325ba2a951da3fe2c72eaad4L592

PR #1132, opened 2020-04-29
fix(safari): Do not reuse inactive m-lines

jallamsetty1:
Audio fails on Safari when we try to re-use audio m-lines whose direction has been set to 'inactive'
when the remote source left the conference.
Disable noisy mic detection on Safari as this causes audio input to fail on Safari on iPadOS

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -419,7 +419,9 @@ JitsiConference.prototype._init = function(options = {}) {
         }
     }
 
-    if (config.enableNoisyMicDetection) {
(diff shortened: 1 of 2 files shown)


jallamsetty1:
> Question: doesn’t this mean that SDP may grow out of proportion if there are many join / leaves?

Unfortunately yes, this is what I was trying to avoid in the first place. I didn't notice issues with video so didn't specifically test audio. We can revert this change once the webkit bug is fixed.

saghul:
Sounds like a plan, thanks for explaining!

PR #1135, opened 2020-04-29
Fix safari audioinput device selection

jallamsetty1:


modules/RTC/RTCUtils.js
```diff
@@ -398,27 +398,36 @@ function newGetConstraints(um = [], options = {}) {
...
 
-        // NOTE(brian): the new-style ('advanced' instead of 'optional')
-        // doesn't seem to carry through the googXXX constraints
-        // Changing back to 'optional' here (even with video using
-        // the 'advanced' style) allows them to be passed through
-        // but also requires the device id to capture to be set in optional
-        // as sourceId otherwise the constraints are considered malformed.
-        if (!constraints.audio.optional) {
-            constraints.audio.optional = [];
+        // Use the standard audio constraints on non-chromium browsers.
```
saghul, line 401: Shouldn't we be more explicit here and use isChromiumBased / else?

JitsiMediaDevices.js
```diff
@@ -136,6 +136,14 @@ class JitsiMediaDevices {
             // Check using the Permissions API.
             this._permissionsApiSupported.then(supported => {
                 if (!supported) {
+                    // Workaround on Safari for audio input device
+                    // selection to work. Safari doesn't support the
+                    // permissions query.
+                    if (browser.isSafari()) {
```
saghul, line 142: This would return true for any type of device. Doesn't it have bad side-effects?

modules/RTC/RTCUtils.js
jallamsetty1: What about RN ?

JitsiMediaDevices.js
jallamsetty1: I have tested with all the audio/video devices I have and doesn't seem to have bad side-effects. My understanding is that we don't need to run permissions query for VIDEO_PERMISSION_NAME and AUDIO_PERMISSION_NAME anymore on any browsers, its legacy code before gUM wasn't standardized before and there were different implementations. Maybe check if its actually needed when we re-write the device selection logic. In the past I have done gUM calls directly w/o checking for permissions and haven't had issues.

jallamsetty1: I have to add I might be wrong as well :)

modules/RTC/RTCUtils.js
saghul: Those are not implemented anyway IIRC

JitsiMediaDevices.js
saghul: I think the original intent was to try and know beforehand if the permissions prompt was going to be displayed before asking.

Yeah, let’s keep it as is and reevaluate!

PR #1147, opened 2020-05-06
fix(E2EE): Apply E2EE on RTCRtpSender when track is replaced on pc

jallamsetty1:
Make sure we inject the encoding function on RTCRtpsenders when tracks are replaced on the peerconnection.
Update the MediaStreamTrack on the JitsiLocalTrack when effects are applied or removed so that
we can find the RTCRtpSender using the MediaStreamTrack

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -964,6 +964,16 @@ JitsiConference.prototype._fireMuteChangeEvent = function(track) {
         actorParticipant = this.participants[actorId];
     }
 
+    // Setup E2EE on the sender that is created for the unmuted track.
+    if (this._e2eeCtx && !track.isMuted() && browser.doesVideoMuteByStreamRemove()) {
+        if (this.p2pJingleSession) {
+            this._setupSenderE2EEForTrack(this.p2pJingleSession, track);
+        }
+        if (this.jvbJingleSession) {
+            this._setupSenderE2EEForTrack(this.jvbJingleSession, track);
+        }
+    }
+
     this.eventEmitter.emit(JitsiConferenceEvents.TRACK_MUTE_CHANGED, track, actorParticipant);
 };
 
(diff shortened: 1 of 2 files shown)


JitsiConference.js
```diff
@@ -1109,6 +1119,17 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
         this.room.setVideoMute(newTrack.isMuted());
     }
 
+    // Setup E2EE on the new track that has been added
```
ashitikov, line 1122: @jallamsetty1 Why not to put that code into _doReplaceTrack ? There is a code which also handles p2p & jvb track replace. So I think it could be a better place to setup sender.
Also, _setupNewTrack before that, did not depend on p2p/jvb jingle sessions.

JitsiConference.js
```diff
@@ -964,6 +964,16 @@ JitsiConference.prototype._fireMuteChangeEvent = function(track) {
         actorParticipant = this.participants[actorId];
     }
 
+    // Setup E2EE on the sender that is created for the unmuted track.
+    if (this._e2eeCtx && !track.isMuted() && browser.doesVideoMuteByStreamRemove()) {
```
paweldomas, line 968: all those little ifs here and there may finally get out of control at some point, and all end up in the JitsiConference which is now 3k lines long

paweldomas: what I mean by that is that it may be better off having e2ee logics in a separate module that works with events coming from other modules

saghul: Makes sense. I'll keep that in mind while working on adding Olm, which will only add more stuff to this. We probably need new events, but that's ok.

JitsiConference.js
jallamsetty1: To me this constitutes as an action that needs to be performed while handling the addition of the new track at the conference level rather than the replace operation. So it makes more sense to add it here.

PR #1149, opened 2020-05-06
ref: disable ICE restart by default

paweldomas:
Adds ICE_FAILED event emitted when ICE fails and disables it by default.

The reason for that it's currently causing issues with signaling when
Octo is enabled. Also when we do an "ICE restart"(which is not a real
ICE restart), the client maintains the TCC sequence number counter, but
the bridge resets it. The bridge sends media packets with TCC sequence
numbers starting from 0.

The 'enableIceRestart' config option can be used to force it, but it's
not recommended.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -84,6 +84,8 @@ const JINGLE_SI_TIMEOUT = 5000;
  * @param {number} [options.config.avgRtpStatsN=15] how many samples are to be
  * collected by {@link AvgRTPStatsReporter}, before arithmetic mean is
  * calculated and submitted to the analytics module.
+ * @param {boolean} [options.config.enableIceRestart=false] - enables the ICE
+ * restart logic.
  * @param {boolean} [options.config.p2p.enabled] when set to <tt>true</tt>
  * the peer to peer mode will be enabled. It means that when there are only 2
  * participants in the conference an attempt to make direct connection will be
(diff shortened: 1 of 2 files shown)


jienco:
Hi @paweldomas, I wanted to ask you if you could explain more about the potential problems of enabling `enableIceRestart`, not to change it in this repo but to enable it in our setup (with our own hosted servers). The reason is that we are getting `iceFailed` errors very frequently, even with Octo disabled, so we want to test enabling ICE restart in our case. Sorry to ping you here but this is the best place where I found information about the `enableIceRestart` and I thought it would be very helpful for us to get your comments on that. Thank you very much in advance.

paweldomas:
The issue you're asking about was answered by Boris on the most recent community call, but I can't find the recording... Can you post a question on the community forum as it seems to be of concerns to many users? I'll try to answer it there. Long story short is that it was done by allocating new channels on the bridge which meant fresh start there, but the peerconnection on the client kept some state which lead to weird issues. You may want to check out this [PR] as it does the same thing, but with a new peerconnection and clear state on the client side. The next step will be to implement a real ice restart on the bridge.

[PR]: https://github.com/jitsi/lib-jitsi-meet/pull/1262

jienco:
Thank you so much @paweldomas for still answering it here, I appreciate it very much. Will look for that community call thank you for that. I will also post the question in the community following your suggestion. Have a great day.

PR #1146, opened 2020-05-05
sdp: remove some jquery usage

fippo:
and make the code use arrow functions with named parameters in .each()/.map.

PR #1151, opened 2020-05-08
test: add a test for jingle to sdp conversion

fippo:
follow-up from #1146. Very high level but this should cover the relevant code paths.

The jingle offer was taken from a current focus offer with a few bits removed:
* ssrc-info element
* web-socket element inside transport
* rtcp-mux element inside transport
* bridge-session element
* bundle element (removed in focus already)


saghul:

PR #1156, opened 2020-05-11
Shuffle ICEServers for loadbalancing

psi-4ward:
Improve traffic loadbalancing when using multiple turnservers

https://chromium.googlesource.com/external/webrtc/+/refs/heads/master/pc/ice_server_parsing.cc#299

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -366,6 +366,15 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
 
                 const options = this.xmpp.options;
 
+                // Shuffle ICEServers for loadbalancing
+                for (let i = iceservers.length - 1; i > 0; i--) {
+                    const j = Math.floor(Math.random() * i);
+                    const temp = iceservers[i];
+
+                    iceservers[i] = iceservers[j];
+                    iceservers[j] = temp;
+                }
+                
                 if (options.useStunTurn) {
                     // we want to filter and leave only tcp/turns candidates
                     // which make sense for the jvb connections
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -374,7 +374,7 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
                     iceservers[i] = iceservers[j];
                     iceservers[j] = temp;
                 }
-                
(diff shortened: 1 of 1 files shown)


saghul:
@psi-4ward Can you please sign the CLA?

psi-4ward:
Did it already. I'm Christoph Wiechert

saghul:

fippo:
I don't think that is somethat that should be done in the client but rather mod_turncredentials.
And one typically uses a DNS hostname in a round-robin fashion so this shouldn't be required at all?

saghul:
> I don't think that is somethat that should be done in the client but rather mod_turncredentials.
> And one typically uses a DNS hostname in a round-robin fashion so this shouldn't be required at all?

We allow an arbirtrary number of STUN / TURN servers to be specified via config, so I'd ssay this is a good idea overall. It will not impact those who use a single one with other types of load balancing (as we do on meet.jit.si) but it will help those who lisst more than one server there.

psi-4ward:
DNS RR Setup is bit more complicated - think of certificate exchange.

Next: Whats the behavior when DNS resolvs to dead/faulty Coturn? ICE does not have alternative candidates in this case.

awlx:
We are running (@FreifunkMUC) this patch on https://meet.ffmuc.net and for the first time all our three turnservers see traffic. I added a graph where you can clearly see that it makes a difference.

Screenshot 2020-05-12 at 10 20 01

fippo:
> Next: Whats the behavior when DNS resolvs to dead/faulty Coturn? ICE does not have alternative candidates in this case.

Specifying multiple turn servers like that isn't a solution to that, it will generate relay candidates for each of them (you can see that in chrome://webrtc-internals) and increases the ice processing time. See the warning Firefox adds when you attempt to use more than four servers.

nils-ohlmeier:
I would suggest that rotating between multiple TURN servers should better get implemented in the browser endpoint. We are happy to take patches for that into Firefox :-)

It makes sense to specify let's two different TURN server with different DNS domains for redundancy purposes (in case DNS for one domain fails or the host goes down etc). But just specifying multiple TURN servers within the same domain just slows things down. And without any randomizing in the browser endpoints only slows things down as the ICE stacks are trying to gather relay candidates from all until they have found a working candidate pair.

PR #1160, opened 2020-05-12
e2ee: derive per-participant keys

fippo:
derived from the participant id in addition to the salt, separated by a null
byte to avoid ambiguity attacks along the lines of
  (someRoom, someParticipant)
  (someRoo, mSomeParticipant)

### DIFF
--- a/modules/e2ee/Worker.js
+++ b/modules/e2ee/Worker.js
@@ -83,6 +83,21 @@ const code = `
         }
 
         /**
+         * Derives a per-participant key.
+         * @param {Uint8Array} keyBytes - Value to derive key from
+         * @param {Uint8Array} salt - Salt used in key derivation
+         */
+        async deriveKey(keyBytes, salt) {
+            const encoder = new TextEncoder();
+            const idBytes = encoder.encode(this._id);
+            // Separate both parts by a null byte to avoid ambiguity attacks.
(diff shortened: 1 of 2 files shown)


modules/e2ee/Worker.js
```diff
@@ -82,6 +82,21 @@ const code = `
...
+         * @param {Uint8Array} keyBytes - Value to derive key from
+         * @param {Uint8Array} salt - Salt used in key derivation
+         */
+        async deriveKey(keyBytes, salt) {
+            const encoder = new TextEncoder();
+            const idBytes = encoder.encode(this._id);
+            // Separate both parts by a null byte to avoid ambiguity attacks.
+            const participantSalt = new Uint8Array(salt.byteLength + idBytes.byteLength + 1);
+            participantSalt.set(salt);
+            participantSalt.set(idBytes, salt.byteLength + 1);
```
saghul, line 96: Maybe set the null byte explictly instead of relying on the array beeing zeroed by default?

fippo: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Uint8Array/Uint8Array -- nulling is in the spec

saghul: 👍

PR #1162, opened 2020-05-12
fix(simulcast): Add only the primary ssrcs to SIM group on FF

jallamsetty1:
In preparation for turning on RTX on Firefox: add only the primary ssrcs to the SIM group if FID groups are present.
Also get rid of comments that are not relevant anymore, we now make sure that the ssrcs are in the correct order of quality (high to low) on Firefox.


jallamsetty1:
> did you test with FF nightly which has RTX behind a flag? Not fully ready yet apparently, see https://bugzilla.mozilla.org/show_bug.cgi?id=1164187

Thanks for the review ! Yes, I was testing with FF nightly with the pref enabled and noticed this issue. Dan has been using my standalone for verifying his fixes against Jitsi. I think I see another issue with simulcast and RTX that I am gonna bring it to his attention.

PR #1163, opened 2020-05-13
sdp: remove SDES mapping

fippo:
removes the SDES mapping. This is a leftover of the time when
Chrome only supported SDES and shouldn't be required anymore.


alvestrand:
Jubilation :-)

PR #1166, opened 2020-05-14
e2ee: update for latest spec changes in Chrome M84

fippo:
which landed in
  https://bugs.chromium.org/p/chromium/issues/detail?id=1069295#c4


fippo:
will wait for tomorrows canary, don't want to build myself. Quite happy with how easy the transition is!

fippo:
almost... `window` wasn't defined in the worker, now using `self`

fippo:
but then RTCEncodedAudioFrame is not defined in the worker scope.

fippo:
updated. We can't have nice things but at least we know we can remove the M83 code once the origin trial expires :-)
Tagging @emcho to keep track of that

PR #1172, opened 2020-05-16
make token fields more noticeable

josuah:
Jitsi meet currently have few logging to indicate that a token is wrong.
Extra visual hint and detail on how to build the token may help figuring
out what is missing.


josuah:
I will consider getting legally bound to the project if I have more serious contribution to add, and let anyone free to reproduce the changes suggested on that branch.  :)

PR #1173, opened 2020-05-18
fix(Firefox): Enable RTX support on Firefox

jallamsetty1:
Add RFC 4588 to the list of advertised features on Firefox.
Use RtxModifier only for cases when simulcast is achieved through SDP munging
This will ensure that RTX is negotiated whenever it is offered by the browser.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2430,7 +2430,7 @@ TraceablePeerConnection.prototype._createOfferOrAnswer = function(
                     dumpSDP(resultSdp));
             }
 
-            if (!this.options.disableRtx && browser.supportsRtx()) {
+            if (!this.options.disableRtx && browser.usesSdpMungingForSimulcast()) {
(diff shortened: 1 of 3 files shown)


fippo:
probably yet another "its firefox and its > 77-ish?

jallamsetty1:
> Can’t we detect RTX support at runtime? Wouldn’t this cause issues for ESR users?

This is what I am exactly doing now. Instead of explicitly enabling/disabling RTX on FF and change the capabilities we send to Jicofo, I am letting Jicofo offer RTX always and then FID groups will be signaled based on the offer that the browser generates. This should work on all versions on FF, I have tested ESR to FF nightly w/ RTX enabled.

jallamsetty1:
> probably yet another "its firefox and its > 77-ish?

Luckily we can skip that check in this case :)

saghul:
This is great @jallamsetty1 thanks for explaining!

fippo:
paging @nils-ohlmeier, *SOMEONE* understands using sdp for negotiatiation!

PR #1177, opened 2020-05-20
Add preferVP9/disableVP9 capability to lib-jitsi-meet

Syonyk:
While @JonathanLennox is working on the server side VP9 SVC support, this set of changes adds the ability to prefer/disable VP9 on the client side.

Similar to prefer/disable H264, this will rewrite the preferred ordering of supported codecs.

In testing, this does appear to properly enable/disable VP9 support, though the server cannot (yet) do proper SVC disassembly.


Syonyk:
To enable this functionality, https://github.com/jitsi/jitsi-meet/pull/6808 is also required.  That adds these to the whitelist.

It is safe to merge these changes without the whitelist entries PR, as this functionality cannot be accessed without the proper config entries.

fippo:
any chance you can use transceiver.setCodecPreferences along the lines of https://github.com/webrtc/samples/pull/1304 when its supported? Progressive enhancement is hard though.

Syonyk:
@fippo I'm almost certainly not the right person to make that set of changes.  I'm a currently bored firmware and kernel developer, not a web dev - it's been... interesting going up back into the stratosphere.  I'm not really fully up to speed on the webrtc interfaces at the moment.

jallamsetty1:
I think we should wait for the server side support work to be done or explicitly add this functionality only for p2p case. Since Safari doesn't support VP9 yet and bridge currently doesn't have a way of switching codecs mid-call, Safari client will not receive video if it joins a call where VP9 has been negotiated because of the preference settings.

Syonyk:
That might be something to mention in configs, but I don't think it's worth blocking VP9 support on Safari (which is relatively new to the whole Jitsi world anyway).  "Use Chrome or it won't work right" has been the general case for quite a while now, and getting the bandwidth of calls down would be quite useful for a lot of people.  I just don't have the time or skills to make the backend changes at this point.

pdarcos:
Hi guys,

Any update on VP9? Would love to try it out and see how it compares in real world use to VP8 or H264.

gpatel-fr:
@pdarcos 
it's possible to use VP9 in Jitsi-meet right now.
in Jicofo sip-communicator.properties set
org.jitsi.jicofo.ENABLE_H264=false
org.jitsi.jicofo.ENABLE_VP8=false
org.jitsi.jicofo.ENABLE_VP9=true

and disable simulcast in config.js

jallamsetty1:
Now that the bridge is able to support VP9 simulcast, code to handle VP9 preference on the client is being added through https://github.com/jitsi/lib-jitsi-meet/pull/1293.
Closing this PR.

PR #1184, opened 2020-05-25
Add server field description and example to the tokens.md documentation

jespino:
I added here the `server` field to the context that wasn't included before, I'm
not 100% sure about the expected usage of this, I guess is for track the
integration who generates the `JWT`.

I added it to the example payload too.


jespino:
Closing this because inactivity.

PR #1185, opened 2020-05-26
fix(presenter): change the max bitrate to 2.5 Mbps when presenter is enabled

jallamsetty1:
This behavior got altered when we correctly update the reference we keep of MediaStreamTrack in JitsiLocalTrack when effects are applied as part of https://github.com/jitsi/lib-jitsi-meet/commit/805eb938ae431cf71d67234fafaa35ae47be6308.
Apply max bitrates only when desktop track is added.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1957,6 +1957,8 @@ TraceablePeerConnection.prototype.setAudioTransferActive = function(active) {
  */
 TraceablePeerConnection.prototype.setMaxBitRate = function(localTrack) {
     const mediaType = localTrack.type;
+    const trackId = localTrack.track.id;
+    const videoType = localTrack.videoType;
 
     // No need to set max bitrates on the streams in the following cases.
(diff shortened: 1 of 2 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1957,23 +1957,21 @@ TraceablePeerConnection.prototype.setAudioTransferActive = function(active) {
...
         return;
     }
     if (!this.peerconnection.getSenders) {
         logger.debug('Browser doesn\'t support RTCRtpSender');
 
         return;
     }
-    const videoType = localTrack.videoType;
-    const trackId = localTrack.track.id;
+    const presenterEnabled = localTrack._originalStream
```
saghul, line 1973: wouldn't this also be true for any potentiall future effect on the video stream?

jallamsetty1: Yeah that is true .. any potential future effect that we apply on the desktop stream will alter the max bit rate. I couldn't find another way to detect this.

PR #1187, opened 2020-05-27
feat(callstats): Add dedicated siteID parameter to allow for injectio…

gimre:
…n of jwt-based siteID

### DIFF
--- a/modules/statistics/CallStats.js
+++ b/modules/statistics/CallStats.js
@@ -376,7 +376,7 @@ export default class CallStats {
                 const match = options.confID.match(/.*\/(.*)\/.*/);
 
                 // if there is no tenant, we will just set '/'
-                configParams.siteID = (match && match[1]) || '/';
+                configParams.siteID = options.siteID || (match && match[1]) || '/';
             }
 
             // userID is generated or given by the origin server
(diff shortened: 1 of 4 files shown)


damencho:
Can you please document it in https://github.com/jitsi/lib-jitsi-meet/blob/master/doc/API.md

modules/statistics/CallStats.js
```diff
@@ -376,7 +376,7 @@ export default class CallStats {
                 const match = options.confID.match(/.*\/(.*)\/.*/);
 
                 // if there is no tenant, we will just set '/'
-                configParams.siteID = (match && match[1]) || '/';
+                configParams.siteID = options.siteID || (match && match[1]) || '/';
```
damencho, line 379: The way it is, siteID will not be used if confID is missing ... I was wondering whether we should separate them, is siteID is passed use it, otherwise if confID is passed extract it from confID, otherwise fallback to `/`

gimre:
added docs

modules/statistics/CallStats.js
gimre: Hmm it looked like it was meant to be a package deal - you either get both or nothing. We might get siteID and no confID this way, is that supported?

damencho: Yeah it is not mentioned in the doc that confID is optional, you are right. If siteID is optional mark it as such in the doc?

gimre: Good point - marked as optional

damencho:
Jenkins add to whitelist.

PR #1216, opened 2020-06-17
fix(media): Add 'ended' event handler on MediaStreamTrack for FF

jallamsetty1:
When screensharing is stopped on FF from the sharing indicator (available in FF nightly), we are currently not detecting it.
Firefox fires a track 'ended' event as opposed to Chrome which fires a mediastream 'inactive' event as well.

### DIFF
--- a/modules/RTC/JitsiTrack.js
+++ b/modules/RTC/JitsiTrack.js
@@ -3,6 +3,7 @@ import EventEmitter from 'events';
 import { getLogger } from 'jitsi-meet-logger';
 import * as JitsiTrackEvents from '../../JitsiTrackEvents';
 import * as MediaType from '../../service/RTC/MediaType';
+import browser from '../browser';
 import RTCUtils from './RTCUtils';
 
 const logger = getLogger(__filename);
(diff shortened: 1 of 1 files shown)


modules/RTC/JitsiTrack.js
```diff
@@ -134,7 +140,7 @@ export default class JitsiTrack extends EventEmitter {
             }
         }
         if (this._streamInactiveHandler) {
-            addMediaStreamInactiveHandler(this.stream, undefined);
+            this._addMediaStreamInactiveHandler(undefined);
```
fippo, line 143: this is a bit confusing, why isn't the streamInactiveHandler passed here as well?

jallamsetty1: we are unregistering the handler here when the stream is getting removed from the conference.

jallamsetty1:
> i'd use track.onended in all browsers. stream.oninactive can happen for other (arguably unlikely) reasons such as all tracks getting removed from the stream.

Thanks for the review! I agree we should be using track.onended but this is something I haven't tested in the rest of the browsers yet. There is more refactoring planned around local/remote tracks and I will make this change then.

modules/RTC/JitsiTrack.js
fippo: ah. and it uses .onended, not addEventListener('ended') so that works. :+1:

PR #1223, opened 2020-06-18
ref/feat: add QualityController

paweldomas:
Refactors the way send/receive video constraints are managed and puts the high level
logic in a separate module. See QualityController.js for high level overview on how
the constraints are managed now.

Adds events to JitsiConference fired whenever it starts new jvb/p2p session.

Adds event to JingleSessionPC.js when remote party signals receive max frame height.
Also adds signaling of the local recv preference for the p2p mode(only existed for JVB).

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -61,6 +61,7 @@ import {
     createJingleEvent,
     createP2PEvent
 } from './service/statistics/AnalyticsEvents';
+import { QualityController } from './modules/qualitycontrol/QualityController';
 import * as XMPPEvents from './service/xmpp/XMPPEvents';
 
 const logger = getLogger(__filename);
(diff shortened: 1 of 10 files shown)


paweldomas:
> I see that you haven't moved the logic for reapplying the constraints when a new sender is created over to the QualityController module. As you recall, this would be needed when video is muted/unmuted (only in case of chrome) and when sources are changed. Do you plan to do this in another PR ?

To be honest I don't know what to do about this. I think this is an internal flaw of the TraceablePeerconnection and the way it deals with senders. It should not care about external top level events, but figure out when the senders change and apply the sender constraint. I don't plan to take this out to the higher level layers.

paweldomas:
What I mean is that I don't think it's a good idea to move that logic to the QualityController module.

jallamsetty1:
> > I see that you haven't moved the logic for reapplying the constraints when a new sender is created over to the QualityController module. As you recall, this would be needed when video is muted/unmuted (only in case of chrome) and when sources are changed. Do you plan to do this in another PR ?
> 
> To be honest I don't know what to do about this. I think this is an internal flaw of the TraceablePeerconnection and the way it deals with senders. It should not care about external top level events, but figure out when the senders change and re-apply the sender constraint to the value it has stored. I don't plan to take this out to the higher level layers.

Yeah I agree, it should be internal to TPC but should be handled in a better way than how it is done now. I can work on that if you think that will be helpful.

PR #1225, opened 2020-06-18
Remove screenshare extension support

jallamsetty1:
- Use getDisplayMedia on browsers where it is supported.
- desktopSharingChromeDisabled and desktopSharingFirefoxDisabled will no longer be supported.

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -117,8 +117,7 @@ const ScreenObtainer = {
 
             window.JitsiMeetScreenObtainer.openDesktopPicker(
                 {
-                    desktopSharingSources: desktopSharingSources
-                        || this.options.desktopSharingChromeSources
+                    desktopSharingSources
                 },
                 (streamId, streamType, screenShareAudio = false) =>
                     onGetStreamResponse(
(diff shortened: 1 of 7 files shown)


jallamsetty1:
@saghul sorry about all the force pushes, I just keep discovering more and more code that is obsolete. I have tested all clients except for spot which I will do tomorrow.

jallamsetty1:
Jenkins pls test this pls

modules/RTC/ScreenObtainer.js
```diff
@@ -188,8 +117,7 @@ const ScreenObtainer = {
 
             window.JitsiMeetScreenObtainer.openDesktopPicker(
                 {
-                    desktopSharingSources: desktopSharingSources
-                        || this.options.desktopSharingChromeSources
+                    desktopSharingSources
```
saghul, line 120: Since this is now only used in electron, can we hardcode this to `[ 'screen', 'window']` ?

jallamsetty1:
Jenkins pls test this pls

modules/RTC/ScreenObtainer.js
jallamsetty1: in a subsequent PR maybe ? that's what we decided in the standup.

PR #1227, opened 2020-06-24
ref: Handles video constraints from the bridge.

gpolitis:


modules/xmpp/JingleSessionPC.js
```diff
@@ -575,19 +573,17 @@ export default class JingleSessionPC extends JingleSession {
...
-                    });
+            // SENDER_VIDEO_CONSTRAINTS_CHANGED events and notify the peer connection
+            this.rtc.addListener(RTCEvents.SENDER_VIDEO_CONSTRAINTS_CHANGED,
+                idealHeight => {
+                    logger.info(`${this} received remote ideal frame height: ${idealHeight}`);
+                    this.remoteRecvMaxFrameHeight = idealHeight;
+                    this.eventEmitter.emit(
+                        MediaSessionEvents.REMOTE_VIDEO_CONSTRAINTS_CHANGED,
+                        this,
+                        idealHeight);
```
jallamsetty1, line 586: Looks like the height passed with this event is not used in the function invoked on the listener. JVB case is fine since you are setting `this.remoteRecvMaxFrameHeight` to this new height value received but @paweldomas I am curious as to how this works in the p2p case. Maybe I am missing something.

gpolitis: that's a great point Jaya.. thanks for pointing that out. I'll try to clean this up tomorrow.

gpolitis: It looks like the resolution argument was added to facilitate the tests. f6a83bb4 should alleviate this but I don't know how to run the tests. @paweldomas could you please confirm that f6a83bb4 is an acceptable solution?

modules/RTC/BridgeChannel.js
```diff
@@ -344,11 +344,11 @@ export default class BridgeChannel {
 
                 break;
             }
-            case 'SelectedUpdateEvent': {
-                const isSelected = obj.isSelected;
+            case 'SenderVideoConstraintsChangedEvent': {
+                const idealHeight = obj.videoConstraints.idealHeight;
```
paweldomas, line 348: is it possible for Websocket data channel to be established and receive this event, before JVB media session is established?

modules/xmpp/JingleSessionPC.js
paweldomas: yes, it's good

modules/RTC/BridgeChannel.js
```diff
@@ -345,10 +345,12 @@ export default class BridgeChannel {
                 break;
             }
             case 'SenderVideoConstraintsChangedEvent': {
-                const idealHeight = obj.videoConstraints.idealHeight;
+                const videoConstraints = obj.videoConstraints;
```
paweldomas, line 348: I really meant to store that value here, but thanks for going the extra mile

PR #1229, opened 2020-06-24
feat: Set degradation preference on video senders

jallamsetty1:
- Use 'maintain-framerate' for camera tracks and 'maintain-resolution' for desktop tracks.
- Apply cached video constraints/preferences on video unmute. Since the sender encodings are available only after the renegotiation is complete (on chrome/safari), constraints/preferences have to be applied after renegotiation is complete

### DIFF
--- a/modules/proxyconnection/ProxyConnectionPC.js
+++ b/modules/proxyconnection/ProxyConnectionPC.js
@@ -233,30 +233,12 @@ export default class ProxyConnectionPC {
         };
 
         /**
-         * A {@code JitsiConference} stub passed to the {@link RTC} module.
-         * @type {Object}
-         */
-        const conferenceStub = {
-            // FIXME: remove once the temporary code below is gone from
-            //  TraceablePeerConnection.
-            // TraceablePeerConnection:359
-            //  this.rtc.conference.on(
-            //         TRACK_ADDED,
-            //         maybeSetSenderVideoConstraints);
-            //     this.rtc.conference.on(
-            //         TRACK_MUTE_CHANGED,
-            //         maybeSetSenderVideoConstraints);
-            // eslint-disable-next-line no-empty-function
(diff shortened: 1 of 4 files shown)


modules/proxyconnection/ProxyConnectionPC.js
```diff
@@ -232,31 +232,13 @@ export default class ProxyConnectionPC {
...
-        };
-
         /**
          * Create an instance of {@code RTC} as it is required for peer
          * connection creation by {@code JingleSessionPC}. An existing instance
          * of {@code RTC} from elsewhere should not be re-used because it is
          * a stateful grouping of utilities.
          */
-        this._rtc = new RTC(conferenceStub, {});
+        this._rtc = new RTC(this, {});
```
paweldomas, line 241: why this change?

paweldomas: https://github.com/jitsi/lib-jitsi-meet/commit/4e9319b10acc9a3abf58d51bd7512174673b5a39

jallamsetty1: I am reverting the change you had to make here when I first added the handlers for TRACK_ADDED and TRACK_MUTE_CHANGED events. https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TraceablePeerConnection.js#L359
This is not needed anymore.

paweldomas: But even without that code, what is `this` referring to? RTC takes conference, so lets keep the `conferenceStub` there, but make it empty and without the FIXME comment.

jallamsetty1: oh I misread the comment in the code there. Added the empty stub.

PR #1233, opened 2020-06-30
feat(recording): JibriQueue support.

hristoterezov:
WIP! Don't merge!

### DIFF
--- a/modules/recording/JibriQueue.js
+++ b/modules/recording/JibriQueue.js
@@ -1,0 +1,244 @@
+/* global $ */
+
+import { getLogger } from 'jitsi-meet-logger';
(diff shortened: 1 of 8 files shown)


modules/recording/JibriQueue.js
```diff
@@ -0,0 +1,233 @@
...
+
+/**
+ * A static counter of the JibriQueue instances used to generate a unique ID for the JibriQueue instances.
+ */
+let id = 0;
+
+/**
+ * Represents a jibri queue.
+ */
+export default class JibriQueue extends EventEmitter {
```
paweldomas, line 17: how about extending `Listenable` which is kind of LJM layer for event emitter instead of extending library import directly?

PR #1236, opened 2020-07-01
cleanup: RTT Monitor is no longer used

jallamsetty1:
This was an experimental feature introduced for calculating the RTT to STUN servers, no longer needed.


jallamsetty1:
Jenkins pls test this pls.

PR #1237, opened 2020-07-02
fix: audio issues caused by insertable streams

paweldomas:
It's been observed that insertable streams are causing audio issues which sound like robotic voice.
The symptoms are lots of concealed events for the audio stream on the receiver side.
It seems to happen more when the browser or the main JS thread is under load.


saghul:
Can we have a config option instead, so we can test easier?

paweldomas:
> Can we have a config option instead, so we can test easier?

Well to be honest I was surprised it landed without a config option already

yanas:
@saghul We should definitely add a config option, but for now instead of disabling it completely can you re-work the implementation in a way so that we enable insertable streams only when someone has actually added a key and is all set to use the feature?

PR #1242, opened 2020-07-08
Takes disabled encodings into account when calculating the local resolution.

gpolitis:
With layer suspension enabled, a 720p track can be "capped" to 180p by parametrising the `RTPSender` in `TraceablePeerConnection.prototype.setSenderVideoConstraint` to enabled/disabled specific encodings. If the track is queried, it reports 720p resolution, but the higher resolution encodings can be disabled.

This can confuse the connection quality indicator/gsm bars to report bad connection because it expects a higher resolution to be streamed. This PR fixes this issue.

Additionally, @bgrozev suggested to report the max enabled resolution in the info bar on a per-participant basis to facilitate debugging. This isn't implemented yet, just the plumping to broadcast the max enabled resolution around.

### DIFF
--- a/modules/connectivity/ConnectionQuality.js
+++ b/modules/connectivity/ConnectionQuality.js
@@ -504,7 +504,8 @@ export default class ConnectionQuality {
         const videoType
             = localVideoTrack ? localVideoTrack.videoType : undefined;
         const isMuted = localVideoTrack ? localVideoTrack.isMuted() : true;
-        const resolution = localVideoTrack ? localVideoTrack.resolution : null;
+        const resolution = localVideoTrack
+            ? Math.min(localVideoTrack.resolution, localVideoTrack.maxEnabledResolution) : null;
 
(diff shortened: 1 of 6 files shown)


modules/connectivity/ConnectionQuality.js
```diff
@@ -278,6 +278,11 @@ export default class ConnectionQuality {
                     this._maybeUpdateUnmuteTime();
                 }
             });
+        conference.on(
+            ConferenceEvents.TRACK_MAX_ENABLED_RESOLUTION_CHANGED,
+            track => {
+                this._localStats.maxEnabledResolution = track.maxEnabledResolution;
```
paweldomas, line 284: is the extra conference event really needed? I suppose `ConnectionQuality` could subscribe to the `RTC` instance directly?

gpolitis:
@paweldomas could you please take another look?

PR #1243, opened 2020-07-08
fix(safari): override min/max video constraints on safari

jallamsetty1:
Override the constraints on Safari because of the following webkit bug.
https://bugs.webkit.org/show_bug.cgi?id=210932
Camera doesn't start on older macOS versions if min/max constraints are specified.


jallamsetty1:
> Can we extract the json parse / stringify into some clone function? Or use lodash deepClone maybe?

done !

jallamsetty1:
> Isn’t package-lock updated?

No it isn't. I specifically checked for it after I ran the build.

PR #1245, opened 2020-07-09
feat: use getSynchronizationSources on the receiver for remote audio levels

jallamsetty1:
Use getSynchronizationSources if it is supported, fallback to using getStats otherwise.
Existing testcases Peer2PeerTest and StartMutedTest check for the audio levels when the remote participants are unmuted so no additional testcases for covering this functionality are needed.


modules/statistics/RTPStatsCollector.js
```diff
@@ -291,31 +291,52 @@ StatsCollector.prototype.errorCallback = function(error) {
  */
 StatsCollector.prototype.start = function(startAudioLevelStats) {
     if (startAudioLevelStats) {
+        const receiverStatsSupported = typeof window.RTCRtpReceiver !== 'undefined'
```
paweldomas, line 294: can we have this defined in the `BrowserCapabilities` class as a method, so that others won't be duplicating it in future?

jallamsetty1: sure

modules/RTC/JitsiTrack.js
```diff
@@ -428,6 +428,10 @@ export default class JitsiTrack extends EventEmitter {
                 JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED,
                 audioLevel,
                 tpc);
+        } else if (this.audioLevel === 0 && audioLevel === 0 && this.isLocal()) {
```
paweldomas, line 431: hm so the audio level equal to 0 means there's no input? what's the value when muted?

jallamsetty1: LocalStatsCollector reports a value of 0.008 when the mic is muted, so we should be good here. I can add a check just in case.

paweldomas: can you mention that in a comment please?

jallamsetty1: Sure

jallamsetty1:

PR #1256, opened 2020-07-16
More then one video

hasevr:
Permit add more than one video/audio streams to TPC (PeerConnection)
Implement isScreenSharing() of JitsiTrack


hasevr:
I'm very sorry but this was my mistake.
I'd like to completely remove my request but I could not find the way to do
it.
I'll ask some extension for lib-jitsi-meet later but now is not the time.
Thank you.

2020年7月16日(木) 20:32 jitsi-jenkins <notifications@github.com>:

saghul:
No worries. AFAICT there is no way to delete a PR.

PR #1262, opened 2020-07-20
Use session-terminate to restart the session when ICE fails

paweldomas:


modules/xmpp/xmpp.js
```diff
@@ -233,13 +224,13 @@ export default class XMPP extends Listenable {
...
-                        this.connection.ping.startInterval(pingJid);
-                    } else {
-                        logger.warn(`Ping NOT supported by ${pingJid}`);
+                    if (!features.has(Strophe.NS.PING)) {
+                        logger.error(
+                            `Ping NOT supported by ${pingJid} - please enable ping in your XMPP server config`);
                     }
 
+                    this.connection.ping.startInterval(pingJid);
+
```
damencho, line 233: So we start the ping even if it is not supported?

paweldomas: Yes, because it's going to be mandatory so it can be used to kill a zombie WebSocket early. If it starts that will keep on brining error in the console and make it harder to be missed.

paweldomas: it's also used to simplify the flow around ice failed handling

paweldomas: I've added extra comment

paweldomas:
@damencho I've added the extra comment and checked the other Caps PR, but it doesn't fix my case, so I left the change here

paweldomas:
jenkins test this please

paweldomas:
jenkins test this please

paweldomas:
jenkins test this please

paweldomas:
jenkins test this please

PR #1263, opened 2020-07-23
feat(prejoin): Expose method to make a precall test

vp8x8:


modules/statistics/PrecallTest.js
```diff
@@ -0,0 +1,107 @@
+import browser from '../browser';
+import Settings from '../settings/Settings';
+import ScriptUtil from '../util/ScriptUtil';
+
+const SCRIPT_URL = 'https://api.callstats.io/static/callstats-ws.min.js';
```
saghul, line 5: Can you please share this with the other CS parts?

modules/statistics/PrecallTest.js
```diff
@@ -0,0 +1,107 @@
...
+ * @returns {Promise<void>}
+ */
+export async function init(options, onResult) {
+    if (_initialized) {
+        throw new Error('Precall Test already initialized');
+    }
+
+    await _loadScript();
+    // eslint-disable-next-line new-cap
+    api = new window.callstats();
```
saghul, line 86: Please make sure this also works on RN. We use a module for that. We might need confirmation that the precall API is available there.

package.json
```diff
@@ -60,7 +60,8 @@
     "postinstall": "webpack -p",
     "test": "karma start karma.conf.js",
     "test-watch": "karma start karma.conf.js --no-single-run",
-    "validate": "npm ls"
+    "validate": "npm ls",
+    "watch": "webpack --config webpack.config.js --watch --mode development"
```
saghul, line 64: I'm curious, how are you using this?

vp8x8: I use this script so the lib is _compiled_ on change and in `jitsi-meet` I just have a symlink to it.

modules/statistics/PrecallTest.js
vp8x8: Sorry, I don't know what *RN* is and how we should use the _precall test_ for it... Could you please give me some more details?

vp8x8: Yes, tested it and this should work on React Native too.

modules/statistics/PrecallTest.js
```diff
@@ -0,0 +1,107 @@
...
+
+    return _initialize(options, onResult);
+}
+
+/**
+ * Executes a pre call test.
+ *
+ * @returns {void}
+ */
+export function execute() {
```
paweldomas, line 96: I'd rather see `execute` return a `Promise` resolved with the results.

I like how you described the `PrecallTestOptions` type and it would be nice to have something similar for the results.

vp8x8: Yeah, it makes sense to have all types and results properly documented.

modules/statistics/PrecallTest.js
```diff
@@ -0,0 +1,126 @@
...
+ * @property rtt {number} - The round trip time.
+ * @property provider {string} - It is usually 'callstats'.
+ *
+ * @returns {Promise<{PrecallTestResults}>}
+ */
+export function execute() {
+    if (!_initialized) {
+        console.log('Precall Test not initialized');
+
+        return;
```
paweldomas, line 111: this should probably be `return Promise.reject`

modules/statistics/PrecallTest.js
```diff
@@ -0,0 +1,126 @@
...
+
+    return _initialize(options);
+}
+
+/**
+ * Executes a pre call test.
+ *
+ * @typedef PrecallTestResults
+ * @type {Object}
+ * @property mediaConnectivity {boolean} - If there is media connectivity or not.
```
paweldomas, line 99: type comes before property name

modules/statistics/PrecallTest.js
```diff
@@ -0,0 +1,126 @@
...
+ * @property throughput {number} - The average throughput.
+ * @property fractionalLoss - The packet loss.
```
paweldomas, line 101: missing type

modules/statistics/PrecallTest.js
```diff
@@ -0,0 +1,126 @@
...
+export function execute() {
+    if (!_initialized) {
+        console.log('Precall Test not initialized');
+
+        return;
+    }
+
+    return new Promise(resolve => {
+        emitter.once(PRECALL_TEST_RESULTS, result => {
+            resolve(result);
```
paweldomas, line 116: how about reject when `!success` at line 56? maybe the status can be processed here

modules/statistics/PrecallTest.js
vp8x8: Yep, good catch.

modules/statistics/PrecallTest.js
vp8x8: lol, my editor validated this although the order was reversed... 🤔

modules/statistics/PrecallTest.js
vp8x8: That's what I though in the first place. 💡

The thing is: every time I call `makePrecallTest` the first result on `preCallTestResults` is a failure with a `null` payload. Then, after a while, I receive a successful result with a proper payload. But I don't know if this behaviour is consistent or will suffer changes in time.

paweldomas: Oh so according to the docs[1] precall test runs continuously and I don't see any docs for the `makePrecallTest` method. Can you check with the CallStats folks about this?

[1]: https://docs.callstats.io/javascript/#callstats-startprecalltests

vp8x8: I just have ignored the first failure (for now)...

vp8x8:
@paweldomas Could you PTAL over the MR again?

PR #1265, opened 2020-07-23
feat: PerformanceObserverStats initial commit

jallamsetty1:
Add a performance stat around long tasks. Chrome supports PerformanceObserver API that lets us
register for long tasks event. Any task that takes longer than 50ms is considered a long task.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -375,6 +375,11 @@ JitsiConference.prototype._init = function(options = {}) {
         Statistics.analytics.addPermanentProperties({
             'callstats_name': this._statsCurrentId
         });
+
+        // Start performance observer for monitoring long tasks
+        if (config.performanceStatsInterval) {
+            this.statistics.attachPerformanceStats(this);
+        }
     }
 
     this.eventManager.setupChatRoomListeners();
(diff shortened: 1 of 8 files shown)


modules/statistics/statistics.js
```diff
@@ -283,6 +290,63 @@ Statistics.prototype.removeByteSentStatsListener = function(listener) {
...
+        () => this.performanceObserverStats.stopObserver());
+};
+
+/**
+ * Obtains the current value of the performance statistics.
+ *
+ * @returns {Object|null} stats object if the observer has been
+ * created, null otherwise.
+ */
+Statistics.prototype.getPerformanceStats = function() {
```
paweldomas, line 334: ah I see you actually wrote a getter here, but it's not used in `JitsiConference`.

JitsiConference.js
```diff
@@ -719,6 +724,16 @@ JitsiConference.prototype.getLocalVideoTrack = function() {
     return this.rtc ? this.rtc.getLocalVideoTrack() : null;
 };
 
+/**
+ * Obtains the performance statistics.
+ * @returns {Object|null}
+ */
+JitsiConference.prototype.getPerformanceStats = function() {
```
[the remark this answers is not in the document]
jallamsetty1: Yes, makes sense.

modules/statistics/statistics.js
jallamsetty1: I will switch to using this getter in `JitsiConference` instead of calling on the `performanceObserverStats` directly.

PR #1267, opened 2020-07-24
Update tokens.md

samlabs821:
From this comment i found that jitsi-meet-tokens replaced by jitsi-meet-prosody
https://github.com/jitsi/jitsi-meet/issues/6217#issuecomment-618246186


damencho:
That is not true, it is just that all modules had been moved to be provided by jitsi-meet-prosody and that is enough if you are doing a manual configuration and manually installing lua dependencies. But jitsi-meet-tokens still have some dependency installing and automatic configuration ...

PR #1276, opened 2020-08-05
WIP - e2ee: introduce per-participant randomly generated keys

saghul:
This the second stage in our E2EE journey.

Instead of using a single pre-shared passphrase for deriving the key used for
E2EE, we now establish a secure E2EE communication channel amongst peers.

This channel is implemented using libolm, using XMPP groupchat or JVB channels
as the transport.

Once the secure E2EE channel has been established each participant will generate
a random 32 byte key and exchange it over this channel.

Keys are rotated (well, just re-created at the moment) when a participant joins
or leaves.


modules/e2ee/E2EEncryption.js
```diff
@@ -64,13 +122,47 @@ export class E2EEncryption {
...
-     *
-     * @param {string} key - the key to be used.
-     * @returns {void}
+     * TODO.
+     */
+    _onParticipantJoined(id) {
+        logger.debug(`Participant ${id} joined`);
+
+        if (this._enabled) {
+            this._rotateKey();
```
fippo, line 131: can you call a .ratchet which for now calls rotateKey?

saghul: 👍

sapkra:
@saghul Do you already have a rough release date for this feature?

JitsiConference.js
```diff
@@ -3374,64 +3381,64 @@ JitsiConference.prototype._sendConferenceLeftAnalyticsEvent = function() {
...
- * Returns whether End-To-End encryption is supported. Note that not all participants
- * in the conference may support it.
+ * Restarts all active media sessions.
  *
- * @returns {boolean}
+ * @returns {void}
  */
-JitsiConference.prototype.isE2EESupported = function() {
-    const config = this.options.config;
+JitsiConference.prototype._restartMeddiaSessions = function() {
```
fippo, line 3388:
```suggestion
JitsiConference.prototype._restartMediaSessions = function() {
```

modules/e2ee/E2EEncryption.js
```diff
@@ -37,6 +60,84 @@ export class E2EEncryption {
...
+        if (enabled === this._enabled) {
+            return;
+        }
+
+        this._enabled = enabled;
+
+        // Need to re-create the peerconnections in order to apply the insertable streams constraint.
+        // TODO: this was necessary due to some audio issues when indertable streams are used
+        // even though encryption is not performed. This should be fixed in the browser eventually.
+        this.conference._restartMeddiaSessions();
```
fippo, line 110:
```suggestion
        this.conference._restartMediaSessions();
```

modules/e2ee/E2EEContext.js
```diff
@@ -23,24 +21,32 @@ const kJitsiE2EE = Symbol('kJitsiE2EE');
...
+     * FIXME: We currently use the MUC room name for this which has the same lifetime
+     * as this context. While not (pseudo)random as recommended in
+     * https://developer.mozilla.org/en-US/docs/Web/API/Pbkdf2Params
+     * this is easily available and the same for all participants.
+     * We currently do not enforce a minimum length of 16 bytes either.
      */
     constructor(options) {
         this._options = options;
 
+        // Figure out the URL for the worker script. Relative URLs are relative to
```
fippo, line 37:
```suggestion
        // Determine the URL for the worker script. Relative URLs are relative to
```

modules/e2ee/E2EEContext.js
```diff
@@ -53,6 +59,19 @@ export default class E2EEcontext {
         });
     }
 
+    /**
+     * Cleans up all state associated with the given participant. This is needed when a
+     * participant leaves the current conference.
+     *
+     * @param {string} participantId - The participant that just left.
+     */
+    cleanup(participantId) {
```
fippo, line 68: didn't you say we don't cleanup earlier today? Or is this just adding the cleanup

modules/e2ee/E2EEncryption.js
```diff
@@ -37,6 +60,84 @@ export class E2EEncryption {
...
+     */
+    setEnabled(enabled) {
+        if (enabled === this._enabled) {
+            return;
+        }
+
+        this._enabled = enabled;
+
+        // Need to re-create the peerconnections in order to apply the insertable streams constraint.
+        // TODO: this was necessary due to some audio issues when indertable streams are used
```
fippo, line 108: add link to https://bugs.chromium.org/p/chromium/issues/detail?id=1103280&

modules/e2ee/E2EEncryption.js
```diff
@@ -37,6 +60,84 @@ export class E2EEncryption {
...
+
+        // Send it to others using the E2EE olm channel.
+        this._olmAdapter.updateKey(key).then(() => {
+            // Set our key so we begin encrypting.
+            this._e2eeCtx.setKey(this.conference.myUserId(), key);
+        });
+    }
+
+    /**
+     * Generates a new 32 byte random key.
```
fippo, line 123: 256 bit? I like key length in bits

modules/e2ee/E2EEncryption.js
```diff
@@ -53,32 +154,89 @@ export class E2EEncryption {
...
+     * @private
      */
-    setKey(key) {
-        this._e2eeCtx.setKey(key);
+    async _rotateKey() {
+        const key = this._generateKey();
+
+        await this._olmAdapter.updateKey(key);
+
+        this._e2eeCtx.setKey(this.conference.myUserId(), key);
```
fippo, line 227: design question: this means we have the keys in userland js, not in the worker?

modules/e2ee/Worker.js
```diff
@@ -1,378 +1,382 @@
...
+ */
+function polyFillEncodedFrameMetadata(encodedFrame, controller) {
+    if (!encodedFrame.getMetadata) {
         encodedFrame.getMetadata = function() {
-          return {
-            // TODO: provide a more complete polyfill based on additionalData for video.
-            synchronizationSource: this.synchronizationSource,
-            contributingSources: this.contributingSources
-          };
+            return {
```
fippo, line 14: we should remove this and drop M83 support. Later

modules/xmpp/ChatRoom.js
```diff
@@ -1063,17 +1063,21 @@ export default class ChatRoom extends Listenable {
                     from, invite.attr('from'), txt, password);
             }
         }
+
         const jsonMessage = $(msg).find('>json-message').text();
```
fippo, line 1067: is this spec-compliant? :rofl:

modules/e2ee/E2EEContext.js
saghul: Oh, looks like past saghul was smart enough to solve it already!

modules/e2ee/E2EEncryption.js
saghul: 👍

modules/e2ee/E2EEncryption.js
saghul: Yes. We generate it outside of the worker and then send it to the worker. In addition, we pass it on the olm adapter, which sends it to everyone else, and keeps it around to send it in ACKs. Now, we can probably get rid of this, since we regenerate the key when a new participant joins anyway, and once we add ratcheting, we'll just send a message to the worker to ratchet.

modules/e2ee/Worker.js
saghul: Agreed. Since it's minimal, and I don't want to add a version check, I thought we can keep it around for a bit longer.

modules/xmpp/ChatRoom.js
saghul: Don't tell anyone.

fippo (approved):
ship it (to beta)

PR #1278, opened 2020-08-07
fix(video-quality): add a check for local video track before applying max bitrates

jallamsetty1:


psi-4ward:
EXACTLY what my fix looks like. Perhaps I should open pr instead of issues 😁

PR #1281, opened 2020-08-11
Promise cleanup in the RTC module

paweldomas:


modules/RTC/TPCUtils.js
```diff
@@ -418,7 +417,7 @@ export class TPCUtils {
                         parameters.encodings.forEach(encoding => {
                             encoding.active = true;
                         });
-                        transceiver.sender.setParameters(parameters);
+                        setParamPromises.push(transceiver.sender.setParameters(parameters));
```
jallamsetty1, line 420: Was this the reason for the onnegotiationneeded loop ? I somehow missed the fact that RTCRtpSender.setParameters is async :(

paweldomas: No worries. It wasn't - the onnegotiationneeded loop was caused on my experimental branch where I was setting sendrecv direction on the recvonly transceiver for the remote stream.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1481,21 +1481,23 @@ TraceablePeerConnection.prototype.containsTrack = function(track) {
 /**
  * Add {@link JitsiLocalTrack} to this TPC.
  * @param {JitsiLocalTrack} track
+ * @returns {Promise<void>} - resolved when done.
```
jallamsetty1, line 1484: We have made `addTrack` return a Promise now but in JingleSessionPC.js where the local tracks are added before offer/answer, we are not waiting for the promise to get resolved.

paweldomas: I'll take a look - must have missed it, thanks!

paweldomas: It seems that the encoding could be passed as `transceiverInit`[1] and `addTrack` won't need it do be async? I'm wondering why is it currently done only for the "isInitiator and not Firefox" case? Is it possible to get rid of [2] and pass the encoding to the constructor?

Also I don't understand why TCPUtils makes assumptions about whether or not tracks are added before 'session-initiate' is sent to another peer[3]. Dealing with problems like this should belong to the jingle session layer and should be of no concern to TCPUtils which belong the RTC layer.

[1]: https://github.com/jitsi/lib-jitsi-meet/blob/d219b6b91b60c6f726b66576f34cb67adab23abd/modules/RTC/TPCUtils.js#L225
[2]: https://github.com/jitsi/lib-jitsi-meet/blob/d219b6b91b60c6f726b66576f34cb67adab23abd/modules/RTC/TraceablePeerConnection.js#L1547
[3]: https://github.com/jitsi/lib-jitsi-meet/blob/d219b6b91b60c6f726b66576f34cb67adab23abd/modules/RTC/TPCUtils.js#L216

jallamsetty1: FF doesn't take the encodings in the `transceiverInit`, they have to be configured after the transceiver is created by calling `setParameters` on the sender. That is the major difference between Chrome/Safari unified vs Firefox unified plan impl. 

In the transceiver model, we do have to determine how to add the tracks based on whether the client is an initiator or a responder. In the responder case, a transceiver is already created when we call sRD. Calling addTransceiver() again would create a new transceiver, so pc.addTrack() needs to be called in this case to make sure the track gets attached to the existing transceiver.
The RTC layer is what add the tracks to the pc, so these decisions have to be made here.

paweldomas: When I'm looking at the code in `JingleSessionPC.setOfferAnswerCycle` the tracks are added, before the remote description is set.

If RTC module relies on whether or not a transceiver exists it should do a check for it by scanning the list on the peerconnection. Same applies to relying on whether or not a remote description has been set - the code should check the `remoteDescription` field.

modules/xmpp/JingleSessionPC.js
```diff
@@ -926,15 +926,19 @@ export default class JingleSessionPC extends JingleSession {
...
-    invite(localTracks) {
+    invite(localTracks = []) {
         if (!this.isInitiator) {
             throw new Error('Trying to invite from the responder session');
         }
         const workFunction = finishedCallback => {
+            const addTracks = [];
+
             for (const localTrack of localTracks) {
                 this.peerconnection.addTrack(localTrack, true /* isInitiator */);
```
jallamsetty1, line 937: I think you are missing the push to `addTracks` here.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Makes sense, as discussed, we can clean this up when working on p2p case in unified since this code path is not being used in the jvb case which is the only active mode for unified plan now.

modules/xmpp/JingleSessionPC.js
paweldomas: 🤦

paweldomas:
@jallamsetty1 I had to force push to resolve the conflicts

PR #1284, opened 2020-08-12
ref: Moves xmpp logs to be accessed from connection.

damencho:
In cases where there is no room like pre-join and lobby screen we still want to be able to debug xmpp messages.

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -261,7 +261,7 @@ export default class XMPP extends Listenable {
                                         }
                                     });
                                 })
-                                .catch(logger.warn('Error getting features from lobby.'));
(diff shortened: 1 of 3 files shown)


modules/xmpp/xmpp.js
```diff
@@ -261,7 +261,7 @@ export default class XMPP extends Listenable {
                                         }
                                     });
                                 })
-                                .catch(logger.warn('Error getting features from lobby.'));
+                                .catch(() => logger.warn('Error getting features from lobby.'));
```
paweldomas, line 264: is the error logged anywhere up? can you do `error => logger.warn('Error getting features from lobby.', error && error.message))` ?

damencho: will do

PR #1289, opened 2020-08-15
Fix: Scale down localVideo which is already smaller than req resulition

psi-4ward:
When disableSimulcast=true the Lib tries to scale-down the local-video track to match `config.resolution`.
If the local-video track is already slower than config.resolution it results in `scaleResolutionDownBy=0` which ist not a valid scaling value.

---
_CLA already signed, see https://github.com/jitsi/lib-jitsi-meet/pull/1156_


jallamsetty1:
Jenkins test this please.

PR #1293, opened 2020-08-17
ref(sdp): Move all instances of a preferred codec up the list in the SDP

jallamsetty1:
1. Make sdp munging logic for changing codec preferences more generic
2. Do not negotiate High profile H264 codecs on when RN clients are in the call.
3. Add options for preferring VP9 over other codecs.

Right now SDP munging is done on the all browsers for changing the order, the impl using setCodecPreferences on the transceiver hasn't be added since it is working only on chrome now in unified plan impl and Safari has a bug. 
Firefox has no support for setCodecPreferences.
That implementation will be added when the safari bug is resolved.
https://bugs.webkit.org/show_bug.cgi?id=215567


modules/xmpp/SDPUtil.js
```diff
@@ -612,57 +603,69 @@ const SDPUtil = {
...
      * it's disabled.
      *
-     * @param {object} videoMLine the video mline object from an sdp as parsed
-     * by transform.parse.
+     * @param {object} mLine the mline object from an sdp as parsed by transform.parse.
      * @param {string} codecName the name of the codec which will be stripped.
      */
-    stripVideoCodec(videoMLine, codecName) {
-        if (!videoMLine || !codecName) {
+    stripCodec(mLine, codecName, highProfile = false) {
```
paweldomas, line 609: The `highProfile` argument seems to be very h264 codec specific. It's also missing JSDoc description. Would it make sense to pass an item matcher function instead? In this example it would be `item => item.config.includes('profile-level-id=64')`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -268,6 +268,30 @@ export default function TraceablePeerConnection(
      */
     this.senderVideoMaxHeight = null;
 
+    // Set the codec preference that will be applied on the SDP
+    // based on the config.js settings.
+    if (this.options.preferH264 || this.options.preferVP9) {
```
paweldomas, line 273: Are `preferH264` and `preferVP9` mutually exclusive? I would probably vote for a generic 'preferredCodec' option rather than adding `preferX`, `preferY`, `preferZ` and so on. The same applies for `disableH264` and `disableVP9` options.

Should we ever want to prefer or disable more than one codec then it would be easy to handle an array in future.

jallamsetty1: Yes, `preferH264` and `preferVP9` are mutually exclusive. Bridge currently doesn't support H264 simulcast, so we would have to disable simulcast if we want to use H264 over jvb connection and our bridge doesn't support unicast VP9. VP9 has other issues like Safari doesn't support it yet and FF has a bug so we would prefer H264 over VP9 if both are present.
I left the `preferH264` option unchanged for backward compatibility.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -268,6 +269,57 @@ export default function TraceablePeerConnection(
      */
     this.senderVideoMaxHeight = null;
 
+    // We currently support preferring/disabling video codecs only.
+    const getCodecMimeType
+        = codec => Object.values(CodecMimeType).find(value => value === codec.toLowerCase());
+
+    // Set the codec preference that will be applied on the SDP based on the config.js settings.
+    if (this.options.preferH264 || this.options.preferredCodec) {
```
paweldomas, line 277: I think picking up the `preferredCodec` value  can be simplified to:

`let preferredCodec = getCodecMimeType(this.options.preferredCodec || (this.options.preferH264 && CodecMimeType.H264));`

Next handle any corner cases:

```
if ((browser.isFirefox() && preferredCodec === CodecMimeType.VP9) {
    preferredCodec = null;
}
```

Then do similar thing to figure out the `disabledCodec` and once you have both it will come down to:

```
if (preferredCodec || disabledCodec) {
this.codecPreference = {
            enable: disabledCodec ? false : true,
            mediaType: MediaType.VIDEO,
            mimeType: disabledCodec ? disabledCodec : preferredCodec
        };
}
```

btw. kudos for defining `CodecMimeType`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1460,6 +1512,44 @@ TraceablePeerConnection.prototype._getSSRC = function(rtcId) {
...
+ * Munges the order of the codecs in the SDP passed based on the preference
+ * set through config.js settings. All instances of the specified codec are
+ * moved up to the top of the list when it is preferred. The specified codec
+ * is deleted from the list if the configuration specifies that the codec be
+ * disabled.
+ * @param {RTCSessionDescription} description that needs to be munged.
+ * @returns {RTCSessionDescription} the munged description.
+ */
+TraceablePeerConnection.prototype._mungeCodecOrder = function(description) {
+    if (!(this.codecPreference && this.codecPreference.mimeType) || browser.supportsCodecPreferences()) {
```
paweldomas, line 1525: The check for `this.codecPreference` should be enough, given that it's initialized only for correct values now

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1869,24 +1959,9 @@ TraceablePeerConnection.prototype.setLocalDescription = function(description) {
...
-            type: localSdp.type,
-            sdp: transform.write(parsedSdp)
-        });
-
-        this.trace('setLocalDescription::postTransform (H264)',
-            dumpSDP(localSdp));
-    }
+    // Munge the order of the codecs based on the preferences set through config.js
+    // eslint-disable-next-line no-param-reassign
+    description = this._mungeCodecOrder(localSdp);
```
paweldomas, line 1964: `localSdp = this._mungeCodecOrder(localSdp);` and the eslint-disable is not necessary here

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2068,6 +2143,10 @@ TraceablePeerConnection.prototype.setMaxBitRate = function(localTrack = null) {
 TraceablePeerConnection.prototype.setRemoteDescription = function(description) {
     this.trace('setRemoteDescription::preTransform', dumpSDP(description));
 
+    // Munge the order of the codecs based on the preferences set through config.js
+    // eslint-disable-next-line no-param-reassign
+    description = this._mungeCodecOrder(description);
```
paweldomas, line 2148: I'm not sure what are those reassigns doing all over the `setRemoteDescription`. By no means it wasn't meant as a guideline. It probably creeped in when eslint was enabled for the first time and the intention was to do least changes possible, so ignore were added everywhere. But you can keep this one since it's all over already, but not something to do in the future.

modules/RTC/TraceablePeerConnection.js
paweldomas: I'm not sure how this ever worked given that `localSdp` instead of the `description` is used by all of the code below

modules/xmpp/JingleSessionPC.js
```diff
@@ -331,12 +331,24 @@ export default class JingleSessionPC extends JingleSession {
...
+
         if (this.isP2P) {
             // simulcast needs to be disabled for P2P (121) calls
             pcOptions.disableSimulcast = true;
             pcOptions.disableH264 = options.p2p && options.p2p.disableH264;
             pcOptions.preferH264 = options.p2p && options.p2p.preferH264;
 
+            // codec preference options for p2p.
+            if (options.p2p) {
+                pcOptions.disabledCodec = options.p2p.disabledCodec;
```
paweldomas, line 348: Can you add the new options description here? https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TraceablePeerConnection.js#L54
Also when opening a PR to jitsi-meet please mark `disableH264` and `preferH264` options as deprecated in the config.js

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Thanks for pointing this out, I think this bug crept in when I moved the logic out to mungeCodecOrder 🤦

PR #1300, opened 2020-08-20
Remove _fixAnswerRFC4145Setup.

JonathanLennox:
We're pretty sure it's no longer necessary with JVB2; and it costs
half an RTT in setup, and stops us being able to choose which
DTLS/SRTP protection profile to use.


jallamsetty1:
Jenkins test this please.

damencho:
Jenkins, test this please.

PR #1311, opened 2020-09-03
Adds utility methods

gpolitis:


JitsiConference.js
```diff
@@ -2407,6 +2407,31 @@ JitsiConference.prototype.isCallstatsEnabled = function() {
...
 };
 
+/**
+ * Finds the SSRC of a given track
+ *
+ * @param track
+ * @returns {number|undefined} the SSRC of the specificed track, otherwise undefined.
+ */
+JitsiConference.prototype.getSsrcByTrack = function(track) {
+    const isP2P = track.isP2P;
```
hristoterezov, line 2417: It seems this is not available for local tracks where you are actually using it. Am I missing something?

Maybe you can obtain the peer connection by using `getActivePeerConnection`?

hristoterezov (changes requested):
Left a comment!

JitsiConference.js
gpolitis: @hristoterezov thanks for taking a look!

> It seems this is not available for local tracks where you are actually using it.

Hmm.. You mean the `isP2P` field, right? That's interesting.. I don't understand all the interactions here, I've simply copied the ssrc lookup code from the `_onTrackAttach` function below.

Here's my concern with your suggestion: if the `track` parameter refers to a local P2P track and the active connection is the JVB one, then the function won't be able to find the track ssrc.

Unfortunately, the current code suffers from similar issues as you've correctly pointed out: if the `track` parameter refers to a local P2P track we'll never search for the SSRC in the P2P peer connection because `track.isP2P` is undefined (falsy). I wonder if this situation causes issues in the `_onTrackAttach` function below?

paweldomas: Only remote tracks are either p2p or not p2p. Local tracks can belong to many peerconnections.

paweldomas: This looks good in general, but I would also suggest using `getActivePeerConnection`

gpolitis: Thanks guys! I patched my code the way you suggested.

JitsiConference.js
```diff
@@ -2407,6 +2407,29 @@ JitsiConference.prototype.isCallstatsEnabled = function() {
     return this.statistics.isCallstatsEnabled();
 };
 
+/**
+ * Finds the SSRC of a given track
+ *
+ * @param track
+ * @returns {number|undefined} the SSRC of the specificed track, otherwise undefined.
+ */
+JitsiConference.prototype.getSsrcByTrack = function(track) {
```
hristoterezov, line 2416: Maybe you can do something like to simplify the code:

```
return track.isLocal() ? this.getActivePeerConnection()?.getLocalSSRC(track) : track.getSSRC();
```

WDYT?

NOTE: If you haven't seen `?.` here are some docs about it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

hristoterezov (approved):
Left a suggestion for simplifying the code but other than that LGTM!

JitsiConference.js
gpolitis: That's neat, I love one-liners 👍

gpolitis:
Closing in favor of https://github.com/jitsi/lib-jitsi-meet/pull/1338

PR #1322, opened 2020-09-18
feat(stats): Add the ability to enable callStats only on a certain % of users

jallamsetty1:
A new config.js setting `testing->callStatsThreshold` is introduced that determines the % of conferences on which callStats will be enabled. It takes a value between 0 and 100.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -38,6 +38,7 @@ import SpeakerStatsCollector from './modules/statistics/SpeakerStatsCollector';
 import Statistics from './modules/statistics/statistics';
 import Transcriber from './modules/transcription/transcriber';
 import GlobalOnErrorHandler from './modules/util/GlobalOnErrorHandler';
+import { hashString } from './modules/util/MathUtil';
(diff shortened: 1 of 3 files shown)


damencho (dismissed):

JitsiConference.js
```diff
@@ -374,6 +374,14 @@ JitsiConference.prototype._init = function(options = {}) {
             });
     this.participantConnectionStatus.init();
 
+    // Add the ability to enable callStats only on a certain percentage of users based
+    // on config.js setting.
+    let disableCallStats = false;
+
+    if (config.testing && config.testing.callStatsUserThreshold) {
+        disableCallStats = Math.random() > config.testing.callStatsUserThreshold;
```
bgrozev, line 382: This will lead to callstats being enabled for some participants in a conference and not others. Are we okay with that? Can we do this pseudo-randomly based on the conference name instead?

jallamsetty1: How do we do based on the conference name ?

bgrozev (approved):

PR #1326, opened 2020-09-23
e2ee: detect transferable streams

fippo:
Since we use transferable streams to transfer the streams to the worker
we need to feature detect them as Chrome 86 will launch Insertable Streams
by default without shipping transferable streams yet (status still unclear,
it might still ship too)

See here for the feature detection used:
  https://groups.google.com/a/chromium.org/g/blink-dev/c/1LStSgBt6AM/m/hj0odB8pCAAJ


saghul (approved):
LTGM! Also 🤮

fippo:
We can keep our fingers crossed that this still makes it to M86. And this still makes the check more robust for other browsers.

PR #1332, opened 2020-09-25
e2ee: refactor to make it clear when to ratchet

fippo:
refactor the signature checking to make it clear when we need to ratchet.
This is similar to
  https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.5.1
but we verify the hash before attempting to decrypt.


fippo:
recommend reviewing with whitespace turned off.
Style-wise I prefer to use promise for decrypt since that can throw

saghul (approved):

PR #1334, opened 2020-09-25
feat(external-storage): Support.

hristoterezov:
Depends on https://github.com/jitsi/js-utils/pull/40


doc/API.md
```diff
@@ -44,6 +44,7 @@ The ```options``` parameter is JS object with the following properties:
     - `enableWindowOnErrorHandler` - boolean property (default false). Enables/disables attaching global onerror handler (window.onerror).
     - `disableThirdPartyRequests` - if true - callstats will be disabled and the callstats API won't be included.
     - `enableAnalyticsLogging` - boolean property (default false). Enables/disables analytics logging.
+    - `externalStorage` - Obejct that implements the Storage interface. If specified this object will be used for storing data instead of `localStorage`.
```
fippo, line 47:
```suggestion
    - `externalStorage` - Object that implements the Storage interface. If specified this object will be used for storing data instead of `localStorage`.
```

jallamsetty1 (dismissed):

jallamsetty1 (approved):

PR #1341, opened 2020-09-29
 e2ee: ratchet the key forward on authentication tag errors

fippo:
similar to what is explained here:
  https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.5.1
but we do it on authentication tag failures since it is not possible
to tell whether decrypt.

### DIFF
--- a/doc/e2ee.md
+++ b/doc/e2ee.md
@@ -39,6 +39,14 @@ We do not encrypt the first few bytes of the packet that form the VP8 payload
 This allows the decoder to understand the frame a bit more and makes it decode the fun looking garbage we see in the video.
 This also means the SFU does not know (ideally) that the content is end-to-end encrypted and there are no changes in the SFU required at all.
 
+## Key Ratcheting
+Unlike described in
+  https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.5.1
+we attempt to ratchet the key forward when we do not find a valid
+authentication tag. Note that we only update the set of keys when
+we find a valid signature which avoids a denial of service attack with invalid signatures.
+
+TODO: if a frame ratchets the key forward it should be signed with the senders private key.
 
 ## Using workers
 
--- a/modules/e2ee/E2EEContext.js
+++ b/modules/e2ee/E2EEContext.js
@@ -23,17 +23,8 @@ const kJitsiE2EE = Symbol('kJitsiE2EE');
 export default class E2EEcontext {
     /**
      * Build a new E2EE context instance, which will be used in a given conference.
-     *
-     * @param {string} options.salt - Salt to be used for key deviation.
-     * FIXME: We currently use the MUC room name for this which has the same lifetime
-     * as this context. While not (pseudo)random as recommended in
-     * https://developer.mozilla.org/en-US/docs/Web/API/Pbkdf2Params
-     * this is easily available and the same for all participants.
-     * We currently do not enforce a minimum length of 16 bytes either.
      */
-    constructor(options) {
-        this._options = options;
-
+    constructor() {
         // Determine the URL for the worker script. Relative URLs are relative to
         // the entry point, not the script that launches the worker.
         let baseUrl = '';
--- a/modules/e2ee/E2EEContext.js
+++ b/modules/e2ee/E2EEContext.js
@@ -139,15 +139,4 @@ export default class E2EEcontext {
             keyIndex
         });
     }
-
-    /**
-     * Ratchet our own key.
-     * @param {string} participantId - the ID of the participant who's key we should ratchet.
-     */
-    ratchet(participantId) {
-        this._worker.postMessage({
-            operation: 'ratchet',
-            participantId
-        });
-    }
 }
--- a/modules/e2ee/E2EEContext.js
+++ b/modules/e2ee/E2EEContext.js
@@ -148,4 +148,15 @@ export default class E2EEcontext {
             keyIndex
         });
     }
+
+    /**
+     * Ratchet our own key.
+     * @param {string} participantId - the ID of the participant who's key we should ratchet.
+     */
+    ratchet(participantId) {
+        this._worker.postMessage({
+            operation: 'ratchet',
+            participantId
+        });
+    }
 }
--- a/modules/e2ee/E2EEncryption.js
+++ b/modules/e2ee/E2EEncryption.js
@@ -30,6 +30,7 @@ export class E2EEncryption {
         this._conferenceJoined = false;
         this._enabled = false;
         this._initialized = false;
+        this._key = undefined;
 
         this._e2eeCtx = new E2EEContext();
         this._olmAdapter = new OlmAdapter(conference);
@@ -125,12 +126,12 @@ export class E2EEncryption {
         }
 
         // Generate a random key in case we are enabling.
-        const key = enabled ? this._generateKey() : false;
+        this._key = enabled ? this._generateKey() : false;
 
         // Send it to others using the E2EE olm channel.
-        this._olmAdapter.updateKey(key).then(index => {
+        this._olmAdapter.updateKey(this._key).then(index => {
             // Set our key so we begin encrypting.
-            this._e2eeCtx.setKey(this.conference.myUserId(), key, index);
+            this._e2eeCtx.setKey(this.conference.myUserId(), this._key, index);
         });
     }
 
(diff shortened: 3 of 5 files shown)


modules/e2ee/E2EEContext.js
```diff
@@ -148,4 +148,15 @@ export default class E2EEcontext {
             keyIndex
         });
     }
+
+    /**
+     * Ratchet our own key.
+     * @param {string} participantId - the ID of the participant who's key we are setting.
```
saghul, line 154:
```suggestion
     * @param {string} participantId - the ID of the participant who's key we should ratchet.
```

modules/e2ee/E2EEncryption.js
```diff
@@ -218,15 +218,16 @@ export class E2EEncryption {
...
      *
      * @private
      */
     async _ratchetKeyImpl() {
         logger.debug('Ratchetting key');
 
-        return this._rotateKey();
+        this._e2eeCtx.ratchet(this.conference.myUserId());
+
+        // TODO: how do we tell the olm adapter which might need to send the current ratchet key
```
saghul, line 229: We can make the ratchet function resolve a promise with the ratchetted key. Then feed it to the olm adapter, without broadcasting it.

modules/e2ee/Worker.js
```diff
@@ -113,12 +108,30 @@ async function deriveKeys(keyBytes) {
     }, material, 128);
 
     return {
+        material,
```
saghul, line 111: Unneeded, see below.

modules/e2ee/Worker.js
```diff
@@ -113,12 +108,30 @@ async function deriveKeys(keyBytes) {
...
+/**
+ * Ratchets a key. See
+ * https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.5.1
+ * @param {CryptoKey} material - base key material
+ * @returns {CrpytoKey} - ratcheted key material
+ */
+async function ratchet(material) {
+    const textEncoder = new TextEncoder();
+
+    return crypto.subtle.deriveBits({
```
saghul, line 127: We can do this in deriveKeys too.

modules/e2ee/Worker.js
```diff
@@ -149,16 +162,30 @@ class Context {
      * @param {CryptoKey} key
      * @param {Number} keyIndex
      */
-    async setKey(key, keyIndex) {
+    async setKey(keyBytes, keyIndex) {
```
saghul, line 165: IF we change this to be bytes, then the docstring needs updating.

modules/e2ee/Worker.js
```diff
@@ -149,16 +162,30 @@ class Context {
...
         } else {
             this._cryptoKeyRing[this._currentKeyIndex] = false;
         }
         this._sendCount = 0n; // Reset the send count (bigint).
     }
 
+    /**
+     * Ratchets a key forward one step.
+     */
+    async ratchet() {
```
saghul, line 182: If we get the ratchet key early, when we derive all keys, this would just be:

```
const keys = this._cryptoKeyRing[this._currentKeyIndex];

this.setKey(keys.ratchetMaterial, this._currentKeyIndex);
```

The benefits of this may become more obvious if we factor all key shenanigans to some helper class.

modules/e2ee/Worker.js
```diff
@@ -418,6 +445,19 @@ onmessage = async event => {
         } else {
             context.setKey(false, keyIndex);
         }
+    } else if (operation === 'ratchet') {
+        const { participantId } = event.data;
+
+        // TODO: can we ensure this is for our own sender key?
```
saghul, line 451: We could bring back the initialize method with the local participant ID.

saghul (commented):
Left some comments!

modules/e2ee/Worker.js
fippo: long term we'll do the sender/receiver context so its ok to leave as a TODO

fippo:
Updated. My current test plan for this is:
1/ join with two clients, enable e2ee
2/ join with a third client
3/ enable e2ee on the third client

I am a bit confused why the third client can ratchet through, it seems that it gets the original keys from the first and second clients? @saghul can you check please?

modules/e2ee/Worker.js
```diff
@@ -72,16 +72,11 @@ const digestLength = {
 
 /**
  * Derives a set of keys from the master key.
- * @param {Uint8Array} keyBytes - Value to derive key from
- * @param {Uint8Array} salt - Salt used in key derivation
+ * @param {CryptoKey} material - Bytes to derive key from
```
fippo, line 75: this should probably have a different name?

modules/e2ee/E2EEncryption.js
fippo: for now we can ratchet here and then tell olm? The alternative is that we do away with the ratchet message to the worker, call setKey with our ratcheted bytes. Less operations + codepaths

modules/e2ee/Worker.js
```diff
@@ -113,12 +112,30 @@ async function deriveKeys(keyBytes) {
...
         authenticationKey,
         saltKey
     };
 }
 
+/**
+ * Ratchets a key. See
+ * https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.5.1
+ * @param {CryptoKey} material - base key material
+ * @returns {CrpytoKey} - ratcheted key material
```
fippo, line 126: that is an arraybuffer and not a CrpytoKey

modules/e2ee/E2EEncryption.js
fippo: we'll go for the "ratchet here, then use setKey" approach

modules/e2ee/Worker.js
fippo: do you still think so?

modules/e2ee/OlmAdapter.js
```diff
@@ -83,6 +83,19 @@ export class OlmAdapter extends Listenable {
         return typeof window.Olm !== 'undefined';
     }
 
+    /**
+     * Updates the current participant key and distributes it to all participants in the conference
+     * by sending a key-info message.
+     *
+     * @param {Uint8Array|boolean} key - The new key.
+     * @retrns {number}
```
fippo, line 91:
```suggestion
     * @returns {number}
```
(why is there no linting rule for such things!)

modules/e2ee/E2EEncryption.js
```diff
@@ -218,15 +219,25 @@ export class E2EEncryption {
...
-     * which is a heavier operation.
      *
      * @private
      */
     async _ratchetKeyImpl() {
         logger.debug('Ratchetting key');
 
-        return this._rotateKey();
+        const material = await crypto.subtle.importKey('raw', this._key, 'HKDF', false, [ 'deriveBits' ]);
+        const newKey = await crypto.subtle.deriveBits({
```
fippo, line 229: can we share this between worker and lib as a module?

modules/e2ee/Worker.js
```diff
@@ -292,16 +325,41 @@ class Context {
...
+
+            for (let distance = 0; distance < ratchetWindow; distance++) {
+                const calculatedTag = await crypto.subtle.sign(authenticationTagOptions,
+                    authenticationKey, encodedFrame.data);
+
+                if (isArrayEqual(new Uint8Array(authTag),
+                        new Uint8Array(calculatedTag.slice(0, digestLength[encodedFrame.type])))) {
+                    valid = true;
+                    if (distance > 0) {
+                        // console.error('Found new authentication key by advancing the ratchet',
```
fippo, line 343: should I remove this?

modules/e2ee/E2EEncryption.js
saghul: Yes, I was thinking we do this in that "Keys" object we've been talking about.

modules/e2ee/Worker.js
saghul: will do

saghul (approved):
🚀

doc/e2ee.md
```diff
@@ -39,6 +39,14 @@ We do not encrypt the first few bytes of the packet that form the VP8 payload
 This allows the decoder to understand the frame a bit more and makes it decode the fun looking garbage we see in the video.
 This also means the SFU does not know (ideally) that the content is end-to-end encrypted and there are no changes in the SFU required at all.
 
+## Key Ratcheting
+Unlike described in
+  https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.5.1
+we attempt to ratchet the key forward when we do not find a valid
+authentication tag. Note that we only update the set of keys when
+we find a valid signature which avoids a denial of service attack with invalid signatures.
```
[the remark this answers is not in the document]
fippo: that is the attack vector which is why I think we should not drop the key until we get an update

modules/e2ee/E2EEncryption.js
fippo: later then :-)

fippo:
we self-reviewed. Don't do this at home :)

fippo:
Ignoring jenkins, its an issue there.

Next steps: buffering frames for keys we do not yet have.

PR #1343, opened 2020-09-30
e2ee: refactor and clean up

fippo:
in particular this allows sharing the ratcheting between the E2EEncryption
module and the Worker.

### DIFF
--- a/modules/e2ee/crypto-utils.js
+++ b/modules/e2ee/crypto-utils.js
@@ -1,0 +1,74 @@
+/**
+ * Derives a set of keys from the master key.
+ * @param {CryptoKey} material - master key to derive from
+ *
+ * See https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.1
+ */
+export async function deriveKeys(material) {
+    const info = new ArrayBuffer();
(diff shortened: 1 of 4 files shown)


modules/e2ee/crypto-utils.js
```diff
@@ -0,0 +1,72 @@
...
+
+/**
+ * Converts a raw key into a WebCrypto key object with default options
+ * suitable for our usage.
+ * @param {ArrayBuffer} keyBytes - raw key
+ * @param {Array} keyUsages - key usages, see importKey documentation
+ * @returns {CryptoKey} - the WebCrypto key.
+ */
+export async function importKey(keyBytes, keyUsages) {
+    return crypto.subtle.importKey('raw', keyBytes, 'HKDF', false, keyUsages);
```
fippo, line 71: we have two key usages:
2x deriveBits, deriveKey
1x deriveBits only
Should we just allow (but don't use) deriveKey) in the main app as well?

saghul: yeah

saghul (approved):

PR #1344, opened 2020-10-01
e2ee: move context to separate file

fippo:
to allow writing tests. Based on top of #1343 but different PR since this is a move-only operation


modules/e2ee/Context.js
```diff
@@ -0,0 +1,325 @@
...
+     * decryption.
+     * @param {Uint8Array|false} key bytes. Pass false to disable.
+     * @param {Number} keyIndex
+     */
+    async setKey(keyBytes, keyIndex) {
+        let newKey;
+
+        if (keyBytes) {
+            // https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey
+            const material = await crypto.subtle.importKey('raw', keyBytes,
```
saghul, line 76: use importKey?

fippo:
I was using an outdated version for moving, thank you!

PR #1345, opened 2020-10-01
e2ee: add support for long key ids

fippo:
similar to the sframe draft but once again with our jframe twist

### DIFF
--- a/doc/e2ee.md
+++ b/doc/e2ee.md
@@ -32,6 +32,15 @@ At a high level the encrypted frame format looks like this:
    +----+Encrypted Portion            Authenticated Portion+---+
 ```
 
+For keys > 7 the X bit will be set and the trailer will look as follows:
+```
+                                                         0 1 2 3 4 5 6 7
++---------------------------+---------------------------+-+-+-+-+-+-+-+-+
+|  CTR... (length=LEN + 1)  |  KID... (length=KLEN + 1) |S|LEN  |1|KLEN |
++---------------------------+---------------------------+-+-+-+-+-+-+-+-+
+```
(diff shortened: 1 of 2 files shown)


doc/e2ee.md
```diff
@@ -32,6 +32,15 @@ At a high level the encrypted frame format looks like this:
    +----+Encrypted Portion            Authenticated Portion+---+
 ```
 
+For keys > 7 the X bit will be set and the trailer will look as follows:
```
saghul, line 35: Which one is the X bit?

modules/e2ee/Context.js
```diff
@@ -209,22 +233,40 @@ export class Context {
...
+            // Extended key index. It is located right before our trailer byte.
+            keyLength = 1 + (data[encodedFrame.data.byteLength - 1] & 0x7);
+
+            // This is an array buffer. We need to convert it back into an (big) integer.
+            const partialBuffer = encodedFrame.data.slice(encodedFrame.data.byteLength - 1 - keyLength,
+                encodedFrame.data.byteLength - 1);
+            const keyBuffer = new Uint8Array(8);
+
+            keyBuffer.set(new Uint8Array(partialBuffer), 8 - keyLength);
+            keyIndex = (new DataView(keyBuffer.buffer)).getBigUint64(0);
```
saghul, line 250: Doesn't this create a problem, because it's BigInt ?

saghul (commented):
I'll be honest: not a fan of the complexity this adds. We can just limit the key ID space.

fippo:
status: still discussing whether we should do this

PR #1346, opened 2020-10-01
e2ee: add unit tests

fippo:
Both for just encode function as well as an end-to-end test.

### DIFF
--- a/modules/e2ee/Context.spec.js
+++ b/modules/e2ee/Context.spec.js
@@ -1,0 +1,174 @@
+import { Context } from './Context';
+import { ratchet, importKey } from './crypto-utils';
+
+/*
+function hexdump(buffer) {
+    const a = new Uint8Array(buffer);
+    let s = '';
+
+    for (let i = 0; i < a.byteLength; i++) {
+        s += '0x';
+        s += a[i].toString(16);
+        s += ' ';
+    }
+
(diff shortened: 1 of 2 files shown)


saghul:
❤️

fippo:
```
   SyntaxError: /home/runner/work/lib-jitsi-meet/lib-jitsi-meet/modules/e2ee/Context.js: Identifier directly after number (60:27)
    58 |         // Must be incremented on every frame that is sent, can be reset on
    59 |         // key changes.
  > 60 |         this._sendCount = 0n;
```
That is an odd failure. This line isn't even in the change?

fippo:
https://github.com/bcoin-org/bcrypto/issues/14#issuecomment-484199450 -- lets see...

modules/e2ee/Context.spec.js
```diff
@@ -0,0 +1,174 @@
...
+    return s.trim();
+}
+*/
+
+/* TODO: more tests
+ * - delta frames
+ * - frame header is not encrypted
+ * - different sendCounts
+ * - different key length
+ * - ratcheting in decodeFunction
```
fippo, line 24: there is a test for this but there should be a test for "it errors when running out of window" :-)

saghul (approved):

PR #1347, opened 2020-10-01
fix: disable google conference flag for screenshare

jallamsetty1:


damencho (approved):

fippo:
I still have questions about why that happens when @orphis CL says it shouldn't:
  https://webrtc-review.googlesource.com/c/src/+/179081

fippo (approved):

jallamsetty1:
> I still have questions about why that happens when @Orphis CL says it shouldn't:
> https://webrtc-review.googlesource.com/c/src/+/179081

So if the conference flag is set on RD but only 1 ssrc is signaled, chrome shouldn't be sending T0 and T1 layers ?

PR #1350, opened 2020-10-02
feat(xmpp) drop the useStunTurn option

saghul:
Always attempt to discover the configured STUN/TURN servers.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -375,23 +375,17 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
                     iceservers[j] = temp;
                 }
 
-                if (options.useStunTurn) {
-                    let filter;
+                let filter;
 
-                    if (options.useTurnUdp) {
-                        filter = s => s.urls.startsWith('turn');
-                    } else {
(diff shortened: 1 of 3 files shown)


modules/xmpp/strophe.jingle.js
```diff
@@ -375,23 +375,17 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
...
+                let filter;
 
-                if (options.p2p && options.p2p.useStunTurn) {
-                    this.p2pIceConfig.iceServers = iceservers;
+                if (options.useTurnUdp) {
+                    filter = s => s.urls.startsWith('turn');
+                } else {
+                    // By default we filter out STUN and TURN/UDP and leave only TURN/TCP.
+                    filter = s => s.urls.startsWith('turn') && (s.urls.indexOf('transport=tcp') >= 0);
                 }
```
damencho, line 385: This changes the behavior where p2p were able to use turn UDP, which does not make sense in jvb connection but make sense for p2p, that's why for the jvb connection we leave only turns TCP.

saghul: No, note how we don't apply the filter there.

damencho: Aaa now I see it 🤦

damencho (approved):

PR #1351, opened 2020-10-02
Pull upstream 20201002

elbow:


elbow:
Apologies Saul I did not mean to send this PR upstream - its changes only relevant for us.

PR #1352, opened 2020-10-02
faet(ScreenObtainer): detect lack of permissions

paweldomas:
Both of the cases where user clicks cancel and when screen recording
permissions were not granted at the OS level are almost the same,
except for a small difference in the error.message.

This was only tested on Chrome 85.0.4183.121/Mac OS 10.15.6


saghul (approved):

gpolitis (approved):

fippo:
@paweldomas can you please file an issue against https://github.com/w3c/mediacapture-screen-share to add an error code so one doesn't have to rely on string matching?

PR #1357, opened 2020-10-05
e2ee: rename constants to upper-case

fippo:
and add a few new ones.


modules/e2ee/Context.js
```diff
@@ -7,7 +7,7 @@ import { isArrayEqual } from './utils';
 // We use a ringbuffer of keys so we can change them and still decode packets that were
 // encrypted with an old key. We use a size of 16 which corresponds to the four bits
 // in the frame trailer.
-const keyRingSize = 16;
+const KEYRINGSIZE = 16;
```
saghul, line 10: KEYRING_SIZE

saghul (approved):

PR #1362, opened 2020-10-07
feat: revoke owner from a participant

zWaR:
Proposing a MR with a new feature of revoking ownership from a participant. The added method is a counterpart of `grantOwner` and is helpful when the affiliation is being changed while participants are already in the conference.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1438,6 +1438,19 @@ JitsiConference.prototype.grantOwner = function(id) {
 };
 
 /**
+ * Revoke owner rights from a participant.
+ * @param {string} id id of the participant to revoke owner rights from.
+ */
+JitsiConference.prototype.revokeOwner = function(id) {
+    const participant = this.getParticipantById(id);
+
+    if (!participant) {
+        return;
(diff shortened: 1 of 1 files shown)


zWaR:
CLA Signed

zWaR:
Hey everyone,

I was thinking about this PR and perhaps it'd be good to give you a bit of a context as of why we're committing this. In our case we developed our own Prosody plugin to handle authorization of the conference participants. In our UI we allow elevated users to set affiliation for other participants . Granting ownership is straight forward, since `lib-jitsi-meet` features `JitsiConference.grantOwner()` method which we use. Revoking ownership however is not as straight forward and required us to implement logic on the Prosody plugin to accomplish it. With a `JitsiConference.revokeOwner` method we'd not need to handle ownership revocation in the Prosody plugin anymore, but instead simply use a call via `lib-jitsi-meet` just like we do for granting the ownership. This would simplify the plugin code and make the overall solution more robust.

Hopefully you find value in adding a feature like that to your code base. Of course any comments or questions are welcome.

PR #1364, opened 2020-10-07
fix(TPC): maxBitratesVideo break screenshare

paweldomas:
In Chrome, if 'maxBitratesVideo' options are specified and 'capScreenshareBitrate' is disabled, then simulcast is enabled
for screenshare. Chrome in plan B, does not send screen stream if 'maxBitrate' are set and simulcast is enabled.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2107,13 +2107,14 @@ TraceablePeerConnection.prototype.setMaxBitRate = function() {
     }
 
     const videoType = localVideoTrack.videoType;
+    const planBScreenSharing = browser.usesPlanB() && videoType === VideoType.DESKTOP;
 
     // Apply the maxbitrates on the video track when one of the conditions is met.
     // 1. Max. bitrates for video are specified through videoQuality settings in config.js
     // 2. Track is a desktop track and bitrate is capped using capScreenshareBitrate option in plan-b mode.
     // 3. The client is running in Unified plan mode.
     if (!((this.options.videoQuality && this.options.videoQuality.maxBitratesVideo)
-        || (browser.usesPlanB() && this.options.capScreenshareBitrate && videoType === VideoType.DESKTOP)
+        || (planBScreenSharing && this.options.capScreenshareBitrate)
         || browser.usesUnifiedPlan())) {
         return Promise.resolve();
     }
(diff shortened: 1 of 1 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2134,14 +2135,24 @@ TraceablePeerConnection.prototype.setMaxBitRate = function() {
...
+                    // FIXME the top 'isSimulcastOn' condition is confusing for screensharing, because
+                    // if capScreenshareBitrate option is enabled then the simulcast is turned off
+                    bitrate = this.options.capScreenshareBitrate
+                        ? presenterEnabled ? this.videoBitrates.high : DESKSTOP_SHARE_RATE
+
+                        // Remove the bitrate config if not capScreenshareBitrate:
+                        // When switching from camera to desktop and videoQuality.maxBitratesVideo were set,
+                        // then the 'maxBitrate' setting must be cleared, because if simulcast is enabled for screen
+                        // and maxBitrates are set then Chrome will not send the screen stream (plan B).
+                        : undefined;
```
jallamsetty1, line 2152: The reason you think it breaks screensharing is because with the old logic, using capScreenshareBitrate=0 and specifying videoBitrates what happens is 
on encoding 0 - max. bitrate corresponding to LD (200 Kbps) gets applied.
on encoding 1 - max. bitrate corresponding to SD (700 Kbps) gets applied.
Since our SDP munging based simulcast produces only 2 encodings for screenshare, 0 with low fps and bitrate and 1 with high fps and high bitrate, the above max bitrates would not be enough to get the key frame for a desktop track out (esp if you are sharing the full screen) which makes it look like its broken.
Instead, can you apply SD and HD bitrates for encodings 0 and 1 and see if its fixes the screenshare ?

paweldomas: It's not what I'm seeing. There are always 3 SSRCs produced regardless camera vs screenshare, but maybe I need to double check that. Now Chrome doesn't send screen if there are any bitrates set and there are multiple SSRCs for simulcast present. With `capScreenshareBitrate=0` and `maxBitrates` in config.js what happens is there are 3 maxBitrates values set which are the leftover from the camera. This is what I remember, but I will double check and try using the values you suggested. Thank you!

jallamsetty1: Yes, there are 3 SSRC's because we munge SDP and add a total of 3 SSRCs for simulcast (same for camera and desktop tracks), but the bridge was receiving only 2 streams if I recall correctly. We don't know which encodings are the active ones looking at client stats since none of them are showing any valid values right now in chrome://webrtc-internals for simulcast case so its hard to tell. Yes, the same bitrates are applied to both camera and desktop tracks if `maxBitratesVideo` are set in config.js.

fippo: This seems a bit similar to https://bugs.chromium.org/p/chromium/issues/detail?id=947495 -- chrome is behaving a bit weird at times

jallamsetty1: @fippo there are so many broken cases like this, it will be great if we can get some clarity from the chrome team about how simulcast/non-simulcast screenshare is supposed to behave.

paweldomas: It doesn't matter what bitrates I set it just doesn't send anything, as long as any values are set. The sender reports 0 resolution and nothing sent:

Screen Shot 2020-10-08 at 7 09 47 PM

Tried this:

Screen Shot 2020-10-08 at 7 08 54 PM

and this:

Screen Shot 2020-10-08 at 7 10 38 PM

no idea.

Clearing `maxBitrate` to `undefined` on every encoding works.

jallamsetty1: Oh I saw the same thing with the webrtc-internals stats but the media was being sent and it was getting received on the other end. In any case let’s make it undefined for now as a workaround for the chrome bug.

jallamsetty1 (approved):

PR #1370, opened 2020-10-09
sdp: allow removing opus fec via config

fippo:
since the fec bitrate is subtracted from the target bitrate, FEC might
decrease the quality / MOS score. See
  https://webrtchacks.com/red-improving-audio-quality-with-redundancy/


modules/xmpp/JingleSessionPC.js
```diff
@@ -1701,6 +1703,9 @@ export default class JingleSessionPC extends JingleSession {
         if (this.webrtcIceUdpDisable) {
```
fippo, line 1703: the style here is a bit odd... could just set the value directly?

fippo:
@bgrozev with this we could easily run a test or two with FEC disable to see if that has any impact on the scores

fippo:
From the data i've seen this is not a good idea. At least not without RED.
Closing.

PR #1372, opened 2020-10-15
fix(JingleSession) log initialization error

saghul:
Also return early to avoid weird errors if the initialization step fails.
Everything is broken at that stage pretty much.


saghul:
@jallamsetty1 PTAL. This would have allowed us to catch that nasty bug much earlier!

jallamsetty1 (approved):
Good catch! Yeah doesn't make sense to continue if initialization fails.

PR #1373, opened 2020-10-15
(feat) e2ee: sign and verify frames

fippo:
Signs and verifies frames. For testing purposes this is done on every frame.
This is not the long-term goal and there needs to be a better strategy.

### DIFF
--- a/doc/e2ee.md
+++ b/doc/e2ee.md
@@ -69,6 +69,22 @@ This allows the decoder to understand the frame a bit more and makes it decode t
 video. This also means the SFU does not know (ideally) that the content is end-to-end encrypted and there are no
 changes in the SFU required at all.
 
+If the signature bit is set on the frame trailer, there is an additional fixed-length signature that is located
+between the counter and the trailing bit. The signature is generated as
+  Signature = Sign(signatureKey, Authentication Tag)
+and covers the current frame. Not every frame is signed but there will be periodic
+signatures on all SSRCs and streams. This prevents the impersonation attacks described in
+  https://tools.ietf.org/html/draft-omara-sframe-00#section-4.4
+Unlike SFrame, we do not cover multiple frames.
+
+We currently use ECDSA with curve P-521 as described on
+  https://developer.mozilla.org/en-US/docs/Web/API/EcKeyGenParams
+and sign/verify with SHA-256:
+  https://developer.mozilla.org/en-US/docs/Web/API/EcdsaParams
+This results in a fixed length signature of 132 bytes.
+We plan to make these options negotiable by exchanging them along with the key as a JWK:
+  https://tools.ietf.org/html/rfc7517
+
 ### Using Web Workers
 
 Insertable Streams are transferable and can be sent from the main JavaScript context to a
--- a/doc/e2ee.md
+++ b/doc/e2ee.md
@@ -75,7 +75,7 @@ between the counter and the trailing bit. The signature is generated as
 and covers the current frame. Not every frame is signed but there will be periodic
 signatures on all SSRCs and streams. This prevents the impersonation attacks described in
   https://tools.ietf.org/html/draft-omara-sframe-00#section-4.4
-Unlike SFrame, we do not cover multiple frames.
+We currently sign every frame, despite the overhead this incurs (132 bytes per frame currently).
 
 We currently use ECDSA with curve P-521 as described on
   https://developer.mozilla.org/en-US/docs/Web/API/EcKeyGenParams
--- a/doc/e2ee.md
+++ b/doc/e2ee.md
@@ -70,12 +70,37 @@ video. This also means the SFU does not know (ideally) that the content is end-t
 changes in the SFU required at all.
 
 If the signature bit is set on the frame trailer, there is an additional fixed-length signature that is located
-between the counter and the trailing bit. The signature is generated as
+between the counter and the trailing bit:
+```
+     +------------+------------------------------------------+^+
+     |unencrypted payload header (variable length)           | |
+   +^+------------+------------------------------------------+ |
+   | |                                                       | |
+   | |                                                       | |
+   | |                                                       | |
+   | |                                                       | |
+   | |                  Encrypted Frame                      | |
+   | |                                                       | |
+   | |                                                       | |
+   | |                                                       | |
+   | |                                                       | |
+   +^+-------------------------------------------------------+ +
+   | |                 Authentication Tag                    | |
+   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
+   | |    CTR... (length=LEN + 1)            |  SIGNATURE    | |
+   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
+   | |    SIGNATURE (fixed length)           |1|LEN  |KID    | |
+   | +---------------------------------------+-+-+-+-+-+-+-+-+^|
+   |                                                           |
+   +----+Encrypted Portion            Authenticated Portion+---+
+```
+
+The signature is generated as
   Signature = Sign(signatureKey, Authentication Tag)
 and covers the current frame. Not every frame is signed but there will be periodic
 signatures on all SSRCs and streams. This prevents the impersonation attacks described in
   https://tools.ietf.org/html/draft-omara-sframe-00#section-4.4
-We currently sign every frame, despite the overhead this incurs (132 bytes per frame currently).
+We currently sign every frame, despite the overhead this incurs.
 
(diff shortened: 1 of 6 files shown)


modules/e2ee/Context.js
```diff
@@ -150,14 +170,25 @@ export class Context {
                 }
             }
 
-            const frameTrailer = new Uint8Array(counterLength + 1);
+
+            let signatureLength = 0;
+
+            if (this._signatureKey) {
+                // TODO: the condition is more complex as we periodically want to
+                //   sign every n-th frame of every substream/ssrc.
```
fippo, line 178: thinking of a `this._shouldSignFrame(encodedFrame.getMetadata())`
which looks at the metdadata.synchroniziationSource and does a time based decision

fippo:
So this works great (at least in the tests). What is missing is exposing the verification status to the application.
E.g. when we get a valid signature we say "you look great". But then who is responsible for checking whether we get a new signature periodically, the Context or the application?

modules/e2ee/Context.js
```diff
@@ -98,6 +116,52 @@ export class Context {
     _setKeys(keys) {
         this._cryptoKeyRing[this._currentKeyIndex] = keys;
         this._sendCount = BigInt(0); // eslint-disable-line new-cap
+        this._lastSignatureAt.clear();
```
fippo, line 119: just to check: if we ratchet because someone joins, we use setKey from the client code which resets calls _setKeys and therefore we reset.
This calls for another test :-)

fippo: test added :-)

modules/e2ee/Context.js
```diff
@@ -220,24 +302,45 @@ export class Context {
...
 
-            // Set authentication tag bytes to 0.
-            const zeros = new Uint8Array(DIGEST_LENGTH[encodedFrame.type]);
+            // Verify the long-term signature of the authentication tag.
+            if (signatureLength) {
+                const signature = data.subarray(data.byteLength - signatureLength - 1, data.byteLength - 1);
+                const validSignature = this._signatureKey
+                    ? await crypto.subtle.verify(this._signatureOptions,
+                        this._signatureKey, signature, authTag)
+                    : false;
```
fippo, line 321: since restructuring the code this can no longer happen. If we receive a (key) frame with a signature but have no key yet, should we buffer? Odd timing problem but...

doc/e2ee.md
```diff
@@ -69,6 +69,22 @@ This allows the decoder to understand the frame a bit more and makes it decode t
 video. This also means the SFU does not know (ideally) that the content is end-to-end encrypted and there are no
 changes in the SFU required at all.
 
+If the signature bit is set on the frame trailer, there is an additional fixed-length signature that is located
```
saghul, line 72: Can you edit the ASCII diagram with this?

doc/e2ee.md
```diff
@@ -69,6 +69,22 @@ This allows the decoder to understand the frame a bit more and makes it decode t
...
+between the counter and the trailing bit. The signature is generated as
+  Signature = Sign(signatureKey, Authentication Tag)
+and covers the current frame. Not every frame is signed but there will be periodic
+signatures on all SSRCs and streams. This prevents the impersonation attacks described in
+  https://tools.ietf.org/html/draft-omara-sframe-00#section-4.4
+We currently sign every frame, despite the overhead this incurs (132 bytes per frame currently).
```
saghul, line 78: Where does the (magical) 132 number come from? Is it the length of the P-521 signature?

modules/e2ee/Context.js
```diff
@@ -100,6 +103,34 @@ export class Context {
...
+        this._signatureOptions = options;
+    }
+
+    /**
+     * Decide whether we should sign a frame.
+     * @returns {boolean}
+     * @private
+     */
+    _shouldSignFrame() {
+        if (!this._signatureKey) {
```
saghul, line 127: `return Boolean(this._signatureKey);`

modules/e2ee/Context.js
```diff
@@ -185,10 +224,20 @@ export class Context {
                     + DIGEST_LENGTH[encodedFrame.type]); // append trailer.
 
                 return crypto.subtle.sign(AUTHENTICATIONTAG_OPTIONS, this._cryptoKeyRing[keyIndex].authenticationKey,
-                    new Uint8Array(newData)).then(authTag => {
+                    new Uint8Array(newData)).then(async authTag => {
```
saghul, line 227: Can we perhaps reorganize this a bit so it uses async / await all the way? It's kinda weird to ready with async / await and promises mixed up.

modules/e2ee/E2EEncryption.js
```diff
@@ -115,14 +123,24 @@ export class E2EEncryption {
...
         if (!this._initialized && enabled) {
+            // Generate a frame signing key pair. Per session currently.
+            this._signatureKeyPair = await crypto.subtle.generateKey(SIGNATURE_OPTIONS,
+                true, [ 'sign', 'verify' ]);
+            this._e2eeCtx.setSignatureKey(this.conference.myUserId(), this._signatureKeyPair.privateKey);
+
+            // Serialize the JWK of the signing key. Using JSON, might be easy to xml-ify.
+            const serializedSigningKey = await crypto.subtle.exportKey('jwk', this._signatureKeyPair.publicKey);
+
+            this.conference.setLocalParticipantProperty('e2ee.signatureKey', JSON.stringify(serializedSigningKey));
```
saghul, line 142: Can you add a TODO here about signing this with the Olm account's id key?

saghul (commented):
LGTM with a couple of (minor) comments, PTAL!

modules/e2ee/Context.js
fippo: i want to make this more complicated in the future so would like to keep this.

modules/e2ee/Context.js
fippo: once we're done with the actual code? Until then just keep on piling up things :-)

doc/e2ee.md
fippo: explained a bit below so removed here

doc/e2ee.md
```diff
@@ -69,6 +69,47 @@ This allows the decoder to understand the frame a bit more and makes it decode t
...
+   | |                                                       | |
+   | |                                                       | |
+   | |                                                       | |
+   +^+-------------------------------------------------------+ +
+   | |                 Authentication Tag                    | |
+   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
+   | |    CTR... (length=LEN + 1)            |  SIGNATURE    | |
+   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
+   | |    SIGNATURE (fixed length)           |1|LEN  |KID    | |
+   | +---------------------------------------+-+-+-+-+-+-+-+-+^|
```
saghul, line 93: S-bit missing?

modules/e2ee/E2EEncryption.js
```diff
@@ -244,11 +263,22 @@ export class E2EEncryption {
...
+        case 'e2ee.signatureKey':
+            logger.debug(`Participant ${participant.getId()} updated their signature key: ${newValue}`);
+            if (newValue) {
+                const parsed = JSON.parse(newValue);
+
+                const importedKey = await crypto.subtle.importKey('jwk', parsed, { name: 'ECDSA',
+                    namedCurve: parsed.crv }, true, parsed.key_ops);
+
+                this._e2eeCtx.setSignatureKey(participant.getId(), importedKey);
+            }
```
saghul, line 280: shouldn't we set the key to false in case there is no new one?

saghul (dismissed):
2 minor comments, LGTM otherwise.

doc/e2ee.md
fippo: no, set to 1 before LEN

modules/e2ee/E2EEncryption.js
fippo: do we want to support removing your signature key?

saghul: Great questions. I'd say the answer is no to both. Maybe we should at least log a warning when an empty key is published?

doc/e2ee.md
saghul: Right.

modules/e2ee/E2EEncryption.js
fippo: for rejecting key updates we need worker support for sending back error message so lets do that late

saghul (approved):
LGTM

PR #1375, opened 2020-10-17
docs for setLocalParticipantProperty, PARTICIPANT_PROPERTY_CHANGED added.

s-xync:
After using lib-jitsi-meet fairly well, I realized there is no way for us to set a custom property on the local user other than the display name. So, I went on a search and found on the jitsi community that we have `setLocalParticipantProperty` and `PARTICIPANT_PROPERTY_CHANGED` that are not given in the official documentation at [API.md](https://github.com/jitsi/lib-jitsi-meet/blob/master/doc/API.md). Knowing that these options are there helped me a lot in writing new features and I think there are others like me who might be using lib-jitsi-meet and don't know about these. So, I decided to add them in the docs.

### DIFF
--- a/doc/API.md
+++ b/doc/API.md
@@ -450,6 +450,10 @@ Throws NetworkError or InvalidStateError or Error if the operation fails.
 
 36. `isHidden` - checks if local user has joined as a "hidden" user. This is a specialized role used for integrations.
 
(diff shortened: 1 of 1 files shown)


s-xync:
Signed the CLA.

jallamsetty1 (approved):

PR #1383, opened 2020-10-23
adds get display media

tmoldovan8x8:


modules/RTC/ScreenObtainer.js
```diff
@@ -204,9 +206,29 @@ const ScreenObtainer = {
 
                     return;
                 }
-
```
saghul, line 207: Unrelated?

modules/RTC/RTCUtils.js
```diff
@@ -1147,20 +1147,19 @@ class RTCUtils extends Listenable {
...
-                            desktop
-                        }),
-                        error => {
-                            if (audioVideo) {
-                                this.stopMediaStream(audioVideo);
-                            }
-                            reject(error);
-                        });
-                });
+                return this._newGetDesktopMedia(this._parseDesktopSharingOptions(options))
```
saghul, line 1150: Let's avoid adding the `_new` functions to the "old" ones. I see that `_newGetDesktopMedia` is functionally equivalent to what we were doing before. Why the change?

saghul (changes requested):

saghul (approved):

PR #1387, opened 2020-10-27
e2ee: sign multiple frames

fippo:
signs multiple frames instead of every single frame. All key frames are signed,
so are the first frames with a new ssrc as well as every frame on which we ratcheted
forward

### DIFF
--- a/doc/e2ee.md
+++ b/doc/e2ee.md
@@ -87,6 +87,8 @@ between the counter and the trailing bit:
    +^+-------------------------------------------------------+ +
    | |                 Authentication Tag                    | |
    | +---------------------------------------+-+-+-+-+-+-+-+-+ |
+   | | additional authentication tags        |number of tags | |
+   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
    | |    CTR... (length=LEN + 1)            |  SIGNATURE    | |
    | +---------------------------------------+-+-+-+-+-+-+-+-+ |
    | |    SIGNATURE (fixed length)           |1|LEN  |KID    | |
--- a/doc/e2ee.md
+++ b/doc/e2ee.md
@@ -70,7 +70,7 @@ video. This also means the SFU does not know (ideally) that the content is end-t
 changes in the SFU required at all.
 
 If the signature bit is set on the frame trailer, there is an additional fixed-length signature that is located
-between the counter and the trailing bit:
+before the counter and the trailing byte:
 ```
      +------------+------------------------------------------+^+
      |unencrypted payload header (variable length)           | |
@@ -89,9 +89,9 @@ between the counter and the trailing bit:
    | +---------------------------------------+-+-+-+-+-+-+-+-+ |
    | | additional authentication tags        |number of tags | |
    | +---------------------------------------+-+-+-+-+-+-+-+-+ |
-   | |    CTR... (length=LEN + 1)            |  SIGNATURE    | |
+   | |    SIGNATURE (fixed length)                           | |
    | +---------------------------------------+-+-+-+-+-+-+-+-+ |
-   | |    SIGNATURE (fixed length)           |1|LEN  |KID    | |
+   | |    CTR... (length=LEN + 1))           |1|LEN  |KID    | |
    | +---------------------------------------+-+-+-+-+-+-+-+-+^|
    |                                                           |
    +----+Encrypted Portion            Authenticated Portion+---+
--- a/modules/e2ee/Context.js
+++ b/modules/e2ee/Context.js
@@ -247,6 +247,35 @@ export class Context {
                 newUint8.set(frameTrailer, frameHeader.byteLength + cipherText.byteLength
                     + DIGEST_LENGTH[encodedFrame.type]); // append trailer.
 
+                let signatureData;
+
+                // Set the pending authentication tags before creating the current authentication tag.
+                // This way they are covered by the current authentication tag.
+                if (signatureLength) {
+                    const numberOfPendingTags = this._pendingAuthenticationTags.has(ssrc)
+                        ? this._pendingAuthenticationTags.get(ssrc).length
+                        : 0;
+
+                    signatureData = new Uint8Array(
+                        (numberOfPendingTags + 1) * DIGEST_LENGTH[encodedFrame.type]);
+
+                    let offset = DIGEST_LENGTH[encodedFrame.type];
+
+                    for (const pendingAuthTag of this._pendingAuthenticationTags.get(ssrc) || []) {
+                        signatureData.set(pendingAuthTag, offset);
+                        offset += pendingAuthTag.byteLength;
+                    }
+
+                    this._pendingAuthenticationTags.set(ssrc, []);
+
+                    // The truncated authentication tag is still zeros at this point.
+                    newUint8.set(signatureData, UNENCRYPTED_BYTES[encodedFrame.type] + cipherText.byteLength);
+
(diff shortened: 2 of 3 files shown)


modules/e2ee/Context.js
```diff
@@ -282,9 +329,12 @@ export class Context {
             // Verify the long-term signature of the authentication tag.
             if (signatureLength) {
```
fippo, line 330: if we don't get a signature we need to check whether we should have received one or set the status to "not trustworthy"

fippo: also if we get a valid signature we can now determine a ratio of valid past frames. If that is too low...

fippo:
this currently leads to this:
```
     +------------+------------------------------------------+^+
     |unencrypted payload header (variable length)           | |
   +^+------------+------------------------------------------+ |
   | |                                                       | |
   | |                                                       | |
   | |                                                       | |
   | |                                                       | |
   | |                  Encrypted Frame                      | |
   | |                                                       | |
   | |                                                       | |
   | |                                                       | |
   | |                                                       | |
   +^+-------------------------------------------------------+ +
   | |                 Authentication Tag                    | |
   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
   | | additional authentication tags        |number of tags | |
   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
   | |    CTR... (length=LEN + 1)            |  SIGNATURE    | |
   | +---------------------------------------+-+-+-+-+-+-+-+-+ |
   | |    SIGNATURE (fixed length)           |1|LEN  |KID    | |
   | +---------------------------------------+-+-+-+-+-+-+-+-+^|
   |                                                           |
   +----+Encrypted Portion            Authenticated Portion+---+
```
which is confusing because the counter is between signature and the data that is signed. Easier to change the definition such that the trailer is (signature, ctr, last byte)

modules/e2ee/Context.js
```diff
@@ -123,12 +125,22 @@ export class Context {
...
+
+        const ssrc = encodedFrame.getMetadata().synchronizationSource;
 
-        return true;
+        if (!this._frameSignatures.has(ssrc)) {
+            return true;
+        }
+
+        // TODO: variable for audio and video?
+        return this._frameSignatures.get(ssrc).length >= 2;
```
fippo, line 143: TODO: decide on what is good here and make sure tests are in sync

doc/e2ee.md
```diff
@@ -87,6 +87,8 @@ between the counter and the trailing bit:
    +^+-------------------------------------------------------+ +
    | |                 Authentication Tag                    | |
    | +---------------------------------------+-+-+-+-+-+-+-+-+ |
+   | | additional authentication tags        |number of tags | |
```
fippo, line 90: This is the number of additional tags, i.e. not including the authentication tag

modules/e2ee/Context.js
```diff
@@ -123,12 +125,22 @@ export class Context {
...
+
+        const ssrc = encodedFrame.getMetadata().synchronizationSource;
 
-        return true;
+        if (!this._frameSignatures.has(ssrc)) {
+            return true;
+        }
+
+        // TODO: variable for audio and video?
+        return this._frameSignatures.get(ssrc).length > 30;
```
fippo, line 143: discuss!

modules/e2ee/Context.js
```diff
@@ -181,14 +194,20 @@ export class Context {
...
+                    + ((this._frameSignatures.get(ssrc) || []).length * DIGEST_LENGTH[encodedFrame.type])
+                : 0;
+
             const frameTrailer = new Uint8Array(counterLength + signatureLength + 1);
 
-            frameTrailer.set(new Uint8Array(counter.buffer, counter.byteLength - counterLength));
+            frameTrailer.set(new Uint8Array(counter.buffer, counter.byteLength - counterLength),
+                frameTrailer.byteLength - (1
+                    + (signatureLength ? this._signatureOptions.byteLength : 0)
+                    + counterLength));
```
fippo, line 210: the reason this is so complicated is that we do
data, signature, additional signatures, counter, signature, trailing byte
If we do this instead
data, signature, additional signatures, signature, counter, trailing byte
it gets easier

modules/e2ee/Context.js
```diff
@@ -227,16 +246,43 @@ export class Context {
...
+                        this._frameSignatures.set(ssrc, []);
+
                         const signature = await crypto.subtle.sign(this._signatureOptions,
-                            this._signatureKey, truncatedAuthTag);
+                            this._signatureKey, signatureData);
 
                         newUint8.set(new Uint8Array(signature), newUint8.byteLength - signature.byteLength - 1);
+
+                        // This count excludes the new authentication tag (which is always there)
+                        newUint8[newUint8.byteLength - 1 - this._signatureOptions.byteLength
```
fippo, line 276: i *think* we can just define this including the authentication tag and save some "+ 1" on the receiver.

modules/e2ee/Context.js
```diff
@@ -270,21 +316,41 @@ export class Context {
         if (this._cryptoKeyRing[keyIndex]) {
             const counterLength = 1 + ((data[encodedFrame.data.byteLength - 1] >> 4) & 0x7);
             const signatureLength = data[encodedFrame.data.byteLength - 1] & 0x80
-                ? this._signatureOptions.byteLength : 0;
+                ? this._signatureOptions.byteLength + 1 : 0;
```
fippo, line 319: this is the length of the signature and the one byte for the count. Maybe use a named constant for the field length?

modules/e2ee/Context.spec.js
```diff
@@ -321,21 +325,61 @@ describe('E2EE Context', () => {
...
                     expect(data.byteLength).toEqual(audioBytes.length);
                     expect(Array.from(data)).toEqual(audioBytes);
-                    done();
+                    if (frameCount === 2) {
+                        done();
+                    }
                 }
             };
             await sender.encodeFunction(makeAudioFrame(), sendController);
+            sender._sendCount = 65537n;
```
fippo, line 349: drive-by test improvement

modules/e2ee/Context.js
```diff
@@ -362,9 +440,7 @@ export class Context {
                 counter,
                 length: CTR_LENGTH
             }, this._cryptoKeyRing[keyIndex].encryptionKey, new Uint8Array(encodedFrame.data,
-                    UNENCRYPTED_BYTES[encodedFrame.type],
-                    encodedFrame.data.byteLength - (UNENCRYPTED_BYTES[encodedFrame.type]
-                    + DIGEST_LENGTH[encodedFrame.type] + counterLength + signatureLength + 1))
+                    UNENCRYPTED_BYTES[encodedFrame.type], authTagOffset - UNENCRYPTED_BYTES[encodedFrame.type])
```
fippo, line 443: this is nicer than in the old version. It says
"start at offset, then decrypt until the authentication tag"

modules/e2ee/Context.js
fippo: tests will be in sync but a good default still needs to be found

modules/e2ee/Context.js
fippo: this is not outdated but a good default needs to be found

modules/e2ee/Context.js
fippo: reviewer note: the definition of signatureLength is different in encode and decode function

fippo (commented):
.

fippo:
overtaken by events

PR #1388, opened 2020-10-28
fix(audio-recorder) remove no longer needed code

saghul:


fippo (approved):
never create empty empty media streams, weird things will happen.
AWAY WITH IT

PR #1390, opened 2020-10-30
feat: Skips pinging when there was a recent server response.

damencho:


modules/xmpp/strophe.ping.js
```diff
@@ -96,7 +99,25 @@ export default class PingConnectionPlugin extends ConnectionPlugin {
...
+            // let's skip the ping
+
+            // server response is measured on raw input and ping response time is measured after all the xmpp
+            // processing is done, and when the last server response is a ping there can be slight misalignment of the
+            // times, we give it 100ms for that processing.
+            if (this._getTimeSinceLastServerResponse() + 100 < new Date() - this._lastServerCheck) {
+                // do this just to keep in sync the intervals so we can detect suspended device
+                this._addPingExecutionTimestamp();
+
+                this._lastServerCheck = new Date();
```
paweldomas, line 131: should it also do `this.failedPings = 0;` here?

paweldomas: LGTM otherwise

damencho: I was wondering about this ... I will check. But now I'm trying to debug as tests don't pass because even when using same link it is like nobody is in the room even though participants had joined the muc 🤷‍♂️ 
And there is no console error ...

paweldomas: I think it should, because the purpose of counting failed ping is to know how long the XMPP connection was unresponsive. So if we determine that by checking when last data was received this is equal to XMPP working.

PR #1393, opened 2020-11-02
Added the email to JitsiParticipant:getIdentity

andywangevertz:
I have the same as issue as the request from https://github.com/jitsi/jitsi-meet/issues/5677.

The following is the simple fix/workaround as currently we don't really use the _identity in JitsiParticipant so just re-use it for the email.

### DIFF
--- a/JitsiParticipant.js
+++ b/JitsiParticipant.js
@@ -277,4 +277,13 @@ export default class JitsiParticipant {
     getBotType() {
         return this._botType;
     }
+
+    /**
+     * Returns the identity/email for the participant.
+     *
+     * @returns {string|undefined} - The identity the participant.
+     */
+    getIdentity() {
+        return this._identity;
+    }
 }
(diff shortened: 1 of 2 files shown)


shubhamsharma4587:
for which api it will work like getEmail, getParticipantsInfo etc.? I'm using getEmail but it's not working☹️

andywangevertz:
> for which api it will work like getEmail, getParticipantsInfo etc.? I'm using getEmail but it's not working☹️

Try getIdentity() to see if it works..

shubhamsharma4587:
I tried api.getIdentity(participant_id) but its not working.

andywangevertz:
This fix is currently in lib-jitsi-meet level only and it will need some work/hook-up on the api level ( I would assume you are using external api) for api.getIdentity or api.getEmail to work.

shubhamsharma4587:
Could you please provide API level changes as well because its tricky to identify where we need to do change

PR #1406, opened 2020-11-18
fix(gum) update permissions prompt detection

quitrk:
Fire PERMISSION_PROMPT_IS_SHOWN when none of the devices have a label
Fire a new SLOW_GET_USER_MEDIA event if the timeout for getUserMedia is exceeded

Update JitsiMeetJS.createLocalTracks to include the options for firing the above events
in the provided options argument. Deprecate the firePermissionPromptIsShownEvent flag in
method's signature


saghul (dismissed):

paweldomas (dismissed):
LGTM I'm wondering if `arePermissionsGrantedForAvailableDevices` should be differentiated between audio and video devices, but this can be improved as next step eventually. Please check with @saghul before merging

saghul (approved):

PR #1417, opened 2020-11-30
ref(JitsiConference): allow only 1 track per media type

paweldomas:
Only one audio or video track is correctly supported by lib-jitsi-meet. This is an attempt to mitigate an issue reported by the community where the user's audio is played in the conference even though the UI shows the user as muted.

Can be reproduced by executing this script in the command line:

`JitsiMeetJS.createLocalTracks({ devices: ['audio']}).then(track => { APP.conference._room.addTrack(track[0])});`

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -942,27 +942,23 @@ JitsiConference.prototype.getTranscriptionStatus = function() {
 
 /**
  * Adds JitsiLocalTrack object to the conference.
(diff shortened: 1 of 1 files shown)


saghul (approved):

paweldomas:
oh wait there's bug there, it used to quickly resolve when it's the same track, but now it will go through

PR #1419, opened 2020-11-30
feat: Function to check is remote track playing/received.

damencho:
Used for some tests in torture.


damencho:
Will be handled in jitsi-meet.

PR #1422, opened 2020-12-01
Remove startBitrate and minBitrate.

bgrozev:


saghul (approved):

daxiondi:
I always use startbitrate and minbitrate. He is very good. What is the reason for deleting it

bgrozev:
> I always use startbitrate and minbitrate. He is very good. What is the reason for deleting it

It has been broken for a long time. How and why do you use it? Would you be able to provide a patch to do it purely client side?

saghul (approved):

bgrozev:
I screwed up while testing, this still works.

PR #1423, opened 2020-12-01
Remove conference-wide RTX/REMB/TCC/opus-red options.

bgrozev:


saghul (dismissed):

saghul (approved):

JonathanLennox:
Does the addition of this new client signaling mean that there'll be a version incompatibility between older jitsi-meet and newer jicofo?  Maybe that's fine, but we should be careful when updating.

bgrozev:
> Does the addition of this new client signaling mean that there'll be a version incompatibility between older jitsi-meet and newer jicofo? Maybe that's fine, but we should be careful when updating.

Yes, good point. Old jitsi-meet and new jicofo will lead to TCC/REMB not being enabled, which is pretty bad. I suppose we should delay merging jitsi/jicofo#643

bgrozev:
The defaults for REMB, TCC and RTX are "enabled", so new jitsi-meet and old jicofo is fine (the client lacks the ability to disable REMB and TCC, which it never did reliably anyway)

JonathanLennox:
> The defaults for REMB, TCC and RTX are "enabled", so new jitsi-meet and old jicofo is fine (the client lacks the ability to disable REMB and TCC, which it never did reliably anyway)

As long as the behavior on Firefox is correct - until recently we disabled RTX on Firefox, didn't we?

jallamsetty1:
> > The defaults for REMB, TCC and RTX are "enabled", so new jitsi-meet and old jicofo is fine (the client lacks the ability to disable REMB and TCC, which it never did reliably anyway)
> 
> As long as the behavior on Firefox is correct - until recently we disabled RTX on Firefox, didn't we?

We explicitly disable RTX for FF. I am creating a PR to do a version check as well because FF fixed the issue that we were seeing with RTX and simulcast in FF84.

bgrozev:
Yes, [we still disable it](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/xmpp.js#L153). The "feature" code for RTX was already there, I only added TCC and REMB

PR #1432, opened 2020-12-09
fix(XMPP): Get the correct domain for XMPP ping.

jallamsetty1:


damencho (approved):

damencho:
Thanks!

damencho:
The desktop sharing test failure is unrelated. We need to update that test.

PR #1434, opened 2020-12-09
ref(JitsiConference): check 1 track limit in replaceTrack

paweldomas:


paweldomas:
wait this PR is wrong, it will go through if it's the same track but it should instantly resolve

PR #1436, opened 2020-12-10
fix: Implement the encodings workaround only on Safari.

jallamsetty1:
Explicitly check if all the encodings report the same scaleResolutionDownBy value before trying to ensure they match the expected values. This makes Chrome VP9 work again.
Chrome VP9 was broken since we were trying to set a scaleResolutionDownBy of '4' on the first encoding which works for VP8 but VP9 SVC works differently, the browser dynamically adjusts the RTP stream associated with the first ssrc based on the encodings configuration so the first encoding should never be scaled down.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -3,7 +3,6 @@ import transform from 'sdp-transform';
 
 import * as MediaType from '../../service/RTC/MediaType';
 import RTCEvents from '../../service/RTC/RTCEvents';
-import VideoType from '../../service/RTC/VideoType';
 import browser from '../browser';
 
 const logger = getLogger(__filename);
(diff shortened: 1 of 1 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -447,20 +446,15 @@ export class TPCUtils {
...
-
-        // Ignore desktop and non-simulcast tracks.
-        if (!(parameters
-            && parameters.encodings
-            && Array.isArray(parameters.encodings)
-            && this.pc.isSimulcastOn()
-            && localVideoTrack
-            && localVideoTrack.videoType !== VideoType.DESKTOP)) {
-            return;
+        const allEqualEncodings
```
saghul, line 449: Maybe compute this inside an if isSafari block? Since it's only used there...

saghul (dismissed):
LGTM with a suggestion.

saghul (approved):

PR #1439, opened 2020-12-11
Changerquest/L33T-2202

EnricoSchw:
In Jitsi the Callstats PreCallTest ist activate by default. I add an config param to make it possible to deactivate the PreCallTest.


EnricoSchw:
Sorry, wrong Project :-)

PR #1442, opened 2020-12-16
Fix of case when findLine returns false

GregOriol:
Hi,

I stumbled upon a case where `findLine` at this position in the code returns `false` instead of a string (I'm not entirely sure what the case is about regarding the actual feature though, I arrived there through the debugger). The problem is that it makes the .substring(7) fail because the function does not exist on `false`. The proposed change fixes this case.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -806,7 +806,7 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
             mediaLines = remoteSDP.media.filter(mls => {
                 const msid = SDPUtil.findLine(mls, 'a=msid');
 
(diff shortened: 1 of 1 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -806,7 +806,7 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
             mediaLines = remoteSDP.media.filter(mls => {
                 const msid = SDPUtil.findLine(mls, 'a=msid');
 
-                return typeof msid !== 'undefined' && streamId === msid.substring(7).split(' ')[0];
+                return typeof msid !== 'undefined' && msid !== false && streamId === msid.substring(7).split(' ')[0];
```
sapkra, line 809: Wouldn't it make more sense to check if `msid` is a `string`?

PR #1449, opened 2020-12-25
Added noTurnsOnSafari option to bypass a bug on Safari

GregOriol:
This PR is to provide a way to bypass a bug on current versions of Safari (as of 14): the bug is that Safari can't connect to tls "turns" servers that are using a Let's Encrypt certificate (and maybe other CAs). So this change feature/fix adds an option to disable the use of turns on Safari that can be enabled if one uses LE certificates.

The bug itself is tracked here https://b.webkit.org/show_bug.cgi?id=219274 and has been marked as resolved, but not shipped yet. Anyway, when shipped older versions would still benefit from it.


fippo:
but why? The current behaviour is that Safari won't gather candidates from `turns` which isn't really distinguishable from failing to connect to the TURN server. The sad but pragmatic solution to this problem is not to use letsencrypt for turns :-(

sapkra:
The linked webkit bug has been marked as `RESOLVED FIXED`.

PR #1450, opened 2021-01-05
 feat: Skips using disco-info for features.

damencho:


modules/xmpp/Caps.js
```diff
@@ -102,10 +106,17 @@ export default class Caps extends Listenable {
      * @param {String} feature the name of the feature.
      * @param {boolean} submit if true - new presence with updated "c" node
      * will be sent.
+     * @param {boolean} external whether this feature was added externally to the library.
      */
-    addFeature(feature, submit = false) {
+    addFeature(feature, submit = false, external = false) {
         this.disco.addFeature(feature);
```
bgrozev, line 112: So external features are added in both presence and caps?

modules/xmpp/Caps.js
```diff
@@ -133,6 +151,28 @@ export default class Caps extends Listenable {
         this.rooms.forEach(room => room.sendPresence());
     }
 
+    /**
+     * Updates the presences in the rooms based on the current values in externalFeatures.
+     * @param {ChatRoom} room the room.
+     * @private
+     */
+    _updateRoomsWithExternalFeatures(room) {
```
bgrozev, line 159: This updates a single room, the name and doc should reflect it.

modules/xmpp/Caps.js
```diff
@@ -133,6 +151,28 @@ export default class Caps extends Listenable {
...
+            const children = [];
+
+            this.externalFeatures.forEach(f => {
+                children.push({
+                    'tagName': 'feature',
+                    attributes: { 'var': f }
+                });
+            });
+
+            room.addToPresence('features', { children });
```
bgrozev, line 172: Just checking: this will not add multiple `features` XML elements, correct?

modules/xmpp/Caps.js
damencho: Yep.

modules/xmpp/Caps.js
damencho: Yes, just one. with the children.

modules/xmpp/ChatRoom.js
```diff
@@ -515,6 +517,26 @@ export default class ChatRoom extends Listenable {
...
                 break;
+            case 'features': {
+                if (!member.features) {
+                    member.features = new Set();
+                }
+                this._extractFeatures(node, member.features);
+                break;
+            }
+            case 'initiator': {
+                // we recognize jigasi by the initiator extension in the presence and its xmlns
```
bgrozev, line 528: Any reason not to have jigasi use a regular `feature`?

modules/xmpp/ChatRoom.js
```diff
@@ -747,26 +778,35 @@ export default class ChatRoom extends Listenable {
...
+
     /**
      * Initialize some properties when the focus participant is verified.
      * @param from jid of the focus
-     * @param mucJid the jid of the focus in the muc
+     * @param features the features reported in jicofo presence
      */
-    _initFocus(from, mucJid) {
+    _initFocus(from, features) {
         // skip if we have queried jicofo already, it will not change
```
bgrozev, line 803: Now that we don't use disco#info can we remove this check?

damencho: Ah, yeah. It was a merge conflict I got and that had sneaked in.

modules/xmpp/ChatRoom.js
bgrozev: Yes, I think that would be better

JitsiConference.js
```diff
@@ -1561,7 +1565,7 @@ JitsiConference.prototype._updateFeatures = function(participant) {
             participant._supportsDTMF = features.has('urn:xmpp:jingle:dtmf:0');
             this.updateDTMFSupport();
 
-            if (features.has('http://jitsi.org/protocol/jigasi')) {
+            if (features.has(JIGASI_XMLNS)) {
```
bgrozev, line 1568: This is no longer used as a XMLNS, perhaps naming it `JIGASI_FEATURE` is more appropriate

bgrozev (approved):

JitsiConferenceEventManager.js
```diff
@@ -596,23 +596,17 @@ JitsiConferenceEventManager.prototype.removeXMPPListeners = function() {
...
-            Strophe.getResourceFromJid(from));
+    const featuresChangedListener = (from, features) => {
+        const participant = conference.getParticipantById(Strophe.getResourceFromJid(from));
 
         if (participant) {
-            conference.eventEmitter.emit(
-                JitsiConferenceEvents.PARTCIPANT_FEATURES_CHANGED,
-                participant);
+            participant._features = features;
+            conference.eventEmitter.emit(JitsiConferenceEvents.PARTCIPANT_FEATURES_CHANGED, participant);
```
saghul, line 604: maybe we could also add a parameter with the previous features? Otherwise calculating the difference is impossible since the previous ones are lost, right?

modules/xmpp/Caps.js
```diff
@@ -79,6 +79,10 @@ export default class Caps extends Listenable {
         this.version = '';
         this.rooms = new Set();
 
+        // We keep track of features added outside the library and we publish them
+        // in the presence of the participant for simplicity, avoiding the disco info request-response
```
saghul, line 83: End sentence in a period.

modules/xmpp/Caps.js
```diff
@@ -102,10 +106,17 @@ export default class Caps extends Listenable {
      * @param {String} feature the name of the feature.
      * @param {boolean} submit if true - new presence with updated "c" node
      * will be sent.
+     * @param {boolean} external whether this feature was added externally to the library.
```
saghul, line 109: Can you elaborate on what makes a feature "external"?

modules/xmpp/Caps.js
saghul: I had the same comment. I'm confused by why we say rooms when we really mean one.

saghul (commented):
Left some comments.

JitsiConferenceEventManager.js
damencho: This is fixed.

modules/xmpp/Caps.js
damencho: External to the library ... Like remote control, jibri ... These are two we set from jitsi-meet. I will try updating the js doc.

JitsiConferenceEventManager.js
damencho: This code has not changed from my PR. We can address it  later if needed.

modules/xmpp/Caps.js
damencho: 👍

JitsiConferenceEventManager.js
saghul: Remember the 11th commandment: Thou shalt leave the code better than it was. :-P

modules/xmpp/Caps.js
saghul: 👍

modules/xmpp/xmpp.js
```diff
@@ -192,7 +198,7 @@ export default class XMPP extends Listenable {
         }
 
         if (E2EEncryption.isSupported(this.options)) {
-            this.caps.addFeature('https://jitsi.org/meet/e2ee');
+            this.caps.addFeature('https://jitsi.org/meet/e2ee', false, true);
```
saghul, line 201: How about we add a constant for this too?

saghul (approved):

saghul (approved):

PR #1451, opened 2021-01-06
capabilities: change capability check style

fippo:
mostly making it shorter. No functional changes.

### DIFF
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -165,7 +165,7 @@ export default class BrowserCapabilities extends BrowserDetection {
      */
     supportsReceiverStats() {
         return window.RTCRtpReceiver
-            && 'getSynchroniziationSourceS' in window.RTCRtpReceiver.prototype;
+            && 'getSynchroniziationSources' in window.RTCRtpReceiver.prototype;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


modules/browser/BrowserCapabilities.js
```diff
@@ -155,16 +156,16 @@ export default class BrowserCapabilities extends BrowserDetection {
...
     }
 
     /**
      * Checks if the current browser supports audio level stats on the receivers.
      */
     supportsReceiverStats() {
-        return typeof window.RTCRtpReceiver !== 'undefined'
-            && Object.keys(RTCRtpReceiver.prototype).indexOf('getSynchronizationSources') > -1;
+        return window.RTCRtpReceiver
+            && 'getSynchroniziationSourceS' in window.RTCRtpReceiver.prototype;
```
saghul, line 168:
```suggestion
            && 'getSynchroniziationSources' in window.RTCRtpReceiver.prototype;
```

saghul (commented):
Found a typo, LGTM otherwise!

modules/browser/BrowserCapabilities.js
fippo: so... why didn't this blow up?

saghul: I don't think we have coverage for that alas.

PR #1453, opened 2021-01-06
Add replace track to doc

aneeshikmat:
I see we have an option to use replace track in conference level which its missing in this API. thanks


aneeshikmat:
> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

Its created...

jallamsetty1 (approved):

PR #1457, opened 2021-01-07
feat: Add codec selection mechanism.

jallamsetty1:
Determine the preferred codec for a given endpoint based on the config.js settings and the codecs supported by the endpoint.
The preferred codec is published in presence and then used by the other endpoints in the call during join/leave to determine
if the codec needs to be changed on the fly. Different codecs can be configured for p2p/jvb connections.
The preferredCodec/disabledCodec settings under videoQuality will have precedence over the older settins like preferH264/disableH264.

### DIFF
--- a/modules/RTC/CodecSelection.js
+++ b/modules/RTC/CodecSelection.js
@@ -1,0 +1,190 @@
+
+import { getLogger } from 'jitsi-meet-logger';
+
+import * as JitsiConferenceEvents from '../../JitsiConferenceEvents';
+import CodecMimeType from '../../service/RTC/CodecMimeType';
+import * as MediaType from '../../service/RTC/MediaType';
+import browser from '../browser';
+
+const logger = getLogger(__filename);
+
+/**
+ * This class handles the codec selection mechanism for the conference based on the config.js settings.
+ * The preferred codec is selected based on the settings and the list of codecs supported by the browser.
+ * The preferred codec is published in presence which is then used by the other endpoints in the
+ * conference to pick a supported codec at join time and when the call transitions between p2p and jvb
+ * connections.
+ */
+export class CodecSelection {
+    /**
+     * Creates a new instance for a given conference.
+     *
+     * @param {JitsiConference} conference the conference instance
+     * @param {*} options
+     * @param {string} options.disabledCodec the codec that needs to be disabled.
+     * @param {boolean} options.enforcePreferredCodec whether codec preference has to be
+     * enforced even when an endpoints that doesn't support the preferred codec joins the call.
+     * Falling back to the standard codec will be skipped when this option is true, endpoints
+     * that do not support the preferred codec may not be able to encode/decode video when this happens.
+     * @param {string} options.jvbCodec the codec that is preferred on jvb connection.
+     * @param {string} options.p2pCodec the codec that is preferred on p2p connection.
+     */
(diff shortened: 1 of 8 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -1808,11 +1816,20 @@ export default class JingleSessionPC extends JingleSession {
...
-            return this._initiatorRenegotiate(remoteDescription);
-        }
+        const promise = this.isInitiator
+            ? this._initiatorRenegotiate(remoteDescription)
+            : this._responderRenegotiate(remoteDescription);
+
+        return promise.then(() => {
+            // Publish the codec info to the other endpoints in the conference if it has changed
+            // as a result of the renegotiation (only on current active session).
+            const codec = this.peerconnection.getConfiguredVideoCodec();
```
paweldomas, line 1826: When participant A supports VP9, but then comes participant B who doesn't, then is it the case that both participants A and B will now be advertising their codec as VP8 after participant A adopts? Is that intentional to never switch back?

paweldomas:
How is switching between p2p and jvb modes handled?

modules/xmpp/JingleSessionPC.js
jallamsetty1: The switch back happens here when participant B leaves or all participants that don't have the preferred codec leave -https://github.com/jallamsetty1/lib-jitsi-meet/blob/codec-preference/modules/RTC/CodecSelection.js#L164.
The idea was to keep the current codec in presence so that we can use that for bitrate adaptation as a next step.

jallamsetty1:
> How is switching between p2p and jvb modes handled?

When ever a new media session is created, the codec is set on the connection based on the connection type and presence updated automatically if it is the active session. https://github.com/jallamsetty1/lib-jitsi-meet/blob/codec-preference/modules/RTC/CodecSelection.js#L113

paweldomas:
> When ever a new media session is created, the codec is set on the connection based on the connection type and presence updated automatically if it is the active session. https://github.com/jallamsetty1/lib-jitsi-meet/blob/codec-preference/modules/RTC/CodecSelection.js#L113

But if P2P session prefers VP9 and there's a switch to the JVB session which supports VP8, will it the presence be updated on the switch to advertise VP8 as preferred?

modules/RTC/CodecSelection.js
```diff
@@ -0,0 +1,207 @@
...
+            : null;
+        let codec = preferredCodec;
+
+        // For a new endpoint joining the call, JitsiConferenceEvents.USER_JOINED event is received before the
+        // media session is created, the supported codecs for all the remote endpoints in the call need to be
+        // compared here before setting the codec on the peerconnection.
+        if (!mediaSession.isP2P) {
+            const remoteParticipants = this.conference.getParticipants().map(participant => participant.getId());
+
+            for (const remote of remoteParticipants) {
```
paweldomas, line 126: instead of maintaining a list of non preferred participants I would probably create a "selector" which selects the codec by going over all remote participants every time any of the input changes. I think it's less error prone to check the state rather than trying to track how things are changing, but it's not a big deal

jallamsetty1: refactored the codec selection logic as per your suggestion.

paweldomas (approved):

PR #1458, opened 2021-01-08
Support the urn:xmpp:extdisco:2 namespace specified by XEP-0215

jbg:
XEP-0215 specifies the use of the namespace `urn:xmpp:extdisco:2` for discovery of external services, but Jitsi only looks them up using the older `urn:xmpp:extdisco:1`. One some XMPP servers this means discovery fails entirely, and with Prosody's `mod_turncredentials` (rather than Jitsi's patched one) this means that some information (e.g. transport) is not provided in the response.

This patch adds support for `urn:xmpp:extdisco:2` whilst still supporting `urn:xmpp:extdisco:1`.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -306,92 +306,104 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
         this.connection.sendIQ(
             $iq({ type: 'get',
                 to: this.xmpp.options.hosts.domain })
-                .c('services', { xmlns: 'urn:xmpp:extdisco:1' }),
-            res => {
-                const iceservers = [];
-
-                $(res).find('>services>service').each((idx, el) => {
-                    // eslint-disable-next-line no-param-reassign
-                    el = $(el);
-                    const dict = {};
-                    const type = el.attr('type');
-
-                    switch (type) {
-                    case 'stun':
-                        dict.urls = `stun:${el.attr('host')}`;
-                        if (el.attr('port')) {
-                            dict.urls += `:${el.attr('port')}`;
-                        }
-                        iceservers.push(dict);
-                        break;
-                    case 'turn':
-                    case 'turns': {
-                        dict.urls = `${type}:`;
-                        const username = el.attr('username');
-
-                        // https://code.google.com/p/webrtc/issues/detail
-                        // ?id=1508
-
-                        if (username) {
-                            const match
-                                = navigator.userAgent.match(
-                                    /Chrom(e|ium)\/([0-9]+)\./);
-
-                            if (match && parseInt(match[2], 10) < 28) {
-                                dict.urls += `${username}@`;
-                            } else {
-                                // only works in M28
-                                dict.username = username;
-                            }
-                        }
-                        dict.urls += el.attr('host');
-                        const port = el.attr('port');
-
-                        if (port) {
-                            dict.urls += `:${el.attr('port')}`;
-                        }
-                        const transport = el.attr('transport');
-
-                        if (transport && transport !== 'udp') {
-                            dict.urls += `?transport=${transport}`;
-                        }
-
-                        dict.credential = el.attr('password')
-                                || dict.credential;
-                        iceservers.push(dict);
-                        break;
+                .c('services', { xmlns: 'urn:xmpp:extdisco:2' }),
+            v2Res => this.onReceiveStunAndTurnCredentials(v2Res),
+            v2Err => {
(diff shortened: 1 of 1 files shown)


damencho:
Does this change means that all requests for turn credentials on a standard jitsi-meet install will work like: the first request will fail and the second for `urn:xmpp:extdisco:1` will pass?

fippo:
nit: xep-0215
someone summon zash or mattj please to explain why there is still a super outdated version around :-)

damencho:
We have a copy of the module in jitsi-meet for few reasons (@fippo maybe this outdated version you mean https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_turncredentials.lua):
1 - We needed a way to specify several turn servers, even though we do not need that anymore
2 - There is no package that contains the available prosody-modules from that repo, so we packed that inside jitsi-meet and also have better control over the versions used

fippo:
clearly we need to reconcile things ;-)

mwild1:
If you no longer need the custom stuff that diverged from [mod_turncredentials](https://modules.prosody.im/mod_turncredentials) in the prosody-modules repo, I recommend switching to or building on top of [mod_external_services](https://modules.prosody.im/mod_external_services), which is now the official way to use XEP-0215. It is in prosody-modules and bundled with Prosody trunk (i.e. will be bundled with Prosody 0.12). It already supports `urn:xmpp:extdisco:2`.

Also, it has an API so if you do need custom behaviour again in the future, you don't need to modify the code :)

jbg:
> Does this change means that all requests for turn credentials on a standard jitsi-meet install will work like: the first request will fail and the second for `urn:xmpp:extdisco:1` will pass?

Yes (but on our jitsi-meet the first request will succeed 😉). Maybe the logic could be reversed to try the old URN first.

damencho:
@jbg Can you fix the lint errors. In the following days I will work on transitioning jitsi-meet to use mod_external_services as advised by Matt and can merge this once that is done.

damencho:
I'm currently trying to migrate from mod_turncredentials to mod_external_services and comparing the outputs with still using `urn:xmpp:extdisco:1` and then on a second phase I will work on supporting `urn:xmpp:extdisco:2` in the client (meging this PR).

Hey @mwild1 in mod_external_services there is that algorithm table and it only defines turn: https://hg.prosody.im/prosody-modules/file/4841cf3fded5/mod_external_services/mod_external_services.lua#l36
Shouldn't there be a listing and for turns? Otherwise, there are no credentials attached to the turns service. 

I'm currently testing with this configuration:
```
external_service_secret = "somesecret";
external_services = {
  { type = "stun", host = "turnserver", port = 443 },
  { type = "turn", host = "turnserver", port = 443, transport = "tcp", secret = true, ttl = 86400 },
  { type = "turns", host = "turnserver", port = 443, transport = "tcp", secret = true, ttl = 86400 },
};
```

mwild1:
Try setting `algorithm = "turn"` on each item and you should be good. It just defaults to looking up by the `"type"` instead.

I.e.

```
external_services = {
  { type = "stun", host = "turnserver", port = 443, algorithm = "turn" },
  { type = "turn", host = "turnserver", port = 443, transport = "tcp", secret = true, ttl = 86400, algorithm = "turn" },
  { type = "turns", host = "turnserver", port = 443, transport = "tcp", secret = true, ttl = 86400, algorithm = "turn" },
};
```

damencho:
Ah :) I see, thank you.

damencho:
The commit was merged as part of https://github.com/jitsi/lib-jitsi-meet/pull/1472

damencho:
Hey @mwild1, we need to be able to access the data that is sent back (the services) for urn:xmpp:extdisco:2 query. We need the data in another module ... and send it back to the client in another stanza with additional data ... 
I don't see such an option at the moment and was wondering what is the correct way to do it ... and/or will you accept such a change for 0.12 ... ? 
Thanks

damencho:
Something like `module:hook("external_service/get_services", get_services);` in mod_external_services?

damencho:
I see, thanks. Yeah, I'm currently experimenting with it. Gathering all data needed and push it to the client with a message early in the connect process (currently I have it on module:hook("resource-bind"). 
The data is some location/region stuff, turn services, enabled components, and some other settings for the muc that will be joined. 
This saves several RTTs and few iq request/responses while connecting ...

damencho:
I did it here
https://github.com/jitsi/jitsi-meet/pull/9052/commits/a5faaf04be8a239be3dcb60aaf97a4a6a88fea82

PR #1461, opened 2021-01-14
Update tokens.md

nathando:
Add a note to current Token Structure that didn't accept numeric values or None (`id` can commonly be number and this will happen easily)

### DIFF
--- a/doc/tokens.md
+++ b/doc/tokens.md
@@ -34,6 +34,7 @@ In addition to the basic claims used in authentication, the token can also provi
   - 'name' is the display name of the user
   - 'email' is the email of the user
   - 'avatar' is the URL of the avatar for the user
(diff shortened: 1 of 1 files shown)


doc/tokens.md
```diff
@@ -34,6 +34,7 @@ In addition to the basic claims used in authentication, the token can also provi
   - 'name' is the display name of the user
   - 'email' is the email of the user
   - 'avatar' is the URL of the avatar for the user
+  Note: As the moment all fields in 'user' need to be a valid string, numeric types or `NoneType` will generate an exception.
```
saghul, line 37: NoneType is not a thing in JS is it? Are you using some Python backend perchance?

nathando: yes i am on Django python

sapkra:

PR #1462, opened 2021-01-14
feat: Detects shard changed when using websockets.

damencho:


modules/xmpp/xmpp.js
```diff
@@ -30,12 +30,19 @@ const logger = getLogger(__filename);
...
  * @param {string} options.enableWebsocketResume - True to enable stream resumption.
  * @param {number} [options.websocketKeepAlive] - See {@link XmppConnection} constructor.
  * @param {Object} [options.xmppPing] - See {@link XmppConnection} constructor.
  * @returns {XmppConnection}
  */
-function createConnection({ enableWebsocketResume, serviceUrl = '/http-bind', token, websocketKeepAlive, xmppPing }) {
+function createConnection({
+    enableWebsocketResume,
+    serviceUrl = '/http-bind',
+    shard, token,
```
paweldomas, line 42: break the line after `shard,`

modules/xmpp/XmppConnection.js
```diff
@@ -260,7 +264,9 @@ export default class XmppConnection extends Listenable {
 
         if (status === Strophe.Status.CONNECTED || status === Strophe.Status.ATTACHED) {
             this._maybeEnableStreamResume();
-            this._maybeStartWSKeepAlive();
+
+            // after connecting immediately check whether shard changed
+            this._maybeStartWSKeepAlive(this._options, 0);
```
paweldomas, line 269: if you need to check shard on the demand it's better extract the fetch and verification part into a separate function and call it here instead of calling "maybe start keep alive" with the intention of "do the fetch request and check if shard changed"

modules/xmpp/XmppConnection.js
```diff
@@ -361,30 +367,41 @@ export default class XmppConnection extends Listenable {
...
+     * @param {Number} [options.websocketKeepAlive=240000] - The websocket keep alive interval.
+     * Pass 0 to immediately check and then proceed with the interval from options.
      *
      * @private
      * @returns {void}
      */
-    _maybeStartWSKeepAlive() {
-        const { websocketKeepAlive } = this._options;
+    _maybeStartWSKeepAlive(
+            { shard, websocketKeepAlive },
```
paweldomas, line 378: passing the shard and options  to override behavior is confusing, I think you did that to be able to do the fetch and check on demand/instantly, but this should be a separate function (see the comment above)

paweldomas (changes requested):

modules/xmpp/xmpp.js
damencho: Ups :)

damencho:
Does it look ok now? 
I still need to test it after the change ... passing the method without binding it will fail, I suppose :)

damencho:
And also I was thinking to avoid the initial request somehow and leave it only if reconnecting and stuff ...

modules/xmpp/XmppConnection.js
```diff
@@ -377,18 +383,35 @@ export default class XmppConnection extends Listenable {
...
+                if (responseShard !== shard) {
+                    logger.error(
+                        `Detected that shard changed from ${shard} to ${responseShard}`);
+                    this.eventEmitter.emit(XmppConnection.Events.CONN_SHARD_CHANGED);
+                }
+            })
+            .catch(error => {
+                logger.error(`Websocket Keep alive failed for url: ${url}`, { error });
+            })
+            .then(() => this._maybeStartWSKeepAlive());
```
paweldomas, line 412: I would remove `.then(() => this._maybeStartWSKeepAlive());` and make the method return a promise, this method should not care about the interval, maybe call it `keepAliveAndCheckShard`

modules/xmpp/XmppConnection.js
```diff
@@ -260,7 +264,9 @@ export default class XmppConnection extends Listenable {
 
         if (status === Strophe.Status.CONNECTED || status === Strophe.Status.ATTACHED) {
             this._maybeEnableStreamResume();
-            this._maybeStartWSKeepAlive();
+
+            // after connecting - immediately check whether shard changed and maybe schedule keep alive
+            this._checkShardAndMaybeScheduleKeepAlive();
```
paweldomas, line 269: Here call `keepAliveAndCheckShard` mentioned in another comment, also leave the comment which explains it's doing the check immediately.
Just after that leave the `this._maybeStartWSKeepAlive();` which is supposed to schedule the keep alive

modules/xmpp/XmppConnection.js
```diff
@@ -377,18 +383,35 @@ export default class XmppConnection extends Listenable {
...
-            this._wsKeepAlive = setTimeout(() => {
-                const url = this.service.replace('wss://', 'https://').replace('ws://', 'http://');
-
-                fetch(url).catch(
-                    error => {
-                        logger.error(`Websocket Keep alive failed for url: ${url}`, { error });
-                    })
-                    .then(() => this._maybeStartWSKeepAlive());
-            }, intervalWithJitter);
+            this._wsKeepAlive = setTimeout(this._checkShardAndMaybeScheduleKeepAlive, intervalWithJitter);
```
paweldomas, line 386:
```
this._wsKeepAlive = setTimeout(() => {
  this.keepAliveAndCheckShard()
      .then(() => this._maybeStartWSKeepAlive());
}
```

paweldomas (approved):

Issue #1465, opened 2021-01-17
WS connection drops after keep alive interval


bukharin:
Hi,

After keepalive interval lib-jitsi-meet drops WS connenction with error `Detected that shard changed from undefined to null `

It seems after [this commit](https://github.com/jitsi/lib-jitsi-meet/commit/cb484cf48ccd3d56cf9bb2e9d452bd7aa644ac8a#diff-6728f0c5879db904f5a1e1a64073e0ddb1ca03bac3c2e20ba73ffa196a32d7d6R416)

damencho:
What do you mean after the keepalive interval? 
This commit adds a check to check the headers coming from the webserver checking for 'x-jitsi-shard'. 
What are your problem and use case, do you reproduce it every time and how do you reproduce it?

bukharin:
@damencho after 60 sec (default keep-alive interval) lib-jitsi-meet fetches `/xmpp-websocket` address. Our installation does not have `x-jitsi-shard` header in the response (null). Also we don't configure `shard` option (undefined).

In the commit above you have a strict comparison (null !== undefined). In this case `li-jitsi-meet` emit `CONFERENCE_FAILED` event with `conference.otherError` code.

damencho:
> In the commit above you have a strict comparison (null !== undefined).

🤦

damencho:
Thank you for the report. Can you test, does https://github.com/jitsi/lib-jitsi-meet/pull/1467 fixes it for you?

bukharin:
@damencho yes it does, thank you

billybobilly:
Leaving a note in case someone needs this too.  In our case, we don't store the shard in header, so it fails on null responseShard.

had to add null test to this line:
 433               if (responseShard !== shard && responseShard !== null) {

Not sure if it's the smartest fix, but worked for us.

PR #1467, opened 2021-01-18
fix: Skip header checking for ws and shard when not configured. Fixes…

damencho:
… #1465.

### DIFF
--- a/modules/xmpp/XmppConnection.js
+++ b/modules/xmpp/XmppConnection.js
@@ -409,6 +409,12 @@ export default class XmppConnection extends Listenable {
 
         return fetch(url)
             .then(response => {
+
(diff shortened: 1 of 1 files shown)


modules/xmpp/XmppConnection.js
```diff
@@ -409,6 +409,12 @@ export default class XmppConnection extends Listenable {
 
         return fetch(url)
             .then(response => {
+
+                // skips header checking if there is no info inn options
```
bukharin, line 413: typo `in`

damencho: Now you reveal who is using the MacBook with the defective keyboard ...

gpolitis (approved):

PR #1470, opened 2021-01-20
feat: Disable JVB via configuration

jersub:
In a context of strong data protection policy, disabling the JVB and forcing the P2P mode would ensure that the hosting provider is not able to access the conference traffic.

This would be an alternative to E2EE, which is not fully ready yet (esp. on native mobile SDKs).

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -540,6 +540,16 @@ JitsiConference.prototype.isJoined = function() {
 };
 
 /**
+ * Tells whether or not the JVB mode is enabled in the configuration.
+ * @return {boolean}
+ */
+JitsiConference.prototype.isJVBEnabled = function() {
+    // FIXME: remove once we have a default config template.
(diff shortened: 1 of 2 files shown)


jersub:
I'll sign the CLA once you confirm by comment that you accept the PR, no problem.

jersub:
CLA signed.

Any feedback for this PR?

Funny fact: I've just found out that @paweldomas has already done such change in branch "disable_jvb_connection" few years ago.

saghul:
This would only work for 2 participants because more requires the JVB.

Using Jitsi Meet for this seems like not the right tool for the job TBH.

You'd need to enforce room limits so conferences don't break if more people join, for example.

If a participant gets disconnected, and rejoins there is a brief window of time in which there are 3 participants. So things will break.

Overall, I don't think this scenario has a place in Jitsi Meet.

PR #1472, opened 2021-01-21
feat: Support the urn:xmpp:extdisco:2 namespace specified by XEP-0215

damencho:
This uses the commit from https://github.com/jitsi/lib-jitsi-meet/pull/1458 and fixes the problems to make the commit mergeable :)

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -297,6 +297,7 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
         //
         // See https://modules.prosody.im/mod_turncredentials.html
         // for a prosody module which implements this.
+        // Or the new implementation https://modules.prosody.im/mod_external_services which will be in prosody 0.12
         //
         // Currently, this doesn't work with updateIce and therefore credentials
         // with a long validity have to be fetched before creating the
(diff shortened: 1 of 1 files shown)


modules/xmpp/strophe.jingle.js
```diff
@@ -345,23 +349,7 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
...
-                // https://code.google.com/p/webrtc/issues/detail
-                // ?id=1508
-
-                if (username) {
-                    const match
-                        = navigator.userAgent.match(
-                            /Chrom(e|ium)\/([0-9]+)\./);
-
-                    if (match && parseInt(match[2], 10) < 28) {
-                        dict.urls += `${username}@`;
```
damencho: @fippo This is correct right, this was some workaround for chrome 27 and below which we no longer need :)? 

The supported chrome version was bumped to 72 recently.

bgrozev (approved):

modules/xmpp/strophe.jingle.js
fippo: kill it with :fire:

modules/xmpp/strophe.jingle.js
```diff
@@ -297,6 +297,7 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
         //
         // See https://modules.prosody.im/mod_turncredentials.html
         // for a prosody module which implements this.
+        // Or the new implementation https://modules.prosody.im/mod_external_services which will be in prosody 0.12
```
fippo, line 300: @mwild1 can you add a deprecation notice to the old url?

PR #1475, opened 2021-01-22
feat: Implement aggresive layer suspension on RN.

jallamsetty1:
RN doesn't support RTCRtpSender yet. Therefore, media is suspended on RN by changing the media direction in the SDP whenever the client receives an ideal height of 0 for sender constraints on the bridge channel.


saghul (approved):

saghul:
This will only stop sending the video if nobody is receiving it right? If they are all simulcast layers will still be sent, correct?

jallamsetty1:
> This will only stop sending the video if nobody is receiving it right? If they are all simulcast layers will still be sent, correct?

Yes, this will stop sending video only when bridge sends an idealHeight of 0 meaning no one is viewing this endpoint, all simulcast layers will be sent for any value > 0.

PR #1479, opened 2021-01-27
fix(ice-restart): Force client reloads when call is migrated.

jallamsetty1:
Force the client to reload when the bridge that is handling the media goes down.
This mitigates issues seen on the bridge because of a client re-joining the call with the same endpointId, BWE issues, etc.
This behavior is configurable through 'enableForcedReload' setting in config.js.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2622,6 +2622,19 @@ JitsiConference.prototype.isConnectionInterrupted = function() {
 };
 
 /**
+ * Handles {@link XMPPEvents.CONNECTION_RESTARTED} event. This happens when the bridge goes down
+ * and Jicofo moves conferences away to a different bridge.
+ * @param {JingleSessionPC} session
+ * @private
+ */
+JitsiConference.prototype._onConferenceRestarted = function(session) {
+    if (!session.isP2P && this.options.config.enableForcedReload) {
+        this.restartInProgress = true;
+        this.eventEmitter.emit(JitsiConferenceEvents.CONFERENCE_FAILED, JitsiConferenceErrors.CONFERENCE_RESTARTED);
+    }
+};
+
+/**
  * Handles {@link XMPPEvents.CONNECTION_INTERRUPTED}
  * @param {JingleSessionPC} session
  * @private
--- a/JitsiConferenceErrors.js
+++ b/JitsiConferenceErrors.js
@@ -28,6 +28,12 @@ export const CONFERENCE_MAX_USERS = 'conference.max_users';
 export const CONNECTION_ERROR = 'conference.connectionError';
 
 /**
+ * Indicates that the client has been forced to restart by jicofo when the
+ * conference was migrated from one bridge to another.
+ */
+export const CONFERENCE_RESTARTED = 'conference.restarted';
+
+/**
  * Indicates that a connection error is due to not allowed,
  * occurred when trying to join a conference.
  */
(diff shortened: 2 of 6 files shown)


paweldomas:
what's the problem with doing ICE restart via 'session-terminate' for this case? you could just make a change for the client to react with session-terminate for transport-replace (so that there's no need to wait for ice failed). is there a plan to remove that flow (it's still used fo E2EE encryption case to enable insertable streams)?

jallamsetty1:
> what's the problem with doing ICE restart via 'session-terminate' for this case? you could just make a change for the client to react with session-terminate for transport-replace (so that there's no need to wait for ice failed). is there a plan to remove that flow (it's still used fo E2EE encryption case to enable insertable streams)?

We are running into other issues on the bridge side where the client re-joins the call on a new media session using the old endpoint id.

paweldomas:
> We are running into other issues on the bridge side where the client re-joins the call on a new media session using the old endpoint id. We won't be removing the session-terminate flow altogether, if the call hasn't been migrated to another bridge, we don't have the bridge issues (which is the case when insertable streams is enabled).

Is it some state not cleaned up after an endpoint is expired?

paweldomas:
Also is this a long term solution or a temporary workaround?

bgrozev:
Yes. When the endpoint is moved to another bridge (or re-created in the same bridge), the bridge's TCC state is reset. This state includes the tcc sequence number, so a new initial one is chosen (I think jvb starts from 0). However, chrome's TCC state is not reset. I think chrome interperts the sequence numbers as too old, so it keep re-sending TCC feedback for the packets received in the previous session. The end result is that until the sequence numbers "catch up", the bridge's BWE drops to practically 0. It's been reported here: https://bugs.chromium.org/p/webrtc/issues/detail?id=12146

I don't think this PR is meant as a permanent solution (right Jaya?). We had talked about finding a less obtrusive way to clear the state (re-create the PC, or call some API that has this as a side effect?), or waiting until it's fixed in the browser.

I'm also not sure if we resolved all the server-side issues that come from moving an endpoint between bridges (there were some bugs around octo since the flow there is very awkward, with some bridges seeing the endpoint change from local to octo, others seeing a change from octo to local).

jallamsetty1:
> Yes. When the endpoint is moved to another bridge (or re-created in the same bridge), the bridge's TCC state is reset. This state includes the tcc sequence number, so a new initial one is chosen (I think jvb starts from 0). However, chrome's TCC state is not reset. I think chrome interperts the sequence numbers as too old, so it keep re-sending TCC feedback for the packets received in the previous session. The end result is that until the sequence numbers "catch up", the bridge's BWE drops to practically 0. It's been reported here: https://bugs.chromium.org/p/webrtc/issues/detail?id=12146
> 
> I don't think this PR is meant as a permanent solution (right Jaya?). We had talked about finding a less obtrusive way to clear the state (re-create the PC, or call some API that has this as a side effect?), or waiting until it's fixed in the browser.
> 
> I'm also not sure if we resolved all the server-side issues that come from moving an endpoint between bridges (there were some bugs around octo since the flow there is very awkward, with some bridges seeing the endpoint change from local to octo, others seeing a change from octo to local).

We asked Huib about the TCC bug today and he said that it's not on their priority list right now. Waiting for it to be fixed in the browser is ruled out. Creating a new PC is possible but then again that wouldn't fix the octo related issues with moving an endpoint between bridges.

damencho (approved):

PR #1481, opened 2021-01-29
Use unified plan for Chrome iOS

### LINKED ISSUE #1480: iOS Chrome not using Unified Plan
## Description

When joining a conference with another participant from Chrome on iOS, the session is immediately terminated. The error in the javascript console:

>ERROR2021-01-27T15:41:37.691Z [modules/xmpp/JingleSessionPC.js] addRemoteStream failed: InvalidAccessError: Failed to set remote offer sdp: Media section has more than one track specified with a=ssrc lines which is not supported with Unified Plan.

## Possible Solution

It appears the browser check to determine if it uses unified plan is not accurate for Chrome on iOS which is based on the webkit engine.

See: https://github.com/jitsi/lib-jitsi-meet/blob/301e3a22dd76341b15d3846ab539e713d8e7569b/modules/browser/BrowserCapabilities.js#L210

I believe the `usesUnifiedPlan` function should be updated to something like this:

```
    usesUnifiedPlan() {
        if (this.isFirefox() || (this.isChrome() && this._bowser.isEngine('webkit')) {
            return true;
        }

        if (this.isSafari() && typeof window.RTCRtpTransceiver !== 'undefined') {
            // https://trac.webkit.org/changeset/236144/webkit/trunk/LayoutTests/webrtc/video-addLegacyTransceiver.html
            // eslint-disable-next-line no-undef
            return Object.keys(RTCRtpTransceiver.prototype).indexOf('currentDirection') > -1;
        }

        return false;
    }
```

# Environment details
iOS 14.4, Chrome 87, docker-jitsi-meet:stable-5142-4

jkowens:
Chrome iOS uses WebKit engine so it uses the Unified Plan. Fixes #1480.

### DIFF
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -65,6 +65,24 @@ export default class BrowserCapabilities extends BrowserDetection {
     }
 
     /**
+     * Checks if the current browser is WebKit based. It's either
+     * Safari or uses WebKit as its engine.
+     *
+     * This includes Chrome and Firefox on iOS
+     *
+     * @returns {boolean}
+     */
+    isWebKitBased() {
+        // https://trac.webkit.org/changeset/236144/webkit/trunk/LayoutTests/webrtc/video-addLegacyTransceiver.html
+        return this._bowser.isEngine('webkit')
+            && typeof navigator.mediaDevices !== 'undefined'
+            && typeof navigator.mediaDevices.getUserMedia !== 'undefined'
+            && typeof window.RTCRtpTransceiver !== 'undefined'
+            // eslint-disable-next-line no-undef
+            && Object.keys(RTCRtpTransceiver.prototype).indexOf('currentDirection') > -1;
+    }
+
+    /**
      * Checks whether current running context is a Trusted Web Application.
      *
      * @returns {boolean} Whether the current context is a TWA.
@@ -208,14 +226,8 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean}
      */
     usesUnifiedPlan() {
-        if (this.isFirefox()) {
+        if (this.isFirefox() || this.isWebKitBased()) {
             return true;
-        }
-
-        if (this.isSafari() && typeof window.RTCRtpTransceiver !== 'undefined') {
-            // https://trac.webkit.org/changeset/236144/webkit/trunk/LayoutTests/webrtc/video-addLegacyTransceiver.html
-            // eslint-disable-next-line no-undef
-            return Object.keys(RTCRtpTransceiver.prototype).indexOf('currentDirection') > -1;
         }
 
         return false;


saghul:
Wouldn't checking if the browser engine is webkit (where we currently just check for Safari) be more resilient? I suspect this same problem will happen with Firefox on iOS for example.

jkowens:
@saghul that's a good point. Firefox correctly uses UnifiedPlan because it checks for Firefox, but other browsers on iOS will  have that issue. I've updated the PR 👍

jallamsetty1:
There are couple of things that we need to take into consideration before enabling unified plan support for chrome/FF on iOS.
gUM is only supported on the latest version of iOS and there are subtle differences in implementation of the unified plan by each of these browsers that we take into account when calling the WebRTC APIs. I would be surprised if everything would work well just by modifying the unified plan check in Browser capabilities.

jkowens:
@jallamsetty1 I believe unified plan is already enabled for Firefox on iOS if I understand the code correctly. This change will enable it for other browsers on iOS as well. 

My testing with FF on iOS hasn't gone so well. I can join a conference using FF from my iPhone, but other participants cannot see my video. I'm guessing FF unified plan differs from FF iOS (webkit) unified plan implementation.

jallamsetty1:
> @jallamsetty1 I believe unified plan is already enabled for Firefox on iOS if I understand the code correctly. This change will enable it for other browsers on iOS as well.
> 
Firefox on iOS is not supported yet - https://github.com/jitsi/jitsi-meet/blob/master/react/features/base/environment/environment.js#L92
We used to show the deeplinking page and not show the `Launch in Web` option for FF/Chrome on iOS but that seems to have recently. 

> My testing with FF on iOS hasn't gone so well. I can join a conference using FF from my iPhone, but other participants cannot see my video. I'm guessing FF unified plan differs from FF iOS (webkit) unified plan implementation?

Yes, I believe so but this is something I will be able to confirm after I test it.

fippo:
I would recommend using the adapter.js version detection. It will detect chrome on ios as webkit -- which is how it behaves.

saghul:
> I would recommend using the adapter.js version detection. It will detect chrome on ios as webkit -- which is how it behaves.

Agreed, we probably should do something similar to what we do with Chromium and introduce a `isWebKitBased` check instead of checking explicitly for Safari.

jkowens:
@saghul I had named the method `usesWebKit`. Would you like me to rename it to `isWebKitBased`? I just wasn't sure if there should be a distinction between browser and engine checks.

saiflakhani:
Is there any update on this PR? Chrome on iOS is a bit of a requirement for us.

saghul:
More testing is required, stats for instance won't work I think.

jallamsetty1:
@jkowens Can you please update your PR to add a check for `navigator.mediaDevices.getUserMedia` in your isWebkitBased() function for iOS based browsers other than Safari. You can also move the Safari unified plan check (i.e., the currentDirection support) under there because we don't support older versions of Safari ? I will then go ahead and update the rest of the code to make this work.

jkowens:
@jallamsetty1 I think I've made most of the changes you've requested, but I'm not sure what I should do exactly with the `navigator.mediaDevices.getUserMedia` check?

jallamsetty1:
> @jallamsetty1 I think I've made most of the changes you've requested, but I'm not sure what I should do exactly with the `navigator.mediaDevices.getUserMedia` check?

You can add something similar to this - https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/browser/BrowserCapabilities.js#L264
This is needed since gUM for WKWebview was added in iOS 14.3 and is not supported in the older versions of iOS.

jkowens:
@jallamsetty1 that makes sense. I've made that update as well, but please let me know if it needs revised. Thanks!

modules/browser/BrowserCapabilities.js
```diff
@@ -266,6 +277,17 @@ export default class BrowserCapabilities extends BrowserDetection {
                     !== 'undefined');
     }
 
+    /**
+     * Checks if the browser supports getUserMedia.
+     * @returns {boolean} {@code true} if the browser supports getUserMedia.
+     */
+    supportsGetUserMedia() {
+        return typeof navigator.getUserMedia !== 'undefined'
+            || (typeof navigator.mediaDevices !== 'undefined'
```
jallamsetty1, line 286: You can directly add this check `typeof navigator.mediaDevices !== 'undefined' && typeof navigator.mediaDevices.getUserMedia !== 'undefined'` to `isWebKitBased()` instead of declaring a function since this function won't be used anywhere else. Sorry if my previous comment was confusing.

jallamsetty1:

modules/browser/BrowserCapabilities.js
jkowens: Got it 👍

jallamsetty1:
@jkowens Can you please rebase against the master ? The test failure is because of a recent commit which is not in your branch.

jkowens:
@jallamsetty1 all set!

jallamsetty1 (approved):

jallamsetty1:
> @jallamsetty1 all set!

Thank you for your patience, this PR has been merged. I will make the other changes necessary to make WKWebview based browsers to work.

PR #1487, opened 2021-02-07
Fixes exceptions which occurs after leaving from conference

### LINKED ISSUE #1476: Cannot read property 'connectionTimes' of null
Hi,

`Cannot read property 'connectionTimes' of null` at https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L2396

I guess it happens after disposing `JitsiConference` (invoke `leave` method), because you delete internal `room` reference, but some of events occurs after that.

As I know this is has no effect on the user, but I think it should be properly handled

bukharin:
Some of events can occurs after application leaves from the conference: `_playCallback` and `_actOnIceFailed`. This PR explicit mark conference as destroyed and check this in the callbacks

* fix: Cannot read property 'supportsRestartByTerminate' of null
* fix: Can't read property connectionTimes of null

Close #1476

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -698,6 +698,7 @@ JitsiConference.prototype.leave = async function() {
     room.removeListener(XMPPEvents.SOURCE_REMOVE, this._updateRoomPresence);
 
     this.eventManager.removeXMPPListeners();
+    this.destroyed = true;
 
     this._signalingLayer.setChatRoom(null);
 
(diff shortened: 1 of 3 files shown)


NewEraCracker (approved):

saghul:
Wouldn't checking if room is null accomplish the same?

bukharin:
@saghul  yes, but I decide to decouple this state. I think it will be better if `JitsiConference` will have explicit public `destroyed` property. Also we may want to close (made private) `room` property as part of internal implementation of `JitsiConference`.

saghul:
That destroyed property can be derived from the room existing or not. Having to sync both properties might yield to bugs in the future if we make changes to one but fail to mark it as destroyed.

Also, in general, I'd rather not make the API larger than it needs to be unless there is a good reason.

I'd rather go with https://github.com/jitsi/lib-jitsi-meet/pull/1834 which is much smaller.

bukharin:
@saghul ok, close this PR

PR #1488, opened 2021-02-08
Rename QualityController to SendVideoController in preparation for using the new 'ReceiverVideoConstraints' message that invokes the new BitrateController in the bridge.

jallamsetty1:
Refactor handling of SENDER_VIDEO_CONSTRAINTS_CHANGED event. Bring it inline with how other events received on the bridge channel are handled.
All the send video constraints for the client will be handled by SendVideoController and the receive video constraints like lastN, selectEndpoints and receive video resolution will be handled by ReceiveVideoController which will be added in another PR.

### DIFF
--- a/modules/RTC/RTC.js
+++ b/modules/RTC/RTC.js
@@ -182,6 +182,8 @@ export default class RTC extends Listenable {
         // The last N change listener.
         this._lastNChangeListener = this._onLastNChanged.bind(this);
 
+        this._senderVideoConstraintsChangedListener = this._senderVideoConstraintsChanged.bind(this);
+
         this._onDeviceListChanged = this._onDeviceListChanged.bind(this);
         this._updateAudioOutputForAudioTracks
             = this._updateAudioOutputForAudioTracks.bind(this);
@@ -207,20 +209,11 @@ export default class RTC extends Listenable {
      * @returns {void}
      */
     destroy() {
-        RTCUtils.removeListener(
-            RTCEvents.AUDIO_OUTPUT_DEVICE_CHANGED,
-            this._updateAudioOutputForAudioTracks
-        );
+        RTCUtils.removeListener(RTCEvents.AUDIO_OUTPUT_DEVICE_CHANGED, this._updateAudioOutputForAudioTracks);
+        RTCUtils.removeListener(RTCEvents.DEVICE_LIST_CHANGED, this._onDeviceListChanged);
 
-        RTCUtils.removeListener(
-            RTCEvents.DEVICE_LIST_CHANGED,
-            this._onDeviceListChanged
-        );
-
-        this.removeListener(
-            RTCEvents.LASTN_ENDPOINT_CHANGED,
-            this._lastNChangeListener
-        );
+        this.removeListener(RTCEvents.LASTN_ENDPOINT_CHANGED, this._lastNChangeListener);
+        this.removeListener(RTCEvents.SENDER_VIDEO_CONSTRAINTS_CHANGED, this._senderVideoConstraintsChangedListener);
 
         if (this._channelOpenListener) {
             this.removeListener(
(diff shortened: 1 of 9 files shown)


modules/RTC/RTC.js
```diff
@@ -340,7 +337,7 @@ export default class RTC extends Listenable {
     _senderVideoConstraintsChanged(senderVideoConstraints) {
         logger.info('Remote max frame height received on bridge channel: ', JSON.stringify(senderVideoConstraints));
         this._senderVideoConstraints = senderVideoConstraints;
-        this.eventEmitter.emit(RTCEvents.SENDER_VIDEO_CONSTRAINTS_CHANGED);
+        this.eventEmitter.emit(RTCEvents.REMOTE_VIDEO_CONSTRAINTS_CHANGED);
```
paweldomas, line 340: it's a little bit confusing that this emits `REMOTE_VIDEO_CONSTRAINTS_CHANGED` when the "senderVideoConstraintsChanged", especially that `this._senderVideoConstraints` is accessible via `getSenderVideoConstraints`

modules/RTC/RTC.js
```diff
@@ -864,8 +861,9 @@ export default class RTC extends Listenable {
             this._channel.close();
             this._channel = null;
 
-            this.removeListener(RTCEvents.LASTN_ENDPOINT_CHANGED,
-                this._lastNChangeListener);
+            this.removeListener(RTCEvents.LASTN_ENDPOINT_CHANGED, this._lastNChangeListener);
```
paweldomas, line 864: why is the cleanup in both `destroy` method and here?

jallamsetty1: Looks like the call in `destroy()` is redundant since `destroy()` is called after `closeBridgeChannel()` when client leaves the room. I will delete the call there. Thanks for pointing it out.

modules/RTC/RTC.js
jallamsetty1: I know, I needed two different events for this, one that gets emitted when the message is received on the bridge channel and the other which notifies JingleSessionPC about this event.
https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/JingleSessionPC.js#L587
I went with the message used there. I am open to any others suggestion you might have.

modules/qualitycontrol/SendVideoController.js
```diff
@@ -1,28 +1,35 @@
...
+        this.rtc = rtc;
         this.conference.on(
             JitsiConferenceEvents._MEDIA_SESSION_STARTED,
             session => this._onMediaSessionStarted(session));
         this.conference.on(
             JitsiConferenceEvents._MEDIA_SESSION_ACTIVE_CHANGED,
             () => this._propagateSendMaxFrameHeight());
+        this.rtc.on(
+            RTCEvents.SENDER_VIDEO_CONSTRAINTS_CHANGED,
+            videoConstraints => this._propagateSendMaxFrameHeight(videoConstraints));
```
paweldomas, line 32: I think it would be a bit cleaner if this part was split into:
```suggestion
            videoConstraints => {
                this._senderVideoConstraints = videoConstraints;
                this._propagateSendMaxFrameHeight());
             }
```

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -0,0 +1,86 @@
...
+export class ReceiveVideoController {
+    /**
+     * Creates a new instance for a given conference.
+     *
+     * @param {JitsiConference} conference the conference instance for which the new instance will be managing
+     * the receive video quality constraints.
+     * @param {RTC} rtc the rtc instance which is responsible for initializing the bridge channel.
+     */
+    constructor(conference, rtc) {
+        this.conference = conference;
```
paweldomas, line 18: It's better to store both `conference` and `rtc` references with the underscore notations, so the are kind of private. I know this is not consistent, but lets keep at least new stuff the right way.

modules/qualitycontrol/SendVideoController.js
paweldomas: also please change `this.senderVideoConstraints` to `this._senderVideoConstraints` in other places in this class

paweldomas (approved):

PR #1490, opened 2021-02-10
fix(example) simplify

saghul:
- Drop unneeded bundled libraries
- Use jQuery from a CDN


damencho (approved):

ralyodio:
WHy don't we remove jquery completely, we're not using it for anything we can do easier with vanilla js.

saghul:
We are using it internally.

It can now indeed be dropped as a external dependency.

PR #1491, opened 2021-02-11
feat: add device id to caps

quitrk:


damencho:
What is the usecase for this? In latest versions we try not to use the disco-info from the clients and only jicofo will query for that information and has it cached, but adding a capability that is different per client will drastically increase signalling between jicofo and prosody as that capability cannot be used from the cache.

damencho (dismissed):

damencho:
Discussing it offline and moved to use the machine id.

damencho (approved):

damencho:
Jenkins add to whitelist.

damencho (approved):

PR #1494, opened 2021-02-13
Add parameter for adding custom constraints to RTCPeerConnection

vladertel:
I tried to add my custom data to rtcstats as it described in here (https://github.com/jitsi/rtcstats/blob/master/README.md#usage) but actually I was not able to do that by using jitsi-meet or lib-jitsi-meet. So here it is!

Also I will file a PR for jitsi-meet with customOptionalConstraints option in config.js

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1950,6 +1950,7 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
     try {
         jingleSession.initialize(this.room, this.rtc, {
             ...this.options.config,
+            customOptionalConstraints: this.connection.options.customOptionalConstraints || [],
             enableInsertableStreams: this._isE2EEEnabled()
         });
     } catch (error) {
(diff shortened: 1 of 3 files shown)


damencho:
@andrei-gavrilescu Can you take a look?

What is the goal with these PRs @vladertel ?

vladertel:
The main reason I need this - is somehow pass client identifier in RTCPeerConnection to gather it later in rtcstats. This allows me to match a connection with a particular user in my ecosystem

vladertel:
cla is already signed

andrei-gavrilescu:
Hello @vladertel, the config options are already being sent via the identity call [here](https://github.com/jitsi/jitsi-meet/blob/684d12115974a65b48f36c7d4fe74d8f17da6ff6/react/features/rtcstats/middleware.js#L68). So all the data should be available in your dump on rtcstats-server side.

PR #1497, opened 2021-02-17
feat: Use the new bridge signaling format.

jallamsetty1:
Translate the 'LastNChangedEvent', 'SelectedEndpointsChangedEvent' and 'ReceiverVideoConstraint' messages into the new 'ReceiverVideoConstraints' message that invokes the new bandwidth allocation algorithm in the bridge that is described here - https://github.com/jitsi/jitsi-videobridge/blob/master/doc/allocation.md. useNewBandwidthAllocationStrategy=true in config.js will invoke the translation in the client.

### DIFF
--- a/modules/qualitycontrol/ReceiveVideoController.js
+++ b/modules/qualitycontrol/ReceiveVideoController.js
@@ -1,5 +1,138 @@
+import { getLogger } from 'jitsi-meet-logger';
 
 import * as JitsiConferenceEvents from '../../JitsiConferenceEvents';
+
+const logger = getLogger(__filename);
+const MAX_HEIGHT_ONSTAGE = 2160;
+const MAX_HEIGHT_THUMBNAIL = 180;
+const LASTN_UNLIMITED = -1;
+
+/**
+ * This class translates the legacy signaling format between the client and the bridge (that affects bandwidth
+ * allocation) to the new format described here https://github.com/jitsi/jitsi-videobridge/blob/master/doc/allocation.md
+ */
+export class ReceiverVideoConstraints {
+    /**
+     * Creates a new instance.
+     */
+    constructor() {
+        // Default constraints used for endpoints that are not explicitly included in constraints.
+        // These constraints are used for endpoints that are thumbnails in the stage view.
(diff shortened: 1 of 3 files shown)


modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -1,6 +1,139 @@
...
+        this._maxFrameHeight = MAX_HEIGHT_ONSTAGE;
+
+        // The endpoint IDs of the participants that are currently selected.
+        this._selectedEndpoints = [];
+
+        this._receiverVideoConstraints = {
+            constraints: {},
+            defaultConstraints: this.defaultConstraints,
+            lastN: this._lastN,
+            selectedEndpoints: this._selectedEndpoints
```
bgrozev, line 36: nit: Also initializee `onStageEndpoints`?

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -80,7 +237,14 @@ export class ReceiveVideoController {
         this._maxFrameHeight = maxFrameHeight;
 
         for (const session of this._conference._getMediaSessions()) {
-            maxFrameHeight && session.setReceiverVideoConstraint(maxFrameHeight);
+            if (session.isP2P || !this._receiverVideoConstraints) {
```
bgrozev, line 240: Is `setReceiverVideoConstraint` actually used in p2p?

jallamsetty1: Yes, JingleSessionPC.setReceiverVideoConstraint sends out a content-modify for requesting a different resolution in p2p mode and uses the legacy format for jvb mode. https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/JingleSessionPC.js#L1311

modules/qualitycontrol/ReceiveVideoController.js
jallamsetty1: Good catch, will initialize `onStageEndpoints` as well.

bgrozev (approved):

PR #1503, opened 2021-02-24
Typescript type definitions and automation of their creation

garyhuntddn:
Auto-generated type definitions including npm script to create them from the jsdocs in the .js files.

Use ```npm run-script types``` in the root folder to re-created the auto-generated files.

Hand-crafted type definitions with as much inference as I could manage at a first glance.

Tool to generate type definitions via an AST and export them to JSON so that they can be compared to see how the auto-generated and the hand-crafted compare.

Use ```npm start``` and ```npm test``` from the types/types-comparer folder to generate two JSON files.

These two JSON files can then be compared to see where we're missing annotations in the jsdocs or where I've got the types wrong in my hand-crafted definition files


types/hand-crafted/JitsiConference.d.ts
```diff
@@ -0,0 +1,142 @@
+import { JitsiConferenceEvents } from './JitsiConferenceEvents';
+import JitsiConnection from './JitsiConnection';
+import JitsiTrackError from './JitsiTrackError';
+import JitsiParticipant from './JitsiParticipant';
+import JitsiLocalTrack from './modules/RTC/JitsiLocalTrack';
+import JitsiRemoteTrack from './modules/RTC/JitsiLocalTrack';
+import JitsiTrack from './modules/RTC/JitsiTrack';
+import Transcriber from './modules/transcription/Transcriber';
```
kierenevans, line 8: I think this needs to be:
```javascript
import Transcriber from './modules/transcription/transcriber';
```
Fixes an issue on linux (case-sensitive filesystem) where it cannot be imported

garyhuntddn: Good spot

types/hand-crafted/modules/RTC/JitsiTrack.d.ts
```diff
@@ -0,0 +1,28 @@
+import JitsiConference from '../../JitsiConference';
+import { MediaType } from '../../service/RTC/MediaType';
+import TraceablePeerConnection from './TraceablePeerConnection';
+
+export default class JitsiTrack {
```
kierenevans, line 5: Might need to be extending EventEmitter from `event` for addListener() and removeListener()
The constructor also aliases these as `addEventListener()` and `removeEventListener()`.

Can't seem to get this recognised locally though

types/hand-crafted/modules/xmpp/JingleSessionPC.d.ts
```diff
@@ -0,0 +1,49 @@
+import JingleSession from './JingleSession';
+import XmppConnection from './XmppConnection';
+import JitsiLocalTrack from '../RTC/JitsiLocalTrack';
+import CodecMimeType from '../../service/RTC/CodecMimeType';
```
kierenevans, line 4: I think this needs to be:
```javascript
import { CodecMimeType } from '../../service/RTC/CodecMimeType';
```

types/hand-crafted/modules/RTC/JitsiTrack.d.ts
garyhuntddn: Will be interesting if we can persuade the jsdocs auto-generated to do the same

types/hand-crafted/modules/xmpp/JingleSessionPC.d.ts
garyhuntddn: Thought I'd got all of those - I switched from types to const enums as they provide better semantics :-) I blame VS code for not picking up the error :-)

Fuzzyma:
@garyhuntddn what would be the proper way of using these types you generated? Since this library is bundled, it feels like a final bundled d.ts file is missing

garyhuntddn:
> @garyhuntddn what would be the proper way of using these types you generated? Since this library is bundled, it feels like a final bundled d.ts file is missing

There's two possibilities - one is that we maintain an independent package containing .d.ts files - the alternative - which I believe is the better option is to use what I've done to gradually improve the jsdoc files such that we can autogenerate them. As part of the build process we can then introduce the automated jsdoc->d.ts step and publish a package each time the main lib.jitsi.meet package is deployed.

kierenevans:
[Here's how I got this PR working for me in an Angular application](https://gist.github.com/kierenevans/77204ba1be12da5ca55d969deb2ec13e)

I'm brand new to the modern JS scene so a lot of this won't be best practise!

It's also a bit clunky and we are not using this in production but it seems to work as a stopgap until this PR is ready to go with easier type imports.

Fuzzyma:
@kierenevans I see. Thanks!
I definitely would go the route and fix all the jsdoc comments. We then can generate all d.ts codes automatically.
This should also take care of a "final" d.ts which then can be imported (or is automatically picked up by typescript). The typings field in package.json should be added in this case

nightvisi0n:
Hey @garyhuntddn, thanks for this awesome addition. I really look forward to when this is merged into master. :raised_hands:

nightvisi0n:
> Use `npm run-script types` in the root folder to re-created the auto-generated files.

Doing so gives me these errors:

```
❯ npm run-script types

> lib-jitsi-meet@0.0.0 types /tmp/lib-jitsi-meet
> npx typescript JitsiMeetJS.js --declaration --allowJs --emitDeclarationOnly --outDir types/generated

npx: installed 1 in 1.639s
modules/transcription/audioRecorder.js:3:1 - error TS9005: Declaration emit for this file requires using private name 'TrackRecorder'. An explicit type annotation may unblock declaration emit.

3 const RecordingResult = require('./recordingResult');
  ~~~~~

modules/transcription/recordingResult.js:13:1 - error TS9005: Declaration emit for this file requires using private name 'RecordingResult'. An explicit type annotation may unblock declaration emit.

13 const RecordingResult = function(blob, name, startTime, wordArray) {
   ~~~~~

modules/transcription/transcriber.js:1:1 - error TS9006: Declaration emit for this file requires using private name 'SphinxService' from module '"/tmp/lib-jitsi-meet/modules/transcription/transcriptionServices/SphinxTranscriptionService"'. An explicit type annotation may unblock declaration emit.

1 const AudioRecorder = require('./audioRecorder');
  ~~~~~

modules/transcription/word.js:7:1 - error TS9005: Declaration emit for this file requires using private name 'Word'. An explicit type annotation may unblock declaration emit.

7 const Word = function(word, begin, end) {
  ~~~~~


Found 4 errors.
```

Is this a result of the incorrect/missing jsdoc comments or am I doing sth wrong?

nightvisi0n:
@garyhuntddn After looking around a bit the biggest "problem" for now are the missing types for `JitsiMeetJS.js`, right?  
I assume the typescript compiler is "confused" by the Object `_mergeNamespaceAndModule` being exported (instead of it's properties).

sapkra:
FYI: I regularly update the types in my fork and already fixed some bugs. You can check it out if you want.

https://github.com/sapkra/lib-jitsi-meet/tree/typescript_type_definitions

saghul:
Integrated in https://github.com/jitsi/lib-jitsi-meet/pull/1682, please see https://github.com/jitsi/lib-jitsi-meet/issues/1683. Let's continue the discussion in either of them. Thanks everyone for your efforts thus far!

PR #1506, opened 2021-02-25
fix(TPC): Remove the existing track instead of overwriting.

jallamsetty1:
When a second remote track of the same mediatype is received for an endpoint, remove the existing track before creating the new remote track.
Ideally, when a remote endpoint changes source, like switching devices, it sends a source-remove (for old ssrc) followed by a source-add (for new ssrc) and Jicofo then should forward these two messages to all the other endpoints in the conference in the same order. However, sometimes, these messages arrive at the client in the reverse order resulting in two remote tracks (of same media type) being created and in case of video, a black strip (that of the first track which has ended) appears over the live track obscuring it. Removing the existing track when that happens will fix this issue.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -884,15 +884,24 @@ TraceablePeerConnection.prototype._createRemoteTrack = function(
     }
 
     if (existingTrack && existingTrack.getTrack() === track) {
-        // Ignore duplicated event which can originate either from
-        // 'onStreamAdded' or 'onTrackAdded'.
+        // Ignore duplicated event which can originate either from 'onStreamAdded' or 'onTrackAdded'.
         logger.info(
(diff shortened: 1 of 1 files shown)


damencho (approved):

paweldomas:
How is this possible exactly given the xmpp is ordered? This means if user sent source-remove followed by source-add then the other endpoints should have received it in the exact same order as well. Unless it gets reorder by jicofo, but that's unlikely given it's processing the packets on ordered queue as well.

damencho:
We have seen this several times, not sure where the reordering is coming from. It happens rarely but still, we see it.

paweldomas:
I'd be curious to see jicofo logs from one of such sessions, because it doesn't seem to be possible - in theory at least, but the must be something if you see this. Anyway this PR is fine, because it still logs an error, but should make things work in case it happens.

PR #1508, opened 2021-02-26
fix(RTC) fix device selection not being available

quitrk:
RTCUtils.enumerateDevices was being called earlier than the RTCUtils.init which assigned that function


saghul (approved):
LGTM but I'd like @hristoterezov to also take a look.

hristoterezov (approved):

PR #1511, opened 2021-03-03
fix(conference): Don't add tracks when client joins audio/video muted.

jallamsetty1:


JitsiConference.js
```diff
@@ -1986,8 +1986,10 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
     // Open a channel with the videobridge.
     this._setBridgeChannel(jingleOffer, jingleSession.peerconnection);
 
-    // Add local tracks to the session
-    const localTracks = this.getLocalTracks();
+    // Do not add the local tracks to pc when client joins audio and video muted.
```
paweldomas, line 1989: what are the reasons for this change and is there a code that adds them back?

jallamsetty1: The intention is not signal the audio/video tracks if the user joins muted until the user unmutes the first time. This will bring down the signaling messages in big calls, especially when everyone joins muted. This also cuts down the sources signaled to each endpoint on join. When user unmutes, jitsi-meet adds those tracks back - https://github.com/jitsi/jitsi-meet/pull/8716

jallamsetty1:
Closing this PR and opening a new PR with the same branch name as that of the jitsi-meet and torture PR so they all get tested together.

PR #1515, opened 2021-03-05
fix(lastN): Return the correct lastN value for the conference.

jallamsetty1:


saghul (dismissed):

jallamsetty1:
@saghul can you pls review this one more time ? Thanks !

damencho (approved):

PR #1516, opened 2021-03-05
feat(dominantSpeaker): Add previous speaker list.

jallamsetty1:
The bridge now sends the list of previous speakers in addition to the dominant speaker in the `DominantSpeakerEndpointChangeEvent` message.

### DIFF
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -292,7 +292,7 @@ export default class BridgeChannel {
             case 'DominantSpeakerEndpointChangeEvent': {
                 const { dominantSpeakerEndpoint, previousSpeakers = [] } = obj;
 
-                logger.info(`New dominant speaker: ${dominantSpeakerEndpoint}.`);
+                logger.debug(`Dominant speaker: ${dominantSpeakerEndpoint}, previous speakers: ${previousSpeakers}`);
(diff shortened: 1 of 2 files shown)


modules/RTC/BridgeChannel.js
```diff
@@ -290,11 +290,10 @@ export default class BridgeChannel {
 
             switch (colibriClass) {
             case 'DominantSpeakerEndpointChangeEvent': {
-                // Endpoint ID from the Videobridge.
-                const dominantSpeakerEndpoint = obj.dominantSpeakerEndpoint;
+                const { dominantSpeakerEndpoint, previousSpeakers = [] } = obj;
 
                 logger.info(`New dominant speaker: ${dominantSpeakerEndpoint}.`);
```
hristoterezov, line 295: I guess it can be useful to add the `previousSpeakers` to the log too. WDYT?

hristoterezov (dismissed):

modules/RTC/BridgeChannel.js
jallamsetty1: Ok, I can make it a debug rather than info since it could be verbose. Is that ok ?

jallamsetty1: @hristoterezov , do you think separate lists for speakers entering/leaving the previous speakers list will be helpful for the application, like we have it for endpoints entering/leaving lastN ?

hristoterezov (approved):

PR #1519, opened 2021-03-08
feat(browser-support): Add support for WKWebview based browsers.

jallamsetty1:
Apple added getUserMedia support for WkWebview based browsers like chrome and Firefox on iOS 14.3. These browsers behave as Safari does on iOS. Therefore, extend the Safari checks to these webkit based browsers as well.


damencho (approved):

jallamsetty1:
Fixes https://github.com/jitsi/lib-jitsi-meet/issues/1480

PR #1521, opened 2021-03-09
Feature/make olm session initialization lazy

tmoldovan8x8:


modules/e2ee/OlmAdapter.js
```diff
@@ -62,39 +62,57 @@ export class OlmAdapter extends Listenable {
...
+        } else {
+            this._sessionInitialization = new Deferred();
+
+            await this._init;
+
+            const promises = [];
+            const localParticipantId = this._conf.myUserId().toString();
+
+            for (const participant of this._conf.getParticipants()) {
+                if (localParticipantId < participant.getId().toString()) {
```
saghul, line 94: No need for the toString()

modules/e2ee/OlmAdapter.js
```diff
@@ -62,39 +62,57 @@ export class OlmAdapter extends Listenable {
...
+                }
+            }
+
+            await Promise.allSettled(promises);
+
+            // TODO: retry failed ones.
+            // TODO: skip participants which don't support E2EE.
+
+            this._sessionInitialization.resolve();
+            this._sessionInitialization = undefined;
```
saghul, line 105: If you set it to undefined you'll keep trying to establish sessions with all participants once a new one joins AFAIS.

modules/e2ee/OlmAdapter.js
```diff
@@ -427,7 +428,38 @@ export class OlmAdapter extends Listenable {
...
+     * when a participant joines an olm session is initialized if participantId is
+     * "smaller" than the localParticipantId.
+     *
+     * @private
+     */
+    async _onParticipantJoined(id, participant) {
+        if (this._conf.isE2EEEnabled()) {
+            await this.initSessions();
+
+            const localParticipantId = this._conf.myUserId().toString();
```
saghul, line 444: No need to toString()

modules/e2ee/OlmAdapter.js
```diff
@@ -427,7 +428,38 @@ export class OlmAdapter extends Listenable {
...
+    }
 
+    /**
+     * Handles a participant joining. Id e2ee is enabled,
+     * when a participant joines an olm session is initialized if participantId is
+     * "smaller" than the localParticipantId.
+     *
+     * @private
+     */
+    async _onParticipantJoined(id, participant) {
```
saghul, line 440: Shouldn't we wait for them to turn on E2EE? It would be done by them, when they receive the e2ee.enabled property for any of the existing participants, but I don't see that code here.

saghul (changes requested):

modules/e2ee/OlmAdapter.js
tmoldovan8x8: I have changed the logic here and onParticipantJoined we just initialise the session with that participant if the condition is met

modules/e2ee/OlmAdapter.js
tmoldovan8x8: We do that on the app side. The joined participant will establish the session with every existing participant that has the id smaller and here we are just initialising the session with that participant if we have the smaller id

modules/e2ee/OlmAdapter.js
```diff
@@ -427,7 +428,39 @@ export class OlmAdapter extends Listenable {
             break;
         }
         }
+    }
 
+    /**
+     * Handles a participant joining. Id e2ee is enabled,
```
saghul, line 434:
```suggestion
     * Handles a participant joining. If e2ee is enabled,
```

modules/e2ee/OlmAdapter.js
saghul: That split doesn't sound right. Maybe we should just add a new config for toggling "auto E2EE".

saghul (commented):
Left some comments, PTAL. Also, the test failures seem legit.

modules/e2ee/E2EEncryption.js
```diff
@@ -130,6 +130,12 @@ export class E2EEncryption {
 
         this._enabled = enabled;
 
+        if (enabled) {
```
saghul, line 133: Let's protect this function from being called multiple times with different values by using some `enabling` Deferred and throwing an "operation in progress" exception. In the UI we will want to disable the control while this function resolves.

modules/e2ee/OlmAdapter.js
```diff
@@ -62,39 +62,58 @@ export class OlmAdapter extends Listenable {
...
     /**
-     * Indicates if olm is supported on the current platform.
-     *
-     * @returns {boolean}
+     * Starts new olm sessions with every other participant that has the participantId "smaller" the localParticipantId.
      */
-    static isSupported() {
-        return typeof window.Olm !== 'undefined';
+    async initSessions() {
+        if (this._sessionInitialization) {
```
saghul, line 84: Replace this check with a throw / assert since it should be impossible for this to happen, and if it does it's a programming error, not a runtime error.

modules/e2ee/OlmAdapter.js
```diff
@@ -62,39 +62,58 @@ export class OlmAdapter extends Listenable {
...
+            for (const participant of this._conf.getParticipants()) {
+                if (localParticipantId < participant.getId()) {
+                    promises.push(this._sendSessionInit(participant));
+                }
+            }
+
+            await Promise.allSettled(promises);
+
+            // TODO: retry failed ones.
+            // TODO: skip participants which don't support E2EE.
```
saghul, line 103: Implement this while we are here please. See participant.getFeatures.

modules/e2ee/OlmAdapter.js
```diff
@@ -155,6 +173,17 @@ export class OlmAdapter extends Listenable {
         return this._keyIndex;
     }
 
+    /**
+     * Updates the current participant key.
+     * @param {Uint8Array|boolean} key - The new key.
+     * @returns {number}
+    */
+    async updateCurrentKey(key) {
```
saghul, line 181: This is not async, can you please remove it while you're here? Also check the use.

modules/e2ee/OlmAdapter.js
```diff
@@ -446,6 +447,46 @@ export class OlmAdapter extends Listenable {
...
+    */
+    async _onParticipantPropertyChanged(participant, name, oldValue, newValue) {
+        switch (name) {
+        case 'e2ee.enabled':
+            if (newValue && this._conf.isE2EEEnabled()) {
+                const localParticipantId = this._conf.myUserId();
+                const participantId = participant.getId();
+
+                if (localParticipantId < participantId) {
+                    await this._init;
```
saghul, line 467: instead, await for the `_sessionInitialization` deferred.

saghul (changes requested):
Left some comments, PTAL!

modules/e2ee/E2EEncryption.js
```diff
@@ -155,10 +169,12 @@ export class E2EEncryption {
...
-        this._olmAdapter.updateKey(this._key).then(index => {
-            // Set our key so we begin encrypting.
-            this._e2eeCtx.setKey(this.conference.myUserId(), this._key, index);
-        });
+        const index = await this._olmAdapter.updateKey(this._key);
+
+        // Set our key so we begin encrypting.
+        this._e2eeCtx.setKey(this.conference.myUserId(), this._key, index);
+
+        this._enabling = undefined;
```
saghul, line 177: I guess you want to resolve it? Otherwise a second call to setEnabled will wait forever.

modules/e2ee/OlmAdapter.js
```diff
@@ -446,6 +448,50 @@ export class OlmAdapter extends Listenable {
...
+    * @param {*} newValue - The property's new value.
+    * @private
+    */
+    async _onParticipantPropertyChanged(participant, name, oldValue, newValue) {
+        switch (name) {
+        case 'e2ee.enabled':
+            if (newValue && this._conf.isE2EEEnabled()) {
+                const localParticipantId = this._conf.myUserId();
+                const participantId = participant.getId();
+
```
saghul, line 466: unnecessary newline?

modules/e2ee/OlmAdapter.js
```diff
@@ -446,6 +448,50 @@ export class OlmAdapter extends Listenable {
...
+                const participantFeatures = await participant.getFeatures();
+
+                if (participantFeatures.has(FEATURE_E2EE) && localParticipantId < participantId) {
+                    if (this._sessionInitialization) {
+                        await this._sessionInitialization;
+                    }
+                    this._sendSessionInit(participant).then(() => {
```
saghul, line 473: use await instead of .then()

saghul (changes requested):
Left some comments, PTAL.

saghul:

modules/e2ee/E2EEncryption.js
```diff
@@ -128,8 +130,20 @@ export class E2EEncryption {
             return;
         }
 
+        if (this._enabling) {
+            await this._enabling;
+        } else {
+            this._enabling = new Deferred();
```
saghul, line 136: You should unconditionally create the Deferred. If we awaited the previous one it means it's resolved, but if *another* call comes in, we want to also await that one.

modules/e2ee/E2EEncryption.js
```diff
@@ -155,10 +169,13 @@ export class E2EEncryption {
...
-            // Set our key so we begin encrypting.
-            this._e2eeCtx.setKey(this.conference.myUserId(), this._key, index);
-        });
+        const index = await this._olmAdapter.updateKey(this._key);
+
+        // Set our key so we begin encrypting.
+        this._e2eeCtx.setKey(this.conference.myUserId(), this._key, index);
+
+        this._enabling.resolve();
+        this._enabling = undefined;
```
saghul, line 178: Don't set this to undefined, the next time we call the function awaiting a resolved deferred is fine.

modules/e2ee/OlmAdapter.js
```diff
@@ -62,39 +62,59 @@ export class OlmAdapter extends Listenable {
...
-     * @returns {boolean}
+     * Starts new olm sessions with every other participant that has the participantId "smaller" the localParticipantId.
      */
-    static isSupported() {
-        return typeof window.Olm !== 'undefined';
+    async initSessions() {
+        if (this._sessionInitialization) {
+            throw new Error('OlmAdapte initSessions called multiple times');
+        } else {
+            this._sessionInitialization = new Deferred();
```
saghul, line 87: Since we are not awaiting it, let's turn it into a boolean.

modules/e2ee/OlmAdapter.js
```diff
@@ -62,39 +62,59 @@ export class OlmAdapter extends Listenable {
...
+                    promises.push(this._sendSessionInit(participant));
+                }
+            }
+
+            await Promise.allSettled(promises);
+
+            // TODO: retry failed ones.
+
+            this._sessionInitialization.resolve();
+            this._sessionInitialization = undefined;
```
saghul, line 107: As mentioned above, let's use a boolean.

saghul (changes requested):
Left some (hopefully) final comments, PTAL, and also look at the failed tests.

modules/e2ee/OlmAdapter.js
tmoldovan8x8: Actually we are awaiting for this in the participantPropertyChanged method.

saghul: Right.

saghul:
Can you please rebase? Looks like I created a conflict, sorry!

PR #1523, opened 2021-03-12
Remove ssrcs on leave

### LINKED ISSUE #1510: degradationPreference property not set correctly in Chromium (and potentially other browsers)
## Description
The degradation preference settings found here: 
https://github.com/jitsi/lib-jitsi-meet/blob/f1ec96678002a88a0a87bb0a6fec1f5d642813d2/modules/RTC/TraceablePeerConnection.js#L29
are not being set correctly in Chromium and thus do not seem to have any effect.

It appears that currently, the property is set in the RTCRtpEnodingParameters:
https://github.com/jitsi/lib-jitsi-meet/blob/f1ec96678002a88a0a87bb0a6fec1f5d642813d2/modules/RTC/TraceablePeerConnection.js#L2138
However, according to the documentation, there is no degradationPreference setting there:
https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpEncodingParameters

Instead, according to the documentation, the degradationPreference property is supposted to be in RTCRtpSendParameters:
https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpSendParameters

This is also where the property is expected in Chromium:
https://github.com/chromium/chromium/blob/1835c63395fa67fe3c7cffe6ac6611133039a0b1/third_party/blink/renderer/modules/peerconnection/rtc_rtp_send_parameters.idl#L16

What I find curious is that degradationPreference is marked as obsolete in RTCRtpSendParameters in the MDN Web Docs linked above. As far as the WebRTC specification is concerned, all I can find is this part of the specification: https://www.w3.org/TR/mst-content-hint/#degradation-preference-when-encoding

I don't know enough about the WebRTC ecosystem to know what the current state of the degradationPreference is supposed to be and will be going forward, all I know is that the current implementation in lib-jitsi-meet does not work with the current implementation in Chromium. I have not checked how and if this is implemented in other browsers.

## Current behavior
Changing the degradation preference does nothing.

## Expected Behavior
Changing the degradation preference leads to either resolution or framerate being maintained.

## Possible Solution
Set the degradationPreference in the RTCRtpSendParameters dictionary instead of RTCRtpEncodingParameters. (I have confirmed that this does indeed resolve the problem, but have not yet checked with other browsers so I am hesitant to create a PR).

## Steps to reproduce
Start a screen share or video stream at a frame rate or resolution respectively that can not be maintained due to CPU/bandwidth constraints. Notice that the frame rate and resolution are lowered respectively. Change the degradation preference from maintain-resolution to maintain-framerate and vice-versa and notice that the behavior does not change.

# Environment details
Tested using Chromium Version 88.0.4324.182 (Official Build) Arch Linux (64-bit)

jallamsetty1:
Remove the ssrcs (associated with remote sources) from the remote desc along with the removal of the remote tracks when an endpoint leaves the call. The source-remove signaling message from Jicofo will no longer be needed in this case and can be dropped by Jicofo.
Properly configure degradation preference on RTCRtpSendParameters instead of RTCRtpEncodingParameters. Fixes #1510.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -1663,23 +1663,26 @@ export default class JingleSessionPC extends JingleSession {
     }
 
     /**
-     * Handles the removal of SSRCs from remote description when an endpoint leaves the call.
+     * Handles the deletion of the remote tracks and SSRCs associated with a remote endpoint.
      *
-     * @param {*} id Endpoint id of the participant that has left the call.
+     * @param {string} id Endpoint id of the participant that has left the call.
+     * @returns {Promise<JitsiRemoteTrack>} Promise that resolves with the tracks that are removed or error if the
+     * operation fails.
      */
     removeRemoteStreamsOnLeave(id) {
-        const removeSsrcInfo = this.peerconnection.getRemoteSourceInfoByParticipant(id);
+        let remoteTracks = [];
 
-        if (!removeSsrcInfo.length) {
-            logger.debug(`No remote SSRCs for participant: ${id} found.`);
-
-            return;
-        }
         const workFunction = finishCallback => {
+            const removeSsrcInfo = this.peerconnection.getRemoteSourceInfoByParticipant(id);
 
+            if (!removeSsrcInfo.length) {
+                logger.debug(`No remote SSRCs for participant: ${id} found.`);
+                finishCallback();
+            }
             const oldLocalSdp = new SDP(this.peerconnection.localDescription.sdp);
             const newRemoteSdp = this._processRemoteRemoveSource(removeSsrcInfo);
 
+            remoteTracks = this.peerconnection.removeRemoteTracks(id);
             this._renegotiate(newRemoteSdp.raw)
                 .then(() => {
                     const newLocalSDP = new SDP(this.peerconnection.localDescription.sdp);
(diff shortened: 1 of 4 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -1662,6 +1662,40 @@ export default class JingleSessionPC extends JingleSession {
         this._addOrRemoveRemoteStream(false /* remove */, elem);
     }
 
+    /**
+     * Handles the removal of SSRCs from remote description when an endpoint leaves the call.
+     *
+     * @param {*} id Endpoint id of the participant that has left the call.
+     */
+    removeRemoteStreamsOnLeave(id) {
+        const removeSsrcInfo = this.peerconnection.getRemoteSourceInfoByParticipant(id);
```
paweldomas, line 1671: This is a little bit problematic, but because of the async queue the sources may not be there when this method is called, but they could be on the queue. I would move this check inside the `workFunction`

jallamsetty1: sounds good, will move it.

jallamsetty1: Also, Jicofo will stop sending the `source-remove` on participant leaving the call so `removeRemoteStream` task will not be queued.

paweldomas (approved):

PR #1527, opened 2021-03-17
Send error message on auth failed with invalid jwt

horymury:
Parse the raw input from xmpp websocket for a failure on auth with jwt and emits the parsed message to consumers.

@saghul @hristoterezov @damencho

### DIFF
--- a/modules/xmpp/XmppConnection.js
+++ b/modules/xmpp/XmppConnection.js
@@ -67,8 +67,8 @@ export default class XmppConnection extends Listenable {
         // The default maxRetries is 5, which is too long.
         this._stropheConn.maxRetries = 3;
 
-        this._lastSuccessTracker = new LastSuccessTracker();
-        this._lastSuccessTracker.startTracking(this, this._stropheConn);
+        this._rawInputTracker = new LastSuccessTracker();
+        this._rawInputTracker.startTracking(this, this._stropheConn);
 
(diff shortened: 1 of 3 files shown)


modules/xmpp/XmppConnection.js
```diff
@@ -351,6 +351,15 @@ export default class XmppConnection extends Listenable {
         return this._lastSuccessTracker.getTimeSinceLastSuccess();
     }
 
+    /**
+     * See {@link LastRequestTracker.getLastFailedMessage}.
+     *
+     * @returns {string|null}
+     */
+    getLastFailedMessage() {
+        return this._lastSuccessTracker.getLastFailedMessage();
```
saghul, line 360: Maybe we can rename the instance variable here, so it doesn't say "success", it feels a bit counter intuitive.

saghul (approved):
Minor comment, LGTM otherwise, nice work! Please remember to adjust the commit message before merging.

modules/xmpp/XmppConnection.js
horymury: > Maybe we can rename the instance variable here, so it doesn't say "success", it feels a bit counter intuitive.

Renamed to _rawInputTracker. Also adjusted the commit message.

saghul (approved):

PR #1530, opened 2021-03-19
feat(e2ee): switch from CTR to GCM

tmoldovan8x8:


modules/e2ee/Context.js
```diff
@@ -161,92 +125,31 @@ export class Context {
...
-            this._sendCount++;
-
-            // Thіs is not encrypted and contains the VP8 payload descriptor or the Opus TOC byte.
-            const frameHeader = new Uint8Array(encodedFrame.data, 0, UNENCRYPTED_BYTES[encodedFrame.type]);
-
-            // Construct frame trailer. Similar to the frame header described in
-            // https://tools.ietf.org/html/draft-omara-sframe-00#section-4.2
-            // but we put it at the end.
-            //                                             0 1 2 3 4 5 6 7
-            // ---------+---------------------------------+-+-+-+-+-+-+-+-+
```
saghul, line 173: Please provide the ASCII diagram for the trailer still.

modules/e2ee/Context.js
```diff
@@ -268,121 +171,38 @@ export class Context {
...
-                // Then set signature bytes to 0.
-                data.set(new Uint8Array(signatureLength), encodedFrame.data.byteLength - (signatureLength + 1));
-            }
-
-            // Set authentication tag bytes to 0.
-            data.set(new Uint8Array(DIGEST_LENGTH[encodedFrame.type]), encodedFrame.data.byteLength
-                - (DIGEST_LENGTH[encodedFrame.type] + counterLength + signatureLength + 1));
-
-            // Do truncated hash comparison of the authentication tag.
-            // If the hash does not match we might have to advance the ratchet a limited number
```
saghul, line 315: We still want to do the ratcheting, but now when decryption fails instead, assuming we have the key.

modules/e2ee/crypto-utils.js
```diff
@@ -11,27 +11,30 @@ export async function deriveKeys(material) {
...
+        name: 'PBKDF2',
         salt: textEncoder.encode('JFrameAuthenticationKey'),
         hash: 'SHA-256',
+        iterations: 100000,
         info
     }, material, {
         name: 'HMAC',
         hash: 'SHA-256'
     }, false, [ 'sign' ]);
     const saltKey = await crypto.subtle.deriveBits({
```
saghul, line 33: I don't think we need the salt key any longer.

modules/e2ee/crypto-utils.js
```diff
@@ -54,9 +57,10 @@ export async function ratchet(material) {
 
     // https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/deriveBits
     return crypto.subtle.deriveBits({
-        name: 'HKDF',
+        name: 'PBKDF2',
```
saghul, line 60: Ditto, this comes from a high entropy source.

modules/e2ee/crypto-utils.js
```diff
@@ -70,5 +74,5 @@ export async function ratchet(material) {
  */
 export async function importKey(keyBytes) {
     // https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto/importKey
-    return crypto.subtle.importKey('raw', keyBytes, 'HKDF', false, [ 'deriveBits', 'deriveKey' ]);
+    return crypto.subtle.importKey('raw', keyBytes, 'PBKDF2', false, [ 'deriveBits', 'deriveKey' ]);
```
saghul, line 77: Ditto.

saghul:
This needs a rebase, now that we removed the signatures.

saghul:
Was the close intentional?

PR #1532, opened 2021-03-22
Feature: Moderator can revoke moderator role to others and himself

BenjaminVega:


JitsiConference.js
```diff
@@ -1470,6 +1470,23 @@ JitsiConference.prototype.grantOwner = function(id) {
...
+ * @param {string} id id of the participant to revoke owner rights to.
+ */
+JitsiConference.prototype.revokeOwner = function(id) {
+    const participant = this.getParticipantById(id);
+    const isMyself = this.myUserId() === id;
+
+    if (isMyself) {
+        this.room.setAffiliation(this.room.myroomjid, 'member');
+    } else if (participant) {
+        this.room.setAffiliation(participant.getJid(), 'member');
```
damencho, line 1485: We use `member` only when lobby is enabled in the rest of the cases it is `none`.

BenjaminVega: Oh, I see. I have pushed a commit to deal with that.
Thanks! 😄

JitsiConference.js
```diff
@@ -1470,6 +1470,24 @@ JitsiConference.prototype.grantOwner = function(id) {
...
 
+/**
+ * Revoke owner rights to the participant or local Participant as
+ * the user might want to refuse to be a moderator.
+ * @param {string} id id of the participant to revoke owner rights to.
+ */
+JitsiConference.prototype.revokeOwner = function(id) {
+    const participant = this.getParticipantById(id);
+    const isMyself = this.myUserId() === id;
+    const role = this.enableLobby() ? 'member' : 'none';
```
damencho, line 1481: This actually enables lobby and returns promise ...

damencho: You should use this.isMembersOnly()

BenjaminVega: fixed

damencho (approved):

damencho:
As a next step, you should add docs for the new method :) here https://github.com/jitsi/handbook/blob/master/docs/dev-guide/ljm-api.md, thank you for your contribution.

PR #1536, opened 2021-03-23
Update presence and skip default values

damencho:
Minimize the size of the initial presence. 

In the future, we can continue optimizing and totally skip the default values instead of setting those, but that needs a bigger refactor. For now it optimizes the initial presence which seems most critical in big conferences.

But overall the other optimization is dropping the namespace which is also some gain to reduce size.

We want to proceed with one cycle of updating environments before merging this, so we can give time to mobile update with the change of having the missing presence tags to be considered as muted (video type missing is already handled and in current mobile clients) as mobile may show some web participants as unmuted while those are actually muted and presence misses the a/v muted tags.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1184,12 +1184,12 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
         }
     }
     if (newTrack.isVideoTrack()) {
-        this.sendCommand('videoType', {
-            value: newTrack.videoType,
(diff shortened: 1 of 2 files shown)


damencho:
We need to update jigasi to send it when starting unmuted before deploying this.

gpolitis (dismissed):

hristoterezov (approved):

PR #1544, opened 2021-03-31
feat(HDAudio): Initial implementation.

muscat1:


modules/RTC/RTCUtils.js
```diff
@@ -420,26 +423,14 @@ function newGetConstraints(um = [], options = {}) {
...
-                { googAutoGainControl: !disableAGC && !disableAP },
-                { googNoiseSuppression: !disableNS && !disableAP },
-                { googHighpassFilter: !disableHPF && !disableAP },
-                { googNoiseSuppression2: !disableNS && !disableAP },
-                { googEchoCancellation2: !disableAEC && !disableAP },
-                { googAutoGainControl2: !disableAGC && !disableAP }
-            );
+            constraints.audio = {
+                autoGainControl: { exact: !disableAGC && !disableAP },
+                channelCount: channelCount,
```
muscat1, line 428: Maybe this should also be exact, what do you guys think?

modules/RTC/RTCUtils.js
```diff
@@ -93,6 +93,9 @@ let disableAGC = false;
 // Disables Highpass Filter
 let disableHPF = false;
 
+// Channel count used to enable stereo.
+let channelCount = null;
```
muscat1, line 97: Should I initialise this with 1, rather than null for the use case when hd audio shouldn't be enabled? (Although I think null might just get ignored)

modules/RTC/RTCUtils.js
```diff
@@ -420,26 +423,14 @@ function newGetConstraints(um = [], options = {}) {
...
-                { googAutoGainControl: !disableAGC && !disableAP },
-                { googNoiseSuppression: !disableNS && !disableAP },
-                { googHighpassFilter: !disableHPF && !disableAP },
-                { googNoiseSuppression2: !disableNS && !disableAP },
-                { googEchoCancellation2: !disableAEC && !disableAP },
-                { googAutoGainControl2: !disableAGC && !disableAP }
-            );
+            constraints.audio = {
+                autoGainControl: { exact: !disableAGC && !disableAP },
+                channelCount,
```
muscat1, line 428: Maybe this should also be exact, what do you guys think?

saghul: I think "ideal" would best. We still want high quality audio even if you have a shitty microphone. We want to allow you to *receive* it too.

modules/RTC/RTCUtils.js
saghul: Please don't use a "global" here.

modules/RTC/RTCUtils.js
```diff
@@ -420,26 +423,14 @@ function newGetConstraints(um = [], options = {}) {
...
+                deviceId: options.micDeviceId,
+                echoCancellation: { exact: !disableAEC && !disableAP },
+                noiseSuppression: { exact: !disableNS && !disableAP },
+                sampleRate: 48000
```
saghul, line 432: This is unnecessary, I asked Google.

modules/RTC/ScreenObtainer.js
```diff
@@ -163,9 +163,19 @@ const ScreenObtainer = {
...
         }
 
+        const { channelCount, disableAGC, disableAP, enableHdAudio } = this.options;
+        const audioProcessingValue = !disableAGC && !disableAP;
+        const audio = enableHdAudio ? {
+            autoGainControl: audioProcessingValue,
+            channelCount,
+            echoCancellation: audioProcessingValue,
+            noiseSuppression: audioProcessingValue,
+            sampleRate: 48000
```
saghul, line 173: Ditto.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,73 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
     return localDescription;
 };
 
+/**
+ * Converts an object to a config-like string.
+ *
+ * @param {Object} obj - The object to be converted into a config string.
+ * @returns {String} - The config string.
+ */
+TraceablePeerConnection.prototype.getConfigFromObject = function(obj) {
```
saghul, line 2117: The name is a bit misleading here. It's related to the fmtp options though.

saghul: It's also only used in a single place, let's inline it.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,73 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
+};
+
+/**
+ * Munges the stereo flag as well as the opusMaxAverageBitrate in the SDP, based
+ * on values set through config.js, if present.
+ *
+ * @param {RTCSessionDescription} description that needs to be munged.
+ * @returns {RTCSessionDescription} the munged description.
+ */
+TraceablePeerConnection.prototype.mungeOpus = function(description) {
```
saghul, line 2134:
```suggestion
TraceablePeerConnection.prototype._mungeOpus = function(description) {
```

saghul (changes requested):
Left some comments, PTAL.

modules/RTC/RTCUtils.js
jallamsetty1: I agree, we should request `ideal` rather than `exact`. Exact would make it a required setting and it would result in a gUM error if it can't be met. Also, if we are getting rid of the old-style of constraints for ChromiumBased clients, why not remove the browser check above and do the same across all browsers ?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,73 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
+    const parsedSdp = transform.parse(description.sdp);
+    const audio = parsedSdp.media.find(mLine => mLine.type === 'audio');
```
jallamsetty1, line 2136: Please note that this will not work for browsers that run Unified plan. On Firefox and Safari, there will be multiple m-lines for type audio, one for each remote audio source and all those m-lines need to be munged, not just the first one.

modules/RTC/RTCUtils.js
```diff
@@ -411,35 +414,15 @@ function newGetConstraints(um = [], options = {}) {
...
-                { googEchoCancellation: !disableAEC && !disableAP },
-                { googAutoGainControl: !disableAGC && !disableAP },
-                { googNoiseSuppression: !disableNS && !disableAP },
-                { googHighpassFilter: !disableHPF && !disableAP },
-                { googNoiseSuppression2: !disableNS && !disableAP },
-                { googEchoCancellation2: !disableAEC && !disableAP },
-                { googAutoGainControl2: !disableAGC && !disableAP }
-            );
+        constraints.audio = {
+            autoGainControl: { exact: !disableAGC && !disableAP },
```
jallamsetty1, line 418: I thought we agreed to specify `ideal` constraints instead of `exact` constraints here. Has that changed ?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,67 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
+ * @param {RTCSessionDescription} description that needs to be munged.
+ * @returns {RTCSessionDescription} the munged description.
+ */
+TraceablePeerConnection.prototype._mungeOpus = function(description) {
+    const parsedSdp = transform.parse(description.sdp);
+    const mLines = parsedSdp.media;
+
+    for (const mLine of mLines) {
+        if (mLine.type === 'audio') {
+            const { payload } = mLine.rtp.find(protocol => protocol.codec === 'opus');
```
jallamsetty1, line 2124: Can you please add mimeType for Opus here https://github.com/jitsi/lib-jitsi-meet/blob/master/service/RTC/CodecMimeType.js and use that for comparison ?

modules/xmpp/JingleSessionPC.js
```diff
@@ -328,6 +328,8 @@ export default class JingleSessionPC extends JingleSession {
         pcOptions.enableInsertableStreams = options.enableInsertableStreams;
         pcOptions.videoQuality = options.videoQuality;
         pcOptions.forceTurnRelay = options.forceTurnRelay;
+        pcOptions.stereo = options.stereo;
```
jallamsetty1, line 331: I have a minor suggestion, will you be able to add these two under a `audioQuality` setting in config.js like how we have video quality settings ?

modules/RTC/RTCUtils.js
muscat1: These were `exact` prior to this PR, I was specifically asking whether `channelCount` should be `exact` or not. I can make these `ideal` as well if that's what we want :D

modules/xmpp/JingleSessionPC.js
muscat1: IIRC, @saghul and I talked about this and said we'd initially add it as hardcoded and then perhaps switch to a configurable object. I have no issue with doing it now, WDYT, Saul?

modules/RTC/TraceablePeerConnection.js
muscat1: Done!

modules/RTC/RTCUtils.js
saghul: Yeah let's make them ideal please. FWIW I don't think it matters much for these options specifically.

modules/RTC/RTCUtils.js
```diff
@@ -93,6 +93,9 @@ let disableAGC = false;
 // Disables Highpass Filter
 let disableHPF = false;
```
saghul, line 94: I think disableHPF is no longer used anywhere, is it? Please get rid off it in that case :-)

modules/RTC/ScreenObtainer.js
```diff
@@ -163,9 +163,18 @@ const ScreenObtainer = {
             getDisplayMedia = navigator.mediaDevices.getDisplayMedia.bind(navigator.mediaDevices);
         }
 
+        const { disableAGC, disableAP, enableHdAudio } = this.options;
+        const audioProcessingValue = !disableAGC && !disableAP;
```
saghul, line 167: Why are you checking for AP and AGC here? If the HD audio option is selected then they need to be false, period.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,67 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
 };
 
+/**
+ * Munges the stereo flag as well as the opusMaxAverageBitrate in the SDP, based
+ * on values set through config.js, if present.
+ *
+ * @param {RTCSessionDescription} description that needs to be munged.
+ * @returns {RTCSessionDescription} the munged description.
+ */
+TraceablePeerConnection.prototype._mungeOpus = function(description) {
```
saghul, line 2118: Maybe first check if stereo or opusMaxAverageBitrate were configured, to avoid the processing otherwise.

modules/xmpp/JingleSessionPC.js
saghul: Yeah, let's group them in audioQuality, me likey!

modules/RTC/ScreenObtainer.js
```diff
@@ -163,9 +163,18 @@ const ScreenObtainer = {
             getDisplayMedia = navigator.mediaDevices.getDisplayMedia.bind(navigator.mediaDevices);
         }
 
+        const { disableAGC, disableAP, enableHdAudio } = this.options;
```
saghul, line 166: `enableHdAudio` is not something LJM should be concerned with, we already pass the individual options from the app.

muscat1: Yeah, was being lazy :P Will base it on `stereo`

saghul (approved):

modules/RTC/ScreenObtainer.js
```diff
@@ -163,9 +163,17 @@ const ScreenObtainer = {
             getDisplayMedia = navigator.mediaDevices.getDisplayMedia.bind(navigator.mediaDevices);
         }
 
+        const { stereo } = this.options;
```
saghul, line 166: read this from audioQuality

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,73 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
+    const { opusMaxAverageBitrate, stereo } = this.options;
```
saghul, line 2119: read these from audioQuality

modules/xmpp/JingleSessionPC.js
```diff
@@ -328,6 +328,8 @@ export default class JingleSessionPC extends JingleSession {
...
+        pcOptions.opusMaxAverageBitrate = options.opusMaxAverageBitrate;
```
saghul, line 332: Use audioQuality

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,73 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
+
+    const parsedSdp = transform.parse(description.sdp);
+    const mLines = parsedSdp.media;
+
+    for (const mLine of mLines) {
+        if (mLine.type === 'audio') {
+            const { payload } = mLine.rtp.find(protocol => protocol.codec === CodecMimeType.OPUS);
+
+            if (!payload) {
+                break;
```
saghul, line 2133: continue

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,73 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
+                sdpChanged = true;
+            }
+
+            if (opusMaxAverageBitrate) {
+                fmtpConfig.opusMaxAverageBitrate = opusMaxAverageBitrate;
+                sdpChanged = true;
+            }
+
+            if (!sdpChanged) {
+                break;
```
saghul, line 2159: continue

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2108,6 +2108,73 @@ TraceablePeerConnection.prototype._adjustLocalMediaDirection = function(
...
+                    config: ''
+                };
+            }
+
+            const fmtpConfig = transform.parseParams(fmtpOpus.config);
+            let sdpChanged = false;
+
+            if (stereo) {
+                fmtpConfig.stereo = 1;
+                sdpChanged = true;
```
saghul, line 2150: you want to reset this for every loop iteration

modules/RTC/TraceablePeerConnection.js
muscat1: derp

modules/RTC/TraceablePeerConnection.js
muscat1: derp

modules/RTC/TraceablePeerConnection.js
muscat1: It was already being initialised with false a few lines above.

saghul (approved):
LGTM, nice work Mihai! @jallamsetty1 can you PTAL too?

jallamsetty1:
> LGTM, nice work Mihai! @jallamsetty1 can you PTAL too?

LGTM to me as well!

jallamsetty1 (approved):

PR #1545, opened 2021-03-31
feat(e2ee) update to SFrame draft -02

saghul:
Drop end to end signature. The impersonation attack vector is deemed out of
scope since insiders are already part of the meeting.


saghul:
Tests do pass, but there seems to be a bug in our Jenkins script.

fippo:
docs need update too. We can change the "S" to "R"(eserved) for now. I think we were looking to increase some key space somewhere, no?

saghul:
Yeah I'll update the diagram. I don't recall the key space thing but I think we can take care of that later. My goal here was to take chance to simplify things so we can move back to GCM easier :-)

fippo (approved):

tmoldovan8x8 (approved):

PR #1550, opened 2021-04-01
fix(codec-selection): Fix codec selection for unified plan browsers.

jallamsetty1:
Make sure the codec order is munged on all the m-lines for unified plan clients. Implement the logic for setting the preferences through RTCRtpTransceiver#setCodecPreferences.
The reason why this wasn't a problem until now is that we don't set any codec other than VP8 as preferred codec for Firefox/Safari. Jicofo in our deployments only offers VP8 and VP9 (more recently), Safari doesn't have support for VP9 and we disabled VP9 on FF because of a known FF bug.
Tested this logic on FF/Safari by modifying Jicofo to offer H.264.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1626,6 +1626,12 @@ TraceablePeerConnection.prototype._mungeCodecOrder = function(description) {
                     type: 'AS',
                     limit: this._isSharingScreen() ? HD_BITRATE : Math.floor(hdBitrate / 1000)
                 } ];
+            } else if (browser.isReactNative()) {
+                // Clear the bandwidth limit in SDP when VP9 is no longer the preferred codec.
+                // This is needed on react native clients only as react-native-webrtc returns the
+                // SDP that the application passed instead of returning the SDP off the native side.
+                // This line automatically gets cleared on web on every renegotiation.
+                mLine.bandwidth = undefined;
             }
         } else if (mLine.type === this.codecPreference.mediaType) {
             SDPUtil.stripCodec(mLine, this.codecPreference.mimeType);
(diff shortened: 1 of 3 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1600,34 +1600,36 @@ TraceablePeerConnection.prototype._mungeCodecOrder = function(description) {
...
+            // we do not want on mobile clients.
+            // Jicofo offers only the baseline code for the jvb connection.
+            // TODO - add check for mobile browsers once js-utils provides that check.
+            if (this.codecPreference.mimeType === CodecMimeType.H264 && browser.isReactNative() && this.isP2P) {
+                SDPUtil.stripCodec(mLine, this.codecPreference.mimeType, true /* high profile */);
+            }
+
+            // Set the max bitrate here on the SDP so that the configured max. bitrate is effective
+            // as soon as the browser switches to VP9.
+            if (this.codecPreference.mimeType === CodecMimeType.VP9) {
```
paweldomas, line 1619: should it also clear the AS line in case VP9 is no longer used?

jallamsetty1: Good catch, yes it should.

jallamsetty1: Actually, the b:AS line doesn't appear automatically after `createOffer` or `createAnswer` and needs to be added back in on every re-negotiate before calling sLD and that's why it doesn't affect the VP9->VP8 bitrate transition on web. However, RN currently just stores the SDP passed the last time and returns it instead of returning what the native side returns. So we need to reset it for RN's sake.

paweldomas: Well okay , but how do you know if any other impl is not retaining the line? I think the app should always set the value it wants rather than relying on specific impl that something is happening or not.

jallamsetty1: Well I don't know if we will support any other webrtc implementation. SDP munging is a hack thats applied over what the webrtc impl produces and all changes that we need are to applied every single time, browser doesn't remember the hacks applied the last time. Having said that I don't mind taking out the browser check.

paweldomas (approved):

PR #1553, opened 2021-04-05
bugfix(e2ee): frees olm session and rotates key when remote participa…

tmoldovan8x8:
…nt disabled e2ee

### DIFF
--- a/modules/e2ee/E2EEncryption.js
+++ b/modules/e2ee/E2EEncryption.js
@@ -9,7 +9,7 @@ import browser from '../browser';
 import Deferred from '../util/Deferred';
 
 import E2EEContext from './E2EEContext';
-import { OlmAdapter } from './OlmAdapter';
+import { OlmAdapter, kOlmData } from './OlmAdapter';
 import { importKey, ratchet } from './crypto-utils';
 
 const logger = getLogger(__filename);
@@ -88,6 +88,8 @@ export class E2EEncryption {
         this._olmAdapter.on(
             OlmAdapter.events.PARTICIPANT_KEY_UPDATED,
             this._onParticipantKeyUpdated.bind(this));
+
+        this._onParticipantE2EEPropertyChanged.bind(this);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


modules/e2ee/E2EEncryption.js
```diff
@@ -263,6 +265,30 @@ export class E2EEncryption {
...
+     * Handles an update in a participant's e2ee.enabled property changed.
+     * If e2ee disabled, frees the olm session and rotates the key.
+     *
+     * @param {JitsiParticipant} participant - The participant.
+     * @param {*} value - The property's previous value.
+     * @private
+     */
+    _onParticipantE2EEPropertyChanged(participant, value) {
+        if (!value && this._enabled) {
+            const olmData = participant[kOlmData] || {};
```
saghul, line 284: Please move this part to some "clearParticipant" function instead of exposing the inner works.

saghul (changes requested):

modules/e2ee/E2EEncryption.js
```diff
@@ -88,6 +88,8 @@ export class E2EEncryption {
         this._olmAdapter.on(
             OlmAdapter.events.PARTICIPANT_KEY_UPDATED,
             this._onParticipantKeyUpdated.bind(this));
+
+        this._onParticipantE2EEPropertyChanged.bind(this);
```
saghul, line 92: Let's inline this function since it's very small and used in a single place please.

modules/e2ee/E2EEncryption.js
```diff
@@ -263,6 +265,25 @@ export class E2EEncryption {
...
+     * Handles an update in a participant's e2ee.enabled property changed.
+     * If e2ee disabled, frees the olm session and rotates the key.
+     *
+     * @param {JitsiParticipant} participant - The participant.
+     * @param {*} value - The property's previous value.
+     * @private
+     */
+    _onParticipantE2EEPropertyChanged(participant, value) {
+        if (!value && this._enabled) {
+            this._olmAdapter.clearParticipantSession(participant);
```
saghul, line 284: I think we should make this return a boolean, indicating that the rotation is necessary.

titusmoldovan: But is there a situation when the rotation is not necessary?

tmoldovan8x8: But is there a situation when the rotation is not necessary?

saghul: It's a bit of a corner case, maybe? I'm thinking a participant who turned e2ee on / off quickly, establishing the session failed. We'd like to avoid rotating too many times I guess. But yeah you can probably skip it.

saghul (approved):
LGTM, love how simple it looks like right now!

PR #1555, opened 2021-04-05
fix(connection-quality): Calculate target bps based on videoQuality settings.

jallamsetty1:
Calculate the target bps based on the video quality settings and the codec configured on the peerconnection.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -702,6 +702,17 @@ TraceablePeerConnection.prototype.getRemoteSourceInfoByParticipant = function(id
 };
 
 /**
+ * Returns the target bitrates configured for the local video source.
+ *
+ * @returns {Object}
+ */
+TraceablePeerConnection.prototype.getTargetBitrates = function() {
+    const currentCodec = this.getConfiguredVideoCodec();
+
+    return this.videoBitrates[currentCodec.toUpperCase()] || this.videoBitrates;
+};
+
+/**
  * Tries to find {@link JitsiTrack} for given SSRC number. It will search both
  * local and remote tracks bound to this instance.
  * @param {number} ssrc
@@ -2321,7 +2332,7 @@ TraceablePeerConnection.prototype.setMaxBitRate = function() {
         }
     } else {
         // Do not change the max bitrate for desktop tracks in non-simulcast mode.
-        let bitrate = this.videoBitrates.high;
+        let bitrate = this.getTargetBitrates()?.high;
 
         if (videoType === VideoType.CAMERA) {
             // Determine the bitrates based on the sender constraint applied for unicast tracks.
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -706,7 +706,7 @@ TraceablePeerConnection.prototype.getRemoteSourceInfoByParticipant = function(id
  *
  * @returns {Object}
  */
-TraceablePeerConnection.prototype.getTargetBitrates = function() {
+TraceablePeerConnection.prototype.getTargetVideoBitrates = function() {
     const currentCodec = this.getConfiguredVideoCodec();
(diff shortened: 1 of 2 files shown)


modules/connectivity/ConnectionQuality.js
```diff
@@ -385,42 +332,22 @@ export default class ConnectionQuality {
         } else {
             // Calculate a value based on the sending bitrate.
 
-            // Figure out if simulcast is in use
+            // Figure out if simulcast is in use.
             const activeTPC = this._conference.getActivePeerConnection();
-            const isSimulcastOn
-                = Boolean(activeTPC && activeTPC.isSimulcastOn());
-
-            const newVideoBitrateCap
```
jallamsetty1, line 393: BandwidthLimiter was removed here and is no longer applicable for target bitrate calculation - https://github.com/jitsi/lib-jitsi-meet/commit/a1b6f1b8b7aa38b5ce7d21bd5bec5eb8d5760467#diff-dae2c7f98ac6a35c23f8c9684b8b0ce1230f101b061e03b98f7b1303d415224e

modules/RTC/TraceablePeerConnection.js
```diff
@@ -701,6 +701,17 @@ TraceablePeerConnection.prototype.getRemoteSourceInfoByParticipant = function(id
     return removeSsrcInfo;
 };
 
+/**
+ * Returns the target bitrates configured for the local video source.
+ *
+ * @returns {Object}
+ */
+TraceablePeerConnection.prototype.getTargetBitrates = function() {
```
saghul, line 709: maybe better call this `getTargetVideoBitrates`, since it's video specific?

modules/connectivity/ConnectionQuality.js
```diff
@@ -256,17 +214,8 @@ export default class ConnectionQuality {
                 this._updateRemoteStats(participant.getId(), payload);
             });
 
-        // Listen to local statistics events originating from the RTC module
-        // and update the _localStats field.
-        // Oh, and by the way, the resolutions of all remote participants are
-        // also piggy-backed in these "local" statistics. It's obvious, really,
```
saghul, line 262: Does the (passive-aggressive) comment no longer apply?

modules/connectivity/ConnectionQuality.js
```diff
@@ -385,42 +332,22 @@ export default class ConnectionQuality {
...
+            // Figure out if simulcast is in use.
             const activeTPC = this._conference.getActivePeerConnection();
-            const isSimulcastOn
-                = Boolean(activeTPC && activeTPC.isSimulcastOn());
-
-            const newVideoBitrateCap
-                = activeTPC && activeTPC.bandwidthLimiter
-                && activeTPC.bandwidthLimiter.getBandwidthLimit('video');
+            const isSimulcastOn = Boolean(activeTPC && activeTPC.isSimulcastOn());
+            const videoQualitySettings = activeTPC.getTargetBitrates();
```
saghul, line 338: What if activeTPC is undefined? You protected yourself from it in the previous line but not here.

saghul (commented):
Left some comments, PTAL!

modules/connectivity/ConnectionQuality.js
jallamsetty1: The comment didn't make sense to me. The local stats are the stats we get from the pc and the remote stats are the ones we get from the remote endpoints through the bridge channel. I don't think we need to keep track of the remote resolution in `_remoteStats` since it is obtained from the local pc `getStats`, that will unnecessarily complicate the logic. WDYT ?

saghul: SGTM!

modules/connectivity/ConnectionQuality.js
jallamsetty1: Good point, will fix it.

saghul (approved):
🚀

PR #1557, opened 2021-04-06
feat(RTC): Signal video type and availability to bridge.

jallamsetty1:


modules/RTC/RTC.js
```diff
@@ -382,6 +385,20 @@ export default class RTC extends Listenable {
...
 
+    /**
+     * Sets the video type and availability for the local video source.
+     *
+     * @param {string} videoType 'camera' for camera, 'desktop' for screenshare and
+     * 'none' for when local video source is muted or removed from the peerconnection.
+     * @returns {void}
+     */
+    setVideoType(videoType) {
+        if (this._videoType !== videoType && this._channel && this._channel.isOpen()) {
```
bgrozev, line 396: Is it okay that calls to `setVideoType` before the channel is open will silently have no effect?

jallamsetty1: Should be fixed now though it makes the client send a redundant message initially on join.  I guess this is necessary for when the bridge channel gets re-initialized during the call because of a network issue.

bgrozev (approved):

PR #1561, opened 2021-04-08
fix: Adds back removed method used by jibri.

damencho:
It was removed by https://github.com/jitsi/lib-jitsi-meet/pull/1512.


modules/xmpp/ChatRoom.js
```diff
@@ -1401,6 +1401,19 @@ export default class ChatRoom extends Listenable {
             errorCallback);
     }
 
+    /**
+     * Adds the key to the presence map, overriding any previous value.
+     * This method is used by jibri.
```
bgrozev, line 1406: Do you want to mark it as deprecated to remind us to remove it once jibri is updated?

damencho: Done

bgrozev (approved):

PR #1564, opened 2021-04-16
fix(stats): Use promise-based getStats on all browsers.

jallamsetty1:
Get rid of the browser specific keys and use the standard spec-compliant fields for stats.
Get the resolution/fps for remote streams from 'inbound-rtp' stats. Use the 'track' stats for the local resolution/fps since these take the active simulcast streams into account.


damencho:
🤯 
I once was there and started doing that, not very successful ... but was for Safari only or something like that... It was a mess :) Congrats on 🔥 the keymaps :)

damencho (approved):

saghul:
❤️❤️❤️🚀🚀🚀

PR #1567, opened 2021-04-19
fix(TPC): Fix error handling for getStats.

jallamsetty1:


paweldomas (approved):

saghul:
Why don't we switch to returning a Promise, like the API we are wrapping? The callbacks are so 2012 😎  and add a (now) unnecessary divergence from the official API.

jallamsetty1:
> Why don't we switch to returning a Promise, like the API we are wrapping? The callbacks are so 2012 😎 and add a (now) unnecessary divergence from the official API.

I agree, makes sense to return a Promise.

jallamsetty1:
> Why don't we switch to returning a Promise, like the API we are wrapping? The callbacks are so 2012 😎 and add a (now) unnecessary divergence from the official API.

@saghul https://github.com/jitsi/lib-jitsi-meet/pull/1585

PR #1569, opened 2021-04-20
JiConOp2

damencho:


modules/xmpp/Caps.js
```diff
@@ -187,28 +187,35 @@ export default class Caps extends Listenable {
...
-                    .each(
-                        (_, el) => identities.add({
-                            type: el.getAttribute('type'),
-                            name: el.getAttribute('name'),
-                            category: el.getAttribute('category')
-                        }));
-                resolve({
-                    features,
-                    identities });
+                resolve(this.parseDiscoInfo(response));
```
bgrozev, line 190: I'm not sure I understand the js well, but does `parseDiscoInfo` need to be defined on the object? It looks like it only operates on its `node` parameter.

modules/xmpp/xmpp.js
```diff
@@ -489,6 +489,20 @@ export default class XMPP extends Listenable {
 
         this.sendDiscoInfo = true;
 
+        if (this.connection._stropheConn && this.connection._stropheConn._addSysHandler) {
+            this.connection._stropheConn._addSysHandler(msg => {
+                this.sendDiscoInfo = false;
```
bgrozev, line 494: Is this a one-time handler? If the XMPP server does not send the expected message, is this going to fire with the next regular `message` stanza we receive?

modules/xmpp/Caps.js
damencho: It is defined in the object and same method is used in another place ... an option is to move it in util folder so can be used from both places ... but we will extract xmpp only and caps only functionality in another place ... so I don't see the point of that.

modules/xmpp/xmpp.js
damencho: this will fire anytime for messages with type conference-info.

modules/xmpp/xmpp.js
```diff
@@ -498,6 +498,15 @@ export default class XMPP extends Listenable {
                 const { features, identities } = this.caps.parseDiscoInfo(msg);
 
                 this._processDiscoInfoIdentities(identities, features);
+
+                // check for shard name in identities
+                if (identities) {
```
bgrozev, line 503: I think this test is redundant (even an empty Set evaluates to `true`), but I don't know if there's a way to express that in js.

damencho: Ups, I will drop it.

modules/xmpp/xmpp.js
bgrozev: Ahh, I missed the other arguments. What makes a message have a type of "conference-info"? It contains a "<conference-info" element?

Nit: we haven't joined a conference yet, is for example "service-info" a more accurate name?

damencho: https://github.com/jitsi/jitsi-meet/pull/9052/commits/1840d4d0bc4bd468c98ee9d7b52d41acba8cd334#diff-51cb31fcf1135ed84b83d4324d1086d258898f786d53cecf56c3d8ab85e07413R28
We set it, ok. Make sense, I will change it.

damencho: Fixed.

modules/xmpp/Caps.js
damencho: I fixed it with the latest commit.

modules/xmpp/xmpp.js
damencho: Done.

bgrozev (approved):

PR #1573, opened 2021-04-22
fix(ReceiverVideoController): Cache the new receiver constraints.

jallamsetty1:


paweldomas:
what problem / bug does it solve?

jallamsetty1:
> what problem / bug does it solve?

Fixes a bug where JitsiConference#getLastN will not return the correct lastN if the client app is using the new receiver constraints - https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3436
This is not being currently used by any of our apps yet, we currently translate the JitsiConference#setLastN and JitsiConference.selectParticipants to use the new receiver constraints but would soon be switching to them hopefully.

hristoterezov (approved):

PR #1577, opened 2021-04-23
fix(RTC) drop the "old gUM" flow

saghul:
Let's f*cking do this for once.

@jallamsetty1 I implemented MST.getSettings here: https://github.com/react-native-webrtc/react-native-webrtc/pull/976 - It just has height, width and frameRate, but I got calls working!

TODO:

- [x] Re-add Electron screen capture code, it was using the old gUM flow just for that O_O
- [x] Decide what to do with the old `config.resolution` setting (should it be a simplified way to set resolution constraints? I'd like that)
- [x] Remember to adjust JM on mobile to use whatever we decide above


modules/RTC/RTC.js
```diff
@@ -223,8 +193,8 @@ export default class RTC extends Listenable {
      * @param {Array<Object>} tracksInfo
      * @returns {Array<JitsiLocalTrack>}
      */
-    static newCreateLocalTracks(tracksInfo) {
-        return _newCreateLocalTracks(tracksInfo);
+    static createLocalTracks(tracksInfo) {
```
jallamsetty1, line 196: Need to call RTC#createLocalTracks here because of the above change - https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/proxyconnection/ProxyConnectionService.js#L279

modules/RTC/ScreenObtainer.js
```diff
@@ -5,7 +5,10 @@ import browser from '../browser';
 
 const logger = require('jitsi-meet-logger').getLogger(__filename);
 
-let gumFunction = null;
+/**
+ * The default frame rate for Screen Sharing.
+ */
+const SS_DEFAULT_FRAME_RATE = 5;
```
jallamsetty1, line 11: The import [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/JingleSessionPC.js#L12) needs to be re-adjusted because of the above change.

modules/RTC/RTCUtils.js
```diff
@@ -377,7 +114,7 @@ function getConstraints(um, options = {}) {
  * @private
  * @returns {Object}
  */
-function newGetConstraints(um = [], options = {}) {
+function getConstraints(um = [], options = {}) {
```
jallamsetty1, line 117:
```
  if (options.resolution) {
            if (Resolutions[options.resolution]) {
                constraints.video.width = {
                    ideal: Resolutions[options.resolution].width
                };
                constraints.video.height = {
                    ideal: Resolutions[options.resolution].height
                };
            }
        }
```
I agree that we should bring back `config.resolution`, and that it should overwrite the video constraints here.

modules/RTC/RTCUtils.js
```diff
@@ -779,10 +360,17 @@ class RTCUtils extends Listenable {
             };
             this.getTrackID = ({ id }) => id;
         } else {
```
jallamsetty1, line 362: Do we want to add a check here for `browser.isSupported()` ? Don't we want to throw the endpoint not webrtc-capable error if its not supported ?

modules/RTC/RTCUtils.js
```diff
@@ -779,10 +360,17 @@ class RTCUtils extends Listenable {
...
-            const message = 'Endpoint does not appear to be WebRTC-capable';
+            this.RTCPeerConnectionType = RTCPeerConnection;
```
jallamsetty1, line 363: Do we want to add a check here for browser.isSupported() ? Don't we want to throw the endpoint not webrtc-capable error if its not supported ?

jallamsetty1:
> Let's f*cking do this for once.
> 
> @jallamsetty1 I implemented MST.getSettings here: [react-native-webrtc/react-native-webrtc#976](https://github.com/react-native-webrtc/react-native-webrtc/pull/976) - It just has height, width and frameRate, but I got calls working!
> 
> TODO:
> 
> * [x]  Re-add Electron screen capture code, it was using the old gUM flow just for that O_O
> * [ ]  Decide what to do with the old `config.resolution` setting (should it be a simplified way to set resolution constraints? I'd like that)
> * [ ]  Remember to adjust JM on mobile to use whatever we decide above

It's high time we finally 🔥 the olg gum flow !! This looks great, I did a first pass and left few comments.

saghul:
Thank you!

modules/RTC/RTC.js
saghul: 👍

modules/RTC/ScreenObtainer.js
saghul: 👍

modules/RTC/RTCUtils.js
saghul: I don't think so. We have a checked in JitsiMeetJS already: isWebRtcSupported, which calls that. I'd say that needs to be called before calling JitsiMeetJS.init. Or check can check within, but not this deep.

saghul:
Got it 💚 ! I tested Electron and RN manually @jallamsetty1. Can you PTAL?

jallamsetty1 (approved):

PR #1578, opened 2021-04-23
fix: Drops the sys message handler after used or connected.

damencho:
Drops the custom type breaking xmpp protocol.

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -278,6 +278,11 @@ export default class XMPP extends Listenable {
 
         this.eventEmitter.emit(XMPPEvents.CONNECTION_STATUS_CHANGED, credentials, status, msg);
         if (status === Strophe.Status.CONNECTED || status === Strophe.Status.ATTACHED) {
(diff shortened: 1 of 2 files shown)


bgrozev (approved):

modules/xmpp/xmpp.js
```diff
@@ -278,6 +278,11 @@ export default class XMPP extends Listenable {
 
         this.eventEmitter.emit(XMPPEvents.CONNECTION_STATUS_CHANGED, credentials, status, msg);
         if (status === Strophe.Status.CONNECTED || status === Strophe.Status.ATTACHED) {
+            // once connected or attached we no longer need this handle, drop it if it exist
+            if (this._sysMessageHandler) {
```
paweldomas, line 282: if it checks for a truthy value here, maybe the it should be cleared? or if it's safe to call it multiple times then do it always/unconditionally?

damencho: Nope, I don't see such check in strophe.

PR #1579, opened 2021-04-23
fix(quality-control): Switch to new receiver constraints by default.

jallamsetty1:
Use the new receiver constraints unless it is explicitly disabled through config.js.


damencho (approved):

jallamsetty1:
Merging since it is a known test failure.

Issue #1089, opened 2020-04-13
5 fps when screen sharing in p2p mode in Chrome


rksm:
## Description
The frame rate when screen sharing gets limited to 5 fps as e.g. described here: https://community.jitsi.org/t/desktop-sharing-frame-rate-config-override-not-working-on-meet-jit-si-fps/20819/3.

This issue does not occur when running the connection over jvb.

This is independent from the constraints passed to getUserMedia or the constrains applied to MediaStreams.

## Possible Solution
Removing `a=x-google-flag:conference` from the SDP that gets patched into via `mungeRemoteDescription` will result in the frame rate not being limited to 5 fps.

## Steps to reproduce
Create a two participant call at https://meet.jit.si/. Share one screen, observe fps count in the stats or chrome://webrtc-internals.

rksm:
Also, disabling simulcast via the server config or with `JitsiMeetJS.init({disableSimulcast: true})` does not change the behavior, the flag is still added.

jallamsetty1:
Hey there, thank you for analyzing and figuring out the solution for this. I will give this a try and see if it fixes the low fps issue with screensharing, thanks !

jallamsetty1:
This has been fixed here - https://github.com/jitsi/lib-jitsi-meet/commit/3794646b5cbe6ceb877acdf0aab999e858d5c630

psi-4ward:
Hi there,
I've run into the 5-FPS thing as well and wondered why Screenshare makes 30 FPS with `disableSimulcast=true`.

`disableSimulcast=false; testing.capScreenshareBitrate=1` brings 5 FPS.

So I think we need to respect capScreenshareBitrate here: https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TraceablePeerConnection.js#L2442 ?

saghul:
I think the patch @jallamsetty1 mentioned was reverted then reworked... and we *just* landed some more fixes in this area yesterday, so I'd recommend re-evaluating with our latest unstable release. If you still see problems please feel free to open a new GH issue.

psi-4ward:
For the docs: https://chromium.googlesource.com/external/webrtc/+/master/media/engine/simulcast.cc#466

psi-4ward:
@saghul: as I see @jallamsetty1 removed the `capScreenshareBitrate` config and derives it from `desktopSharingFrameRate`.

[...]

PR #1583, opened 2021-04-29
Enable unified-plan for chrome

psi-4ward:
@jallamsetty1, @smilingthax

Follow up from: #1089 

As you can see we didn't do much.
This main thing after enabling it was that chrome complains about modifying _read-only parameter_  encodings. Looks like `setMaxBitrate()` does this for us later in the flow so the `videoQuality.maxBitratesVideo{low,standard,high}` from config.js are set on pc-senders.

Also `addStream()` is deprecated so we used `addTrack()` here.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -244,12 +244,19 @@ export class TPCUtils {
             const stream = localTrack.getOriginalStream();
 
             if (stream) {
-                this.pc.peerconnection.addStream(localTrack.getOriginalStream());
-
-                return this.setEncodings(localTrack).then(() => {
+                if (browser.isChromiumBased()) {
+                    // stream.getTracks().forEach(t => this.pc.peerconnection.addTrack(t, stream) );
+                    this.pc.peerconnection.addTrack(localTrack.getTrack(), stream);
                     this.pc.localTracks.set(localTrack.rtcId, localTrack);
                     transceiver.direction = 'sendrecv';
-                });
+                } else {
+                    this.pc.peerconnection.addStream(localTrack.getOriginalStream());
+
+                    return this.setEncodings(localTrack).then(() => {
+                        this.pc.localTracks.set(localTrack.rtcId, localTrack);
+                        transceiver.direction = 'sendrecv';
+                    });
+                }
             }
 
             return Promise.resolve();
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -231,7 +231,7 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean}
      */
     usesUnifiedPlan() {
-        if (this.isFirefox() || this.isWebKitBased()) {
+        if (this.isFirefox() || this.isWebKitBased() || this.isChromiumBased()) {
             return true;
         }
 
@@ -326,7 +326,7 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean}
      */
     supportsSdpSemantics() {
-        return this.isChromiumBased();
+        return false;
     }
 
     /**


psi-4ward:
> Hi, thanks for your contribution!
> If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.

CLA already signed but I think this PR is not ready to merge so far :)
Should give you a hint what works for us for now.

jallamsetty1:
Thanks for the PR but like you said it is not ready for merging yet. I know of few other issues that need to be fixed before switching chrome over to unified plan. Auto p2p is disabled in unified plan and there is a known issue with track replace. I will be working on fixing these in the coming few weeks so that we are ready for the switch. I will also check out whats happening with setting the bitrates, it shouldn't be behaving any differently because as per spec maxBitrate is a member of RTCRtpEncodingParameters.

psi-4ward:
> Auto p2p is disabled in unified plan and there is a known issue with track replace.

We don't use p2p at all.

>  I will also check out whats happening with setting the bitrates, it shouldn't be behaving any differently because as per spec maxBitrate is a member of RTCRtpEncodingParameters.

Yes but probably the call get's invoked to early? 

With 
```javascript
this.pc.peerconnection.addStream(localTrack.getOriginalStream());

return this.setEncodings(localTrack).then(() => {
    this.pc.localTracks.set(localTrack.rtcId, localTrack);
    transceiver.direction = 'sendrecv';
});
```
in [ addTrackUnmute()](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TPCUtils.js#L246)

I get `[modules/xmpp/JingleSessionPC.js] <Object.callback>:  Replace track error on JingleSessionPC[p2p=false,initiator=false,sid=aovpi90qbv98l]: DOMException: Read-only field modified in setParameters().` when I enable the Webcam/Screenshare for the first time. 

The video track does not have any encodings set at this time:
```
// track: MediaStreamTrack {kind: "video", id: "a24c7881-b602-408b-bbd3-a226420c7777", label: "PC Camera (0ac8:303b)", enabled: true, muted: false, …}
APP.conference._room.rtc.peerConnections.get(1).peerconnection.getSenders()[1].getParameters().encodings
[]
```


Not setting the encodings gives still the expected result:
```JSON
[
  {
    "active": true,
    "maxBitrate": 200000,
    "networkPriority": "low",
    "priority": "low"
  },
  {
    "active": true,
    "maxBitrate": 500000,
    "networkPriority": "low",
    "priority": "low"
  },
  {
    "active": true,
    "maxBitrate": 1000000,
    "networkPriority": "low",
    "priority": "low"
  }
]
```

```
    videoQuality: {
        maxBitratesVideo: {
            low: 200000,
            standard: 500000,
            high: 1e6
        },
    },
```

smilingthax:
> The video track does not have any encodings set at this time: [...]

More specifically, chrome does not seem to create the corresponding codecs (populate the `encodings` array) until the SDP exchange (?). Especially: It is not there yet when (e.g.) the `.addTrack` promise resolves. But before these array entries exist, chrome will not allow `.setParameters` to set/modify them.

jallamsetty1:
@psi-4ward , @smilingthax,
Thanks for your inputs. Unfortunately we won't be able to take this PR since we want to switch Chrome to Unified plan only when we have verified that all the bugs with Unified plan impl have been fixed. Additionally, we would want to put this behind a config.js flag and also leave react native to operate in plan-b mode. Therefore, I will have to close this PR. Thanks for understanding. I have fixed the above mentioned addTrackUnmute() issue [here](https://github.com/jitsi/lib-jitsi-meet/pull/1596).

psi-4ward:
No problem. The intention was to submit our changes and our good experiences with it. But would be nice to have a track on your open issues with unifid plan.

I had a mixes meeting with about 10 ppl using this changes including Ios and Android apps. No problems so far.

PR #1585, opened 2021-04-29
fix(stats): Return promise for getStats.

jallamsetty1:
Switch to returning a Promise for getStats. Reset frame rate stat to 0 when video is suspended as a result of endpoint falling out of last-n.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -355,9 +355,8 @@ export default function TraceablePeerConnection(
 
     if (this.maxstats) {
         this.statsinterval = window.setInterval(() => {
-            this.getStats(stats => {
-                if (stats.result
-                    && typeof stats.result === 'function') {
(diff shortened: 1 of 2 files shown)


saghul (approved):
❤️ 🚀

modules/RTC/TraceablePeerConnection.js
```diff
@@ -355,9 +355,8 @@ export default function TraceablePeerConnection(
 
     if (this.maxstats) {
         this.statsinterval = window.setInterval(() => {
-            this.getStats(stats => {
-                if (stats.result
-                    && typeof stats.result === 'function') {
+            this.getStats().then(stats => {
+                if (typeof stats?.result === 'function') {
```
saghul, line 359: Question: when is this the case? I thought this was with ancient stats?

jallamsetty1: Nope, these are being used by torture testcases.

PR #1587, opened 2021-04-30
Show how to load correct server config directly

jbemmel:


damencho:
We don't want to tie this to particular deployment... its an example to put on your server ...

saghul:
Same as the previous PR. We could just change the target server in the example options though.

PR #1588, opened 2021-04-30
Use 'config' directly from beta server

jbemmel:
Example code uses outdated, broken config options

### DIFF
--- a/doc/example/example.js
+++ b/doc/example/example.js
@@ -1,5 +1,10 @@
 /* global $, JitsiMeetJS */
 
+// Use 'config' options loaded from beta jitsi server
+const options = config;
+
+/*
+ * These options are slightly outdated, and result in XMPP connection errors
 const options = {
     hosts: {
         domain: 'jitsi-meet.example.com',
@@ -10,6 +15,7 @@ const options = {
     // The name of client node advertised in XEP-0115 'c' stanza
     clientNode: 'http://jitsi.org/jitsimeet'
 };
+*/
 
 const confOptions = {
     openBridgeChannel: true


damencho:
What's the purpose of this?

saghul:
Jitsi Meet and lib-jitsi-meet share options, but this is not a great example for others because they need not have a full Jitsi Meet installation. When using the low level API configuration should be explicit.

jbemmel:
The options evolve over time, and the example is currently broken. By
tying it to the beta server there is a better chance it keeps working, if
you prefer not to do that we could also elaborate in a comment where to get
an example of the latest config options, such that users can troubleshoot
themselves

I lost quite a bit of time figuring out this basic misalignment, and there
are a few examples in the forums of users struggling with the same

Regards,
Jeroen

saghul:
The example is not broken, the server at meet.jit.si is currently rejecting those requests, which is different. The current example should just work if you change the server to point to beta. That would be an acceptable change. The example contains barely any config, precisely for that reason.

damencho:
I'm not sure when beta will start rejecting that as on meet (if not already doing it) ...

jbemmel:
[image: image.png]

It only starts working when I copy some of the settings in
https://beta.meet.jit.si/config.js, websocket value is the same but it
seems that some other flags are required too

// Copied from https://beta.meet.jit.si/config.js
const options =
{
hosts: {
domain: 'beta.meet.jit.si',
muc: 'conference.beta.meet.jit.si', // FIXME: use XEP-0030
focus: 'focus.beta.meet.jit.si',
},
disableSimulcast: false,
resolution: 720,
constraints: {
video: {
height: {
ideal: 720,
max: 720,
min: 180
},
width: {
ideal: 1280,
max: 1280,
min: 320
}
}
},
useStunTurn: true, // use XEP-0215 to fetch TURN servers for the JVB
connection
useTurnUdp: true,
bosh: '//beta.meet.jit.si/http-bind', // FIXME: use xep-0156 for that
websocket: 'wss://beta.meet.jit.si/xmpp-websocket', // FIXME: use xep-0156
for that
websocketKeepAliveUrl: 'https://beta.meet.jit.si/_unlock',
// The name of client node advertised in XEP-0115 'c' stanza
clientNode: 'http://jitsi.org/jitsi-at-scale-agg',
enableP2P: false, // flag to control P2P connections
channelLastN: 20, // The default value of the channel attribute last-n.
lastNLimits: {
5: 20,
30: 15,
50: 10,
70: 5,
90: 2,
800: 1 // JvB: Added for ***@***.***
},
};

PR #1591, opened 2021-05-02
Jitsi@scale Hackathon submission: Enable HTML Media Element streaming

jbemmel:
This patch adds support for a new virtual device type 'htmlmedia', to enable streaming from 1 or more \<audio\>,\<video\> or \<canvas\> elements on the page

```
JitsiMeetJS.createLocalTracks({ devices: [ 'htmlmedia' ], 
                                htmlMediaElements: [ document.querySelector('canvas') ],
                                htmlMediaFrameRate: 10 }) 
```

This allows us to (for example) create a web application that aggregates multiple video streams onto a \<canvas\> element, and sends the aggregate stream back into the conference - to achieve better scale

Note: Only modules/RTC/RTCUtils.js should be part of this PR, the other files got added erroneously

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -1221,7 +1221,8 @@ class RTCUtils extends Listenable {
                    new Error("'htmlmedia' requested but no htmlMediaElements"));
              }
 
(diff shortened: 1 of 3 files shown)


jbemmel:
Abandoned - will submit a clean PR shortly

PR #1594, opened 2021-05-03
Fix: Screenshare with Audio broken

psi-4ward:
On Screenshare with Audio the constraint is 

```js
    "audio": {
        "optional": {},
        "mandatory": {
            "chromeMediaSource": "desktop"
        }
    },
```

Electron complains with `Failed to execute 'getUserMedia' on 'MediaDevices': The object must have a callable @@iterator property.` cause of empty optional object.

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -105,7 +105,7 @@ const ScreenObtainer = {
             channelCount: 2,
             echoCancellation: false,
             noiseSuppression: false
-        } : false;
+        } : true;
 
         return audio;
     },
@@ -138,11 +138,11 @@ const ScreenObtainer = {
                         let audioConstraints = false;
 
                         if (screenShareAudio) {
-                            audioConstraints = { };
+                            audioConstraints = {};
 
-                            if (this._getAudioConstraints()) {
-                                audioConstraints.optional = {
-                                    ...this._getAudioConstraints()
+                            if (typeof this._getAudioConstraints() !== 'boolean') {
+                                audioConstraints = {
+                                    optional: this._getAudioConstraints()
                                 };
                             }
 
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -105,7 +105,7 @@ const ScreenObtainer = {
             channelCount: 2,
             echoCancellation: false,
             noiseSuppression: false
-        } : true;
+        } : false;
 
         return audio;
     },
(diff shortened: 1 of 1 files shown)


psi-4ward:
CLA already signed :)

modules/RTC/ScreenObtainer.js
```diff
@@ -105,7 +105,7 @@ const ScreenObtainer = {
             channelCount: 2,
             echoCancellation: false,
             noiseSuppression: false
-        } : true;
```
saghul, line 108: This change looka bogus. It would break the non-electron case. In Electron's case we are only calling this function when we know audio is going to be shared.

modules/RTC/ScreenObtainer.js
```diff
@@ -138,11 +138,13 @@ const ScreenObtainer = {
                         let audioConstraints = false;
 
                         if (screenShareAudio) {
-                            audioConstraints = {
-                                optional: {
+                            audioConstraints = { };
+
+                            if (this._getAudioConstraints()) {
```
saghul, line 143: Get the result of this and check with typeof if it's boolean and if so, don't destructure it.

saghul (changes requested):
Left some comments, PTAL.

modules/RTC/ScreenObtainer.js
```diff
@@ -138,11 +138,11 @@ const ScreenObtainer = {
                         let audioConstraints = false;
 
                         if (screenShareAudio) {
-                            audioConstraints = { };
+                            audioConstraints = {};
 
-                            if (this._getAudioConstraints()) {
-                                audioConstraints.optional = {
-                                    ...this._getAudioConstraints()
+                            if (typeof this._getAudioConstraints() !== 'boolean') {
```
saghul, line 143: Can you please avoid calling the function twice? Take the value and then work with it please.

saghul (approved):

saghul:

jallamsetty1:
Merging since its a known unrelated failure.

psi-4ward (commented):
> Merging since its a known failure that is unrelated.

Don't understand what you mean with _unrelated_ but thanks.

jallamsetty1:
> > Merging since its a known failure that is unrelated.
> 
> Don't understand what you mean with _unrelated_ but thanks.

My bad, I meant the torture test failure is unrelated to your change and it is a known issue that has been failing consistently for the past week or so. Thats why I didn't wait for the tests to be all green and went ahead with the merge.

PR #1596, opened 2021-05-03
fix(TPC): Use addTrack instead of addStream in Unified-plan impl.

jallamsetty1:
Do not try to configure encodings on the sender until they are available. This fixes an issue on Chrome (running in unified-plan mode) where unmuting the local source throws a 'Read-only field modified in setParameters()' error.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -191,12 +191,12 @@ export class TPCUtils {
     /**
     * Adds {@link JitsiLocalTrack} to the WebRTC peerconnection for the first time.
     * @param {JitsiLocalTrack} track - track to be added to the peerconnection.
-    * @param {boolean} isInitiator - boolean that indicates if the endpoint is offerer
-    * in a p2p connection.
-    * @returns {void}
+    * @param {boolean} isInitiator - boolean that indicates if the endpoint is offerer in a p2p connection.
+    * @returns {Promise<void>} - resolved when done, rejected otherwise.
     */
     addTrack(localTrack, isInitiator) {
         const track = localTrack.getTrack();
+        let promise = Promise.resolve();
 
         if (isInitiator) {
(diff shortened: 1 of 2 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -395,6 +412,12 @@ export class TPCUtils {
             .find(t => t.sender && t.sender.track && t.sender.track.kind === track.getType());
         const parameters = transceiver.sender.getParameters();
 
+        // Resolve if the encodings are not available yet. This happens immediately after the track is added to the
+        // peerconnection on chrome in unified-plan. It is ok to ignore and not report the error here since the
+        // action that triggers 'addTrack' (like unmute) will also configure the encodings and set bitrates after that.
+        if (!parameters?.encodings?.length) {
```
paweldomas, line 418: 👍

modules/RTC/TPCUtils.js
```diff
@@ -191,12 +191,12 @@ export class TPCUtils {
     /**
     * Adds {@link JitsiLocalTrack} to the WebRTC peerconnection for the first time.
     * @param {JitsiLocalTrack} track - track to be added to the peerconnection.
-    * @param {boolean} isInitiator - boolean that indicates if the endpoint is offerer
-    * in a p2p connection.
-    * @returns {void}
+    * @param {boolean} isInitiator - boolean that indicates if the endpoint is offerer in a p2p connection.
+    * @returns {Promise<void>} - resolved when done, rejected otherwise.
```
paweldomas, line 195: This code is not really async. Note that you're returning either resolve or reject immediately. Is that necessary?

paweldomas (changes requested):

modules/RTC/TPCUtils.js
jallamsetty1: Yeah, I didn't have to return a promise here. I am catching the error in TPC#addTrack now.

paweldomas (approved):

PR #1597, opened 2021-05-03
fix(conference): Enable p2p for unified plan clients.

jallamsetty1:


damencho (approved):

saghul:
So it Just Works (TM) now? Kewl!

jallamsetty1:
> So it Just Works (TM) now? Kewl!

Well, isn't it amazing when you think something is broken but its fixed already :) Yeah its working fine now. Remove p2p tracks and adding the jvb tracks and vice versa was what was broken and that seems to have been fixed now. There was some work done late last summer in that area.

PR #1598, opened 2021-05-04
Add dependency for promise.allSettled.

jallamsetty1:
Older chrome versions like M72 do not support Promise.allSettled.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1896,6 +1896,119 @@
         }
       }
     },
+    "array.prototype.map": {
+      "version": "1.0.3",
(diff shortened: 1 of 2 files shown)


damencho (approved):

saghul:
Doesn't it need to be imported so the polyfill works?

jallamsetty1:
> Doesn't it need to be imported so the polyfill works?

🤦‍♀️ I didn't have a repro case ... promise.allSettled is used in a bunch of different files, should I import it in JitsiMeetJS.js or is there a better alternative ?

saghul:
Yes, I'd say import it at the top level of the entry point, that should be enough.

Also, I guess we can remove the dependency from Jitsi Meet now?

PR #1600, opened 2021-05-05
fix: Drops unused nick from messages. Fixes #1592

### LINKED ISSUE #1592: JitsiConference.sendMessage() throws error when trying to send custom 'json-message'
## Description
When I try to send an object through the method `sendMessage()` of the class `JitsiConference` I receive an error.
The error also appears to be present at the current `meet.jit.si` deployment.

```
lib-jitsi-meet.min.js:1 Uncaught TypeError: Cannot read property 'appendChild' of null
    at Builder.c (lib-jitsi-meet.min.js:1)
    at E.sendMessage (lib-jitsi-meet.min.js:10)
    at ie.sendTextMessage (lib-jitsi-meet.min.js:10)
    at ie.sendMessage (lib-jitsi-meet.min.js:10)
    at <anonymous>:1:6
```
The documentation hints that this should be possible:

```
// JitsiConference.js:2636
/**
 * Sends a message to a given endpoint (if 'to' is a non-empty string), or
 * broadcasts it to all endpoints in the conference.
 * @param {string} to The ID of the endpoint/participant which is to receive
 * the message, or '' to broadcast the message to all endpoints in the
 * conference.
 * @param {string|object} message the message to send. If this is of type
 * 'string' it will be sent as a chat message. If it is of type 'object', it
 * will be encapsulated in a format recognized by jitsi-meet and converted to
 * JSON before being sent.
 * @param {boolean} sendThroughVideobridge Whether to send the message through
 * jitsi-videobridge (via the COLIBRI data channel or web socket), or through
 * the XMPP MUC. Currently only objects can be sent through jitsi-videobridge.
 */
JitsiConference.prototype.sendMessage = function(
        message,
        to = '',
        sendThroughVideobridge = false) { ...
```

## Current behavior
- The payload is correctly converted into a JSON string and the `elementType` is set to `json-element`
- The `Strophe.Builder` throws an error and the payload is not sent.

## Expected Behavior
- The payload is correctly converted into a JSON string and the `elementType` is set to `json-element`
- The `Strophe.Builder` throws no error and the payload is sent.

## Steps to reproduce

1. goto https://meet.jit.si/
2. start a random conference and join the room
3. open the developer console
4. send a message through the conference

`window.APP.store.getState()["features/base/conference"].conference.sendMessage({foo: "bar"})`

# Environment details
Chrome, importing `lib-jitsi-meet` from self-hosted server.

damencho:


modules/xmpp/ChatRoom.js
```diff
@@ -1113,10 +1095,10 @@ export default class ChatRoom extends Listenable {
         if (txt) {
             if (type === 'chat') {
                 this.eventEmitter.emit(XMPPEvents.PRIVATE_MESSAGE_RECEIVED,
-                        from, nick, txt, this.myroomjid, stamp);
+                        from, txt, this.myroomjid, stamp);
```
saghul, line 1098: Won't this require changes in Jitsi Meet?

saghul (approved):
LGTM with a question.

modules/xmpp/ChatRoom.js
damencho: Nope these are the xmpp events internal to the lib and the change is done

saghul: Great!

PR #1602, opened 2021-05-05
fix(JingleSession): Add a unique identifier for source on Firefox.

jallamsetty1:
In certain cases, when RTCPeerconnection#addTrack is used, the browser doesn't produce the streamid in the 'a=msid' line which is being used by Jicofo as a unique identifier for the associated ssrc. Whenever the msid is missing in the local description, generate an ID using the local endpoint id.

### DIFF
--- a/modules/sdp/SDP.js
+++ b/modules/sdp/SDP.js
@@ -118,7 +118,7 @@ SDP.prototype.containsSSRC = function(ssrc) {
 };
 
 // add content's to a jingle element
-SDP.prototype.toJingle = function(elem, thecreator, localId) {
+SDP.prototype.toJingle = function(elem, thecreator, localEndpointId) {
     // https://xmpp.org/extensions/xep-0338.html
     SDPUtil.findLines(this.session, 'a=group:').forEach(line => {
         const parts = line.split(' ');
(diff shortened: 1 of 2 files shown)


damencho (dismissed):

modules/sdp/SDP.js
```diff
@@ -118,7 +118,7 @@ SDP.prototype.containsSSRC = function(ssrc) {
 };
 
 // add content's to a jingle element
-SDP.prototype.toJingle = function(elem, thecreator) {
+SDP.prototype.toJingle = function(elem, thecreator, localId) {
```
paweldomas, line 121: can you call it `localEndpointId`?

damencho (approved):

paweldomas (approved):
Approving because it's urgent, but it would better fit here next to other similar transformations:
https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TraceablePeerConnection.js#L1506

jallamsetty1:
> Approving because it's urgent, but it would better fit here next to other similar transformations:
> https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TraceablePeerConnection.js#L1506

Like discussed, this may not be great from a performance standpoint since these transformations will be done every time we fetch the local desc which the client is doing now periodically for figuring out connection quality and stuff. We will be eventually needing fewer of these transformations once we switch to Unified plan completely.

PR #1611, opened 2021-05-11
fix(JingleSession): Move the ssrc identifier generation to LocalSdpMunger.

jallamsetty1:
When a msid attribute is missing in the 'a=ssrc' line, use the local endpoint id as an indentifier. Move this generation logic to LocalSdpMunger. Also suppress notifying Jicfo of a ssrc change on Firefox when the change is a result of the media being suspended on the jvb connection.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -273,7 +273,7 @@ export default function TraceablePeerConnection(
      * sending SSRC updates on attach/detach and mute/unmute (for video).
      * @type {LocalSdpMunger}
      */
-    this.localSdpMunger = new LocalSdpMunger(this);
+    this.localSdpMunger = new LocalSdpMunger(this, this.rtc.getLocalEndpointId());
 
     /**
      * TracablePeerConnection uses RTC's eventEmitter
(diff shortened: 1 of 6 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2901,9 +2901,7 @@ TraceablePeerConnection.prototype._processLocalSSRCsMap = function(ssrcMap) {
 
             // eslint-disable-next-line no-negated-condition
             if (newSSRCNum !== oldSSRCNum) {
-                if (!oldSSRCNum) {
-                    logger.error(`Overwriting SSRC for ${track} ${trackMSID} in ${this} with: `, newSSRC);
-                }
+                oldSSRCNum && logger.error(`Overwriting SSRC for ${track} ${trackMSID} in ${this} with: `, newSSRC);
```
jallamsetty1, line 2904: Fixing a mistake in the previous commit, too lazy to open another PR for this :)

modules/sdp/LocalSdpMunger.js
```diff
@@ -175,11 +177,14 @@ export default class LocalSdpMunger {
...
 
                     if (streamAndTrackIDs.length === 2) {
-                        const streamId = streamAndTrackIDs[0];
+                        let streamId = streamAndTrackIDs[0];
                         const trackId = streamAndTrackIDs[1];
 
-                        ssrcLine.value
-                            = `${streamId}-${pcId} ${trackId}-${pcId}`;
+                        // eslint-disable-next-line max-depth
+                        if (streamId === '-') {
```
paweldomas, line 184: can you put the old comment here which describes the firefox case?

paweldomas: also the old code which was creating id which was media specific was better than just the endpoint id

paweldomas: I don't know about the lipsync logic, but it's been broken anyway maybe there's a better way of doing it

paweldomas (approved):

PR #1614, opened 2021-05-13
fix(TPC): Return default codec if the local sdp is not available.

jallamsetty1:
Get the correct media type when generating the source identifier.


paweldomas (approved):
Is it possible it may not work as expected because of the timing issue? I feel like it should try again later once the local sdp becomes available?

paweldomas:
Anyway this fixes worse bug so lets land it first

PR #1616, opened 2021-05-17
destroys olm session on disabling e2ee

tmoldovan8x8:


modules/e2ee/Context.js
```diff
@@ -266,16 +267,6 @@ export class Context {
             }
 
             // TODO: notify the application about error status.
-
-            // TODO: For video we need a better strategy since we do not want to based any
-            // non-error frames on a garbage keyframe.
-            if (encodedFrame.type === undefined) { // audio, replace with silence.
```
saghul, line 272: Why are we removing this? It would nice to keep it. Can you check why it's no longer working?

saghul (approved):

Issue #1612, opened 2021-05-12
Safari: RTCRtpTransceiver for video not found


bukharin:
## Description

Safari:

If application mutes local track and then try to remove it - an error occurs:

`RTCRtpTransceiver for video not found`

``` javascript
await localTrack.mute();
await jitsiConference.removeTrack(localTrack); // error RTCRtpTransceiver for video not found
```

## Expected Behavior

Should remove track without error

## Possible Solution

As I know `lib-jitsi-meet` in Safari removes track from `Transceiver` when app mutes it. May be we should not reject operation if Transceiver not found [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TPCUtils.js#L302) ?

# Environment details

Tested in Safari >14.1

PR #1618, opened 2021-05-17
fix(RTC): Set transceiver direction after RTCRtpSender#replaceTrack.

jallamsetty1:
This fixes the issue where TRACK_REMOVED event is not fired when a remote track is removed from the peerconnection.
Fixes https://github.com/jitsi/lib-jitsi-meet/issues/1612 and https://github.com/jitsi/jitsi-meet/issues/8482.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -2,7 +2,6 @@ import { getLogger } from 'jitsi-meet-logger';
 import transform from 'sdp-transform';
 
 import * as MediaType from '../../service/RTC/MediaType';
-import RTCEvents from '../../service/RTC/RTCEvents';
 import browser from '../browser';
 
 const logger = getLogger(__filename);
@@ -316,7 +315,7 @@ export class TPCUtils {
     replaceTrack(oldTrack, newTrack) {
         if (oldTrack && newTrack) {
             const mediaType = newTrack.getType();
-            const stream = newTrack.stream;
+            const stream = newTrack.getOriginalStream();
 
             // Ignore cases when the track is replaced while the device is in a muted state,like
             // replacing camera when video muted or replacing mic when audio muted. These JitsiLocalTracks
@@ -332,7 +331,7 @@ export class TPCUtils {
             const transceiver = oldTrack.isVideoTrack() && browser.doesVideoMuteByStreamRemove && oldTrack.isMuted()
                 ? this._findTransceiver(mediaType)
                 : this._findTransceiver(mediaType, oldTrack);
-            const track = mediaType === MediaType.AUDIO ? stream.getAudioTracks()[0] : stream.getVideoTracks()[0];
+            const track = newTrack.getTrack();
 
             if (!transceiver) {
                 return Promise.reject(new Error('replace track failed'));
@@ -350,9 +349,6 @@ export class TPCUtils {
 
                     this.pc._addedStreams.push(stream);
                     this.pc.localSSRCs.set(newTrack.rtcId, ssrc);
-                    this.pc.eventEmitter.emit(RTCEvents.LOCAL_TRACK_SSRC_UPDATED,
-                        newTrack,
-                        this.pc._extractPrimarySSRC(ssrc));
                 });
         } else if (oldTrack && !newTrack) {
             return this.removeTrackMute(oldTrack)
@@ -371,8 +367,6 @@ export class TPCUtils {
                     this.pc.localSSRCs.delete(oldTrack.rtcId);
                 });
         } else if (newTrack && !oldTrack) {
-            const ssrc = this.pc.localSSRCs.get(newTrack.rtcId);
-
             return this.addTrackUnmute(newTrack)
                 .then(() => this.setEncodings(newTrack))
                 .then(() => {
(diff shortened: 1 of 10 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -316,7 +316,7 @@ export class TPCUtils {
     replaceTrack(oldTrack, newTrack) {
         if (oldTrack && newTrack) {
             const mediaType = newTrack.getType();
-            const stream = newTrack.getOriginalStream();
+            const stream = newTrack.stream;
```
paweldomas, line 319: `.stream` is not really public and should use `getOriginalStream` getter

modules/RTC/TPCUtils.js
```diff
@@ -328,11 +328,11 @@ export class TPCUtils {
...
-            const track = mediaType === MediaType.AUDIO
-                ? stream.getAudioTracks()[0]
-                : stream.getVideoTracks()[0];
-            const transceiver = this.pc.peerconnection.getTransceivers()
-                .find(t => t.receiver.track.kind === mediaType && !t.stopped);
+
+            const transceiver = oldTrack.isVideoTrack() && browser.doesVideoMuteByStreamRemove && oldTrack.isMuted()
+                ? this._findTransceiver(mediaType)
+                : this._findTransceiver(mediaType, oldTrack);
+            const track = mediaType === MediaType.AUDIO ? stream.getAudioTracks()[0] : stream.getVideoTracks()[0];
```
paweldomas, line 335: why do you need to fetch the `track` this way? shouldn't it be accessible from the `newTrack` via `getTrack()` getter?
Btw. naming of jitsi tracks and webrtc media stream/tracks is very confusing. Maybe it would help to rename the arguments to `oldJitsiTrack` and `newJitsiTrack`.

modules/RTC/TPCUtils.js
```diff
@@ -357,14 +357,35 @@ export class TPCUtils {
...
+
+                    // Remove the old track from the list of local tracks.
                     this.pc.localTracks.delete(oldTrack.rtcId);
                     this.pc.localSSRCs.delete(oldTrack.rtcId);
                 });
         } else if (newTrack && !oldTrack) {
             const ssrc = this.pc.localSSRCs.get(newTrack.rtcId);
 
             return this.addTrackUnmute(newTrack)
+                .then(() => this.setEncodings(newTrack))
```
paweldomas, line 377: is `setEncodings` needed for the `if (oldTrack && newTrack)` case as well?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -315,14 +315,15 @@ export default function TraceablePeerConnection(
         this.peerconnection.onremovestream
             = event => this._remoteStreamRemoved(event.stream);
     } else {
-        this.peerconnection.ontrack = event => {
-            const stream = event.streams[0];
+        this.onTrack = evt => {
```
paweldomas, line 318: why this change?

modules/sdp/LocalSdpMunger.js
```diff
@@ -189,15 +190,22 @@ export default class LocalSdpMunger {
...
             }
         }
+
+        // If the msid attribute is missing, then remove the ssrc from the transformed description so that a
+        // source-remove is signaled to Jicofo. This happens when the direction of the transceiver (or m-line)
+        // is set to 'inactive' or 'recvonly' on Firefox.
+        const msid = mediaSection.ssrcs.find(s => s.attribute === 'msid');
+
+        if (!msid && browser.isFirefox()) {
+            mediaSection.ssrcs = undefined;
```
paweldomas, line 207: it would be really great to have unit tests for corner cases like this one

modules/RTC/TPCUtils.js
jallamsetty1: This wasn't an intentional change, was debugging something and forgot to take it out. Thanks for pointing it out.

modules/RTC/TPCUtils.js
jallamsetty1: It's not needed if there was a source already and only the stream is being replaced on the sender. However, if the client starts muted and the source is being added on unmute, then the encodings need to be configured.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: To avoid running into issues when something else like callstats/rtc stats accidentally overrides these methods.

modules/sdp/LocalSdpMunger.js
```diff
@@ -203,9 +203,29 @@ export default class LocalSdpMunger {
...
+            const sources = [ ...new Set(mediaSection.mLine?.ssrcs?.map(s => s.id)) ];
+
+            for (const source of sources) {
+                const msidExists = mediaSection.ssrcs
+                    .find(ssrc => ssrc.id === source && ssrc.attribute === 'msid');
+
+                !msidExists && mediaSection.ssrcs.push({
+                    id: source,
+                    attribute: 'msid',
+                    value: `${this.localEndpointId}-${mediaSection.mLine?.type}-${pcId} ${trackId}-${pcId}`
```
paweldomas, line 226: can we extract msid generation into a function? it seems to be duplicated with the logic at line 190?

paweldomas (approved):

PR #1619, opened 2021-05-18
fix(caps): Disable TCC on Firefox.

jallamsetty1:
There is a known issue with Firefox where the BWE gets halved on every renegotiation causing the low upload bitrates from the Firefox clients.


damencho (approved):

jallamsetty1:
Jenkins test this please.

PR #1622, opened 2021-05-27
fix(TPC): Do not configure encodings on chromium immediately after replaceTrack.

jallamsetty1:
Avoid configuring the encodings on chromium immediately after replace track since the encoding params are read-only until the renegotation is done.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -79,12 +79,19 @@ export class TPCUtils {
     _findTransceiver(mediaType, localTrack = null) {
         let transceiver = null;
 
-        if (localTrack) {
-            transceiver = this.pc.peerconnection.getTransceivers()
-                .find(t => t.sender?.track?.id === localTrack.getTrackId());
-        } else if (mediaType) {
+        // Check if the local track has been removed from the peerconnection already.
+        const trackRemoved = !localTrack
+            || (localTrack
+                && browser.doesVideoMuteByStreamRemove()
+                && localTrack.isVideoTrack()
+                && localTrack.isMuted());
+
+        if (trackRemoved) {
             transceiver = this.pc.peerconnection.getTransceivers()
                 .find(t => t.receiver?.track?.kind === mediaType);
+        } else if (localTrack) {
+            transceiver = this.pc.peerconnection.getTransceivers()
+                .find(t => t.sender?.track?.id === localTrack.getTrackId());
         }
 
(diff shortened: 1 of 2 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -379,8 +383,15 @@ export class TPCUtils {
                         transceiver.direction = TransceiverDirection.SENDRECV;
                     }
 
-                    // Add the new track to the list of local tracks.
-                    this.pc.localTracks.set(newTrack.rtcId, newTrack);
+                    // Avoid configuring the encodings on chromium since the encoding params are read-only
+                    // until the renegotation is done after the track is replaced on the sender.
+                    const promise = browser.isChromiumBased() ? Promise.resolve() : this.setEncodings(newTrack);
```
paweldomas, line 388: isn't that the case for any browser? what does the spec say? I thought the encodings are set via "transceiver init" or something like that

jallamsetty1: No, it doesn't seem to be the case on all the browsers. When the track is added initially, encodings are specified in transceiverInit for Chrome/Safari but FF ignores config transceiverInit so the encodings are configured after addTrack. This case is for when a track is removed the pc and then added back to the pc using replaceTrack. This is the case where chrome makes them read-only until the reneg happens.

modules/RTC/TPCUtils.js
```diff
@@ -79,12 +79,19 @@ export class TPCUtils {
...
         let transceiver = null;
 
-        if (localTrack) {
-            transceiver = this.pc.peerconnection.getTransceivers()
-                .find(t => t.sender?.track?.id === localTrack.getTrackId());
-        } else if (mediaType) {
+        // Check if the local track has been removed from the peerconnection already.
+        const trackRemoved = !localTrack
+            || (localTrack
+                && browser.doesVideoMuteByStreamRemove()
```
paweldomas, line 85: we're using this combination of conditions so often that it feels would make sense to add a helper method to jitsi track "was removed for video mute?". Just a suggestion doesn't need to happen in this PR. I just think it would be easier to read one condition rather than having to connect 3 every time.

paweldomas (approved):

modules/RTC/TPCUtils.js
jallamsetty1: Yeah makes sense Pawel! Will add it in the next one.

PR #1632, opened 2021-06-08
fix(recording): Send participant id when recording starts/stops

vp8x8:
This fixes the issue where undefined was sent instead of a participant on
recording stop/start. Appropriate handling has been done in jitsi-meet.

### DIFF
--- a/modules/recording/JibriSession.js
+++ b/modules/recording/JibriSession.js
@@ -40,7 +40,7 @@ export default class JibriSession {
     /**
      * Returns the initiator of the session instance.
      *
-     * @returns {JitsiParticipant|undefined} The participant that started the session.
+     * @returns {JitsiParticipant|string} The participant that started the session.
      */
     getInitiator() {
         return this._initiator;
@@ -67,7 +67,7 @@ export default class JibriSession {
     /**
      * Returns the jid of the participant that stopped the session.
      *
-     * @returns {JitsiParticipant|undefined} The participant that stopped the session.
+     * @returns {JitsiParticipant|string} The participant that stopped the session.
      */
     getTerminator() {
         return this._terminator;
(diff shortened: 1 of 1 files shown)


damencho:
@vp8x8 what is the point of the change? To be able to detect and when the local participant is doing the change?

damencho (changes requested):
You also need to change the docs here https://github.com/jitsi/lib-jitsi-meet/blob/713db5d8c34a4a799cf33be7f4e1f5f99ed497c4/modules/recording/JibriSession.js#L127 that it is changed from participant to a jid ... 
And here https://github.com/jitsi/lib-jitsi-meet/blob/713db5d8c34a4a799cf33be7f4e1f5f99ed497c4/modules/recording/JibriSession.js#L119 and all docs around those initiator/terminator members of the class

vp8x8:
> @vp8x8 what is the point of the change? To be able to detect and when the local participant is doing the change?

Yes exactly.

damencho:
Are you ready with the jitsi-meet part, maybe create yhe PR there first. If we metge this, we block updating jitsi-meet till you fix it ... As this is change in the API ... Oh now I wonder are there people using it already ... As we will break their code ...

damencho:
I will look around later to check how do we handle similar cases in other places ... Can we go wothout breaking the API ...

vp8x8:
The PR is ready https://github.com/jitsi/jitsi-meet/pull/9362
Is there any other way to correlate the id with the participant? If there is I could maybe use it...

damencho:
> The PR is ready [jitsi/jitsi-meet#9362](https://github.com/jitsi/jitsi-meet/pull/9362)
> Is there any other way to correlate the id with the participant? If there is I could maybe use it...

Well we do not have JitsiParticipant object for the local participant. That is your problem right?

Other than that the participant id is part of the jid ... Is this what you mean?

vp8x8:
I have updated the PR.

damencho (approved):

PR #1633, opened 2021-06-08
Fix media direction p2p

jallamsetty1:
Introduce an enum for media direction.
For p2p connections, the media direction needs to be adjusted after every source-add/source-remove is processed based on the availability of local sources.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -5,6 +5,7 @@ import { getLogger } from 'jitsi-meet-logger';
 import transform from 'sdp-transform';
 
 import * as CodecMimeType from '../../service/RTC/CodecMimeType';
+import MediaDirection from '../../service/RTC/MediaDirection';
 import * as MediaType from '../../service/RTC/MediaType';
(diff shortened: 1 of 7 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2417,12 +2448,14 @@ TraceablePeerConnection.prototype.setRemoteDescription = function(description) {
...
                 dumpSDP(description));
-
-            // eslint-disable-next-line no-param-reassign
-            description = this.tpcUtils.ensureCorrectOrderOfSsrcs(description);
         }
     }
 
+    if (this._usesUnifiedPlan) {
+        // eslint-disable-next-line no-param-reassign
+        description = this.tpcUtils.ensureCorrectOrderOfSsrcs(description);
```
damencho, line 2456: Is this operation ok when simulcast is off?

jallamsetty1: Yes, this is needed even when simulcast is turned off as there will still be a FID group for the single source and the primary ssrc needs to be present before the secondary ssrc in the SDP.

damencho (approved):

PR #1636, opened 2021-06-10
fix(TPC): fix local resolution/fps stats.

jallamsetty1:
Browsers do not generate a 'msid' attribute for ssrcs in unified plan mode, use mediaType as a key for the TrackSSRCInfo map.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1156,23 +1156,13 @@ TraceablePeerConnection.prototype._removeRemoteTrackById = function(
 };
 
 /**
- * @typedef {Object} SSRCGroupInfo
- * @property {Array<number>} ssrcs group's SSRCs
- * @property {string} semantics
- */
-/**
- * @typedef {Object} TrackSSRCInfo
- * @property {Array<number>} ssrcs track's SSRCs
- * @property {Array<SSRCGroupInfo>} groups track's SSRC groups
- */
-/**
- * Returns map with keys msid and <tt>TrackSSRCInfo</tt> values.
- * @param {Object} desc the WebRTC SDP instance.
+ * Returns a map with keys msid/mediaType and <tt>TrackSSRCInfo</tt> values.
+ * @param {RTCSessionDescription} desc the local description.
  * @return {Map<string,TrackSSRCInfo>}
  */
-function extractSSRCMap(desc) {
+TraceablePeerConnection.prototype._extractSSRCMap = function(desc) {
     /**
-     * Track SSRC infos mapped by stream ID (msid)
+     * Track SSRC infos mapped by stream ID (msid) or mediaType (unfied-plan)
      * @type {Map<string,TrackSSRCInfo>}
      */
     const ssrcMap = new Map();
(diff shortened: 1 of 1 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1224,25 +1222,31 @@ function extractSSRCMap(desc) {
             }
         }
         for (const ssrc of mLine.ssrcs) {
-            if (ssrc.attribute !== 'msid') {
+            if (!this._usesUnifiedPlan && ssrc.attribute !== 'msid') {
```
paweldomas, line 1225: two negated conditions are a bit hard to read here, is it the same as "do it for  every ssrc in unified or only those with msid in plan B" ? how about doing filter on `mLine.ssrcs` only in plan B one line above ? this way there would be no continue and no eslint ignore.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1224,25 +1222,31 @@ function extractSSRCMap(desc) {
...
+            const key = this._usesUnifiedPlan ? mLine.type : ssrc.value;
+            const ssrcNumber = ssrc.id;
+            let ssrcInfo = ssrcMap.get(key);
 
             if (!ssrcInfo) {
                 ssrcInfo = {
                     ssrcs: [],
                     groups: [],
-                    msid
+                    msid: key
```
paweldomas, line 1241: is this `msid` field used later on (can it be removed from this structure) ? it's misleading to put not an msid under the name `msid`

damencho (approved):

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2911,16 +2914,15 @@ TraceablePeerConnection.prototype._processLocalSSRCsMap = function(ssrcMap) {
...
+                oldSSRCNum && logger.error(`${this} Overwriting SSRC for track=${track}] with ssrc=${newSSRC}`);
                 this.localSSRCs.set(track.rtcId, newSSRC);
                 this.eventEmitter.emit(RTCEvents.LOCAL_TRACK_SSRC_UPDATED, track, newSSRCNum);
             }
         } else if (!track.isVideoTrack() && !track.isMuted()) {
             // It is normal to find no SSRCs for a muted video track in
             // the local SDP as the recv-only SSRC is no longer munged in.
             // So log the warning only if it's not a muted video track.
-            logger.warn(`${this} No SSRCs found in the local SDP for track[track=${track},id=${trackMSID}]`);
+            logger.warn(`${this} No SSRCs found in the local SDP for track=${track}`);
```
paweldomas, line 2925: can you also include the `sourceIdentifier` in the log? just to know what id was not found

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1224,25 +1222,31 @@ function extractSSRCMap(desc) {
...
 
-            const ssrcNumber = ssrc.id;
-
-            ssrcInfo.ssrcs.push(ssrcNumber);
+            // Avoid duplicates.
+            if (!ssrcInfo.ssrcs.find(s => s === ssrcNumber)) {
+                ssrcInfo.ssrcs.push(ssrcNumber);
+            }
 
             if (groupsMap.has(ssrcNumber)) {
```
paweldomas, line 1251: do I understand correctly that this executes multiple times in unified if there are more than one attributes for one SSRC? does it not result in the same group pushed multiple times to `ssrcInfo.groups` ?

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Fixed it in this [PR](https://github.com/jitsi/lib-jitsi-meet/pull/1638)

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Fixed it in this [PR](https://github.com/jitsi/lib-jitsi-meet/pull/1638)

modules/RTC/TraceablePeerConnection.js
jallamsetty1: I checked if its being used anywhere before opening this PR but couldn't find anything using this field. Do you know why it was added in the first place if its not being used anywhere ? RN will be running in plan-b mode for a while so I didn't want to mess with that just in case.

PR #1638, opened 2021-06-10
fix(TPC): Filter ssrcs differently while extracting the SSRC map from SDP.

jallamsetty1:
Use 'msid' for plan-b clients and 'cname' for unified-plan clients.


paweldomas (approved):
thanks for doing the change!

PR #1639, opened 2021-06-11
feat(replace-participant): Replace participant with same jwt in the conf

horymury:
Replaces a participant with the same jwt relevant data into a conference when this is specifically asked on joining the conference.


tmoldovan8x8 (approved):

tmoldovan8x8:
Look good to me!

tmoldovan8x8 (approved):

PR #1642, opened 2021-06-17
fix: Send videoType bridge message for mute/unmute on FF.

jallamsetty1:
Set the initial video type state to NONE to cover cases when the endpoint joins video muted.

### DIFF
--- a/modules/RTC/JitsiTrack.js
+++ b/modules/RTC/JitsiTrack.js
@@ -4,7 +4,6 @@ import { getLogger } from 'jitsi-meet-logger';
 
 import * as JitsiTrackEvents from '../../JitsiTrackEvents';
 import * as MediaType from '../../service/RTC/MediaType';
-import VideoType from '../../service/RTC/VideoType';
 import browser from '../browser';
 
 import RTCUtils from './RTCUtils';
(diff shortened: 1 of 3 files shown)


modules/RTC/JitsiTrack.js
```diff
@@ -174,6 +175,13 @@ export default class JitsiTrack extends EventEmitter {
         }
     }
 
+    /**
+     * Returns the video type (camera or desktop) of this track.
+     */
+    getVideoType() {
```
hristoterezov, line 181: If this type will be used only for the messages to the bridge I'd add "bridge" to the name. Also if you don't plan to be public I'd add "_".

Is this the case?

jallamsetty1: > If this type will be used only for the messages to the bridge I'd add "bridge" to the name. Also if you don't plan to be public I'd add "_".
> 
> Is this the case?

It doesn't have to be specific to the bridge message, I made it generic enough so that it can be used for getting the videoType. It is a public method and I am calling that instead of using `track.videoType`.

hristoterezov: What I'm worried about is that currently we use `track.videoType` everywhere including jitsi-meet and also `track.videoType` can't be `'none'` for local video track. IMHO it gets a little confusing whether to use `track.videoType` or `track.getVideoType()` and what result to expect.

jallamsetty1: I can modify it to return the `videoType` and add the muted check wherever the bridge message is sent. What should it return for the audio tracks though ? Is `none` acceptable there ?

hristoterezov (approved):

PR #1646, opened 2021-06-28
fix(VAD): Do not enable VAD on Chrome on iOS.

jallamsetty1:
This causes the audio to not work.


jallamsetty1:
The failing lobby and lastN test on FF are not related to the changes in the PR.

PR #1651, opened 2021-07-07
Small fix in tokens doc.

damencho:


Freddie-m (approved):
This is better. Fixes the confusion that the previous phrase could have caused.

Freddie-m (approved):
This works. Thanks for editing.

raluca8x8 (approved):

PR #1652, opened 2021-07-07
fix(VADAudioAnalyser): NPE error evaluating this._vadEmitter.on

paweldomas:
The crash occurs every time the audio track is muted/unmuted, if the emitter fails to initialize ("Failed to start VADAudioAnalyser" warning is printed).

@andrei-gavrilescu I'm not sure if it's the right way to fix this. It feels like something more should be taken out of the event chain in case the vad emitter fails to start.

### DIFF
--- a/modules/detection/VADAudioAnalyser.js
+++ b/modules/detection/VADAudioAnalyser.js
@@ -100,9 +100,11 @@ export default class VADAudioAnalyser extends EventEmitter {
      * @returns {void}
      */
     _startVADEmitter() {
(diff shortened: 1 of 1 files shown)


damencho (approved):

andrei-gavrilescu:
> The crash occurs every time the audio track is muted/unmuted, if the emitter fails to initialize ("Failed to start VADAudioAnalyser" warning is printed).
> 
> @andrei-gavrilescu I'm not sure if it's the right way to fix this. It feels like something more should be taken out of the event chain in case the vad emitter fails to start.

It's fine as is imo.

andrei-gavrilescu (approved):

PR #1654, opened 2021-07-10
feat: allow json messages from focus

wernf:
This patch allows participants to receive json messages from the focus user. This is a dependency of the breakout rooms feature. See [this](https://github.com/jitsi/jitsi-meet/pull/9131) jitsi-meet pull request.

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -413,13 +413,13 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
 
     chatRoom.addListener(XMPPEvents.JSON_MESSAGE_RECEIVED,
         (from, payload) => {
-            const id = Strophe.getResourceFromJid(from);
-            const participant = conference.getParticipantById(id);
+            const _id = Strophe.getResourceFromJid(from);
+            const participant = conference.getParticipantById(_id);
 
-            if (participant) {
+            if (participant || _id === 'focus') {
                 conference.eventEmitter.emit(
                     JitsiConferenceEvents.ENDPOINT_MESSAGE_RECEIVED,
-                    participant, payload);
+                    participant || { _id }, payload);
             } else {
                 logger.warn(
                     'Ignored XMPPEvents.JSON_MESSAGE_RECEIVED for not existing '


JitsiConferenceEventManager.js
```diff
@@ -413,13 +413,13 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
 
     chatRoom.addListener(XMPPEvents.JSON_MESSAGE_RECEIVED,
         (from, payload) => {
-            const id = Strophe.getResourceFromJid(from);
-            const participant = conference.getParticipantById(id);
+            const _id = Strophe.getResourceFromJid(from);
+            const participant = conference.getParticipantById(_id);
 
-            if (participant) {
+            if (participant || _id === 'focus') {
```
saghul, line 419: Do we have a better way to detect this?

wernf: This is now gone. The new NON_PARTICIPANT_MESSAGE_RECEIVED event is used instead.

modules/xmpp/ChatRoom.js
```diff
@@ -1837,7 +1841,9 @@ export default class ChatRoom extends Listenable {
             }
             eventEmitter.on(XMPPEvents.MUC_LEFT, onMucLeft);
             this.doLeave();
-        });
+        }));
+
+        return Promise.all(promises);
```
saghul, line 1846: Promise.all will resolve when the first one fails. I'd say we want Promise.allSettled here.

modules/xmpp/Lobby.js
```diff
@@ -244,14 +248,6 @@ export default class Lobby {
 
                     this.mainRoom.eventEmitter.emit(XMPPEvents.MUC_DESTROYED, reason);
                 });
-
-            // If participant retries joining shared password while waiting in the lobby
```
saghul, line 248: Why is this no longer needed?

modules/xmpp/Lobby.js
```diff
@@ -307,7 +315,7 @@ export default class Lobby {
         if (memberRoomJid) {
             const jid = this.lobbyRoom.members[memberRoomJid].jid;
             const msgToSend
-                = $msg({ to: this.mainRoom.roomjid })
+                = $msg({ to: this.getMainRoomJidFromRoomJid(this.mainRoom.roomjid) })
```
saghul, line 318: So, what are we doing here? Requesting access to a breakout rooom which has the lobby enabled through the main room?

saghul (commented):
Left a couple of comments, I like how small this change feels here.

modules/xmpp/ChatRoom.js
saghul: Let's leave it as-is for now.

saghul:
Finally back at this. I rebased and force-pushed.

PR #1662, opened 2021-07-27
fix issuer documentation

lfuelling:
Hi,

it seems like the default value of `asap_accepted_issuers` and/or `asap_accepted_audiences` is no longer `{ "*" }`.

This PR corrects the corresponding docs.


lfuelling:
Why do I have to disclose my mailing address for fixing a Markdown file?

"I hereby release this contribution under public domain." <- Isn't that enough?

sapkra:

lfuelling:
Thanks for accepting this :)

PR #1664, opened 2021-07-29
fix(LocalSdpMunger): do not fake video sdp when screen sharing

paweldomas:
... is stopped. If there's any sRD/sLD cycle happening
while the screen sharing stops, the local track
returns muted and it was injecting fake video SSRCs
which results in invalid SSRC description.

At the time when it happens there's error log in jicofo:
"Error adding SSRCs from X"
and the client will get bad-request error in response
to the source-add request.

### DIFF
--- a/modules/sdp/LocalSdpMunger.js
+++ b/modules/sdp/LocalSdpMunger.js
@@ -4,6 +4,7 @@ import { getLogger } from 'jitsi-meet-logger';
 
 import MediaDirection from '../../service/RTC/MediaDirection';
 import * as MediaType from '../../service/RTC/MediaType';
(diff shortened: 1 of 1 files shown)


jallamsetty1:
@paweldomas, does the repro case involve a camera source added automatically when the screenshare is being removed ? I don't understand when the local video track will come back as muted. If no camera source is automatically added, tpc.getLocalTracks('video') will return an empty array. If a camera source has been added, does that come back as muted ? If the new ssrc hasn't being signed to Jicofo, is the active field on the associated MediaStream set to false ?

jallamsetty1 (approved):

PR #1665, opened 2021-07-31
support to upgrade role via jwt token.

jiyeyuran:


damencho:
@jiyeyuran Can you give some details, what does this change do?

jiyeyuran:
With this feature, the `authenticateAndUpgradeRole` function on React Native can work with `token`.

saghul:
Ah, is this because mobile doesn't reload the page? I think that might be difference in flow here.

jiyeyuran:
Because the `tokenAuthUrl` in `config.js` is not working on mobile, we want to use the `authenticateAndUpgradeRole` funcation to upgrade the role by Token.

saghul:
Right. How are you calling that at the moment?

saghul:
@damencho In general this SGTM, WDYT?

jiyeyuran:
We want to reuse the login UI of xmpp, so we modified the function.

```
export function authenticateAndUpgradeRole(
        id: string,
        password: string,
        conference: Object) {
    return (dispatch: Dispatch<any>, getState: Function) => {
        const { password: roomPassword } = getState()['features/base/conference'];
        const { apiUrl } = getState()['features/base/config'];
        const process = fetch(`${apiUrl}/login`, {
            method: 'POST',
            headers: {
                'Content-Type': 'application/json'
            },
            body: JSON.stringify({
                username: id.split('@')[0],
                password,
                roomName: conference.getName()
            })
        })
        .then(response => response.json())
        .then(data => data.jwt)
        .then(token => conference.authenticateAndUpgradeRole({
            token,
            roomPassword,
            onLoginSuccessful() {
            // When the login succeeds, the process has completed half
            // of its job (i.e. 0.5).
                return dispatch(_upgradeRoleFinished(process, 0.5));
            }
        }))
        .then(() => dispatch(_upgradeRoleFinished(process, 1)))
        .catch(error => {
            // The lack of an error signals a cancellation.
            if (error.authenticationError || error.connectionError) {
                logger.error('authenticateAndUpgradeRole failed', error);
            }

            dispatch(_upgradeRoleFinished(process, error));
        });

        dispatch(_upgradeRoleStarted(process));

        return process;
    };
}
```

saghul:
Can you please make all cabbages part of the PR so we can see the full picture? Also, what are the steps to test the fix?

jiyeyuran:
1. Join the conference as an anonymous user.
2. The moderator of the conference quit.
3. The anonymous user wants to be the moderator but without jumping out of the conference UI.
4. The anonymous user requests our private API to get JWT token.
5. call `conference.authenticateAndUpgradeRole` with the token. The problem is that the function `authenticateAndUpgradeRole` can only accept the XMPP username and password even the `XMPP` class allows the jwt token as the second parameter.

damencho:
This sounds to me more like your service could just change the affiliation of the participant server-side (in prosody), and the user will receive the new role ... as like it was granted by someone else.

saghul:
I  agree that's a cleaner way to go about it.

jiyeyuran:
Unfortunately, no one in our team knows how to talk with prosody from the server-side. Learning from [AuthHandler.js](https://github.com/jitsi/jitsi-meet/blob/master/modules/UI/authentication/AuthHandler.js) in the `jitsi-meet` project, We think fetching the jwt token, and upgrading the role from js is easier.

damencho:
Then disconnecting and connecting again shouldn't that work with current code base?

damencho:
I'm not sure how this will work if you already are connected and you create a second connection ... this sound suspicious to me?

damencho:
The prosody thing is as easy as creating an HTTP endpoint like this https://github.com/jitsi/jitsi-meet/blob/af010728279b5c9cff0fd75f201a0f1022d094a2/resources/prosody-plugins/mod_muc_census.lua#L76
which will be accessed only by your service and changing the affiliation is done like: https://github.com/jitsi/jitsi-meet/blob/af010728279b5c9cff0fd75f201a0f1022d094a2/resources/prosody-plugins/mod_muc_allowners.lua#L102
There are ton of examples in that folder ...

PR #1669, opened 2021-08-04
Respect custom callstats script url for precall test & add config to toggle callstats collect ip

pangrr:


pangrr:
I think PRs from a bloomberg fork should pass your CLA checks. Could you confirm?

saghul:
Can you please break this PR into 2? I think we can land the first commit right away, but I'd like to discuss the second one some more.

I think it might be a good idea to pass an entire `configParams` object instead of each individual option: https://docs.callstats.io/docs/javascript that would allow any LJM user to set any of those parameters.

pangrr:
> Can you please break this PR into 2? I think we can land the first commit right away, but I'd like to discuss the second one some more.
> 
> I think it might be a good idea to pass an entire `configParams` object instead of each individual option: https://docs.callstats.io/docs/javascript that would allow any LJM user to set any of those parameters.

#1670 for the first part.

pangrr:
#1672 for part 2

PR #1679, opened 2021-08-10
Fix Spot wireless screen sharing feature

saghul:
See each individual commit message for details.


jallamsetty1 (approved):
Typically a PC that is created with offerToReceive audio/video as constraints has both mlines, even if no local sources are added. Looks like the pc created by Spot's wireless screenshare feature doesn't use those options. Nice catch! I wonder if we should change it there so that we don't run into any other unexpected issues.

saghul:
> Typically a PC that is created with offerToReceive audio/video as constraints has both mlines, even if no local sources are added. Looks like the pc created by Spot's wireless screenshare feature doesn't use those options. Nice catch! I wonder if we should change it there so that we don't run into any other unexpected issues.

Yeah we can do that! If you can send a PR I can test it on a Spot setup.

PR #1681, opened 2021-08-11
fix(ProxyConnectionPC): request to receive audio/video.

jallamsetty1:
Offering to receive audio/video during pc creation ensures that there are mlines for audio and video in the SDP, making this change to prevent any unexpected issues (in wireless screensharing for spot) with sdp->jingle translations after the unified plan switch.

### DIFF
--- a/modules/proxyconnection/ProxyConnectionPC.js
+++ b/modules/proxyconnection/ProxyConnectionPC.js
@@ -41,8 +41,8 @@ export default class ProxyConnectionPC {
         this._options = {
             iceConfig: {},
             isInitiator: false,
-            receiveAudio: false,
-            receiveVideo: false,
+            receiveAudio: true,
+            receiveVideo: true,
             ...options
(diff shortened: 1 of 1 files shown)


saghul:
I had a second look at this, and I don't think we want to do this change. This PC is only used to send the "wireless" screen-share from a participant who is not in the conference to the Spot TV. This it's unidirectional and video only. We don't want to receive anything in this PC.

jallamsetty1:
> I had a second look at this, and I don't think we want to do this change. This PC is only used to send the "wireless" screen-share from a participant who is not in the conference to the Spot TV. This it's unidirectional and video only. We don't want to receive anything in this PC.

Are you saying that this is a p2p connection only with the Spot TV ? If that is the case then we should be fine.

saghul:
Yep it's a P2P connection between the remote and the TV, just to send a video stream to it.

PR #1684, opened 2021-08-13
added support for DSCP marking outbound media packets on Chromium

jbg:
Chromium-based browsers support DSCP marking of outbound media packets, but it's disabled by default because some middleboxes on the wider Internet might drop packets with DSCP marks set. On controlled networks (e.g. corporate networks) DSCP can be very helpful for prioritising videoconferencing traffic.

This patch allows it to optionally be enabled with a `enableDSCP` option.

Relevant community discussion: https://community.jitsi.org/t/looking-for-a-way-to-configure-udp-ports-used-in-peer-to-peer-for-1-to-1-sessions-p2p4121/103706/6

### DIFF
--- a/modules/RTC/RTCUtils.js
+++ b/modules/RTC/RTCUtils.js
@@ -76,6 +76,9 @@ let disableNS = false;
 // Disables Automatic Gain Control
 let disableAGC = false;
 
+// Enabled DSCP marking
+let enableDSCP = false;
+
 // Enables stereo.
 let stereo = null;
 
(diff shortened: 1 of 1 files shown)


jbg:
CLA already signed.

modules/RTC/RTCUtils.js
```diff
@@ -377,7 +377,9 @@ class RTCUtils extends Listenable {
             ? { optional: [
                 { googScreencastMinBitrate: 100 },
                 { googCpuOveruseDetection: true }
-            ] }
+            ] + this.options.enableDSCP ? [
+                    { googDscp: true }
+                ] : [] }
```
damencho, line 382: I don't think this syntax is correct, I tested it and it throws an error when enableDSCP is missing ...

jbg: D'oh! Should be fixed now, and changed it to be handled the same way as the other options along with the log message when its value is set.

jbg:
I'm marking this as a draft, because in our testing it doesn't seem to consistently cause DSCP marking to be applied, and based on some research it looks like `RtpSender.setParameters` may need to be used to set the desired marking on a per-sender basis. I'll update this PR once I've investigated further.

PR #1694, opened 2021-08-22
[feat] Audiofeedback prevention

theunafraid:
The purpose of this PR is to make use of a RNN to prevent the audiofeedback loop and be able to act on it BEFORE "howl" without affecting the user experience. More info here https://github.com/theunafraid/audiofeedback-prevention .

This needs much more work and did not have time for it, model must be trained using much more data and did not have time to find or sort it(used [TCD VoIP Dataset](https://qxlab.ucd.ie/index.php/tcd-voip-dataset/) and [MLCommons](https://mlcommons.org/en/) . At the time i began, did not find anything similar(trained model) that works in a similar way to this(classifier).


daimoc:
Hi @theunafraid do you think you'll fix your PR ? This audiofeedback loop looks a nice feature to impove the Jitsi user experience.

theunafraid:
Hi @daimoc  Yes, not sure exactly when, in the next few months(3-4), will have to make time for it.

PR #1695, opened 2021-08-23
feat: JSON encoded sources.

bgrozev:


modules/xmpp/strophe.jingle.js
```diff
@@ -22,6 +22,144 @@ const logger = getLogger(__filename);
...
+
+    // This is where we'll add "source" and "ssrc-group" elements. Create them elements if they don't exist.
+    let videoRtpDescription = getOrCreateRtpDescription(iq, 'video');
+    let audioRtpDescription = getOrCreateRtpDescription(iq, 'audio');
+
+    for (let owner in json.sources) {
+        let ownerSources = json.sources[owner];
+        // The video sources, video ssrc-groups, audio sources and audio ssrc-groups are encoded in that order in the
+        // elements of the array.
+        let videoSources = ownerSources && ownerSources.length > 0 && ownerSources[0];
```
jallamsetty1, line 59: What does the json message look like when a given remote endpoint doesn't have video sources but only audio source ? Or when the endpoint doesn't have any video ssrc-groups, like Firefox with simulcast disabled ?

modules/xmpp/strophe.jingle.js
```diff
@@ -22,6 +22,144 @@ const logger = getLogger(__filename);
...
+ * @param {*} owner the endpoint ID of the owner of the source.
+ * @param {*} sourceCompactJson the compact JSON representation of the source.
+ * @returns the created "source" XML element.
+ */
+function createSourceExtension(owner, sourceCompactJson) {
+    let node = $build('source', {xmlns: 'urn:xmpp:jingle:apps:rtp:ssma:0', ssrc: sourceCompactJson.s})
+    if (sourceCompactJson.m) {
+        node.c('parameter', {name: 'msid', value: sourceCompactJson.m}).up()
+    }
+    if (sourceCompactJson.c) {
```
jallamsetty1, line 98: I think we should be ok with skipping this. WDYT ?
`cname` is not used anywhere, once the ssrc is added to remote desc and sRD is done, the track event fired by the browser uses the `msid` as the stream id which is then used for creating/identifying the remote track.

modules/xmpp/strophe.jingle.js
```diff
@@ -22,6 +22,144 @@ const logger = getLogger(__filename);
...
+            for (let i = 0; i < audioSources.length; i++) {
+                audioRtpDescription.appendChild(createSourceExtension(owner, audioSources[i]))
+            }
+        }
+        if (audioSsrcGroups && audioSsrcGroups.length > 0) {
+            for (let i = 0; i < audioSsrcGroups.length; i++) {
+                audioRtpDescription.appendChild(createSsrcGroupExtension(audioSsrcGroups[i]))
+            }
+        }
+    }
```
jallamsetty1, line 84: Since we are already extracting all the source information from the json message, is it possible to log it here as a debug message so that we can get rid of the log that logs the full iq ?

modules/xmpp/strophe.jingle.js
```diff
@@ -22,6 +22,144 @@ const logger = getLogger(__filename);
...
+    return node.node;
+}
+
+/**
+ * Convert the short string representing SSRC group semantics in compact JSON format to the standard representation (i.e. convert "f" to "FID"
+ *  and "s" to "SIM").
+ * @param {*} str the compact JSON format representation of an SSRC group's semantics.
+ * @returns the SSRC group semantics corresponding to [str].
+ */
+function getSementics(str) {
```
jallamsetty1, line 124: `getSemantics` perhaps ?

modules/xmpp/strophe.jingle.js
```diff
@@ -132,6 +270,16 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
 
         // see http://xmpp.org/extensions/xep-0166.html#concepts-session
 
+        const jsonMessages = $(iq).find('jingle>json-message');
```
jallamsetty1, line 273: I don't see that we are doing the same [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/JingleSessionPC.js#L1632) for source-adds. Is there a reason for that ? Is Jicofo going to batch up sources in source-add and use the json format or is that coming later ?

modules/xmpp/strophe.jingle.js
bgrozev: I'd be more than happy to remove it everywhere :)

modules/xmpp/strophe.jingle.js
bgrozev: I don't understand what you mean by "the same". Here we read `<json-message>` and populate the `<content>`s with the sources regardless of the message type (session-initiate, source-add, etc). The JingleSessionPC code then operates as before on `<content>`s containing `<source>`s. 

In the first version Jicofo will encode sources in JSON for all session-initiate and source-add messages. In a later version some source-add messages will be combined (batched).

modules/xmpp/strophe.jingle.js
bgrozev: Ops, will fix.

modules/xmpp/strophe.jingle.js
bgrozev: I'd rather do it in a separate PR though -- starting with Jicofo.

modules/xmpp/strophe.jingle.js
jallamsetty1: My bad, yeah I see that the content is modified for all message types 🤦‍♀️

modules/xmpp/strophe.jingle.js
jallamsetty1: Will you still need to make changes in Jicofo if the client stops sending it ? That way we don't have to worry about mobile catching up with latest LJM version.

bgrozev: No, if it is not used by the client then the changes can be done independently.

jallamsetty1 (approved):

PR #1697, opened 2021-08-24
sdp: switch port to 9 and rtp protocol to UDP/TLS/RTP/SAVPF

fippo:
the protocol was switched back in Chrome 44 in 2016:
  https://bugs.chromium.org/p/webrtc/issues/detail?id=2796
the port switch more than five years ago too.


fippo:
wondering if that (in particular the port change; the protocol isn't mapped to jingle) creates woes on the bridge

fippo:
oh wait. the port in the m-line isn't mapped to jingle either. It is only mapped from jingle (and that is still a hack :-)

modules/sdp/SDP.js
```diff
@@ -557,15 +557,15 @@ SDP.prototype.jingle2media = function(content) {
...
 
-    media.port = '1';
+    media.port = '9';
     if (content.attr('senders') === 'rejected') {
         // estos hack to reject an m-line.
         media.port = '0';
     }
     if (transport.find('>fingerprint[xmlns="urn:xmpp:jingle:apps:dtls:0"]').length) {
-        media.proto = sctp.length ? 'DTLS/SCTP' : 'RTP/SAVPF';
+        media.proto = sctp.length ? 'DTLS/SCTP' : 'UDP/TLS/RTP/SAVPF';
```
nils-ohlmeier, line 566: While you are editing this could you also update `DTLS/SCTP` into `UDP/DTLS/SCTP` please (which is also supported by all major browsers for quite some time now)?

nils-ohlmeier (changes requested):
Thanks for the patch!
LGTM, but it would be great to update the data channel protocol as well at the same time :-)

modules/sdp/SDP.js
fippo: I was wondering about that but what is the state of sctp support even?

nils-ohlmeier: You mean in Jitsi Meet? It is not being used on m.j.s, but people can still turn it own for their own deployments AFAIK. So that is why I think it's worth updating.

fippo: agree, will do a followup change ;-)

fippo:
jenkins please test this

fippo:

JonathanLennox (approved):

PR #1698, opened 2021-08-26
fix(replace-track) Add not null check for newTrack

robertpin:
Fixes error caused by stop screen sharing call (newTrack is null) and video moderation on (isVideoMutedByFocus is true)


saghul:
@damencho 👍

robertpin:
Jenkins test this please.

damencho (approved):

PR #1702, opened 2021-08-31
fix(browser-support): Add audio track to pc always on mobile Safari.

jallamsetty1:
On mobile Safari, if a user joins audio and video muted, the browser doesn't decode the incoming audio. Workaround is to always add the audio track to pc and mute it if needed.


damencho (approved):

jallamsetty1:
Merging since the failing tests are unrelated and are known to fail on Firefox.

PR #1704, opened 2021-09-07
Reduce verbose logs take X

saghul:


modules/xmpp/moderator.js
```diff
@@ -102,7 +102,6 @@ Moderator.prototype.isSipGatewayEnabled = function() {
 };
 
 Moderator.prototype.onMucMemberLeft = function(jid) {
-    logger.info(`Someone left is it focus ? ${jid}`);
```
damencho, line 105: Hum, we do not have a log for someone leaving the room if we remove this one ...

saghul: I see plenty of them. You mean in LJM itself?

damencho: I meant in general ... but yeah, you are right `USER 6c59086d LEFT` from jitsi-meet, sorry

damencho (approved):

PR #1707, opened 2021-09-08
Perf metrics

saghul:
See the first commit for a thorough explainer.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2015,10 +2015,7 @@ JitsiConference.prototype._onIncomingCallP2P = function(
 /**
  * Handles an incoming call event.
  */
-JitsiConference.prototype.onIncomingCall = function(
-        jingleSession,
-        jingleOffer,
-        now) {
+JitsiConference.prototype.onIncomingCall = function(jingleSession, jingleOffer) {
     // Handle incoming P2P call
     if (jingleSession.isP2P) {
         this._onIncomingCallP2P(jingleSession, jingleOffer);
@@ -2035,21 +2032,19 @@ JitsiConference.prototype.onIncomingCall = function(
 
             return;
         }
-        this._acceptJvbIncomingCall(jingleSession, jingleOffer, now);
+        this._acceptJvbIncomingCall(jingleSession, jingleOffer);
     }
 };
 
 /**
  * Accepts an incoming call event for the JVB jingle session.
  */
-JitsiConference.prototype._acceptJvbIncomingCall = function(
-        jingleSession,
-        jingleOffer,
-        now) {
-
+JitsiConference.prototype._acceptJvbIncomingCall = function(jingleSession, jingleOffer) {
     // Accept incoming call
     this.jvbJingleSession = jingleSession;
-    this.room.connectionTimes['session.initiate'] = now;
+
+    window.performance.mark('jitsi.xmpp.session-initiate');
+
     this._sendConferenceJoinAnalyticsEvent();
 
     if (this.wasStopped) {
@@ -2067,12 +2062,8 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
         serverRegion);
 
     this._maybeClearSITimeout();
-    Statistics.sendAnalytics(createJingleEvent(
-        ACTION_JINGLE_SI_RECEIVED,
-        {
-            p2p: false,
-            value: now
-        }));
+
+    Statistics.sendAnalytics(createJingleEvent(ACTION_JINGLE_SI_RECEIVED, { p2p: false }));
 
     try {
         jingleSession.initialize(this.room, this.rtc, {
@@ -2506,13 +2497,6 @@ JitsiConference.prototype.isStartAudioMuted = function() {
  */
 JitsiConference.prototype.isStartVideoMuted = function() {
     return this.startVideoMuted;
-};
-
-/**
- * Returns measured connectionTimes.
- */
-JitsiConference.prototype.getConnectionTimes = function() {
-    return this.room.connectionTimes;
 };
 
 /**
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -2,6 +2,7 @@ import { getLogger } from 'jitsi-meet-logger';
 
 import RTCEvents from '../../service/RTC/RTCEvents';
 import { createBridgeChannelClosedEvent } from '../../service/statistics/AnalyticsEvents';
+import perfMetrics from '../perf-metrics/perfMetrics';
 import Statistics from '../statistics/statistics';
 import GlobalOnErrorHandler from '../util/GlobalOnErrorHandler';
 
@@ -276,10 +277,16 @@ export default class BridgeChannel {
      * Set events on the given RTCDataChannel or WebSocket instance.
      */
     _handleChannel(channel) {
+        // This handler gets called right after the channel is created, so we can mark the
+        // start here.
+        perfMetrics.markStart(perfMetrics.MEASURES.BRIDGE_CHANNEL_OPEN);
+
         const emitter = this._eventEmitter;
 
         channel.onopen = () => {
             logger.info(`${this._mode} channel opened`);
+
+            perfMetrics.markEnd(perfMetrics.MEASURES.BRIDGE_CHANNEL_OPEN);
 
             // Code sample for sending string and/or binary data.
             // Sends string message to the bridge:
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -22,6 +22,7 @@ import {
     createNoDataFromSourceEvent
 } from '../../service/statistics/AnalyticsEvents';
 import browser from '../browser';
+import perfMetrics from '../perf-metrics/perfMetrics';
 import Statistics from '../statistics/statistics';
 
 import JitsiTrack from './JitsiTrack';
@@ -454,9 +455,17 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * @returns {Promise}
      */
     _queueSetMuted(muted) {
+        const mark = perfMetrics.MEASURES[`${this.getType().toUpperCase()}_${muted ? 'MUTE' : 'UNMUTE'}`];
+
+        perfMetrics.markStart(mark);
+
         const setMuted = this._setMuted.bind(this, muted);
 
         this._prevSetMuted = this._prevSetMuted.then(setMuted, setMuted);
+
+        this._prevSetMuted.then(() => {
+            perfMetrics.markEnd(mark);
+        });
 
         return this._prevSetMuted;
     }
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -209,36 +209,22 @@ export default class JitsiRemoteTrack extends JitsiTrack {
     _playCallback() {
         const type = this.isVideoTrack() ? 'video' : 'audio';
 
-        const now = window.performance.now();
+        window.performance.mark(`jitsi.rtc.${type}.render`);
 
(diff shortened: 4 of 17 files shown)


modules/RTC/BridgeChannel.js
```diff
@@ -276,11 +277,17 @@ export default class BridgeChannel {
...
      */
     _handleChannel(channel) {
+        // This handler gets called right after the channel is created, so we can mark the
+        // start here.
+        perfMetrics.markStart(perfMetrics.METRICS.BRIDGE_CHANNEL_OPEN);
+
         const emitter = this._eventEmitter;
 
         channel.onopen = () => {
             logger.info(`${this._mode} channel opened`);
```
hristoterezov, line 287: Maybe now this log is redundant?

hristoterezov (commented):
LGTM!

hristoterezov (dismissed):

JitsiConference.js
```diff
@@ -2067,12 +2062,8 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(
...
 
     this._maybeClearSITimeout();
-    Statistics.sendAnalytics(createJingleEvent(
-        ACTION_JINGLE_SI_RECEIVED,
-        {
-            p2p: false,
-            value: now
-        }));
+
+    Statistics.sendAnalytics(createJingleEvent(ACTION_JINGLE_SI_RECEIVED, { p2p: false }));
```
hristoterezov, line 2066: Why remove the timestamp from the event?

JitsiConference.js
```diff
@@ -2508,13 +2499,6 @@ JitsiConference.prototype.isStartVideoMuted = function() {
     return this.startVideoMuted;
 };
 
-/**
- * Returns measured connectionTimes.
- */
-JitsiConference.prototype.getConnectionTimes = function() {
-    return this.room.connectionTimes;
```
hristoterezov, line 2515: I see this method is used in jitsi-meet. Aren't we going to brake something? 

I think it will be good to expose those measurements somehow. WDYT? Maybe they can be used for performance testing, etc.

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -209,36 +209,22 @@ export default class JitsiRemoteTrack extends JitsiTrack {
...
-        // The conference can be started without calling GUM
-        // FIXME if there would be a module for connection times this kind
-        // of logic (gumDuration or ttfm) should end up there
-        const gumStart = window.connectionTimes['obtainPermissions.start'];
-        const gumEnd = window.connectionTimes['obtainPermissions.end'];
-        const gumDuration
-            = !isNaN(gumEnd) && !isNaN(gumStart) ? gumEnd - gumStart : 0;
+        // TTFM: calculated as the time from the session-initiate sent by the
+        // bridge until first remote render.
+        window.performance.measure(ttfmMarkName, 'jitsi.xmpp.session-initiate');
```
hristoterezov, line 218: shouldn't the `ttfmMarkName` be the end and the session initiate the start?

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -209,36 +209,22 @@ export default class JitsiRemoteTrack extends JitsiTrack {
...
 
-        // Subtract the muc.joined-to-session-initiate duration because jicofo
-        // waits until there are 2 participants to start Jingle sessions.
-        const ttfm = now
-            - (this.conference.getConnectionTimes()['session.initiate']
-                - this.conference.getConnectionTimes()['muc.joined'])
-            - gumDuration;
```
hristoterezov, line 230: Why we don't need to take into account the gumDuration and muc joined anymore?

modules/perf-metrics/observer.js
```diff
@@ -0,0 +1,70 @@
...
+        if (observer) {
+            dispose();
+        }
+        observer = new PerformanceObserver((list, observer_) => {
+            const metrics = [];
+
+            list.getEntries().forEach(entry => {
+                if (observer === observer_ && entry.name.startsWith(JITSI_METRIC_PREFIX)) {
+                    metrics.push(entry);
+                    logger.debug(`${entry.name.replace(JITSI_METRIC_PREFIX, '')} took ${entry.duration}ms`);
```
hristoterezov, line 29: It will be useful to include the time from navigation start to the end marker as well.

For example during our recent performance testing we were interested in how much time does it take from the navigation start to moment when the session initiate packet is received.

modules/xmpp/JingleSessionPC.js
```diff
@@ -464,15 +465,15 @@ export default class JingleSessionPC extends JingleSession {
...
          */
         this.peerconnection.oniceconnectionstatechange = () => {
-            const now = window.performance.now();
             let isStable = false;
 
             if (!this.isP2P) {
-                this.room.connectionTimes[
-                    `ice.state.${this.peerconnection.iceConnectionState}`]
-                    = now;
+                window.performance.mark(`jitsi.ice.state.${this.peerconnection.iceConnectionState}`);
```
hristoterezov, line 471: Do we use this mark somewhere?

modules/xmpp/JingleSessionPC.js
```diff
@@ -464,15 +465,15 @@ export default class JingleSessionPC extends JingleSession {
...
             }
-            logger.log(`(TIME) ICE ${this.peerconnection.iceConnectionState} ${this.isP2P ? 'P2P' : 'JVB'}:\t`, now);
+
+            logger.log(`ICE ${this.peerconnection.iceConnectionState} ${this.isP2P ? 'P2P' : 'JVB'}`);
```
hristoterezov, line 474: it will be still useful to have `now` printed. Especially since we still use it.

modules/xmpp/xmpp.js
```diff
@@ -271,13 +271,11 @@ export default class XMPP extends Listenable {
...
      */
     connectionHandler(credentials = {}, status, msg) {
-        const now = window.performance.now();
         const statusStr = Strophe.getStatusString(status).toLowerCase();
 
-        this.connectionTimes[statusStr] = now;
-        logger.log(
-            `(TIME) Strophe ${statusStr}${msg ? `[${msg}]` : ''}:\t`,
-            now);
+        window.performance.mark(`jitsi.xmpp.strophe.${statusStr}`);
```
hristoterezov, line 276: do we use this mark?

service/statistics/AnalyticsEvents.js
```diff
@@ -565,3 +565,17 @@ export const createBridgeChannelClosedEvent = function(code, reason) {
...
 };
+
+/**
+ * Creates an event with performance metrics information. Each element in the metrics
+ * array will have a name and duration.
+ *
+ * @param {Array<object>} metrics - Array of metrics objects.
+ */
+export const createPerfMetricsEvent = function(metrics) {
+    return {
```
hristoterezov, line 576: I'm not sure if sending separate events per metric entity would be better. I'm not really sure if we would be able to filter those metrics in Amplitude since all the info (name, timestamp, etc) is in the array. 

Maybe we can send separate events with the action field the name of the mark and then all the values? WDYT?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -454,10 +455,18 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+        const mark = perfMetrics.MEASURES[`${this.getType().toUpperCase()}_${muted ? 'MUTE' : 'UNMUTE'}`];
+
+        perfMetrics.markStart(mark);
+
         const setMuted = this._setMuted.bind(this, muted);
 
         this._prevSetMuted = this._prevSetMuted.then(setMuted, setMuted);
 
+        this._prevSetMuted.then(() => {
+            perfMetrics.markEnd(mark);
```
hristoterezov, line 467: Kudos for introducing this one!

modules/xmpp/strophe.jingle.js
```diff
@@ -314,7 +314,7 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
 
         switch (action) {
         case 'session-initiate': {
-            logger.log('(TIME) received session-initiate:\t', now);
+            logger.log('Received session-initiate');
```
hristoterezov, line 317: Actually this timestamp was very useful during the last performance testing we were performing. It will be good to have it!

hristoterezov (changes requested):
Great work!

Left some questions and suggestions!

As an additional suggestion, I'm wondering if it is possible to encapsulate all the use cases that we have in the perf-metrics module so that we don't need to use the Performance API outside of it. And do you think it makes sense? 

For example measuring the time between 2 different marks and extracting the result. You use this for TTFM.

modules/RTC/BridgeChannel.js
saghul: It's still good to know, regardless of the measurement, IMHO.

JitsiConference.js
saghul: Because it was never a timestamp :-) performance.now() does not return timestamps, it produces monotonically increasing marks from an arbitrary point in time, so only the difference between them is relevant. Also, some actions had it some didn't so I opted for that. Last, we are not actually using this event so its usefulness is questionable I guess.

JitsiConference.js
saghul: I plan to fix JM too. The `perfMetrics` object is exposed in JitsiMeetJS, but the metrics can be obtained with performance.getEntries(). I can create a helper for that.

modules/RTC/JitsiRemoteTrack.js
saghul: No, the API is `performance.measure(name, startMarkName, endMarkName)` if the end mark name is omitted then `now()` is taken.

modules/perf-metrics/observer.js
saghul: Do you think that would be useful wholesale or for specific metrics?

modules/xmpp/JingleSessionPC.js
saghul: Not really, but I didn't want to remove it. We may want to measure the time in between phases I guess? There is some of that already going on, but it's a bit involved and I didn't want to fall into that rabbit-hole :-P

modules/xmpp/JingleSessionPC.js
saghul: As I mentioned in https://github.com/jitsi/lib-jitsi-meet/pull/1707#discussion_r706075330 that now() is not a timestamp. Also, the logger itself shows a timestamp, do we need another?

modules/xmpp/xmpp.js
saghul: Not at the moment.

modules/xmpp/strophe.jingle.js
saghul: How were you using it? I'm genuinely curious, because performance.now() is not a timestamp, so in isolation it doesn't add much! Maybe I can add a metric about start to session-initiate?

service/statistics/AnalyticsEvents.js
saghul: Right now metrics will come in batches, not sure if 1 event per metric would be overkill. But doable. If we opt for individual events I can flatten the thing and make it like so:

```
    return {
        type: TYPE_OPERATIONAL,
        action: 'perf-metrics',
        metric: 'jitsi.foo.bar',
        duration: xxxxx
    };
```

Do you think we should go for that?

saghul:
Thanks for the thorough review!

> As an additional suggestion, I'm wondering if it is possible to encapsulate all the use cases that we have in the perf-metrics module so that we don't need to use the Performance API outside of it. And do you think it makes sense?
> 
> For example measuring the time between 2 different marks and extracting the result. You use this for TTFM.

I asked myself this question :-) I chose to create the smallest possible helper, but I do see the value in making it a bit more "integrated". I'll give it a try in a separate commit to see how it feels like.

PR #1709, opened 2021-09-10
ref(JitsiConference): don't crash on wrong oldTrack

paweldomas:
If oldTrack was not previously in the conference this will
lead to a crash in onLocalTrackRemoved where oldTrack doesn't
have `muteHandler` and `audioLevelHandler` listeners defined.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1143,8 +1143,10 @@ JitsiConference.prototype.removeTrack = function(track) {
  * @returns {Promise} resolves when the replacement is finished
  */
 JitsiConference.prototype.replaceTrack = function(oldTrack, newTrack) {
+    const oldTrackBelongsToConference = this === oldTrack?.conference;
+
     // First do the removal of the oldTrack at the JitsiConference level
-    if (oldTrack) {
+    if (oldTrackBelongsToConference) {
         if (oldTrack.disposed) {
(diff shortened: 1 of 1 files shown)


paweldomas:
It's either that or the method should throw a meaningful error. @saghul @jallamsetty1 WDYT?
Allowing wrong old track to not crash makes it easier to deal with corner cases from https://github.com/jitsi/jitsi-meet/pull/9920

saghul:
Maybe we can just print a warning as in "this may be fine if you know what you're doing"

JitsiConference.js
```diff
@@ -1143,8 +1143,10 @@ JitsiConference.prototype.removeTrack = function(track) {
  * @returns {Promise} resolves when the replacement is finished
  */
 JitsiConference.prototype.replaceTrack = function(oldTrack, newTrack) {
+    const oldTrackBelongsToConference = this === oldTrack?.conference;
+
     // First do the removal of the oldTrack at the JitsiConference level
-    if (oldTrack) {
+    if (oldTrackBelongsToConference) {
```
jallamsetty1, line 1149: Can we print an error if the oldTrack is not null and it doesn't belong to the conference ?

jallamsetty1 (approved):

PR #1711, opened 2021-09-14
Rename iceConfig to pcConfig and other logging enhancements

jallamsetty1:
Please look at the individual commits for more information.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -3,6 +3,7 @@
 import { getLogger } from 'jitsi-meet-logger';
 import { $iq, Strophe } from 'strophe.js';
 
+import * as MediaType from '../../service/RTC/MediaType';
 import {
     ACTION_JINGLE_TR_RECEIVED,
     ACTION_JINGLE_TR_SUCCESS,
@@ -31,10 +32,10 @@ const logger = getLogger(__filename);
  *
  * @param {*} iq the IQ to which source descriptions will be added.
  * @param {*} jsonMessageXml The XML node for the "json-message" element.
- * @returns nothing.
+ * @returns {Map<string, Array<string>} The audio and video ssrcs extracted from the JSON-encoded message with remote
+ * endpoint id as the key.
  */
 function expandSourcesFromJson(iq, jsonMessageXml) {
-
     let json;
 
     try {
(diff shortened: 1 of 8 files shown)


modules/xmpp/strophe.jingle.js
```diff
@@ -354,20 +373,24 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
...
             break;
         }
         case 'transport-info': {
+            logger.debug(`Received ${action} from ${fromJid}`, iq);
             this.eventEmitter.emit(
                 XMPPEvents.TRANSPORT_INFO, sess, $(iq).find('>jingle'));
             break;
         }
         case 'session-terminate': {
+            logger.debug(`Received ${action} from ${fromJid}`, iq);
```
paweldomas, line 393: have to be careful with logging objects on mobile

jallamsetty1: This has always been the case before, instead of logging every IQ, I am just logging the IQs for all messages except `session-initiate`, `source-add` and `source-remove`.

modules/xmpp/strophe.jingle.js
```diff
@@ -331,7 +331,7 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
                 `Marking session from ${fromJid
                 } as ${isP2P ? '' : '*not*'} P2P`);
 
-            const iceConfig = isP2P ? this.p2pIceConfig : this.jvbIceConfig;
+            const pcConfig = isP2P ? this.p2pIceConfig : this.jvbIceConfig;
```
saghul, line 353: Do we not want to rename `this.p2pIceConfig` and `this.jvbIceConfig` ? Me thinks `this.pcConfig.p2p` and `this.pcConfig.jvb` looks moar better :-P

modules/xmpp/strophe.jingle.js
```diff
@@ -407,12 +431,10 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
                 sess.sendTransportReject();
             });
             break;
-        case 'addsource': // FIXME: proprietary, un-jingleish
```
saghul, line 410: nice!

saghul (commented):
Left some comments, thanks for doing this!

modules/xmpp/strophe.jingle.js
jallamsetty1: Actually, `this.p2pIceConfig `and `this.jvbIceConfig` have the ice servers info, that is why I didn't change them.

saghul: Oh, nevermind then!

modules/xmpp/strophe.jingle.js
jallamsetty1: Done.

saghul (approved):
Beautiful!

PR #1715, opened 2021-09-16
chore(deps): bump sdp-interop to get another ICE restart fix

nils-ohlmeier:


saghul:
You need to run npm install so package-lock is updated too

nils-ohlmeier:
I don't get why the change to package-lock.json doesn't show up. Locally I see it. Abandoning to open a new clean PR instead.

PR #1718, opened 2021-09-20
Task/add external key handler

tmoldovan8x8:


modules/e2ee/Context.js
```diff
@@ -39,7 +39,7 @@ export class Context {
     /**
      * @param {string} id - local muc resourcepart
```
saghul, line 40: Please remove the stray comment.

modules/e2ee/Context.js
```diff
@@ -80,10 +82,11 @@ export class Context {
      */
     _setKeys(keys, keyIndex = -1) {
         if (keyIndex >= 0) {
-            this._cryptoKeyRing[keyIndex] = keys;
-        } else {
-            this._cryptoKeyRing[this._currentKeyIndex] = keys;
+            this._currentKeyIndex = keyIndex % this._cryptoKeyRing.length;
```
saghul, line 85: This is duplicated from `setKeyBytes`, let's unify the logic somehow.

modules/e2ee/Context.js
```diff
@@ -250,17 +253,25 @@ export class Context {
 
             encodedFrame.data = newData;
         } catch (error) {
-            if (ratchetCount < RATCHET_WINDOW_SIZE) {
-                material = await importKey(await ratchet(material));
+            if (!this._externallyManaged) {
```
saghul, line 256: You can reverse the if condition, return early, and save an indentation level.

modules/e2ee/Context.js
```diff
@@ -250,17 +253,25 @@ export class Context {
...
+                if (ratchetCount === 0) {
```
saghul, line 257: Please leave a thick comment above this, explaining why we are doing it that way, but more importantly, this is not really related to the spirit of this PR, so let's make it its own commit at least.

modules/e2ee/E2EEContext.js
```diff
@@ -133,12 +133,29 @@ export default class E2EEcontext {
...
         this._worker.postMessage({
-            operation: 'setKey',
+            operation: 'setKeyBytes',
             participantId,
             key,
             keyIndex
         });
     }
+
+    setKey(participantId, key, keyIndex, unique = false) {
```
saghul, line 145: dcstring

modules/e2ee/E2EEContext.js
```diff
@@ -133,12 +133,29 @@ export default class E2EEcontext {
...
+        this._worker.postMessage({
+            operation: 'setKey',
+            key,
+            keyIndex,
+            participantId,
+            unique
+        });
+    }
+
+    setKeyManagementMode(externallyManaged) {
```
saghul, line 155: Maybe we can call this "initialize" instead, and pass an options object, so we have some extensibility for the future.

JitsiConference.js
```diff
@@ -260,7 +260,7 @@ export default function JitsiConference(options) {
     if (this.isE2EESupported()) {
         logger.info('End-to-End Encryprtion is supported');
 
-        this._e2eEncryption = new E2EEncryption(this);
+        this._e2eEncryption = new E2EEncryption(this, this.options.config.externallyManagede2ee);
```
saghul, line 263: We don't need to pass it explicitly, we can read it from the conference.

modules/e2ee/E2EEncryption.js
```diff
@@ -96,10 +32,13 @@ export class E2EEncryption {
      * @returns {boolean}
      */
     static isSupported(config) {
+        if (!config.externallyManagede2ee && !OlmAdapter.isSupported()) {
```
saghul, line 35: Maybe we should create a dedicated "e2ee" section instead. With a "mode" option.

modules/e2ee/E2EEncryption.js
```diff
@@ -118,238 +57,10 @@ export class E2EEncryption {
...
-     * @private
-     */
-    _trackMuteChanged(track) {
-        if (browser.doesVideoMuteByStreamRemove() && track.isLocal() && track.isVideoTrack() && !track.isMuted()) {
-            for (const session of this.conference._getMediaSessions()) {
-                this._setupSenderE2EEForTrack(session, track);
-            }
-        }
+    setEncryptionKey(keyInfo) {
+        this._keyHandler.setKey && this._keyHandler.setKey(keyInfo);
```
saghul, line 64:
```suggestion
        this._keyHandler?.setKey(keyInfo);
```

modules/e2ee/KeyHandler.js
```diff
@@ -0,0 +1,173 @@
...
+
+        this.enabled = enabled;
+
+        if (!enabled) {
+            for (const participant of this.conference.getParticipants()) {
+                this.e2eeCtx.cleanup(participant.getId());
+            }
+        }
+
+        await this.setEnabledExtras(enabled);
```
saghul, line 79: Maybe we can call this `_setEnabled`.

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -0,0 +1,185 @@
...
+        }
+
+        // Generate a random key in case we are enabling.
+        this._key = enabled ? this._generateKey() : false;
+
+        // Send it to others using the E2EE olm channel.
+        const index = await this._olmAdapter.updateKey(this._key);
+
+        // Set our key so we begin encrypting.
+        this.e2eeCtx.setKeyBytes(this.conference.myUserId(), this._key, index ?? 0);
```
saghul, line 78: Can you check in which case we'd lack the index?

modules/e2ee/OlmAdapter.js
```diff
@@ -626,3 +646,5 @@ function safeJsonParse(data) {
         return {};
     }
 }
+
+OlmAdapter.events = OlmAdapterEvents;
```
saghul, line 650: Do we use these events outside of the adapter anymore? If not let's get rid of them!

modules/e2ee/Worker.js
```diff
@@ -7,18 +7,29 @@ import { Context } from './Context';
 
 const contexts = new Map(); // Map participant id => context
 
+let singleKey = false;
```
saghul, line 10: We don't need this. We can check if there is a shared context and infer it.

modules/e2ee/Worker.js
```diff
@@ -7,18 +7,29 @@ import { Context } from './Context';
...
  * @param {string} participantId - The participant whose context we need.
  * @returns {Object} The context.
  */
 function getParticipantContext(participantId) {
-    if (!contexts.has(participantId)) {
-        contexts.set(participantId, new Context(participantId));
-    }
+   if (singleKey) {
+       if (!singleContext) {
+          singleContext = new Context();
```
saghul, line 22: Instead of doing it here, implicitly, let's be explicit and have a `initialize` method, which sets it  up.

modules/e2ee/Worker.js
```diff
@@ -52,15 +63,22 @@ onmessage = async event => {
         const context = getParticipantContext(participantId);
 
         handleTransform(context, operation, readableStream, writableStream);
-    } else if (operation === 'setKey') {
+    } else if (operation === 'setKeyBytes') {
```
saghul, line 66: This looks almost a carbon copy of setKey. How about this: a subclass of Context, called SharedContext, which overrides setKey, does the needful, and calls _setKey.

modules/e2ee/Context.js
```diff
@@ -39,7 +39,7 @@ export class Context {
     /**
      * @param {string} id - local muc resourcepart
      */
-    constructor(id) {
+    constructor(externallyManaged = false) {
```
saghul, line 42: We could create the context with shared keys as follows: `new Context({ shareKey: true });

Then, internally, we wouldn't derive any keys or ratchet. We just use what we are given in setKey.

saghul (commented):
Round 1!

modules/e2ee/E2EEContext.js
```diff
@@ -24,7 +24,7 @@ export default class E2EEcontext {
     /**
      * Build a new E2EE context instance, which will be used in a given conference.
      */
-    constructor() {
+    constructor(options) {
```
saghul, line 27: Document, maybe provide a default value?

modules/e2ee/E2EEContext.js
```diff
@@ -45,6 +45,13 @@ export default class E2EEcontext {
 
         this._worker = new Worker(blobUrl, { name: 'E2EE Worker' });
         this._worker.onerror = e => logger.onerror(e);
+
+        const { shareKey } = options;
```
saghul, line 49: sharedKey

modules/e2ee/E2EEncryption.js
```diff
@@ -118,238 +60,16 @@ export class E2EEncryption {
...
-     * @param {JitsiLocalTrack} track - the track for which muted status has changed.
-     * @private
-     */
-    _trackMuteChanged(track) {
-        if (browser.doesVideoMuteByStreamRemove() && track.isLocal() && track.isVideoTrack() && !track.isMuted()) {
-            for (const session of this.conference._getMediaSessions()) {
-                this._setupSenderE2EEForTrack(session, track);
-            }
-        }
+     * @param {Object} keyInfo whether to enable E2EE or not.
```
saghul, line 69: Let's describe the layout of the object please.

modules/e2ee/KeyHandler.js
```diff
@@ -0,0 +1,168 @@
...
+
+        this._enabling && await this._enabling;
+
+        this._enabling = new Deferred();
+
+        this.enabled = enabled;
+
+        if (!enabled) {
+            for (const participant of this.conference.getParticipants()) {
+                this.e2eeCtx.cleanup(participant.getId());
```
saghul, line 78: Maybe we can take the opportunity and add some `cleanupAll` here?

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -0,0 +1,183 @@
...
+        this._key = undefined;
+        this._conferenceJoined = false;
+
+        this._olmAdapter = new OlmAdapter(conference);
+
+        this._rotateKey = debounce(this._rotateKeyImpl, DEBOUNCE_PERIOD);
+        this._ratchetKey = debounce(this._ratchetKeyImpl, DEBOUNCE_PERIOD);
+
+        // Olm signalling events.
+        this._olmAdapter && this._olmAdapter.on(
```
saghul, line 37: no need to check for olmAdapter, it will exist for sure, you just created it a couple of lines above

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -0,0 +1,183 @@
...
+    async _onParticipantPropertyChanged(participant, name, oldValue, newValue) {
+        switch (name) {
+        case 'e2ee.idKey':
+            logger.debug(`Participant ${participant.getId()} updated their id key: ${newValue}`);
+            break;
+        case 'e2ee.enabled':
+            if (!newValue && this.enabled) {
+                this._olmAdapter && this._olmAdapter.clearParticipantSession(participant);
+
+                this._rotateKey();
```
saghul, line 99: Not related to this change, but can you please remind me: why did we do this? Looks like we are rotating the key when a participant disables e2ee while we have it enabled? Why is that?

modules/e2ee/Worker.js
```diff
@@ -59,7 +71,7 @@ onmessage = async event => {
         if (key) {
             context.setKey(key, keyIndex);
         } else {
-            context.setKey(false, keyIndex);
+            context.setKey((false, keyIndex));
```
saghul, line 74: Is this right?

modules/e2ee/OlmAdapter.js
saghul: Do we still need these then?

tmoldovan8x8: Yep, it is still used in ManagedKeyHandler

JitsiConference.js
```diff
@@ -3661,6 +3661,19 @@ JitsiConference.prototype.toggleE2EE = function(enabled) {
     this._e2eEncryption.setEnabled(enabled);
 };
 
+/**
+ * Sets the key for End-to-End encryption.
+ *
+ * @param keyInfo: {Object {
+                encryptionKey: CryptoKey
```
saghul, line 3668: Check how we document other inner options, usually it's a param line for each, with `keyInfo.XXX` as each item.

modules/e2ee/Context.js
```diff
@@ -260,9 +268,18 @@ export class Context {
                 return await this._decryptFrame(
                     encodedFrame,
                     keyIndex,
+                    initialKey || this._cryptoKeyRing[this._currentKeyIndex],
                     ratchetCount + 1);
             }
 
+            /*
+               Since the key it is first send and only afterwards actually used for encrypting, there were
```
saghul, line 276: Can you please format this log with a `*` at the beginning of each line? LIke our other multiline logs please.

modules/e2ee/E2EEContext.js
```diff
@@ -45,6 +45,13 @@ export default class E2EEcontext {
 
         this._worker = new Worker(blobUrl, { name: 'E2EE Worker' });
         this._worker.onerror = e => logger.onerror(e);
+
+        const { sharedKey } = options;
```
saghul, line 49: Maybe destructure this in the constructor line?

modules/e2ee/E2EEContext.js
```diff
@@ -24,7 +24,7 @@ export default class E2EEcontext {
     /**
      * Build a new E2EE context instance, which will be used in a given conference.
      */
-    constructor() {
+    constructor(options = {}) {
```
saghul, line 27: Document `options`

modules/e2ee/E2EEncryption.js
```diff
@@ -118,238 +58,19 @@ export class E2EEncryption {
...
-     * @private
-     */
-    _trackMuteChanged(track) {
-        if (browser.doesVideoMuteByStreamRemove() && track.isLocal() && track.isVideoTrack() && !track.isMuted()) {
-            for (const session of this.conference._getMediaSessions()) {
-                this._setupSenderE2EEForTrack(session, track);
-            }
-        }
+     * @param keyInfo: {Object {
+                encryptionKey: CryptoKey
```
saghul, line 68: Ditto log formatting.

modules/e2ee/E2EEncryption.js
```diff
@@ -118,238 +58,19 @@ export class E2EEncryption {
...
+                index: Number
+        }}
+     * @returns {void}
+    */
+    setEncryptionKey(keyInfo) {
```
saghul, line 73: Checking if setKey is available reads a bit weird. How about this: let's make the base class throw an error that says "not implemented". Olm won't override it. The other one will. This way someone will notice if they are using the API wrong.

modules/e2ee/ExternallyManagedKeyHandler.js
```diff
@@ -0,0 +1,26 @@
...
+     * Sets the key for End-to-End encryption.
+     *
+     * @param keyInfo: {Object {
+                encryptionKey: CryptoKey
+                index: Number
+        }}
+     * @returns {void}
+    */
+    setKey(keyInfo) {
+        this.e2eeCtx.setKey(undefined, { encryptionKey: keyInfo.encryptionKey }, keyInfo.index);
```
saghul, line 24: Shouldn't the index be part of the second argument?

modules/e2ee/KeyHandler.js
```diff
@@ -0,0 +1,166 @@
...
+const logger = getLogger(__filename);
+
+/**
+ * Abstract class that integrates {@link E2EEContext} with a key management system.
+ */
+export class KeyHandler extends Listenable {
+    /**
+     * Build a new KeyHandler instance, which will be used in a given conference.
+     */
+    constructor(conference, options = {}) {
```
saghul, line 22: Please document options.

saghul (changes requested):
Did another round!

modules/e2ee/ExternallyManagedKeyHandler.js
tmoldovan8x8: It is the third parameter, used from different places, I would not change it for this.

saghul: Ah right! I was mislead by the fact that we have more than one item in the key object in the olm case.

saghul:
There seems to be a conflict, can you rebase?

saghul:
Please rebase again... sorry my bad this time.

tmoldovan8x8:

PR #1719, opened 2021-09-20
Fix codec renegotiation and reverse the order of the ssrcs being signaled from Firefox to Jicofo.

jallamsetty1:
Impose VP9 bitrates only when VP9 is negotiated. If Jicofo doesn't offer VP9 but the client expresses a preference for VP9, VP9 bitrates were being imposed before.
This fixes an issue where the bridge doesn't forward the HD stream from Firefox to other users in the call. The order of the ssrcs produced by the browser is from Highest resolution to lowest whereas the bridge assumes it to be from lowest to highest as is the case with Chrome and Safari.

### DIFF
--- a/modules/sdp/SDP.js
+++ b/modules/sdp/SDP.js
@@ -143,10 +143,16 @@ SDP.prototype.toJingle = function(elem, thecreator) {
         }
 
         let ssrc;
+        const simGroup = SDPUtil.findLine(this.media[i], 'a=ssrc-group:SIM');
         const assrcline = SDPUtil.findLine(this.media[i], 'a=ssrc:');
 
         if (assrcline) {
-            ssrc = assrcline.substring(7).split(' ')[0]; // take the first
+            if (browser.isFirefox() && simGroup) {
+                // Use the first ssrc from the SIM group since the order of the SSRCs need to be reversed for Firefox.
+                ssrc = simGroup.substring(17).split(' ')[0];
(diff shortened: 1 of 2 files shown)


modules/sdp/SDP.js
```diff
@@ -143,9 +143,14 @@ SDP.prototype.toJingle = function(elem, thecreator) {
...
         let ssrc;
+        const simGroup = SDPUtil.findLine(this.media[i], 'a=ssrc-group:SIM');
         const assrcline = SDPUtil.findLine(this.media[i], 'a=ssrc:');
 
         if (assrcline) {
+            if (browser.isFirefox() && simGroup) {
+                // Use the first ssrc from the SIM group since the order of the SSRCs need to be reversed for Firefox.
+                ssrc = simGroup.substring(17).split(' ')[0];
+            }
             ssrc = assrcline.substring(7).split(' ')[0]; // take the first
```
JonathanLennox, line 154: Don't you need an `else` here?  It looks like the second `ssrc` assignment overrides the first.

jallamsetty1: 🤦‍♀️ I thought I did that but wonder how the test I ran was showing the correct ssrcs in the signaling ..

jallamsetty1: @JonathanLennox fixed it.

damencho (approved):

PR #1721, opened 2021-09-21
fix(e2ee) restore initial key when RATCHET_WINDOW_SIZE reached

tmoldovan8x8:


modules/e2ee/Context.js
```diff
@@ -263,6 +275,8 @@ export class Context {
                     ratchetCount + 1);
             }
 
+            this._setKeys(this._initialKey);
+
```
saghul, line 279: let's reset the key here then?

saghul (dismissed):
LGTM with a suggestion

saghul (approved):

PR #1724, opened 2021-09-23
feat: add facial-expressions in speaker stats

gabiborlea:
Receives facial expressions and their duration via xmpp or data channel then it stores them in speaker stats


damencho:
Also, if you haven't added dependencies in package.json the lock file should not change. But I see the opposite in your two PRs. 
Probably you are using too new npm, @saghul which version of npm, 6 for now, right?

quitrk (dismissed):

quitrk (approved):

andrei-gavrilescu (approved):

PR #1725, opened 2021-09-23
feat: generates source names

paweldomas:
This is the first step in adding support for multiple
streams per endpoint. Each source(stream?) needs to have
an identifier. For now always generate the 0 index name
until the machinery for sending more than 1 stream is put
in place.

### DIFF
--- a/modules/sdp/LocalSdpMunger.js
+++ b/modules/sdp/LocalSdpMunger.js
@@ -309,17 +309,53 @@ export default class LocalSdpMunger {
 
         if (audioMLine) {
             this._transformMediaIdentifiers(audioMLine);
+            this._injectSourceNames(audioMLine);
         }
 
         const videoMLine = transformer.selectMedia('video');
 
         if (videoMLine) {
             this._transformMediaIdentifiers(videoMLine);
+            this._injectSourceNames(videoMLine);
         }
 
         return new RTCSessionDescription({
             type: sessionDesc.type,
             sdp: transformer.toRawSDP()
         });
+    }
+
+    /**
+     * Injects source names. Source names are need to for multiple streams per endpoint support. The final plan is to
+     * use the "mid" attribute for source names, but because the SDP to Jingle conversion still operates in the Plan-B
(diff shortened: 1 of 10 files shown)


paweldomas:
I'm wondering if this should be behind a feature flag already?

modules/sdp/SDPUtil.js
```diff
@@ -261,6 +261,29 @@ const SDPUtil = {
...
     },
+
+    /**
+     * Gets the source name out of the name attribute "a=ssrc:254321 name:name1" and removes it from
+     * the {@code ssrcLines} array, so that it doesn't leak further up as source>parameter element in jingle.
+     *
+     * @param {string[]} ssrcLines
+     * @returns {string | undefined}
+     */
+    getAndRemoveSourceNameLine(ssrcLines) {
```
paweldomas, line 272: I'm going to get rid of the removal side effect and actually filter out all parameters except for `msid` during the to jingle conversion

modules/sdp/LocalSdpMunger.js
```diff
@@ -309,17 +309,53 @@ export default class LocalSdpMunger {
 
         if (audioMLine) {
             this._transformMediaIdentifiers(audioMLine);
+            this._injectSourceNames(audioMLine);
```
jallamsetty1, line 312: Don't we want to do this only for Unified plan implementation of the client ?

paweldomas: No. The plan-B client even if won't support sending more than one source at the same time, should still use the new signaling format.

jallamsetty1: In that case, can we put this behind a flag so that we can have it disabled it on our prod environments. Also will older version of Jicofo throw a parsing error if it is talking to a new client ?

paweldomas: I can put it behind the flag.

No, Jicofo will not throw an error because it's just one extra piece of information that it's going to include if available or ignore otherwise.

jallamsetty1 (approved):

PR #1732, opened 2021-09-29
feat(share-tab-audio): Add share audio flag on desktop share.

tudor-phd:


modules/RTC/ScreenObtainer.js
```diff
@@ -214,9 +214,10 @@ const ScreenObtainer = {
             getDisplayMedia = navigator.mediaDevices.getDisplayMedia.bind(navigator.mediaDevices);
         }
 
-        const { desktopSharingFrameRate } = this.options;
+        const { desktopSharingFrameRate, desktopSharingAudio } = this.options;
         const video = typeof desktopSharingFrameRate === 'object' ? { frameRate: desktopSharingFrameRate } : true;
-        const audio = this._getAudioConstraints();
+        // eslint-disable-next-line no-negated-condition
+        const audio = typeof desktopSharingAudio !== 'undefined' ? desktopSharingAudio : this._getAudioConstraints();
```
paweldomas, line 220: what if `desktopSharingAudio` is set to `flase`?

damencho:
Jenkins add to whitelist.

modules/RTC/ScreenObtainer.js
jallamsetty1, line 220: Why are adding a new `desktopSharingAudio` option ? Doesn't this affect audio-only screenshare feature ? Are making sure this option is set to true for that feature ?

jallamsetty1:
@tudor-phd What happens when the audio constraint is set to false for getDisplayMedia() ? Does the share audio checkbox not show up in the screen selection dialog ? If that is the case are you planning to pass this new option set to false while creating a desktop track for virtual background ?

tudor-phd:
> @tudor-phd What happens when the audio constraint is set to false for getDisplayMedia() ? Does the share audio checkbox not show up in the screen selection dialog ? If that is the case are you planning to pass this new option set to false while creating a desktop track for virtual background ?

Yes and yes
Screenshot 2021-09-29 at 18 22 09
.

tudor-phd:
> @tudor-phd What happens when the audio constraint is set to false for getDisplayMedia() ? Does the share audio checkbox not show up in the screen selection dialog ? If that is the case are you planning to pass this new option set to false while creating a desktop track for virtual background ?

Also if the value will be undefined/not specified it will be ignored and set to true by default. The remaining cases are setting to: true / false
Example of use:
Screenshot 2021-09-30 at 09 12 40
Screenshot 2021-09-30 at 09 13 13
:

modules/RTC/ScreenObtainer.js
```diff
@@ -214,9 +214,10 @@ const ScreenObtainer = {
             getDisplayMedia = navigator.mediaDevices.getDisplayMedia.bind(navigator.mediaDevices);
         }
 
-        const { desktopSharingFrameRate } = this.options;
+        const { desktopSharingFrameRate, desktopSharingAudio } = this.options;
         const video = typeof desktopSharingFrameRate === 'object' ? { frameRate: desktopSharingFrameRate } : true;
-        const audio = this._getAudioConstraints();
+        // eslint-disable-next-line no-negated-condition
```
saghul, line 219: Please reverse this instead of disabling the eslint rule.

JitsiMeetJS.js
```diff
@@ -301,9 +303,12 @@ export default _mergeNamespaceAndModule({
     createLocalTracks(options = {}, oldfirePermissionPromptIsShownEvent) {
         let promiseFulfilled = false;
 
-        const { firePermissionPromptIsShownEvent, fireSlowPromiseEvent, ...restOptions } = options;
+        const { firePermissionPromptIsShownEvent, fireSlowPromiseEvent, desktopSharingAudio, ...restOptions } = options;
         const firePermissionPrompt = firePermissionPromptIsShownEvent || oldfirePermissionPromptIsShownEvent;
 
+        if (typeof desktopSharingAudio !== undefined) {
```
saghul, line 309: Why is this necessary here? IIRC we initialize it in RTC.js already.

saghul (commented):
Left a comment.

JitsiMeetJS.js
tudor-phd: I saw that in RTCUtils.js we call screenObtainer.init(options); -> but looks like we pass all options at the start -> then we want to change an option. This is the reason that I call again screenObtainer.init(options); with the new desktopSharingAudio option. @jallamsetty1 What do you think about this? Would you have another approach to fix this?

jallamsetty1: But doesn't it mean that once someone sets their screenshare as their virtual background, this audio share option won't be available for regular screensharing and audio-only share ?

tudor-phd: No. Already test that. For example, I tried to apply the option just on the virtual background. Worked. After that, I wanted to test this case that you specified and I clicked the screen share option from the bottom toolbar to see the behaviour on browser tab sharing. The audio share checkbox was there.

jallamsetty1: I think we should find a better way or workaround for this change in Chrome's behavior where the audio share gets checked by default whenever audio is requested as this affects regular screenshare as well.

saghul:
> > I won't block this PR because Jaya approved it, but I will state my strong -1.
> > 
> > That option bears no meaning in lib-jitsi-meet. Virtual backgrounds are part of JM, all this library needs to offer is a way for the user of the API to specify some audio constraints.
> > 
> > As it stands it's a leaky abstraction.
> > 
> > If we come up with another case which needs audio sharing disabled (maybe presenter mode?) the flag name would be even more confusing.
> 
> In that case, can we make "audio: false" the default constraints and pass option (shareDesktopAudio or something similar) for explicitly enabling it ?

This is exactly how this should be implemented IMHO.

PR #1739, opened 2021-10-04
ref(TPC) Configure encodings in one go.

jallamsetty1:
Configure the active state, maxBitrate and degradation preference in one go instead of configuring them separately.


modules/RTC/TPCUtils.js
```diff
@@ -273,29 +274,48 @@ export class TPCUtils {
...
+        const encodingsState = this.localStreamEncodingsConfig
+        .map(encoding => height / encoding.scaleResolutionDownBy)
+        .map((frameHeight, idx) => {
+            let active = localVideoTrack.getVideoType() === VideoType.CAMERA
+
+                // Keep the LD stream enabled even when the LD stream's resolution is higher than of the requested
+                // resolution. This can happen when camera is captured at resolutions higher than 720p but the
+                // requested resolution is 180. Since getParameters doesn't give us information about the resolutions
+                // of the simulcast encodings, we have to rely on our initial config for the simulcast streams.
+                ? newHeight > 0 && this.localStreamEncodingsConfig[idx]?.scaleResolutionDownBy === 4.0
```
paweldomas, line 297: Is the problem that JVB sends constraints based off 720p being the max?
Can you give one example that the exception handles?

Another question:
The idx can't be used to tell which layer is LD, SD or HD, because it's reversed for Firefox?
Referencing quality layers through the scale down factor as a number is hard to read. I'm wondering if constants like "SD_SCALE_FACTOR", "HD_SCALE_FACTOR" would make it easier?

modules/RTC/TPCUtils.js
```diff
@@ -273,29 +274,48 @@ export class TPCUtils {
...
+                : true;
+
+            // Disable the lower spatial layers for screensharing in Unified plan when low fps screensharing is in
+            // progress. There is no way to enable or disable simulcast during the call since we are re-using the same
+            // sender. Safari is an exception here since it does not send the desktop stream at all if only the high
+            // resolution stream is enabled.
+            if (this.pc.isSharingLowFpsScreen()
+                && this.pc.usesUnifiedPlan()
+                && !browser.isWebKitBased()
+                && this.localStreamEncodingsConfig[idx].scaleResolutionDownBy !== 1.0) {
```
paweldomas, line 311: What happens if other layers are not disabled? Why do they need to be disabled exactly?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2568,48 +2432,77 @@ TraceablePeerConnection.prototype.setSenderVideoConstraint = function(frameHeigh
...
+                // Firefox doesn't follow the spec and lets application specify the degradation preference on the
+                // encodings.
+                browser.isFirefox() && (parameters.encodings[encoding].degradationPreference = preference);
+
+                // Max bitrates are configured on the encodings only for VP8.
+                if (this.getConfiguredVideoCodec() === CodecMimeType.VP8
+                    && (this.options?.videoQuality?.maxBitratesVideo
+                        || this.isSharingLowFpsScreen()
+                        || this._usesUnifiedPlan)) {
+                    parameters.encodings[encoding].maxBitrate = this.isSharingLowFpsScreen() && !browser.isWebKitBased()
```
paweldomas, line 2464: can the max bitrate calculation be extracted to a separate method just like the encoding's active states are?

modules/xmpp/JingleSessionPC.js
```diff
@@ -2110,14 +2082,12 @@ export default class JingleSessionPC extends JingleSession {
...
                             logger.debug(`${this} replaceTrack worker: configuring video stream`);
 
-                            // FIXME set all sender parameters in one go?
-                            // Set the degradation preference on the new video sender.
-                            return this.peerconnection.setSenderVideoDegradationPreference()
-                                .then(() => this.peerconnection.setSenderVideoConstraint())
-                                .then(() => this.peerconnection.setMaxBitRate());
+                            // Configure the video encodings and set the active state, max bitrate and degradation
+                            // preference on them.
+                            return this.setSenderVideoConstraint();
```
paweldomas, line 2090: I think this method should be called differently as it's no longer only about senderVideoConstraints. Then the comment above would not be needed to explain that it's all of the encoding parameters being reconfigured. It's a little bit weird that sometimes this method is called without the actual constraint, I think if it was split into `setSenderVideoConstraint` that takes the actual constraint and then another one which only reconfigures based on the current state it would be a little cleaner.

modules/RTC/TPCUtils.js
jallamsetty1: For low fps sharing, i.e., when user prefers the resolution over fps, we disable the lower resolution streams since sending all 3 streams results in the browser suspending the high resolution stream in low b/w and cpu cases and we hit this more often on the low end machines. Suspending the lower streams ensures the highest resolution stream is available always.

modules/RTC/TPCUtils.js
```diff
@@ -273,29 +290,83 @@ export class TPCUtils {
...
+        const desktopShareBitrate = this.pc.options?.videoQuality?.desktopBitrate || DESKTOP_SHARE_RATE;
+        const presenterEnabled = localVideoTrack._originalStream
+            && localVideoTrack._originalStream.id !== localVideoTrack.getStreamId();
+
+        const encodingsBitrates = this.localStreamEncodingsConfig
+        .map(encoding => {
+            const bitrate = this.pc.isSharingLowFpsScreen() && !browser.isWebKitBased()
+
+                // For low fps screensharing, set a max bitrate of 500 Kbps when presenter is not turned on.
+                // FIXME the top 'isSimulcastOn' condition is confusing for screensharing, because
```
paweldomas, line 355: can you remove the FIXME comment about 'isSimulcastOn' as it's no longer relevant?

paweldomas (approved):

modules/RTC/TPCUtils.js
jallamsetty1: Ah good point!

PR #1742, opened 2021-10-06
fix(IceFailedHandling) force client reload when ICE fails locally.

jallamsetty1:


jallamsetty1:
> LGTM

Thanks for the review!

PR #1744, opened 2021-10-07
Webpack 5

saghul:


package.json
```diff
@@ -40,31 +40,28 @@
     "@babel/plugin-proposal-export-namespace-from": "7.0.0",
     "@babel/plugin-proposal-nullish-coalescing-operator": "7.13.8",
     "@babel/plugin-proposal-optional-chaining": "7.2.0",
-    "@babel/plugin-transform-flow-strip-types": "7.0.0",
     "@babel/preset-env": "7.1.0",
-    "@babel/preset-flow": "7.0.0",
```
hristoterezov, line 45: We remove flow with this PR?

I thought this was don from other PR. Or the previous one was only preparation?

saghul: The previous one was the first attempt, then I broke it in 2. This is part 1, TypeScript comes right after we merge this.

PR #1757, opened 2021-10-15
Source names in presence

paweldomas:
Use source names in presence to advertise track's muted state and the video type.

For now, if the source name signaling flag is enabled, both legacy and new <SourceInfo> element will be used at the same time. This is to be able to interoperate with legacy clients and start testing the new format at the same time. Whenever possible <SourceInfo> will be used as main source of truth with the fallback to legacy <audiomuted/>, <videomuted/> and <videotype/> elements (only if a given client doesn't <SourceInfo>).


JitsiConference.js
```diff
@@ -1280,6 +1286,44 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
...
+        this._signalingLayer.setTrackVideoType(
+            getSourceNameForJitsiTrack(
+                this.myUserId(),
+                track.getType(),
+                0
+            ),
+            track.videoType);
+    }
+
+    const videoTypeTagName = 'videoType';
```
paweldomas, line 1302: legacy video type handling unchanged

JitsiConference.js
```diff
@@ -1280,6 +1286,44 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
...
+    if (FeatureFlags.isSourceNameSignalingEnabled()) {
+        // TODO When legacy signaling part is removed, remember to adjust signalingLayer.setTrackMuteStatus, so that
+        // it triggers sending the presence (it only updates it for now, because the legacy code below sends).
+        this._signalingLayer.setTrackMuteStatus(
+            getSourceNameForJitsiTrack(this.myUserId(), localTrack.getType(), 0),
+            isMuted
+        );
+    }
+
+    if (localTrack.isAudioTrack()) {
```
paweldomas, line 1320: legacy mute status handling unchanged

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -51,40 +98,211 @@ export default class SignalingLayerImpl extends SignalingLayer {
...
+                }
+            } else {
+                throw new Error(`Unsupported media type: ${mediaType}`);
+            }
+
+            return mediaInfo;
         }
-        logger.error('Requested peer media info, before room was set');
+
+        return legacyGetPeerMediaInfo();
```
paweldomas, line 298: if the feature flag is disabled, the legacy handling was left unchanged, but moved into a method, so can be re-used as part of the new flow (temporarily)

PR #1762, opened 2021-10-20
fix(SS): missing.

hristoterezov:
The issue was caused by incorrect ordering of the mixedmslabel mline.
sdp-interop is expecting that this mline was the first one but in
reality we were placing it last. This was causing the direction of a
user related m line to be sendrecv instead of sendonly. In certain
situation this issue was causing 2 MediaStreamTracks associated with a
single MediaStream/participant.


jallamsetty1 (approved):

paweldomas:
@hristoterezov good catch!

PR #1764, opened 2021-10-21
fix: Missed "cname" attribute in SDP jingle

mnasyrov:
Added `cname` attribute to SDP jingle.

The missed `cname` causes receiving PeerConnection's track without a media stream in not-latest Chromium-based browsers.

### DIFF
--- a/modules/sdp/SDP.js
+++ b/modules/sdp/SDP.js
@@ -205,6 +205,7 @@ SDP.prototype.toJingle = function(elem, thecreator) {
 
                 for (const [ availableSsrc, ssrcParameters ] of ssrcMap) {
                     const sourceName = SDPUtil.parseSourceNameLine(ssrcParameters);
+                    const { msid, cname } = SDPUtil.parseSsrcAttributes(ssrcParameters);
 
                     elem.c('source', {
                         ssrc: availableSsrc,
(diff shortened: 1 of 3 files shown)


saghul:

paweldomas:
```
Uncaught TypeError: Cannot destructure property 'id' of 'undefined' as it is undefined.
    at Object.Yt.isReactNative.getStreamID (lib-jitsi-meet.min.js:2)
    at Function.getStreamID (lib-jitsi-meet.min.js:2)
    at Xo._remoteTrackAdded (lib-jitsi-meet.min.js:2)
    at Xo._usesUnifiedPlan.onTrack (lib-jitsi-meet.min.js:2)
    at RTCPeerConnection.s (lib-jitsi-meet.min.js:2)
```
What is this stack trace? Are you running custom lib-jitsi-meet?
Cname is not meant to be a stream ID

jallamsetty1:
@mnasyrov, Can you please give us an example of where this breaks ? Jicofo doesn't advertise cnames anymore as msid is the only param that needs to be set on the browser in the a:ssrc line  and that value corresponds with the stream id. All browsers have already dropped generation of the cnames in the local description.

jallamsetty1:
Just read through the original issue. @mnasyrov Can you please grab the remote description that is being set on the peerconnection before the track event is fired ?

paweldomas:
It is also important to see if 'isReactNative' flows are not activated, because this has many different consequences in different places of the app.

mnasyrov:
@jallamsetty1 jicofo 6173 is used

agustinbranaf:
@mnasyrov interested in finding a solution to this since we use chromium in our App, did you ever find a fix/workaround to this issue?

mnasyrov:
> @mnasyrov interested in finding a solution to this since we use chromium in our App, did you ever find a fix/workaround to this issue?

We use a custom lib-jitsi patched by this PR locally with a previous version of Jicofo. Soon, I hope, we'll update requirements for client browsers, and discard this fix as well.

PR #1772, opened 2021-10-27
feat: SourceVideoTypeMessage message

paweldomas:
Sends the source name to video type mapping to the jvb in a new bridge channel message.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -55,6 +55,7 @@ import {
 import BridgeVideoType from './service/RTC/BridgeVideoType';
 import CodecMimeType from './service/RTC/CodecMimeType';
 import * as MediaType from './service/RTC/MediaType';
+import RTCEvents from './service/RTC/RTCEvents';
 import { getSourceNameForJitsiTrack } from './service/RTC/SignalingLayer';
 import VideoType from './service/RTC/VideoType';
 import {
@@ -400,6 +401,9 @@ JitsiConference.prototype._init = function(options = {}) {
     if (!this.rtc) {
         this.rtc = new RTC(this, options);
         this.eventManager.setupRTCListeners();
+        if (FeatureFlags.isSourceNameSignalingEnabled()) {
+            this._registerRtcListeners(this.rtc);
+        }
     }
 
     this.receiveVideoController = new ReceiveVideoController(this, this.rtc);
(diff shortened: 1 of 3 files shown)


JitsiConference.js
```diff
@@ -718,6 +722,20 @@ JitsiConference.prototype._getMediaSessions = function() {
     return sessions;
 };
 
+/**
+ * Registers event listeners on the RTC instance.
+ * @param {RTC} rtc - the RTC module instance used by this conference.
+ * @private
+ * @returns {void}
+ */
+JitsiConference.prototype._registerRtcListeners = function(rtc) {
```
jallamsetty1, line 731: There is already a listener registered for RTCEvents.DATA_CHANNEL_OPEN event here - https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConferenceEventManager.js#L532. Can we add this functionality there instead of adding another listener here ?

paweldomas: Yeah I was thinking exactly about that, but JitsiConferenceEventManager feels more like an event forwarder the way I understand it. I didn't want to add more complex logic there, especially one related to the streams. Let me know if that make sense, otherwise I will move it there.

jallamsetty1: It does make sense to have it here then. Is there a reason why you chose not to do it the way its done for legacy vidyoType in RTC.js and add the listener here instead ?

paweldomas: The reason for doing it here was the fact that `_sendBridgeVideoTypeMessage` is in the `JitsiConference`. In the new flow the video type is not stored in the RTC module, but taken from the local tracks state. Maybe eventually it could be moved to the RTC module, but looking at the existing usages of `_sendBridgeVideoTypeMessage` it's not a simple change.

jallamsetty1 (approved):

modules/RTC/RTC.js
```diff
@@ -151,7 +152,12 @@ export default class RTC extends Listenable {
         this._updateAudioOutputForAudioTracks
             = this._updateAudioOutputForAudioTracks.bind(this);
 
-        // The default video type assumed by the bridge.
+        /**
+         * The default video type assumed by the bridge.
+         * @deprecated this will go away with multiple streams support
+         * @type {BridgeVideoType}
+         * @private
+         */
```
paweldomas, line 160: the default type assumed by the JVB is actually CAMERA, but I'm a bit afraid to touch it. This value belong to the existing flow and is not used with the multi stream support

PR #1776, opened 2021-11-03
fix(e2ee): fix loading web worker when using a relative path inside a blob for the E2EE context

DanielMcAssey:
When loading a web worker inside a blob using importScripts you cannot use a relative path because of CORS issues as it resolves to a blob: path.

Please see: https://www.html5rocks.com/en/tutorials/workers/basics/#toc-inlineworkers-loadingscripts

### DIFF
--- a/modules/e2ee/E2EEContext.js
+++ b/modules/e2ee/E2EEContext.js
@@ -37,14 +37,22 @@ export default class E2EEcontext {
             baseUrl = `${ljm.src.substring(0, idx)}/`;
         }
 
-        // Initialize the E2EE worker. In order to avoid CORS issues, start the worker and have it
-        // synchronously load the JS.
         const workerUrl = `${baseUrl}lib-jitsi-meet.e2ee-worker.js`;
-        const workerBlob
-            = new Blob([ `importScripts("${workerUrl}");` ], { type: 'application/javascript' });
-        const blobUrl = window.URL.createObjectURL(workerBlob);
+        
+        if (baseUrl && baseUrl.length > 0) {
+            // Initialize the E2EE worker. In order to avoid CORS issues, start the worker and have it
+            // synchronously load the JS.
+            const workerBlob
+                = new Blob([ `importScripts("${workerUrl}");` ], { type: 'application/javascript' });
+            const blobUrl = window.URL.createObjectURL(workerBlob);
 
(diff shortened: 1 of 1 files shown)


modules/e2ee/E2EEContext.js
```diff
@@ -37,14 +37,22 @@ export default class E2EEcontext {
...
 
-        this._worker = new Worker(blobUrl, { name: 'E2EE Worker' });
+        if (baseUrl && baseUrl.length > 0) {
+            // Initialize the E2EE worker. In order to avoid CORS issues, start the worker and have it
+            // synchronously load the JS.
+            const workerBlob
+                = new Blob([ `importScripts("${workerUrl}");` ], { type: 'application/javascript' });
+            const blobUrl = window.URL.createObjectURL(workerBlob);
+
+            this._worker = new Worker(blobUrl, { name: 'E2EE Worker' });
```
saghul, line 49: Let's turn workerBlob into a let and reassign it in this case. Then we can have a single worker construction.

saghul (commented):
Left a comment, PTAL. How are you using lib-jitsi-meet that prompted this change?

modules/e2ee/E2EEContext.js
DanielMcAssey: Done!

DanielMcAssey:
We actually use this internally, however because we have a segregated network, we periodically download and build the lib-jitsi-meet library and include it from the same domain, rather from the jitsi domain. Weird use case I know!

saghul (approved):

saghul:

DanielMcAssey:
Lobby test failure on an unrelated component? May need to rerun build

DanielMcAssey:

saghul:

DanielMcAssey:
Having no luck today! I cant imagine my change would have affected the test that has failed?

saghul:
Not your fault, likely a flaky test. I'll take a look.

DanielMcAssey:
Ive made a tiny tweak in checking for the baseUrl. I noticed that a slash is added regardless if there is any source, so instead of checking for length greater than 0, just check that it isn't just a forward slash.

saghul:

saghul (approved):

PR #1777, opened 2021-11-03
fix(tracks) Add metadata to jitsi tracks

robertpin:
Add timestamp to all tracks, and display surface to screen sharing tracks

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -81,6 +81,19 @@ export default class JitsiLocalTrack extends JitsiTrack {
             this._startStreamEffect(effect);
         }
 
+        const displaySurface = videoType === VideoType.DESKTOP
+            ? track.getSettings().displaySurface
+            : null;
+
+        /**
+         * Track metadata.
+         */
+        this.metadata = {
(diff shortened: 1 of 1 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -81,6 +81,19 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+        const displaySurface = videoType === VideoType.DESKTOP
+            ? track.getSettings().displaySurface
+            : null;
+
+        /**
+         * Track metadata.
+         */
+        this.metadata = {
+            timestamp: Date.now(),
+            ...displaySurface ? { displaySurface } : {}
```
saghul, line 93: Parentheses for extra readability?

saghul (approved):
One nit, LGTM otherwise.

modules/RTC/JitsiLocalTrack.js
robertpin: Tried. Eslint doesn't like them, says they're unnecessary

saghul: ah right

Issue #1778, opened 2021-11-03
update sctp jingle mapping


fippo:
follow-up from #1697 where @nils-ohlmeier said we should update the protocol from DTLS/SCTP to UDP/DTLS/SCTP.

It is a bit more complicated than that single transportToJingle looks for `a=sctpmap:` lines and there are gone now (I think, not sure if browsers still respond to the "legacy" variant)

So we need an updated specification that maps these lines:
```
m=application 9 UDP/DTLS/SCTP webrtc-datachannel (9 more lines)
...
a=sctp-port:5000
a=max-message-size:262144
```

PR #1779, opened 2021-11-03
fix(sdp): update data channel to RFC format

nils-ohlmeier:
This PR switches to generating sctp-port based SDP.
It also adds support for parsing sctp-port, but can handle
fallback to old-style sctpmap as well.
Fixes: #1778


fippo (approved):
but you're mapping the two forms to the same jingle format?
Which is ok since this isn't used in p2p and only deals with offers from the focus

nils-ohlmeier:
Yes I am mapping it to same message, because the new format has no other information in it except the message size. 

Why do you think it could cause problems in P2P mode?

fippo:
I don't think p2p connections even use sctp

PR #1789, opened 2021-11-11
feat: Enables the moderator to {dis/en}able bridge-side rtcstats.

gpolitis:


damencho:
Will we disable and callstats when this is set?

gpolitis:
> Will we disable and callstats when this is set?

We can, if we want to. Alternatively we can propagate a separate flag for callstats. I don't have a preference, so I'm open to suggestions :-)

damencho:
Well if this is called rtcstatsEnabled, then we need to add and one for callstats.

damencho:
Can you add one so we can do the same and for callstats, anyway you will do the dance with PRs in all projects, can you add to those PRs and callstats disabling. 
We already detect callstats. being disabled ...

damencho:
https://github.com/gimre/lib-jitsi-meet/blob/c24130622ef3cf24078b0e5244a09e1d9c37dfd7/modules/statistics/statistics.js#L166

gpolitis:
@damencho this is ready, could you please take a look?

damencho (approved):

gpolitis:
@damencho I realized that the default in the client is actually `false` (the default in the config.js is misleading and I have opened a PR for that here https://github.com/jitsi/jitsi-meet/pull/10425) so I had to adjust this patch. Could you please take another look?

damencho (approved):

damencho (approved):

PR #1790, opened 2021-11-15
fix(browser) Mark safari <14 as unsupported

quitrk:


modules/browser/BrowserCapabilities.js
```diff
@@ -102,6 +103,7 @@ export default class BrowserCapabilities extends BrowserDetection {
      */
     isSupported() {
         return (this.isChromiumBased() && this._getChromiumBasedVersion() >= MIN_REQUIRED_CHROME_VERSION)
+            || (this.isSafari() && this._getSafariVersion() >= MIN_REQUIRED_SAFARI_VERSION)
```
saghul, line 106: Watch out, `isWebKitBased` will return true.

quitrk: i tested in safari 15 and it didnt..

quitrk: updated

horymury (approved):

PR #1794, opened 2021-11-18
Save track source name to JitsiRemoteTrack

dennisbb2:
DRAFT pull request.  Changes have yet to be reviewed, validated and tested.

### DIFF
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -73,7 +73,7 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.ownerEndpointId = ownerEndpointId;
         this.muted = muted;
         this.isP2P = isP2P;
-        this.sourceName = sourceName;
+        this._sourceName = sourceName;
 
         logger.debug(`New remote track added: ${this}`);
 
@@ -197,7 +197,7 @@ export default class JitsiRemoteTrack extends JitsiTrack {
      * @returns {string} the track's source name
      */
      getSourceName() {
-        return this.sourceName;
+        return this._sourceName;
     }
 
     /**
@@ -324,6 +324,6 @@ export default class JitsiRemoteTrack extends JitsiTrack {
      */
     toString() {
         return `RemoteTrack[userID: ${this.getParticipantId()}, type: ${this.getType()}, ssrc: ${
-            this.getSSRC()}, p2p: ${this.isP2P}, sourceName: ${this.sourceName}, status: ${this._getStatus()}]`;
+            this.getSSRC()}, p2p: ${this.isP2P}, sourceName: ${this._sourceName}, status: ${this._getStatus()}]`;
     }
 }
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -190,13 +190,13 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         return this.ssrc;
     }
 
-    
+
     /**
      * Returns the tracks source name
      *
      * @returns {string} the track's source name
      */
-     getSourceName() {
+    getSourceName() {
         return this._sourceName;
     }
 
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -38,6 +38,7 @@ export default class JitsiRemoteTrack extends JitsiTrack {
      * @param {boolean} muted the initial muted state
      * @param {boolean} isP2P indicates whether or not this track belongs to a
      * P2P session
+     * @param {String} sourceName the source name signaled for the track
      * @throws {TypeError} if <tt>ssrc</tt> is not a number.
      * @constructor
      */
@@ -51,7 +52,8 @@ export default class JitsiRemoteTrack extends JitsiTrack {
             videoType,
             ssrc,
             muted,
-            isP2P) {
+            isP2P,
+            sourceName) {
         super(
             conference,
             stream,
@@ -71,6 +73,7 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.ownerEndpointId = ownerEndpointId;
         this.muted = muted;
         this.isP2P = isP2P;
+        this.sourceName = sourceName;
 
         logger.debug(`New remote track added: ${this}`);
 
@@ -185,6 +188,16 @@ export default class JitsiRemoteTrack extends JitsiTrack {
      */
     getSSRC() {
         return this.ssrc;
+    }
+
+    
+    /**
+     * Returns the tracks source name
+     *
+     * @returns {string} the track's source name
+     */
+     getSourceName() {
+        return this.sourceName;
     }
 
     /**
@@ -311,6 +324,6 @@ export default class JitsiRemoteTrack extends JitsiTrack {
      */
     toString() {
         return `RemoteTrack[userID: ${this.getParticipantId()}, type: ${this.getType()}, ssrc: ${
-            this.getSSRC()}, p2p: ${this.isP2P}, status: ${this._getStatus()}]`;
+            this.getSSRC()}, p2p: ${this.isP2P}, sourceName: ${this.sourceName}, status: ${this._getStatus()}]`;
     }
 }
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -898,8 +898,11 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
         return;
     }
 
+
+    let sourceName;
+
     if (FeatureFlags.isSourceNameSignalingEnabled()) {
-        let sourceName = this.signalingLayer.getTrackSourceName(trackSsrc);
+        sourceName = this.signalingLayer.getTrackSourceName(trackSsrc);
 
         // If source name was not signaled, we'll generate one which allows testing signaling
         // when mixing legacy(mobile) with new clients.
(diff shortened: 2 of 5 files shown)


modules/xmpp/SignalingLayerImpl.js
```diff
@@ -391,4 +401,33 @@ export default class SignalingLayerImpl extends SignalingLayer {
...
+    
+    /**
+     * Saves the source name for a track identified by it's ssrc.
+     * @param {number} ssrc the ssrc of the target track.
+     * @param {SourceName} sourceName the track's source name to save.
+     * @throws TypeError if <tt>ssrc</tt> is not a number
+     */
+    setTrackSourceName(ssrc, sourceName) {
+        if (typeof ssrc !== 'number') {
+            throw new TypeError(`SSRC(${ssrc}) must be a number`);
```
paweldomas, line 420: 👍

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -391,4 +401,33 @@ export default class SignalingLayerImpl extends SignalingLayer {
...
+        }
+        
+        // Now signaling layer instance is shared between different JingleSessionPC instances, so although very unlikely
+        // an SSRC conflict could potentially occur. Log a message to make debugging easier.
+        if (this.sourceNames.has(ssrc)) {
```
paweldomas, line 425: lets also check for the value? Jaya recently pushed a change in regards to the SSRC owner handling

paweldomas: see https://github.com/jitsi/lib-jitsi-meet/commit/5ae76612b411aefa9c9d3448091feca634ba6687

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -52,6 +52,16 @@ export default class SignalingLayerImpl extends SignalingLayer {
...
+
+        /**
+         * A map that stores the source name of a track identified by it's ssrc.
+         * We store the mapping when jingle is received, and later is used
+         * onaddstream webrtc event where we have only the ssrc
+         * FIXME: This map got filled and never cleaned and can grow during long
+         * conference
+         * @type {Map<number, string>} maps SSRC number to source name
+         */
+         this.sourceNames = new Map();
```
paweldomas, line 64: lets make it `this._sourceNames` to make it "private"

modules/xmpp/JingleSessionPC.js
```diff
@@ -879,6 +879,11 @@ export default class JingleSessionPC extends JingleSession {
                 this._signalingLayer.setSSRCOwner(
                     ssrc, Strophe.getResourceFromJid(this.remoteJid));
             } else {
+                // Only set sourceName for non-P2P case
```
paweldomas, line 882: lets do this only if feature flag is enabled

modules/RTC/TraceablePeerConnection.js
```diff
@@ -896,7 +897,16 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
         return;
     }
 
-    logger.info(`${this} creating remote track[endpoint=${ownerEndpointId},ssrc=${trackSsrc},type=${mediaType}]`);
+    let sourceName = this.signalingLayer.getTrackSourceName(trackSsrc);
```
paweldomas, line 900: lets execute the logic only if feature flag is enabled and assign `undefined` otherwise

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -311,6 +324,6 @@ export default class JitsiRemoteTrack extends JitsiTrack {
      */
     toString() {
         return `RemoteTrack[userID: ${this.getParticipantId()}, type: ${this.getType()}, ssrc: ${
-            this.getSSRC()}, p2p: ${this.isP2P}, status: ${this._getStatus()}]`;
+            this.getSSRC()}, p2p: ${this.isP2P}, sourceName: ${this.sourceName}, status: ${this._getStatus()}]`;
```
paweldomas, line 327: what do you think about changing it to ` n: ${this.sourceName}`? I'm thinking the shorter the better, given that it's getting crowded in here already

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -71,6 +73,7 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.ownerEndpointId = ownerEndpointId;
         this.muted = muted;
         this.isP2P = isP2P;
+        this.sourceName = sourceName;
```
paweldomas, line 76: lets call it `this._sourceName` to make it "private"

modules/RTC/TraceablePeerConnection.js
paweldomas: the goal is really to not execute any of the new logic when the feature flag is off

For the modified log message below I think it's fine if it will say "undefined" without the feature flag.

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -52,6 +52,16 @@ export default class SignalingLayerImpl extends SignalingLayer {
          * @private
          */
         this._remoteSourceState = { };
+
+        /**
+         * A map that stores the source name of a track identified by it's ssrc.
+         * We store the mapping when jingle is received, and later is used
+         * onaddstream webrtc event where we have only the ssrc
+         * FIXME: This map got filled and never cleaned and can grow during long
```
paweldomas, line 60: can you try cleaning it up here:
https://github.com/jitsi/lib-jitsi-meet/blob/6bff32f86e0615e78c5a2aca9f64f8bc08a8a0f9/modules/xmpp/SignalingLayerImpl.js#L246
maybe it's possible to get SSRC owner and then remove from the map based on that?

modules/RTC/JitsiRemoteTrack.js
dennisbb2: I saw similar single letter naming in the compact json structure.  Was concerned single letter naming conventions (granted this is logging here and not variable naming) could be confusing when debugging (greps for 'sourceName' as such).  I can make that adjustment if you're ok with it.

modules/xmpp/SignalingLayerImpl.js
dennisbb2: ok will take a look

modules/RTC/JitsiRemoteTrack.js
paweldomas: ok lets keep it this way

modules/xmpp/SignalingLayerImpl.js
dennisbb2: updated code here.  please check new logic.  it iterates through the ssrcOwners map to find all relevant ssrc to delete based on endpointId.  this works right now only because ssrcOwners is not cleaned up, and its a full iteration each time and ssrcOwners grows indefinitely atm.  if ssrcOwners is cleaned up in a different PR, then care needs to be taken here as the ordering of the cleanup would matter.

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -234,6 +244,14 @@ export default class SignalingLayerImpl extends SignalingLayer {
             const endpointId = Strophe.getResourceFromJid(jid);
 
             delete this._remoteSourceState[endpointId];
+
+            if (FeatureFlags.isSourceNameSignalingEnabled()) {
+                for (const [key, value] of this.ssrcOwners.entries()) {
+                    if (value === endpointId) {
+                        delete this._sourceNames[key];
```
paweldomas, line 251: great, can you put a comment here with a note that this may stop working if ssrcOwners are cleaned up before this code executes?

modules/xmpp/SignalingLayerImpl.js
paweldomas: great, thanks!

modules/xmpp/SignalingLayerImpl.js
jallamsetty1: Are we deleting the entry from `this._sourceNames` map when a `source-remove` is received from Jicofo ?

paweldomas: oh that's a good point, but we weren't doing that for the SSRC owners either

jallamsetty1: Yes, we need to clean up SSRC owners as well. I didn't notice this until now.

jallamsetty1:
Also, please follow the commit message guidelines from https://github.com/jitsi/handbook/blob/master/docs/dev-guide/contributing.md.

jallamsetty1:
Jenkins, please add this to whitelist

modules/xmpp/SignalingLayerImpl.js
dennisbb2: I was thinking the same thing , needing to delete it for source-remove , but did not think that was within scope of this PR. Should that be a follow up PR or should we include that here?

modules/RTC/TraceablePeerConnection.js
dennisbb2: I don't think lint is allowing me to have it as undefined?

modules/xmpp/SignalingLayerImpl.js
jallamsetty1: I think it's ok to do it in a follow up PR.

modules/RTC/TraceablePeerConnection.js
paweldomas: lets add ignore exception

modules/RTC/TraceablePeerConnection.js
```diff
@@ -896,7 +898,19 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
         return;
     }
 
-    logger.info(`${this} creating remote track[endpoint=${ownerEndpointId},ssrc=${trackSsrc},type=${mediaType}]`);
+    if (FeatureFlags.isSourceNameSignalingEnabled()) {
+        let sourceName = this.signalingLayer.getTrackSourceName(trackSsrc);
```
paweldomas, line 902: this needs to be declared outside of the `if` and assigned to `undefined` by default

paweldomas (approved):
LGTM, @jallamsetty1 can you please take a look?

jallamsetty1 (approved):

PR #1796, opened 2021-11-19
fix(TPC): Select 1 as the default scale factor for p2p.

jallamsetty1:
This fixes an issue where a user is not able to unmute their video if the MediaStreamTrack associated with the camera stream returns a null value for the track height.


hristoterezov:
What is the difference between this version and the old version? They seem equivalent to me. Am I missing something?

Also I see that the changes are in the `if` where the height is more than 0. How is this related to height `null`?

jallamsetty1:
Before, if the MediaStreamTrack height came back as null, the scaleFactor would be null and the browser doesn't accept that value and throw an error. Now, the default value is set to 1. Only if the height is not null and > requested height, a new scale factor is calculated, otherwise it remains as 1.

hristoterezov (approved):

PR #1798, opened 2021-11-21
fix(rtc): fix sending video type on unmute

DanielMcAssey:
When initially adding a muted desktop track to the conference, the video type was not being sent to the video bridge when subsequently unmuting, thus the stream was not being forwarded to clients as the video type was set to NONE.

This fixes the logic that when the track is unmuted the BridgeVideoType is sent, otherwise when muted it checks if the stream is muted by removing it from the PC, similar to how it was done before.


damencho:
Can you test with latest master, I think this is now fixed.

damencho:
After https://github.com/jitsi/lib-jitsi-meet/pull/1814

PR #1807, opened 2021-12-02
Fixup leave

saghul:


damencho (approved):

JitsiConference.js
```diff
@@ -703,6 +698,11 @@ JitsiConference.prototype.leave = function() {
                     participant => this.onMemberLeft(participant.getJid()));
 
                 throw error;
+            })
+            .then(() => {
+                if (this.rtc) {
```
paweldomas, line 703: how does that make sure RTC is always destroyed? it will not be in the case of the catch above because it throws at the end, am I missing something?

saghul: Damn you're right, it slipped my mind.

saghul: https://github.com/jitsi/lib-jitsi-meet/pull/1808

PR #1812, opened 2021-12-07
feat: Refactor updating presence for audio/video mute and video type.

damencho:
This brings few changes and fixes. Thew initial presence will always mis audioMuted and videoMuted value, those will be added on session accept. All updates to presence are done on sessionAccept, on source-add or source-remove and the only one not signalled when camera track is replaced by video track and vice versa.
This change brings more presence updates when number of participants are below startAudioMuted/startVideoMuted, but when participants are above those numbers we get less presence update. This is important for big meetings.
Fixes wrong videoMuted state, as replace track and mute are both executed in promise, sometimes the replace-track one finishes first and when the mute one is resolved there is no conference object in the track to be able to update the presence (hitting this when we pass the startAudioMuted threshold).

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1253,6 +1253,11 @@ JitsiConference.prototype.replaceTrack = function(oldTrack, newTrack) {
                 this._sendBridgeVideoTypeMessage(newTrack);
             }
 
+            // updates presence when we replace the video tracks desktop with screen and screen with desktop
+            if (oldTrackBelongsToConference && oldTrack?.isVideoTrack() && this.jvbJingleSession) {
+                this.jvbJingleSession.updateRoomPresence();
+            }
+
             if (newTrack !== null && (this.isMutedByFocus || this.isVideoMutedByFocus)) {
                 this._fireMuteChangeEvent(newTrack);
             }
@@ -1319,18 +1324,7 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
         }
     }
 
-    let videoTypeChanged = false;
-
-    if (newTrack.isVideoTrack()) {
-        videoTypeChanged = this._setNewVideoType(newTrack);
-    }
     this.rtc.addLocalTrack(newTrack);
-
-    // ensure that we're sharing proper "is muted" state
-    if (this._setTrackMuteStatus(newTrack, newTrack.isMuted()) || videoTypeChanged) {
-        // send presence if it was changed with vide type or mute status
-        this.room.sendPresence();
-    }
 
     newTrack.muteHandler = this._fireMuteChangeEvent.bind(this, newTrack);
     newTrack.audioLevelHandler = this._fireAudioLevelChangeEvent.bind(this);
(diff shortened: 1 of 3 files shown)


damencho:
@jallamsetty1 and @paweldomas Can you take a look, please? I tried simplifying it as much as I can :) 

@paweldomas I was not sure what to do with JitsiConference._sendBridgeVideoTypeMessage, does it make sense to be moved around ... or something. I can do changes with your guidance or you can do any adjustments after we merge this and test it on your side...

paweldomas:
I will take a look, but what bothers me from a high level perspective is that the decision about sending presence update was moved to JingleSession and that there can be two jingle sessions at the same time (p2p and the jvb one) and that a jingle session may not have enough context to make the right decisions.

damencho:
> I will take a look, but what bothers me from a high level perspective is that the decision about sending presence update was moved to JingleSession and that there can be two jingle sessions at the same time (p2p and the jvb one) and that a jingle session may not have enough context to make the right decisions.

@jallamsetty1 wdyt?

We can create internal ljm events for the success of source-add, source-remove and session-accept. Use those events in JitsiConference and do the calculation based on that for the peerconnection that fires it? 
And basically move the logic back to JitsiConference.
Does that approach cover what bothers you?

paweldomas:
@damencho I will try to go through this in details as I only skimmed through the PR so far. I'll try to do it today

jallamsetty1:
> > I will take a look, but what bothers me from a high level perspective is that the decision about sending presence update was moved to JingleSession and that there can be two jingle sessions at the same time (p2p and the jvb one) and that a jingle session may not have enough context to make the right decisions.
> 
> @jallamsetty1 wdyt?
> 
> We can create internal ljm events for the success of source-add, source-remove and session-accept. Use those events in JitsiConference and do the calculation based on that for the peerconnection that fires it? And basically move the logic back to JitsiConference. Does that approach cover what bothers you?

Since rejects for session-accept, source-add and source-remove are not that common and come into play only when the sender limits are reached, we could send a second presence update after a reject is received like we discussed. But then that wouldn't address the muted by focus case right ?

damencho:
> Since rejects for session-accept, source-add and source-remove are not that common and come into play only when the sender limits are reached, we could send a second presence update after a reject is received like we discussed. But then that wouldn't address the muted by focus case right ?

We can call update presence, on reject and it will send presence if it is needed. 
What do you mean that wouldn't address the muted by focus case? This PR is addressing those inconsistencies, as described in the commit message, or at least, I tried explaining it :)

jallamsetty1:
> > Since rejects for session-accept, source-add and source-remove are not that common and come into play only when the sender limits are reached, we could send a second presence update after a reject is received like we discussed. But then that wouldn't address the muted by focus case right ?
> 
> We can call update presence, on reject and it will send presence if it is needed. What do you mean that wouldn't address the muted by focus case? This PR is addressing those inconsistencies, as described in the commit message, or at least, I tried explaining it :)

I mis-read your comment before, I thought we were reverting to sending the presences after the tracks are replaced on pc and not waiting for source-add, source-reject and session-accept to be accepted, my bad. This new approach makes sense to me.

damencho:
> We can create internal ljm events for the success of source-add, source-remove and session-accept. Use those events in JitsiConference and do the calculation based on that for the peerconnection that fires it?
> And basically move the logic back to JitsiConference.
> Does that approach cover what bothers you?

@paweldomas @jallamsetty1 I implemented the above here https://github.com/jitsi/lib-jitsi-meet/pull/1814
@jallamsetty1 I haven't added updating presence on rejection as it also needs removing the tracks and I know you will be doing that ... but with the current code it will be easy to add it.

PR #1814, opened 2021-12-07
Refactor presence update 2

damencho:


jallamsetty1:
LGTM !!

JitsiConference.js
```diff
@@ -1348,21 +1350,21 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
 
 /**
  * Sets the video type.
- * @param track
+ * @param trackVideoType
  * @return <tt>true</tt> if video type was changed in presence.
  * @private
  */
-JitsiConference.prototype._setNewVideoType = function(track) {
+JitsiConference.prototype._setNewVideoType = function(trackVideoType) {
```
paweldomas, line 1357: can you keep the track argument here? with multi stream support the track will be needed here

JitsiConference.js
```diff
@@ -1371,28 +1373,28 @@ JitsiConference.prototype._setNewVideoType = function(track) {
...
 /**
  * Sets mute status.
- * @param localTrack
+ * @param mediaType
  * @param isMuted
  * @param <tt>true</tt> when presence was changed, <tt>false</tt> otherwise.
  * @private
  */
-JitsiConference.prototype._setTrackMuteStatus = function(localTrack, isMuted) {
+JitsiConference.prototype._setTrackMuteStatus = function(mediaType, isMuted) {
```
paweldomas, line 1392: can you keep the local track argument here? it will be needed for multi stream support

JitsiConference.js
```diff
@@ -3596,6 +3598,44 @@ JitsiConference.prototype._stopP2PSession = function(options = {}) {
...
 
+/**
+ * Updates room presence if needed and send the packet in case of a modification.
+ * @param {JingleSessionPC} jingleSession the session firing the event, contains the peer connection which
+ * tracks we will check.
+ * @param {Object|null} ctx a context object we can distinguish multiple calls of the same pass of updating tracks.
+ */
+JitsiConference.prototype._updateRoomPresence = function(jingleSession, ctx) {
+    if (ctx && ctx.skip) {
+        ctx.skip = true;
```
paweldomas, line 3609: is it not possible to skip the call in the place which sets the `ctx`?

paweldomas: actually I don't see it being set anywhere?

damencho: https://github.com/jitsi/lib-jitsi-meet/pull/1814/files#diff-2ab65ee8ec5b7568335948135d9d2a5b7010c054a1d821f9774a7b542689b282R2525

damencho: It is just the case if you have a source add and source remove in one call of update ssrc to send one presence, not two.

JitsiConference.js
damencho: I added it back, but it is unused for now.

There is a case where you set it back to default, there is no track.

JitsiConference.js
damencho: done, but the case with going from screen share back to video muted is handled only for presence, the case when there is no video track.

JitsiConference.js
paweldomas: but there it only defines the `ctx` object, where is the `ctx.skip` being set?

modules/xmpp/JingleSessionPC.js
```diff
@@ -954,7 +954,15 @@ export default class JingleSessionPC extends JingleSession {
                 // FIXME we may not care about RESULT packet for session-accept
                 // then we should either call 'success' here immediately or
                 // modify sendSessionAccept method to do that
-                this.sendSessionAccept(success, failure);
+                this.sendSessionAccept(() => {
+                    success();
+
+                    this.room.eventEmitter.emit(XMPPEvents.SESSION_ACCEPT, this);
+                }, () => {
+                    failure();
```
paweldomas, line 962: how is the error handled here?

modules/xmpp/JingleSessionPC.js
```diff
@@ -2512,12 +2520,24 @@ export default class JingleSessionPC extends JingleSession {
             );
         const removedAnySSRCs = sdpDiffer.toJingle(remove);
 
+        // context a common object for one run of ssrc update (source-add and source-remove) so we can match them if we
+        // need to
+        const ctx = {};
```
paweldomas, line 2525: what is the exact problem that `ctx` solves? It's hard to follow and passing context objects is an unusual pattern. I'm wondering if there's another of fixing that?

modules/xmpp/JingleSessionPC.js
```diff
@@ -2538,7 +2558,16 @@ export default class JingleSessionPC extends JingleSession {
             logger.info(`${this} Sending source-add`);
             logger.debug(add.tree());
             this.connection.sendIQ(
-                add, null, this.newJingleErrorHandler(add), IQ_TIMEOUT);
+                add,
+                () => {
+                    this.room.eventEmitter.emit(XMPPEvents.SOURCE_ADD, this, ctx);
+                },
+                () => {
+                    this.newJingleErrorHandler(add);
```
paweldomas, line 2566: newJingleErrorHandler creates a method which needs to be called with an error

modules/xmpp/JingleSessionPC.js
```diff
@@ -2512,12 +2520,24 @@ export default class JingleSessionPC extends JingleSession {
...
             logger.debug(remove.tree());
             this.connection.sendIQ(
-                remove, null,
-                this.newJingleErrorHandler(remove), IQ_TIMEOUT);
+                remove,
+                () => {
+                    this.room.eventEmitter.emit(XMPPEvents.SOURCE_REMOVE, this, ctx);
+                },
+                () => {
+                    this.newJingleErrorHandler(remove);
```
paweldomas, line 2536: newJingleErrorHandler creates a method which needs to be called with an error

JitsiConference.js
damencho: It is set here, once it passes this method and will skip the second invocation.

damencho: Oh, I need to set it after the if 🤦‍♂️

modules/xmpp/JingleSessionPC.js
damencho: If you do source add and source remove in one pass to send just one presence

JitsiConference.js
damencho: updated.

modules/xmpp/JingleSessionPC.js
damencho: I don't know, this is the callback that was executed before.

damencho: I had added the error to the callback.

modules/xmpp/JingleSessionPC.js
damencho: Fixed.

modules/xmpp/JingleSessionPC.js
damencho: Fixed.

JitsiConference.js
paweldomas: I'm not a big fan of this solution, but should do for now... source-remove/source-add should be merged into a single notification.

paweldomas (approved):

damencho:
Unrelated known failure, merging.

PR #1817, opened 2021-12-08
fix(browser-capabilities) expose supported mobile browsers

quitrk:


modules/browser/BrowserCapabilities.js
```diff
@@ -112,6 +113,26 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+     */
+    isSupportedAndroidBrowser() {
+        return this.isChromiumBased() || this.isFirefox();
+    }
+
+    /**
+     * Returns whether the browser is supported for iOS
+     * @returns {boolean} true if the browser is supported for iOS devices
+     */
+    isSupportedIOSBrowser() {
```
saghul, line 128: What does this return for say Brave on iOS 12?

saghul (approved):
LGTM with a comment. Also please re-target this to master. Then we can cherry-pick onto a release branch.

modules/browser/BrowserCapabilities.js
quitrk: webkit on ios, so it will return true. this is just moved logic from jitsi-meet, and adding the extra safari version check

PR #1820, opened 2021-12-09
feat(JingleSessionPC) Handle source-add rejects by Jicofo.

jallamsetty1:


damencho:
Hum the desktop sharing test failed. Can you check it ... I will rerun to see is it consistent

JitsiConference.js
```diff
@@ -1311,6 +1314,24 @@ JitsiConference.prototype._doReplaceTrack = function(oldTrack, newTrack) {
...
+ * @returns {void}
+ */
+JitsiConference.prototype._removeLocalSourceOnReject = function(jingleSession, error, mediaType) {
+    if (!jingleSession) {
+        return;
+    }
+    logger.warn(`Source-add rejected on ${jingleSession}, reason="${error?.reason}", message="${error?.msg}"`);
+    const track = this.getLocalTracks(mediaType)[0];
+
+    this.eventEmitter.emit(JitsiConferenceEvents.TRACK_UNMUTE_REJECTED, track);
```
paweldomas, line 1332: Is it that this event goes back to jitsi-meet and then in response it calls the track removal methods on the JitsiConference? If so I think the removal should happen here and the event should be just a notification that the track was removed from the conference, so the track can be disposed (or not depending on the app's choice). Otherwise the feature is not contained in lib-jitsi-meet.

jallamsetty1: No, it's just not about track removal. The track needs to be removed from redux and only then it will result in a replaceTrack and source-add being sent the next time when the user is allowed to unmute. Just removing the track in LJM will result in the track not being signaled on the next unmute.

damencho (approved):

JitsiConference.js
paweldomas: It can be removed from redux on this event

PR #1826, opened 2021-12-14
fix(xmpp): fix race in p2p set codecs

nils-ohlmeier:


modules/xmpp/JingleSessionPC.js
```diff
@@ -1060,15 +1060,29 @@ export default class JingleSessionPC extends JingleSession {
...
+                        return this._responderRenegotiate(remoteDescription)
+                        .then(() => finishedCallback(), error => finishedCallback(error));
+                    };
+
+                    logger.debug(`${this} Queued responderRenegotiate task`);
+                    this.modificationQueue.push(
+                        workFunction,
+                        error => {
+                            if (error) {
+                                logger.error(`${this} failed to renegotiate a decoder for muted ednpoint ${error}`);
```
jallamsetty1, line 1081: nit: typo

modules/xmpp/JingleSessionPC.js
```diff
@@ -1060,15 +1060,29 @@ export default class JingleSessionPC extends JingleSession {
...
+                            } else {
+                                logger.debug(`${this} renegotiate a decoder for muted ednpoint`);
```
jallamsetty1, line 1083: same as above

jallamsetty1 (commented):
Good catch! Thanks for tracking it down. LGTM !!

PR #1829, opened 2021-12-20
fix: Drop unused region info from presence.

damencho:
We set the region using a custom extension https://github.com/jitsi/jitsi-xmpp-extensions/blob/master/src/main/java/org/jitsi/xmpp/extensions/jitsimeet/RegionPacketExtension.java in ChatRoom.initPresenceMap, and we use that in jicofo.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -167,16 +167,6 @@ export default class ChatRoom extends Listenable {
             });
         }
 
-        if (options.deploymentInfo && options.deploymentInfo.userRegion) {
-            this.presMap.nodes.push({
-                'tagName': 'region',
-                'attributes': {
-                    id: options.deploymentInfo.userRegion,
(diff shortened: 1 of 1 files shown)


bgrozev:
Removing it from here will result in the region not being available as a `JitsiParticipant` property. It is currently used for [e2eping](https://github.com/jitsi/jitsi-meet/blob/master/react/features/connection-indicator/statsEmitter.js#L43), which is something we could remove altogether. But I think it's also being used somehow to display the region in the UI (I can't trace the code).

damencho:
I was searching for its usage and couldn't find one, don't we use just the other one ...

damencho:
I will look around again. Do we want to drop e2eeping?

damencho:
I managed to fix and the region missing in the UI https://github.com/jitsi/jitsi-meet/pull/10676
Now I started to wonder should we drop the other one, although we have an xmpp extension for it ... and implement in jicofo to read both for some time ... wdyt?

hristoterezov (approved):

PR #1840, opened 2022-01-10
fix: p2p reject reason

sapkra:
Additionally I was wondering if it is still a problem to have a P2P connections between Chrome, Firefox and Safari?


NewEraCracker:
@sapkra Not sure what is the status right now. But roughly a month ago I tested by removing Firefox from the hardcoded disabled list and, although call establish over P2P with a 2nd Chrome participant, I had audio quality issues - sounding like a robot.

saghul (approved):

PR #1841, opened 2022-01-10
fix: add required events dependency for browser environments

sapkra:
Currently lib-jitsi-meet imports `events` but it's not defined in the package.json. The IDE thinks it imports the built-in module of node.js but webpack actually resolves it as the `events` package which is installed as a dependency of webpack itself. This PR clarifies which module is actually imported.


saghul (approved):

saghul:

saghul:
OOps, sorry this one got lost, can you PL rebase?

PR #1843, opened 2022-01-10
fix: ensure mucNickname

gcascio:
Fixes the error:

`Strophe: error: Cannot read properties of null (reading 'substr')`

when the jid has no node part.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -326,14 +326,14 @@ JitsiConference.resourceCreator = function(jid, isAuthenticatedUser) {
         // We try to use the first part of the node (which for anonymous users
         // on prosody is a UUID) to match the previous behavior (and maybe make
         // debugging easier).
-        mucNickname = Strophe.getNodeFromJid(jid).substr(0, 8)
+        mucNickname = Strophe.getNodeFromJid(jid)?.substr(0, 8)
             .toLowerCase();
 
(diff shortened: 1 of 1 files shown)


damencho:
What is the use case here? How to reproduce this? How it is possible to not have a resource there?

gcascio:
I could not reproduce this yet. This PR is a result of going through my sentry issues and the mentioned error occurred a couple of times. A cause could be some issues with the JVB which popped up around the same time. Nonetheless I thought this case could be avoided here. If there is no real need to catch this (my understanding of jitsi is only superficial) feel free to close this.

damencho:
> A cause could be some issues with the JVB which popped up around the same time.

Jvb never connects to the client room.

sapkra:
Even though it should never happen, if all goes well, it is still not coded correctly. `getNodeFromJid` can return `null`, so when we use it we should also handle this return value.

https://github.com/strophe/strophejs/blob/cb6e658969b3ca2636400ecad86f4f512cfbd25d/src/core.js#L660

damencho:
@sapkra I agree. 

@gcascio Can you sign the CLA or someone adds you to the list of contributors in the company CLA.

gcascio:
@damencho sure, I just signed the individual CLA.

damencho:
Jenkins test this please

saghul (approved):

PR #1850, opened 2022-01-13
feat(RTC): added DTLS transport monitoring

nils-ohlmeier:
This PR adds monitoring and logging of the DTLS transport object.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -192,6 +192,13 @@ export default function TraceablePeerConnection(
     this.remoteUfrag = null;
 
     /**
+     * The DTLS transport object for the PeerConnection.
+     * Note: this assume only one shared transport exists because we bundled
+     *       all streams on the same underlying transport.
+     */
+    this._dtlsTransport = null;
+
+    /**
      * The signaling layer which operates this peer connection.
      * @type {SignalingLayer}
      */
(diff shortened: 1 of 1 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2289,6 +2296,31 @@ TraceablePeerConnection.prototype._mungeOpus = function(description) {
...
+
+    if (senders.length !== 0 && senders[0].transport) {
+        this._dtlsTransport = senders[0].transport;
+
+        this._dtlsTransport.onerror = error => {
+            logger.error(`${this} DtlsTransport error: ${error}`);
+        };
+
+        this._dtlsTransport.onstatechange = () => {
+            this.trace('dtlsTransport.onstatechange', this._dtlsTransport.state);
```
jallamsetty1, line 2318: Do we need to handle the cases when the DTLS transport state changes to `failed` ? Does the ICE connection state also change when that happens ? If that is the case, we are good but otherwise I am wondering the client might end up in a broken state where it expects media to be sent and received. Do we need to attempt to reload the client in those cases ?

nils-ohlmeier: So if the ICE succeeded and DTLS fails, ICE will stay in the success stated. In other words: a DTLS failure doesn't result in any other state changes.

A very good point. Yeah we could/should attempt to do something if we encounter a failure. I would propose that I'll do an action part as a follow up to this. Because honestly I'm not sure how the failures will look like. I have not tried to test a failure locally so far.
I would think the `onerror` callback is where we would want to hook up any actions.

jallamsetty1 (approved):

saghul:
Does this code run on mobile? We have no getSenders() there.

nils-ohlmeier:
> Does this code run on mobile? We have no getSenders() there.

I did not exclude any platforms. So I would assume the answer is yes. What is the best way to test this?

jallamsetty1:
> Does this code run on mobile? We have no getSenders() there.

Good point, I forgot about that :(

PR #1852, opened 2022-01-17
fix: recover from reservation error

shawnchin:
Ref: https://community.jitsi.org/t/ux-when-reservation-fails/110095/9?u=shawn

Right now, users are not able to retry join after focus returns reservation-error; strophe emuc plugin caches room object and no longer allows re-creation because it thinks room already created.

This fix adds a listener for reservation errors and expunges room object from emuc cache.

**Update:** This issue is also present for `XMPPEvents.ROOM_MAX_USERS_ERROR`. When max users reached, user sees error and can no longer retry join.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -3007,6 +3007,19 @@ JitsiConference.prototype._onIceConnectionRestored = function(session) {
 };
 
 /**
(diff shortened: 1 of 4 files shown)


damencho:
Jenkins test this please

shawnchin:
This is fixed by https://github.com/jitsi/lib-jitsi-meet/pull/1878

PR #1853, opened 2022-01-18
fix(TPC) Do not change the tranceiver dir when the connection is not …

jallamsetty1:
…active.

Fixes https://github.com/jitsi/jitsi-meet/issues/10754.


jallamsetty1:
Jenkins test this please

damencho (approved):

PR #1855, opened 2022-01-19
feat(multi-stream-support) Add track streaming status

WillLiang918:
This change allow us handle the streaming status of each individual track when the feature flag is enabled. The emitted streaming status is mostly used for the connection indicator in the UI. 

Related jitsi-meet PR: https://github.com/jitsi/jitsi-meet/pull/10934 

Todo/Follow up item:
- Look into moving some of the `TrackStreamingStatus#onForwardedSourcesChanged` logic out into `RTC#_onForwardedSourcesChanged`. The reason being is that since TrackStreamingStatus is only initialized when a track is rendered, there is a corner case where we are not capturing the timestamp of when it entered forwarded sources.
- Add specs, this can be part of another PR.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1613,6 +1613,14 @@ JitsiConference.prototype.getLastN = function() {
 };
 
 /**
+ * Obtains the forwarded sources list in this conference.
+ * @return {Array<string>|null}
+ */
+JitsiConference.prototype.getForwardedSources = function() {
+    return this.rtc.getForwardedSources();
+};
+
+/**
  * Selects a new value for "lastN". The requested amount of videos are going
  * to be delivered after the value is in effect. Set to -1 for unlimited or
  * all available videos.
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2136,8 +2136,7 @@ JitsiConference.prototype.onRemoteTrackRemoved = function(removedTrack) {
                 // considered equal the result of splice can be ignored.
                 participant._tracks.splice(i, 1);
 
-                this.eventEmitter.emit(
-                    JitsiConferenceEvents.TRACK_REMOVED, removedTrack);
+                this.eventEmitter.emit(JitsiConferenceEvents.TRACK_REMOVED, removedTrack);
 
                 if (this.transcriber) {
                     this.transcriber.removeTrack(removedTrack);
--- a/JitsiConferenceEvents.ts
+++ b/JitsiConferenceEvents.ts
@@ -154,6 +154,16 @@ export enum JitsiConferenceEvents {
     LAST_N_ENDPOINTS_CHANGED = 'conference.lastNEndpointsChanged',
 
     /**
+     * The forwarded sources set is changed.
+     *
+     * @param {Array<string>|null} leavingForwardedSources the sourceNames of all the tracks which are leaving forwarded
+     * sources
+     * @param {Array<string>|null} enteringForwardedSources the sourceNames of all the tracks which are entering forwarded
+     * sources
+     */
+    FORWARDED_SOURCES_CHANGED = 'conference.forwardedSourcesChanged',
+
+    /**
      * Indicates that the room has been locked or unlocked.
      */
     LOCK_STATE_CHANGED = 'conference.lock_state_changed',
--- a/JitsiConferenceEvents.ts
+++ b/JitsiConferenceEvents.ts
@@ -491,6 +491,7 @@ export const JVB121_STATUS = JitsiConferenceEvents.JVB121_STATUS;
 export const KICKED = JitsiConferenceEvents.KICKED;
 export const PARTICIPANT_KICKED = JitsiConferenceEvents.PARTICIPANT_KICKED;
 export const LAST_N_ENDPOINTS_CHANGED = JitsiConferenceEvents.LAST_N_ENDPOINTS_CHANGED;
+export const FORWARDED_SOURCES_CHANGED = JitsiConferenceEvents.FORWARDED_SOURCES_CHANGED;
 export const LOCK_STATE_CHANGED = JitsiConferenceEvents.LOCK_STATE_CHANGED;
 export const SERVER_REGION_CHANGED = JitsiConferenceEvents.SERVER_REGION_CHANGED;
 export const _MEDIA_SESSION_STARTED = JitsiConferenceEvents._MEDIA_SESSION_STARTED;
--- a/JitsiConferenceEvents.ts
+++ b/JitsiConferenceEvents.ts
@@ -156,9 +156,9 @@ export enum JitsiConferenceEvents {
     /**
      * The forwarded sources set is changed.
      *
-     * @param {Array<string>|null} leavingForwardedSources the sourceNames of all the tracks which are leaving forwarded
+     * @param {Array<string>} leavingForwardedSources the sourceNames of all the tracks which are leaving forwarded
      * sources
-     * @param {Array<string>|null} enteringForwardedSources the sourceNames of all the tracks which are entering forwarded
+     * @param {Array<string>} enteringForwardedSources the sourceNames of all the tracks which are entering forwarded
      * sources
      */
     FORWARDED_SOURCES_CHANGED = 'conference.forwardedSourcesChanged',
--- a/JitsiTrackEvents.js
+++ b/JitsiTrackEvents.js
@@ -43,3 +43,16 @@ export const NO_DATA_FROM_SOURCE = 'track.no_data_from_source';
  * the microphone that is currently selected.
  */
 export const NO_AUDIO_INPUT = 'track.no_audio_input';
+
+/**
+ * Event fired when we detect local problem with the video track.
+ * First argument is the sourceName of the track and the second is a string indicating if the connection is currently
+ * - active - the connection is active.
+ * - inactive - the connection is inactive, was intentionally interrupted by the bridge because of low BWE or because
+ *   of the endpoint falling out of last N.
+ * - interrupted - a network problem occurred.
+ * - restoring - the connection was inactive and is restoring now.
+ *
+ * The current status value can be obtained by calling JitsiRemoteTrack.getTrackStreamingStatus().
+ */
+export const TRACK_STREAMING_STATUS_CHANGED = 'track.streaming_status_changed';
--- a/JitsiTrackEvents.js
+++ b/JitsiTrackEvents.js
@@ -45,7 +45,7 @@ export const NO_DATA_FROM_SOURCE = 'track.no_data_from_source';
 export const NO_AUDIO_INPUT = 'track.no_audio_input';
 
 /**
- * Event fired when we detect local problem with the video track.
+ * Event fired whenever video track's streaming changes.
  * First argument is the sourceName of the track and the second is a string indicating if the connection is currently
  * - active - the connection is active.
  * - inactive - the connection is inactive, was intentionally interrupted by the bridge because of low BWE or because
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -359,11 +359,25 @@ export default class BridgeChannel {
                 break;
             }
             case 'LastNEndpointsChangeEvent': {
-                // The new/latest list of last-n endpoint IDs (i.e. endpoints for which the bridge is sending video).
-                const lastNEndpoints = obj.lastNEndpoints;
+                if (!FeatureFlags.isSourceNameSignalingEnabled()) {
+                    // The new/latest list of last-n endpoint IDs (i.e. endpoints for which the bridge is sending
+                    // video).
+                    const lastNEndpoints = obj.lastNEndpoints;
 
-                logger.info(`New forwarded endpoints: ${lastNEndpoints}`);
-                emitter.emit(RTCEvents.LASTN_ENDPOINT_CHANGED, lastNEndpoints);
+                    logger.info(`New forwarded endpoints: ${lastNEndpoints}`);
+                    emitter.emit(RTCEvents.LASTN_ENDPOINT_CHANGED, lastNEndpoints);
+                }
+
+                break;
+            }
+            case 'ForwardedSources': {
+                if (FeatureFlags.isSourceNameSignalingEnabled()) {
+                    // The new/latest list of forwarded sources
+                    const forwardedSources = obj.forwardedSources;
+
+                    logger.info(`New forwarded sources: ${forwardedSources}`);
+                    emitter.emit(RTCEvents.FORWARDED_SOURCES_CHANGED, forwardedSources);
+                }
 
                 break;
             }
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -1,5 +1,7 @@
 import * as JitsiTrackEvents from '../../JitsiTrackEvents';
 import { createTtfmEvent } from '../../service/statistics/AnalyticsEvents';
+import TrackStreamingStatusImpl, { TrackStreamingStatusMap } from '../connectivity/TrackStreamingStatus';
+import FeatureFlags from '../flags/FeatureFlags';
 import Statistics from '../statistics/statistics';
 
 import JitsiTrack from './JitsiTrack';
@@ -74,6 +76,19 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.muted = muted;
         this.isP2P = isP2P;
         this._sourceName = sourceName;
+        this._trackStreamingStatus = null;
+        this._trackStreamingStatusImpl = null;
+
+        /**
+         * This holds the timestamp indicating when remote video track entered forwarded sources set. Track entering
+         * forwardedSources will have streaming status restoring and when we start receiving video will become active,
+         * but if video is not received for certain time {@link DEFAULT_RESTORING_TIMEOUT} that track streaming status
+         * will become interrupted.
+         */
+        this._enteredForwardedSourcesTimestamp = null;
+
+        this.addEventListener = this.on = this._addEventListener.bind(this);
+        this.removeEventListener = this.off = this._removeEventListener.bind(this);
 
         logger.debug(`New remote track added: ${this}`);
 
@@ -107,6 +122,60 @@ export default class JitsiRemoteTrack extends JitsiTrack {
     }
 
     /**
+     * Overrides addEventListener method to init TrackStreamingStatus instance when there are listeners for the
+     * {@link JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED} event.
+     *
+     * @param {string} event - event name
+     * @param {function} handler - event handler
+     */
+    _addEventListener(event, handler) {
+        super.addListener(event, handler);
+
+        if (FeatureFlags.isSourceNameSignalingEnabled()
+            && event === JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED
+            && this.listenerCount(JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED)
+            && !this._trackStreamingStatusImpl
+        ) {
+            const config = this.conference.options.config;
+
+            this._trackStreamingStatus = TrackStreamingStatusMap.ACTIVE;
+
+            this._trackStreamingStatusImpl = new TrackStreamingStatusImpl(
+                this.rtc,
+                this.conference,
+                this,
+                {
+                    // These options are not public API, leaving it here only as an entry point through config for
+                    // tuning up purposes. Default values should be adjusted as soon as optimal values are discovered.
+                    p2pRtcMuteTimeout: config._p2pConnStatusRtcMuteTimeout,
+                    rtcMuteTimeout: config._peerConnStatusRtcMuteTimeout,
+                    outOfForwardedSourcesTimeout: config._peerConnStatusOutOfLastNTimeout
+                });
+
+            this._trackStreamingStatusImpl.init();
+            logger.debug(`Initializing track streaming status: ${this._sourceName}`);
+        }
+    }
+
+    /**
+     * Overrides removeEventListener method to dispose TrackStreamingStatus instance.
+     *
+     * @param {string} event - event name
+     * @param {function} handler - event handler
+     */
+    _removeEventListener(event, handler) {
+        super.removeListener(event, handler);
+
+        if (FeatureFlags.isSourceNameSignalingEnabled()
+            && event === JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED
+            && !this.listenerCount(JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED)
+        ) {
+            this.disposeTrackStreamingStatus();
+            logger.debug(`Disposing track streaming status: ${this._sourceName}`);
+        }
+    }
+
+    /**
      * Callback invoked when the track is muted. Emits an event notifying
      * listeners of the mute event.
      *
@@ -130,6 +199,19 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         logger.debug(`"onunmute" event(${Date.now()}): ${this}`);
 
         this.rtc.eventEmitter.emit(RTCEvents.REMOTE_TRACK_UNMUTE, this);
+    }
+
+    /**
+     * Removes attached event listeners and dispose TrackStreamingStatus .
+     *
+     * @returns {Promise}
+     */
+    dispose() {
+        if (FeatureFlags.isSourceNameSignalingEnabled()) {
+            this.disposeTrackStreamingStatus();
+        }
+
+        return super.dispose();
     }
 
     /**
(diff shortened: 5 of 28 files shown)


modules/RTC/BridgeChannel.js
```diff
@@ -358,11 +359,26 @@ export default class BridgeChannel {
                 break;
             }
             case 'LastNEndpointsChangeEvent': {
-                // The new/latest list of last-n endpoint IDs (i.e. endpoints for which the bridge is sending video).
-                const lastNEndpoints = obj.lastNEndpoints;
+                if (!FeatureFlags.isSourceNameSignalingEnabled()) {
```
WillLiang918, line 362: I've added this check to indirectly stop emitting the `JitsiConferenceEvents.LAST_N_ENDPOINTS_CHANGED` and `JitsiConferenceEvents.PARTICIPANT_CONN_STATUS_CHANGED` legacy events when the feature flag is enabled.

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,758 @@
...
+            : this.conference.isP2PActive() ? this.p2pRtcMuteTimeout : this.outOfLastNTimeout;
+    }
+
+    /**
+     * Initializes <tt>TrackStreamingStatus</tt> and bind required event
+     * listeners.
+     */
+    init() {
+        // Handles P2P status changes
+        this._onP2PStatus = this.figureOutConnectionStatus.bind(this);
```
WillLiang918, line 293: Since `TrackStreamingStatus` is scoped to a single track, we no longer need the `ParticipantConnectionStatus#refreshConnectionStatusForAll` method. Each instances will now handle refreshing its own connection status.

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,758 @@
...
+     * the track for whom the values above were provided.
+     * @private
+     */
+    static _getNewStateForJvbMode(
+            isInForwardedSources,
+            isRestoringTimedout,
+            isVideoMuted,
+            isVideoTrackFrozen) {
+
+        // We are currently not checking the endpoint connection status received from the JVB.
```
WillLiang918, line 108: The logic to store (`connStatusFromJvb`) and check the endpoint connection status from JVB have been removed.

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,758 @@
...
+        this.rtc = rtc;
+        this.conference = conference;
+        this.track = track;
+
+        /**
+         * This holds the timeout callback ID scheduled using
+         * window.setTimeout.
+         * @type {number}
+         */
+        this.trackTimer = null;
```
WillLiang918, line 185: `this.trackTimers`, `this.rtcMutedTimestamp`, `this.enteredLastNTimestamp`, `this.restoringTimers`, and `this.connectionStatusMap` are now flattened to no longer hold a mapping of endpointId to value.

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,758 @@
...
+            this._onTrackRtcMuted = this.onTrackRtcMuted.bind(this);
+            this.rtc.addListener(
+                RTCEvents.REMOTE_TRACK_MUTE, this._onTrackRtcMuted);
+
+            this._onTrackRtcUnmuted = this.onTrackRtcUnmuted.bind(this);
+            this.rtc.addListener(
+                RTCEvents.REMOTE_TRACK_UNMUTE, this._onTrackRtcUnmuted);
+
+            // Listened which will be bound to JitsiRemoteTrack to listen for
+            // signalling mute/unmute events.
```
WillLiang918, line 314: In the legacy code, the `JitsiTrackEvents.TRACK_MUTE_CHANGED` and `JitsiTrackEvents.TRACK_VIDEOTYPE_CHANGED` event listeners are registered onto a track when the `JitsiConferenceEvents.TRACK_ADDED` event is emitted. This logic has been moved into the `init` method of `TrackStreamingStatus`.

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,758 @@
...
+    dispose() {
+        if (browser.supportsVideoMuteOnConnInterrupted()) {
+            this.rtc.removeListener(
+                RTCEvents.REMOTE_TRACK_MUTE,
+                this._onTrackRtcMuted);
+            this.rtc.removeListener(
+                RTCEvents.REMOTE_TRACK_UNMUTE,
+                this._onTrackRtcUnmuted);
+
+            this.track.off(
```
WillLiang918, line 353: The logic to handle the `JitsiConferenceEvents.TRACK_REMOVED` event has been moved into the `dispose` method of this class.

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,758 @@
...
+            this.track._setConnectionStatus(newStatus);
+
+            logger.debug(
+                `Emit track connection status(${Date.now()}) ${sourceName}: ${
+                    newStatus}`);
+
+            // Log the event on CallStats
+            Statistics.sendLog(
+                JSON.stringify({
+                    id: 'track.conn.status',
```
WillLiang918, line 395: Is there another part in the stack that needs to be updated in order for CallStats to handle this new id?

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,738 @@
...
+        this._onUserLeft = this.onUserLeft.bind(this);
+        this.conference.on(JitsiConferenceEvents.USER_LEFT, this._onUserLeft);
+
+        // On some browsers MediaStreamTrack trigger "onmute"/"onunmute"
+        // events for video type tracks when they stop receiving data which is
+        // often a sign that remote user is having connectivity issues
+        if (browser.supportsVideoMuteOnConnInterrupted()) {
+
+            this._onTrackRtcMuted = this.onTrackRtcMuted.bind(this);
+            this.rtc.addListener(
```
WillLiang918, line 306: Will it make more sense to move all this logic to register rtc event listeners (`REMOTE_TRACK_MUTE, REMOTE_TRACK_UNMUTE, LASTN_VALUE_CHANGED`) into `JitsiConference`? Maybe in `_registerRtcListeners` since it is already behind the feature flag?

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,738 @@
...
+
+    /**
+     * On change in forwarded sources set check all leaving and entering track to
+     * change their corresponding statuses.
+     *
+     * @param {Array<string>} leavingForwardedSources - The array of sourceName leaving forwarded sources.
+     * @param {Array<string>} enteringForwardedSources - The array of sourceName entering forwarded sources.
+     * @private
+     */
+    _onForwardedSourcesChanged(leavingForwardedSources = [], enteringForwardedSources = []) {
```
WillLiang918, line 548: I don't think this will work as I expect it to. I'll need to move this logic somewhere else, perhaps `JitsiConferenceEventManager`, and have it iterate through all the leaving and entering forwarded sources.

modules/RTC/BridgeChannel.js
paweldomas: yes, that's good

modules/RTC/TrackStreamingStatus.js
paweldomas: I think it's okay to register here. Otherwise the `JitsiConference` would have to call into this class which would tie them together (right now `JitsiConference` is not aware of this class - I think). Why do you think it's better to do it there? Is it because the events appear to be global in a way?

modules/RTC/TrackStreamingStatus.js
paweldomas: And when a track is removed this instance will be disposed anyway, right?

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,758 @@
...
+        this.onUserLeft(this.track.getParticipantId());
+        this.figureOutConnectionStatus();
+    }
+
+    /**
+     * Changes connection status.
+     * @param newStatus
+     */
+    _changeConnectionStatus(newStatus) {
+        if (this.track.getConnectionStatus() !== newStatus) {
```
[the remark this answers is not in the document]
paweldomas: yes, lets call it streaming status

modules/RTC/TrackStreamingStatus.js
paweldomas: I don't think this part is particularly important (it only logs a custom event on CS timeline), but lets change the `id` to be `track.streaming.status`

modules/RTC/TrackStreamingStatus.js
paweldomas: Ah because the timestamps will not be accurate since this instance may disappear. Lets try to think about it... Maybe we need to store the timestamps somewhere (the place which emits the forwarded sources changed event) and then expose that via a getter? Or better have a static part/helper class of `TrackStreamingStatus` which would keep track of the timestamps (mostly because they are not used by any other logic) ?

JitsiTrackEvents.js
```diff
@@ -43,3 +43,18 @@ export const NO_DATA_FROM_SOURCE = 'track.no_data_from_source';
...
  */
 export const NO_AUDIO_INPUT = 'track.no_audio_input';
+
+/**
+ * Event fired when we detect local problem with the video track.
+ * First argument is the sourceName of the track and
+ * the seconds is a string indicating if the connection is currently
+ * - active - the connection is active
+ * - inactive - the connection is inactive, was intentionally interrupted by
+ * the bridge
```
jallamsetty1, line 53: You might want to mention that the media was stopped by the bridge for this ep because of low BWE.

jallamsetty1 (commented):
Great work @WillLiang918. Left few minor comments. One complaint that I have is there is a lot of code duplication.

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,738 @@
...
+                = this.onSignallingMuteChanged.bind(this);
+            this.track.on(
+                JitsiTrackEvents.TRACK_MUTE_CHANGED,
+                this._onSignallingMuteChanged);
+
+            // Used to send an analytics event when the video type changes.
+            this._onTrackVideoTypeChanged
+                = this.onTrackVideoTypeChanged.bind(this);
+            this.track.on(
+                JitsiTrackEvents.TRACK_VIDEOTYPE_CHANGED,
```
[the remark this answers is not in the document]
paweldomas: Technically they still can change. We're not freezing the video type yet.

JitsiConferenceEvents.js
```diff
@@ -152,6 +152,16 @@ export const PARTICIPANT_KICKED = 'conference.participant_kicked';
  */
 export const LAST_N_ENDPOINTS_CHANGED = 'conference.lastNEndpointsChanged';
 
+/**
+ * The forwarded sources set is changed.
+ *
+ * @param {Array<string>|null} leavingForwardedSources the sourceNames of all the tracks
```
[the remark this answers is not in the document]
WillLiang918: Absolutely! I'll update this for all the new code

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -318,6 +392,27 @@ export default class JitsiRemoteTrack extends JitsiTrack {
...
+    }
+
+    /**
+     * Return track's connectivity status.
+     *
+     * @returns {string} the connection status
+     * <tt>TrackStreamingStatus</tt> of the track.
+     * {@link TrackStreamingStatus}.
+     */
+    getConnectionStatus() {
```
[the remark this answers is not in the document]
WillLiang918: I'll update the naming throughout the code. I'll thinking about the no listener scenario a bit more

modules/RTC/TrackStreamingStatus.js
```diff
@@ -0,0 +1,738 @@
+import { getLogger } from '@jitsi/logger';
```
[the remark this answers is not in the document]
WillLiang918: I see, it might make more sense to have this grouped under connectivity since ParticipantConnectionStatus also lives there

modules/RTC/TrackStreamingStatus.js
WillLiang918: Yes in a way it does feel like these are global events so I was hesitant on having them in here but I see both of your points now

modules/RTC/TrackStreamingStatus.js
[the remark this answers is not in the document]
WillLiang918: I'll add a note to remove this when we start freezing the video type

modules/RTC/TrackStreamingStatus.js
WillLiang918: I will double check to confirm if that is true. If not, I can dispose it where we are emitting `JitsiConferenceEvents.TRACK_REMOVED`

modules/RTC/TrackStreamingStatus.js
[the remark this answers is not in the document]
WillLiang918: 👍 I will rename it throughout the new code

modules/connectivity/TrackStreamingStatus.ts
```diff
@@ -0,0 +1,711 @@
+import { getLogger } from '@jitsi/logger';
+
+import * as JitsiConferenceEvents from '../../JitsiConferenceEvents';
+import * as JitsiTrackEvents from '../../JitsiTrackEvents';
+import RTCEvents from '../../service/RTC/RTCEvents';
+import { createTrackStreamingStatusEvent } from '../../service/statistics/AnalyticsEvents';
+import JitsiConference from '../../types/hand-crafted/JitsiConference';
+import JitsiRemoteTrack from '../../types/hand-crafted/modules/RTC/JitsiRemoteTrack';
+import RTC from '../../types/hand-crafted/modules/RTC/RTC';
```
WillLiang918, line 9: I'm importing the hand-crafted types for `JitsiConference`, `JitsiRemoteTrack` and `RTC` here because I ran into issues with the auto generated ones. I want to limit the scope of this PR since it is already growing quite large

types/hand-crafted/JitsiConference.d.ts
```diff
@@ -10,6 +10,8 @@ import JitsiVideoSIPGWSession from './modules/videosipgw/JitsiVideoSIPGWSession'
 import TraceablePeerConnection from './modules/RTC/TraceablePeerConnection';
 import { MediaType } from './service/RTC/MediaType';
 
+type JitsiConferenceEventsType = `${JitsiConferenceEvents}`
```
WillLiang918, line 13: I had to use template literal types here to get it working as expected since the parameter is an enum and we are passing string literals as arguments

tsconfig.json
```diff
@@ -14,6 +14,7 @@
       "dom"
     ],
     "outDir": "dist/esm/",
+    "moduleResolution": "Node"
```
WillLiang918, line 17: I added this to resolved scoped modules (`@jitsi/logger`)

modules/RTC/TrackStreamingStatus.js
WillLiang918: I don't think I saw any logic in LJM that called `dispose` when a tracked was removed. I wasn't sure if this is intended so I added logic to clear `_trackStreamingStatusImpl` and `_trackStreamingStatus` where the event was emitted

modules/RTC/TrackStreamingStatus.js
WillLiang918: I was looking into using a static method/helper class and it was starting to get a bit messy since it would still need to call `figureOutStreamingStatus` for each instances that is leaving/entering or when `!browser.supportsVideoMuteOnConnInterrupted()`. If our only concern is having inaccurate timestamps and performance isn't an issue, can we just emit the timestamp with the forwarded sources?

modules/connectivity/TrackStreamingStatus.ts
WillLiang918: I can also cast `rtc`, `conference`, and `track` to the `any` type for now if we shouldn't be importing these hand-crafted types

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -74,6 +76,12 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.muted = muted;
         this.isP2P = isP2P;
         this._sourceName = sourceName;
+        this._connectionStatus = TrackStreamingStatus.ACTIVE;
+
+        this.addEventListener = this._addEventListener.bind(this);
+        this.removeEventListener = this.off = this._removeEventListener.bind(this);
+
+        this.trackStreamingStatus = null;
```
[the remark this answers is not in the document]
WillLiang918: I think `_trackStreamingStatusImpl` is a good name =)

modules/RTC/JitsiRemoteTrack.js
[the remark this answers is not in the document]
WillLiang918: I think when there are no listener registered it makes more sense to be set to `null` since `_trackStreamingStatusImpl` hasn't been initialized yet to handle the streaming status. I'm not sure where else this would be used but it might lead to some confusion if this always active

modules/connectivity/TrackStreamingStatus.ts
```diff
@@ -0,0 +1,683 @@
...
+
+    /**
+     * If video track frozen detection through RTC mute event is supported, we wait some time until video track is
+     * considered frozen. But because when the track falls out of forwarded sources it is expected for the video to
+     * freeze this timeout must be significantly reduced in "out of forwarded sources" case.
+     *
+     * Basically this value is used instead of {@link rtcMuteTimeout} when
+     * track is not in forwarded sources.
+     */
+    outOfForwardedSourcesTimeout: number;
```
WillLiang918, line 95: Something I've noticed is that after changing this file's extension to .ts and running `npm run build`, all the class properties types were removed in the auto generated type declarations file. I had to explicitly define them here

modules/RTC/TrackStreamingStatus.js
paweldomas: We can actually make it so that the `JitsiRemoteTrack` holds the timestamps when it entered/left the forwarded sources and this can happen any time the `RTC` module emits the forwarded sources event?

JitsiConference.js
```diff
@@ -1905,6 +1905,10 @@ JitsiConference.prototype.onMemberLeft = function(jid) {
 
     // Fire the event before renegotiation is done so that the thumbnails can be removed immediately.
     tracksToBeRemoved.forEach(track => {
+        if (FeatureFlags.isSourceNameSignalingEnabled() && track.getTrackStreamingStatus()) {
+            track.disposeTrackStreamingStatus();
```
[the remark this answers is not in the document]
WillLiang918: I was confused about the difference between having an instance of a track completely removed and having it just disposed where the `disposed` field is true. However in any case, I think it is good practice to call `dispose` whenever a track is removed. I don't think there should be any side effects but I'll still keep it behind the feature flag for now

JitsiConference.js
```diff
@@ -2103,8 +2107,11 @@ JitsiConference.prototype.onRemoteTrackRemoved = function(removedTrack) {
                 // considered equal the result of splice can be ignored.
                 participant._tracks.splice(i, 1);
 
-                this.eventEmitter.emit(
-                    JitsiConferenceEvents.TRACK_REMOVED, removedTrack);
+                if (FeatureFlags.isSourceNameSignalingEnabled() && removedTrack.getTrackStreamingStatus()) {
```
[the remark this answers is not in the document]
WillLiang918: Similar to other comment, I'll update

JitsiConference.js
```diff
@@ -1925,6 +1925,10 @@ JitsiConference.prototype.onMemberLeft = function(jid) {
 
     // Fire the event before renegotiation is done so that the thumbnails can be removed immediately.
     tracksToBeRemoved.forEach(track => {
+        if (FeatureFlags.isSourceNameSignalingEnabled()) {
+            track.dispose();
```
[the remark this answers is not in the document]
WillLiang918: Initially I just wanted to cleanup the code related to track streaming status, this was an oversight on my end. After taking a deeper look into the jitsi-meet code, we already have listeners to the `JitsiConferenceEvents.TRACK_REMOVED` event which can be use to trigger a cleanup by removing listeners for `JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED`. I'll remove this

JitsiConference.js
```diff
@@ -2123,8 +2127,11 @@ JitsiConference.prototype.onRemoteTrackRemoved = function(removedTrack) {
                 // considered equal the result of splice can be ignored.
                 participant._tracks.splice(i, 1);
 
-                this.eventEmitter.emit(
-                    JitsiConferenceEvents.TRACK_REMOVED, removedTrack);
+                if (FeatureFlags.isSourceNameSignalingEnabled()) {
```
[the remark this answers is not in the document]
WillLiang918: Similar to above, this will be removed

modules/RTC/JitsiRemoteTrack.js
```diff
@@ -132,6 +199,17 @@ export default class JitsiRemoteTrack extends JitsiTrack {
         this.rtc.eventEmitter.emit(RTCEvents.REMOTE_TRACK_UNMUTE, this);
     }
 
+    /**
+     * Removes attached event listeners and dispose TrackStreamingStatus .
+     *
+     * @returns {Promise}
+     */
+    dispose() {
+        this.disposeTrackStreamingStatus();
```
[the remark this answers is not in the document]
WillLiang918: We are checking if `this._trackStreamingStatusImpl` is defined in `disposeTrackStreamingStatus` which is only possible when FeatureFlags.isSourceNameSignalingEnabled is true. However, that could change and it would be good practice to also add it here

jallamsetty1:
Jenkins test this please.

modules/connectivity/TrackStreamingStatus.ts
```diff
@@ -0,0 +1,651 @@
...
+const DEFAULT_RTC_MUTE_TIMEOUT = 10000;
+
+/**
+ * The time to wait a track to be restored. Track which was out of forwarded sources should be inactive and when
+ * entering forwarded sources it becomes restoring and when data is received from bridge it will become active, but if
+ * no data is received for some time we set status of that track streaming to interrupted.
+ */
+const DEFAULT_RESTORING_TIMEOUT = 10000;
+
+export const TrackStreamingStatusMap: TrackStreamingStatusMapType = {
```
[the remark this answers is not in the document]
WillLiang918: I'll take another look into this. I think I was having issues when I first tried it because `TrackStreamingStatusMap` was being imported into non `.ts` files where `enum`s aren't supported

modules/connectivity/TrackStreamingStatus.ts
```diff
@@ -0,0 +1,651 @@
...
+import browser from '../browser';
+import Statistics from '../statistics/statistics';
+
+/** Track streaming statuses. */
+type TrackStreamingStatus = 'active' | 'inactive' | 'interrupted' | 'restoring';
+interface TrackStreamingStatusMapType {
+    [key: string]: TrackStreamingStatus
+}
+
+type VideoType = 'camera' | 'desktop';
```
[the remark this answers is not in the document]
jallamsetty1: Why can't we import VideoType from service/RTC/VideoType.js here directly ?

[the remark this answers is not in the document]
WillLiang918: We definitely can! I found that type after Pawel pointed this out. I'll push another set of changes out shortly after testing it locally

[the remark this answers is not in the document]
WillLiang918: Wait do you mean the VideoType enum in types/hand-crafted/services/RTC/VideoType.d.ts or the VideoType object in VideoType.js?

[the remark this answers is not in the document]
WillLiang918: I'm going to take a quick look to see if we can do something similar to the recent changes to /JitsiConferenceEvents.ts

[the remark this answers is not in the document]
WillLiang918: I'm going to revert my changes and adding a TODO to use the VideoType enum in https://github.com/jitsi/lib-jitsi-meet/pull/1916/files

JitsiConferenceEvents.ts
```diff
@@ -153,6 +153,16 @@ export enum JitsiConferenceEvents {
      */
     LAST_N_ENDPOINTS_CHANGED = 'conference.lastNEndpointsChanged',
 
+    /**
+     * The forwarded sources set is changed.
+     *
+     * @param {Array<string>|null} leavingForwardedSources the sourceNames of all the tracks which are leaving forwarded
```
paweldomas, line 159: For both params, the type is always an array (not null)

paweldomas (approved):
It's looking good! Left just one comment about JS docs

paweldomas (approved):

JitsiConferenceEvents.ts
WillLiang918: 👍  updated

PR #1856, opened 2022-01-20
fix(RTC) added safety checks before getSenders() calls

nils-ohlmeier:
When using getSenders() in another PR I learned that it is not available on all of our clients yet. This PR adds safety checks to prevent existing code from making the same mistake.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1954,7 +1954,9 @@ TraceablePeerConnection.prototype.removeTrack = function(localTrack) {
  * was found.
  */
 TraceablePeerConnection.prototype.findSenderByKind = function(mediaType) {
-    return this.peerconnection.getSenders().find(s => s.track && s.track.kind === mediaType);
+    if (this.peerconnection.getSenders) {
+        return this.peerconnection.getSenders().find(s => s.track && s.track.kind === mediaType);
+    }
 };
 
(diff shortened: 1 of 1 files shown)


jallamsetty1 (approved):

paweldomas:
I think maybe it was for the better to let it crash, because the existing usages silently return when no sender is returned, so we may end up scratching our heads to figure out why things are not working vs a crash on undefined getSenders is obvious. It blends together the failure cases of a sender not found and getSenders not supported on the platform.

nils-ohlmeier:
@paweldomas I see your point. But shouldn't it be the responsibility of the caller to check for error, in this case nothing getting returned?
If you think it's helpful I take a look at adding error handling to the callers of these functions.

paweldomas:
I was just saying that now the api of the changed methods does not allow to distinguish between unsupported case and a not found case, because both return `undefined`. I think usually the code that relies on the `getSenders` method should be doing a check for "is transceivers api supported" and if not supported try using another strategy (if possible). Now the fact that this code was crashing allowed to find those cases pretty easily, but if they start silently doing nothing it will be harder to find. I don't think it's a huge deal given there's not too many usage. I would probably rather see the change reverted than trying to spend time on sophisticated error handling. As long as there is some client monitoring in place for unhandled exceptions, it's usually easy to spot such crashes early and fix (while still on staging environment).

PR #1857, opened 2022-01-21
fix(TPC) add muted tracks if p2p is disabled (fix #1851)

### LINKED ISSUE #1851: With P2P disabled, unmuting the first participant's camera does not work if the second participant joined while it was muted
### Description:

If P2P is disabled, and the second participant joins while the first participant's camera is muted, the first participant can no longer unmute their camera (the video tracks do not get added).

### Steps to reproduce:

1. Open https://meet.jit.si/testingp2pdisabled#config.prejoinPageEnabled=false&config.p2p.enabled=false
2. Mute the camera.
3. Open the same page in a second browser tab (or second browser, or second computer).
4. Unmute the camera in the first browser tab.

### Expected behavior:

The camera should unmute and be visible to the second participant.

### Actual behavior:

The camera is not unmuted — the video tracks are not re-added — although the UI shows that it is unmuted.

### Version information:

Reproducible on 6726 but not on 6433. Currently reproducible on meet.jit.si.
Browser / app version: Chrome 97.0.4692.71 arm64
Operating System: macOS 12.0.1

Also tested with Firefox 96 on Mac and Linux — behaviour is the same.

### Additional information:

I've filed this on lib-jitsi-meet because it appears to affect custom UIs based on lib-jitsi-meet as well, I don't believe it's specific to jitsi-meet. This lib-jitsi-meet commit (cc @jallamsetty1) appears to introduce the issue: https://github.com/jitsi/lib-jitsi-meet/commit/e566291864046cadba2ef0af7fc2e31b513e3b4b as rolling back to an earlier commit resolves it.

NewEraCracker:
Fixes #1851. Please review.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1645,7 +1645,7 @@ TraceablePeerConnection.prototype.addTrack = function(track, isInitiator = false
 
     if (this._usesUnifiedPlan) {
         logger.debug(`${this} TPC.addTrack using unified plan`);
-        if (webrtcStream) {
+        if (webrtcStream || !this.isP2P) {
             try {
                 this.tpcUtils.addTrack(track, isInitiator);
             } catch (error) {


alan-albuquerque:
hi, any news on that?

jallamsetty1:
This fix will not work since we do not want to signal muted sources to Jicofo. I am out today, will take a look at the issue tomorrow.

jallamsetty1:
Jenkins test this please.

jallamsetty1:
I was wrong, actually this fix will work. Muted source will only be signaled for the first user that mutes themselves after joining the call and before the peer joins. The other alternative would be to force renegotiation on every unmute which is not desirable. Thanks @NewEraCracker for the fix and the PR.

jallamsetty1:
@NewEraCracker Can you please rebase your branch ? The build is failing since we have made some changes recently to the way we build LJM.

jallamsetty1:
Jenkins test this please.

jallamsetty1 (approved):

PR #1859, opened 2022-01-25
feat(JitsiConference) added end conference

wfleischer:
This PR is part of a feature which allows moderators to end the conference for all participants.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -732,6 +732,31 @@ JitsiConference.prototype.leave = async function(reason) {
 };
 
 /**
+ * Returns <tt>true</tt> if end conference support is enabled in the backend.
+ *
+ * @returns {boolean} whether end conference is supported in the backend.
+ */
+JitsiConference.prototype.isEndConferenceSupported = function() {
+    return Boolean(this.room && this.room.xmpp.endConferenceComponentAddress);
+};
+
+/**
+ * Ends the conference.
+ */
+JitsiConference.prototype.end = async function() {
+    if (!this.isEndConferenceSupported()) {
+        logger.warn('Cannot end conference: is not supported.');
+
+        return;
+    }
+    if (!this.room) {
+        throw new Error('The conference has been already left');
+    }
+
+    this.room.end();
+};
+
+/**
  * Returns the currently active media session if any.
  *
  * @returns {JingleSessionPC|undefined}
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -743,7 +743,7 @@ JitsiConference.prototype.isEndConferenceSupported = function() {
 /**
  * Ends the conference.
  */
-JitsiConference.prototype.end = async function() {
+JitsiConference.prototype.end = function() {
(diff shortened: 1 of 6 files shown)


JitsiConference.js
```diff
@@ -730,6 +730,31 @@ JitsiConference.prototype.leave = async function() {
...
+ * @returns {boolean} whether end conference is supported in the backend.
+ */
+JitsiConference.prototype.isEndConferenceSupported = function() {
+    return Boolean(this.room && this.room.xmpp.endConferenceComponentAddress);
+};
+
+/**
+ * Ends the conference.
+ */
+JitsiConference.prototype.end = async function() {
```
saghul, line 745: This function is not async, doesn't return a promise or await anything

modules/xmpp/ChatRoom.js
```diff
@@ -1861,6 +1861,18 @@ export default class ChatRoom extends Listenable {
 
         return Promise.allSettled(promises);
     }
+
+    /**
+     * Ends the conference.
```
saghul, line 1866: Maybe clarify that this affects all participants, and not just the local one.

JitsiConference.js
saghul: Should we protect ourselves from running this from a breakout room? I'd say we do. Also, what should happen if there are active breakout rooms?

saghul (commented):
Left some comments, PTAL.

JitsiConference.js
wfleischer: > This function is not async, doesn't return a promise or await anything

Good catch. Thanks.

wfleischer: > Should we protect ourselves from running this from a breakout room? I'd say we do. Also, what should happen if there are active breakout rooms?

We had the end meeting button hidden on the client while in a breakout room. But I totally agree that we should handle this case here, too. 

The breakout room module is designed to end the breakout rooms, when the main room gets destroyed. So participants in breakout rooms will get the same notification that the conference has ended as those in the main room. Of course the moderator can decide to gracefully close the breakout rooms beforehand, if he would like to announce to everyone that the conference will be ended. Does this make sense?

modules/xmpp/ChatRoom.js
wfleischer: I will add a note that this will end the conference for all participants. (The end function is in contrast to the leave function.)

JitsiConference.js
saghul: I think it does, yeah!

modules/xmpp/ChatRoom.js
saghul: 👍

saghul:
This is good to go, can you rebase it one last time pl?

wfleischer:
> This is good to go, can you rebase it one last time pl?

Thanks, @saghul! This should be ready now.

saghul:

saghul (approved):

PR #1862, opened 2022-01-26
fix(ProxyConnection) set signaling layer

saghul:
Regression introduced in
https://github.com/jitsi/lib-jitsi-meet/commit/5c9b90511120ae42660176afc0715f8c8886ef68


quitrk (approved):

paweldomas (approved):
Thanks for the fix!

jallamsetty1 (approved):

PR #1867, opened 2022-01-31
fix: Add private message handler for breakout_rooms

steve-vsee:
Add the _onPrivateMessage handler when breakout_rooms is found in disco info.

I think this is a bug.  If only breakout_rooms rooms is enabled then, the _onPrivateMessage handler never gets called for breakout room messages.

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -463,7 +463,8 @@ export default class XMPP extends Listenable {
 
         if (this.avModerationComponentAddress
             || this.speakerStatsComponentAddress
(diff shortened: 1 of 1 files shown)


damencho:
Jenkins test this please.

damencho:
@steve-vsee have you signed the CLA as I cannot find it?

saghul:
Wait this is very weird. How are getting BR messages now then?

saghul:
Ah wait it works because we always have one of the other components. Good catch. LGTM.

saghul (approved):

damencho:
Thanks for your contribution.

PR #1868, opened 2022-02-01
feat(multi-stream-support) Handle SDP munging for multiple local/remote streams per ep.

jallamsetty1:
A new SDPSimulcast class is added for handling SDP munging for local and remote descriptions to handle multiple streams for both local and remote endpoints. This new class will be used only for unified plan implentation. The sdp-simucast npm package will be used for plan-b and deprecated later when all the clients switch to unified plan.


jallamsetty1:
There is an issue with the unit tests because I am using TS for the new class, working on fixing that issue, will commit the unit tests once that is done.

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+            const numSsrcs = new Set(media.ssrcs?.map(ssrcInfo => ssrcInfo.id));
+            const numGroups = media.ssrcGroups?.length ?? 0;
+            let primarySsrc: number;
+
+            // Do not munge if the description has no ssrcs or if simulcast is already enabled.
+            if (numSsrcs.size === 0 || numSsrcs.size > 2 || (numSsrcs.size === 2 && numGroups === 0)) {
+                continue;
+            }
+            if (numSsrcs.size === 1) {
+                primarySsrc = parseInt((media.ssrcs[0]?.id).toString());
```
paweldomas, line 211: again doing toString to parse it back, what will happen when `media.ssrcs[0]?.id` is `undefined`? Will it convert to string and parse as `NaN`?

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+            } else {
+                const fidGroup = media.ssrcGroups.filter(group => group.semantics === 'FID')?.[0];
+
+                primarySsrc = parseInt(fidGroup.ssrcs.split(' ')[0]);
```
paweldomas, line 215: `fidGroup` can be `undefined` (from the `?.[0]` code)

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+            let primarySsrc: number;
+
+            // Do not munge if the description has no ssrcs or if simulcast is already enabled.
+            if (numSsrcs.size === 0 || numSsrcs.size > 2 || (numSsrcs.size === 2 && numGroups === 0)) {
+                continue;
+            }
+            if (numSsrcs.size === 1) {
+                primarySsrc = parseInt((media.ssrcs[0]?.id).toString());
+            } else {
+                const fidGroup = media.ssrcGroups.filter(group => group.semantics === 'FID')?.[0];
```
paweldomas, line 213: because looking for the first one `find` should do the job instead of `?.[0]` ?

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    /**
+     * Munges the given media description by removing the SSRCs and related FID groups for the higher layer streams.
+     *
+     * @param description
+     * @returns
+     */
+    mungeRemoteDescription(description: Description) : Description {
+        if (!(description && description?.sdp !== '')) {
```
paweldomas, line 242: nit: the negated condition is a little bit hard to read. How about `if (!description || !description.sdp)`  ? This will account for `sdp` being `undefined` or empty.

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+        });
+    }
+
+    /**
+     * Munges the given media description by removing the SSRCs and related FID groups for the higher layer streams.
+     *
+     * @param description
+     * @returns
+     */
+    mungeRemoteDescription(description: Description) : Description {
```
paweldomas, line 241: I think Jicofo strips simulcast from remote SDP now? https://github.com/jitsi/jicofo/commit/adbc2b5ef520ddd71759336eec15259e84697d32

paweldomas: (I'm not reviewing this method since it's ok to remove unless missing something?)

paweldomas (commented):
Looking good, left few comments

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+     * @param mLine
+     * @param ssrc
+     * @param attributeName
+     * @returns
+     */
+    _getSsrcAttribute(mLine: transform.MediaDescription, ssrc: number, attributeName: string) : string {
+        return mLine.ssrcs
+            .filter(ssrcInfo => parseInt(ssrcInfo.id?.toString()) === ssrc)
+            .filter(ssrcInfo => ssrcInfo.attribute === attributeName)
+            .map(ssrcInfo => ssrcInfo.value)[0];
```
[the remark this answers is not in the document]
jallamsetty1: I copied the code as is from the sdp-simulcast npm module :) will change it.

modules/sdp/SdpSimulcast.ts
jallamsetty1: same reason as above, code copied from npm module, will change it.

modules/sdp/SdpSimulcast.ts
jallamsetty1: This is behind a flag and not the default behavior still and some of our deployments don't have the flag enabled. Will remove it after its made default, until then we need it.

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+     * Returns the requested attribute value for a SSRC from a given media description.
+     *
+     * @param mLine
+     * @param ssrc
+     * @param attributeName
+     * @returns
+     */
+    _getSsrcAttribute(mLine: transform.MediaDescription, ssrc: number, attributeName: string) : string {
+        return mLine.ssrcs
+            .filter(ssrcInfo => parseInt(ssrcInfo.id?.toString()) === ssrc)
```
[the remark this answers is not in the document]
jallamsetty1: ssrcInfo.id is of type `string | number ` and therefore coercion is necessary, otherwise TS complains.

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,307 @@
...
+     * @returns
+     */
+    _parseSimLayers(mLine: transform.MediaDescription) : Array<number> {
+        const simGroup = mLine.ssrcGroups?.find(group => group.semantics === 'SIM');
+
+        if (simGroup) {
+            return simGroup.ssrcs.split(' ').map(ssrc => parseInt(ssrc));
+        }
+
+        return [ parseInt(mLine.ssrcs?.[0]?.id?.toString()) ];
```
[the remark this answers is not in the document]
jallamsetty1: same reason as above

modules/RTC/TraceablePeerConnection.js
```diff
@@ -260,14 +261,20 @@ export default function TraceablePeerConnection(
...
+    if (this._usesUnifiedPlan) {
+        this.simulcast = new SdpSimulcast({ numOfLayers: SIM_LAYER_RIDS.length });
+    } else {
+        const Simulcast = require('@jitsi/sdp-simulcast');
+
+        this.simulcast = new Simulcast(
+            {
+                numOfLayers: SIM_LAYER_RIDS.length,
+                explodeRemoteSimulcast: false,
+                usesUnifiedPlan: this._usesUnifiedPlan
```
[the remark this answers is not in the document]
jallamsetty1: Yes it is but didn't want to make changes to the sdp-simulcast for this so I left it as is, I can pass false instead.

modules/sdp/SdpSimulcast.ts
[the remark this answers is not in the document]
paweldomas: can ssrc's id technically be anything else than a number? maybe we should adjust the sdp-transform library?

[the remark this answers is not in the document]
jallamsetty1: I am reading the types from [this](https://github.com/clux/sdp-transform) library. Its ok to go with the type coercion here instead of spending to submit a PR there.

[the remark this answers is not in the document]
paweldomas: What do you mean by type coercion in this case exactly? I'm only worried that parsing is not really a type coercion, but it has performance implications. I'd rather put a ts ignore here or do a cast (you can write `as number` I think).

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,313 @@
...
+                const fidGroup = fidGroups.find(group => group.ssrcs.includes(primarySsrc));
+                const secondarySsrc = fidGroup.ssrcs.split(' ')[1];
+
+                for (const ssrcGroup of media.ssrcGroups) {
+                    if (ssrcGroup !== fidGroup) {
+                        mungedSsrcGroups.delete(ssrcGroup);
+                    }
+                }
+                for (const ssrc of media.ssrcs) {
+                    if (ssrc.id.toString() !== primarySsrc
```
paweldomas, line 287: seems like the number is preferred overall, so maybe `primarySsrc` can be converter via `Number()` instead of `toString()` on the `id`?

paweldomas: the same applies to `secondarySsrc`

paweldomas (dismissed):
Just 2 minor comments and also can you define a type for the `options`? There was a comment about it. There's no reason for it to be `any`.

modules/sdp/SdpSimulcast.ts
jallamsetty1: The inferred type for `primarySsrc` and `secondarySsrc` is `string` since they are derived from the media.ssrcGroups and hence I am doing it the other way.

jallamsetty1:
> Just 2 minor comments and also can you define a type for the `options`? There was a comment about it. There's no reason for it to be `any`.

I missed the comment about options, will define a type for it.

modules/sdp/SdpSimulcast.ts
```diff
@@ -0,0 +1,317 @@
...
+
+/**
+ * This class handles SDP munging for enabling simulcast for local video streams in Unified plan. A set of random SSRCs
+ * are generated for the higher layer streams and they are cached for a given mid. The cached SSRCs are then reused on
+ * the subsequent iterations while munging the local description. This class also handles imploding of the simulcast
+ * SSRCs for remote endpoints into the primary FID group in remote description since Jicofo signals all SSRCs relevant
+ * to a given endpoint.
+ */
+export default class SdpSimulcast {
+    private _options: any;
```
paweldomas, line 25: this should also have the options type or maybe this declaration is not needed (not sure)

paweldomas (approved):

modules/sdp/SdpSimulcast.ts
jallamsetty1: It is fine but let me push another commit for this.

paweldomas (approved):

modules/sdp/SdpSimulcast.ts
paweldomas: What I've seen is that too many `any` types are making local development slow as this seems to be impacting TypeScript performance, so it's better to avoid that (if you have TS checks enabled in IDE).

modules/sdp/SdpSimulcast.ts
paweldomas: Right, but if they were wrapped in `Number` then it would be more consistent to work on the numbers only. Anyway not a big deal.

PR #1870, opened 2022-02-03
feat(multi-stream-support) Add the support for multiple local video streams.

jallamsetty1:
This feature is behind 'sendMultipleVideoStreams' config.js flag and is currently supported only on clients running in Unified plan mode.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1095,6 +1095,23 @@ JitsiConference.prototype.addTrack = function(track) {
             return Promise.resolve(track);
         }
 
+        if (FeatureFlags.isMultiStreamSupportEnabled() && mediaType === MediaType.VIDEO) {
+            const addTrackPromises = [];
+
+            this.p2pJingleSession && addTrackPromises.push(this.p2pJingleSession.addTrack(track));
+            this.jvbJingleSession && addTrackPromises.push(this.jvbJingleSession.addTrack(track));
+
+            return Promise.all(addTrackPromises)
+                .then(() => {
+                    this._setupNewTrack(track);
+
+                    // TODO Update presence and sent videoType message.
+                    if (this.isMutedByFocus || this.isVideoMutedByFocus) {
+                        this._fireMuteChangeEvent(track);
+                    }
+                });
+        }
+
         return Promise.reject(new Error(`Cannot add second ${mediaType} track to the conference`));
     }
 
@@ -1327,6 +1344,8 @@ JitsiConference.prototype._removeLocalSourceOnReject = function(jingleSession, e
  * @param {JitsiLocalTrack} newTrack the new track being created
  */
 JitsiConference.prototype._setupNewTrack = function(newTrack) {
+    const mediaType = newTrack.getType();
+
     if (newTrack.isAudioTrack() || (newTrack.isVideoTrack() && newTrack.videoType !== VideoType.DESKTOP)) {
         // Report active device to statistics
         const devices = RTC.getCurrentlyAvailableMediaDevices();
@@ -1336,6 +1355,16 @@ JitsiConference.prototype._setupNewTrack = function(newTrack) {
         if (device) {
             Statistics.sendActiveDeviceListEvent(RTC.getEventDataForActiveDevice(device));
         }
+    }
+
+    // Create a source name for this track if it doesn't exist.
+    if (FeatureFlags.isSourceNameSignalingEnabled() && !newTrack.getSourceName()) {
+        const sourceName = getSourceNameForJitsiTrack(
+            this.myUserId(),
+            mediaType,
+            this.getLocalTracks(mediaType)?.length);
+
+        newTrack.setSourceName(sourceName);
     }
 
     this.rtc.addLocalTrack(newTrack);
@@ -1363,13 +1392,7 @@ JitsiConference.prototype._setNewVideoType = function(track) {
     if (FeatureFlags.isSourceNameSignalingEnabled() && track) {
         // FIXME once legacy signaling using 'sendCommand' is removed, signalingLayer.setTrackVideoType must be adjusted
         // to send the presence (not just modify it).
-        this._signalingLayer.setTrackVideoType(
-            getSourceNameForJitsiTrack(
-                this.myUserId(),
-                track.getType(),
-                0
-            ),
-            track.videoType);
+        this._signalingLayer.setTrackVideoType(track.getSourceName(), track.videoType);
 
         // TODO: Optimize to detect whether presence was changed, for now always report changed to send presence
         return true;
@@ -1403,10 +1426,7 @@ JitsiConference.prototype._setTrackMuteStatus = function(mediaType, localTrack, 
     if (FeatureFlags.isSourceNameSignalingEnabled()) {
         // TODO When legacy signaling part is removed, remember to adjust signalingLayer.setTrackMuteStatus, so that
         // it triggers sending the presence (it only updates it for now, because the legacy code below sends).
-        this._signalingLayer.setTrackMuteStatus(
-            getSourceNameForJitsiTrack(this.myUserId(), mediaType, 0),
-            isMuted
-        );
+        this._signalingLayer.setTrackMuteStatus(localTrack?.getSourceName(), isMuted);
     }
 
     if (!this.room) {
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -163,6 +163,9 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // correspond to the id of a matching device from the available device list.
         this._realDeviceId = this.deviceId === '' ? undefined : this.deviceId;
 
+        // The source name that will be signaled for this track.
+        this._sourceName = null;
+
         this._trackMutedTS = 0;
 
         this._onDeviceListWillChange = devices => {
@@ -669,6 +672,15 @@ export default class JitsiLocalTrack extends JitsiTrack {
     }
 
     /**
+     * Returns the source name associated with the jitsi track.
+     *
+     * @returns {string | null} source name
+     */
+    getSourceName() {
+        return this._sourceName;
+    }
+
+    /**
      * Returns if associated MediaStreamTrack is in the 'ended' state
      *
      * @returns {boolean}
@@ -859,6 +871,15 @@ export default class JitsiLocalTrack extends JitsiTrack {
                 logger.error('Failed to switch to the new stream!', error);
                 throw error;
             });
+    }
+
+    /**
+     * Sets the source name to be used for signaling the jitsi track.
+     *
+     * @param {string} name The source name.
+     */
+    setSourceName(name) {
+        this._sourceName = name;
     }
 
     /**
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -351,20 +351,33 @@ export class TPCUtils {
 
         // If old track exists, replace the track on the corresponding sender.
         if (oldTrack) {
-            transceiver = this.pc.peerconnection.getTransceivers()?.find(t => t.sender?.track === oldTrack.getTrack());
+            transceiver = this.pc.peerconnection.getTransceivers().find(t => t.sender.track === oldTrack.getTrack());
 
-        // Find the first recvonly transceiver when a second track of the same media type is being added to the pc. As
-        // part of the track addition, a new m-line was added to the remote description with direction set to recvonly.
-        } else if (this.pc.getLocalTracks(mediaType)?.length && !newTrack?.conference) {
-            transceiver = this.pc.peerconnection.getTransceivers()?.find(
-                t => t.receiver?.track?.kind === mediaType
+        // Find the first recvonly transceiver when more than one track of the same media type is being added to the pc.
+        // As part of the track addition, a new m-line was added to the remote description with direction set to
+        // recvonly.
+        } else if (this.pc.getLocalTracks(mediaType)?.length && !newTrack.conference) {
+            transceiver = this.pc.peerconnection.getTransceivers().find(
+                t => t.receiver.track.kind === mediaType
                 && t.direction === MediaDirection.RECVONLY
                 && t.currentDirection === MediaDirection.INACTIVE);
 
-        // If a track of a given media type is added for the first time, replace the track on the first sender. This
-        // will be called when the user joins the call muted but unmutes during the call.
+        // For unmute operations, find the transceiver based on the track index in the source name if present, otherwise
+        // it is assumed to be the first local track that was added to the peerconnection.
         } else {
-            transceiver = this.pc.peerconnection.getTransceivers()?.find(t => t.receiver?.track?.kind === mediaType);
+            transceiver = this.pc.peerconnection.getTransceivers().find(t => t.receiver.track.kind === mediaType);
+
+            const sourceName = newTrack.getSourceName();
+
+            if (sourceName) {
+                const trackIndex = Number(sourceName.split('-')[1].substring(1));
+
+                if (trackIndex) {
+                    transceiver = this.pc.peerconnection.getTransceivers()
+                        .filter(t => t.receiver.track.kind === mediaType
+                            && t.direction !== MediaDirection.RECVONLY)[trackIndex];
+                }
+            }
         }
 
         if (!transceiver) {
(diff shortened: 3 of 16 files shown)


JitsiConference.js
```diff
@@ -1403,10 +1426,7 @@ JitsiConference.prototype._setTrackMuteStatus = function(mediaType, localTrack,
     if (FeatureFlags.isSourceNameSignalingEnabled()) {
         // TODO When legacy signaling part is removed, remember to adjust signalingLayer.setTrackMuteStatus, so that
         // it triggers sending the presence (it only updates it for now, because the legacy code below sends).
-        this._signalingLayer.setTrackMuteStatus(
-            getSourceNameForJitsiTrack(this.myUserId(), mediaType, 0),
-            isMuted
-        );
+        this._signalingLayer.setTrackMuteStatus(localTrack?.getSourceName(), isMuted);
```
paweldomas, line 1429: what's the point of calling this method if `localTrack` is `undefined`?

modules/RTC/TPCUtils.js
```diff
@@ -339,21 +339,41 @@ export class TPCUtils {
...
      */
     replaceTrack(oldTrack, newTrack) {
         const mediaType = newTrack?.getType() ?? oldTrack?.getType();
-        const transceiver = this.findTransceiver(mediaType, oldTrack);
         const track = newTrack?.getTrack() ?? null;
+        let transceiver;
+
+        // If old track exists, replace the track on the corresponding sender.
+        if (oldTrack) {
+            transceiver = this.pc.peerconnection.getTransceivers()?.find(t => t.sender?.track === oldTrack.getTrack());
```
paweldomas, line 354: how about putting source names as `mid` and using mlines as slots for sending local tracks that can be re-used? wouldn't that make finding the transceiver easier?

paweldomas: when I'm looking at how many `?.` there are, it seems there's so many things that can go wrong there

modules/sdp/LocalSdpMunger.js
```diff
@@ -205,15 +206,29 @@ export default class LocalSdpMunger {
...
 
-                    if (streamAndTrackIDs.length === 2) {
-                        ssrcLine.value
-                            = this._generateMsidAttribute(
-                                mediaSection.mLine?.type,
-                                streamAndTrackIDs[1],
-                                streamAndTrackIDs[0]);
-                    } else {
-                        logger.warn(`Unable to munge local MSID - weird format detected: ${ssrcLine.value}`);
+                    let streamId = streamAndTrackIDs[0];
```
paweldomas, line 209: `streamId` seems to be used only in the feature flagged conditional, maybe it can be moved inside the `if` body?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -163,6 +163,9 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // correspond to the id of a matching device from the available device list.
         this._realDeviceId = this.deviceId === '' ? undefined : this.deviceId;
 
+        // The source name that will be signaled for this track.
+        this._sourceName = null;
```
paweldomas, line 167: I think it may have to be the conference or a "sender" nested inside the conference that remembers which source name is used for which local tracks, because `JitsiLocalTrack` is technically allowed to be in multiple "conferences" or at least peer connection (it may be the case for Spot, but will have to check that). Anyway lets try to start this way.

modules/RTC/TPCUtils.js
```diff
@@ -339,21 +339,41 @@ export class TPCUtils {
...
+        // Find the first recvonly transceiver when a second track of the same media type is being added to the pc. As
+        // part of the track addition, a new m-line was added to the remote description with direction set to recvonly.
+        } else if (this.pc.getLocalTracks(mediaType)?.length && !newTrack?.conference) {
+            transceiver = this.pc.peerconnection.getTransceivers()?.find(
+                t => t.receiver?.track?.kind === mediaType
+                && t.direction === MediaDirection.RECVONLY
+                && t.currentDirection === MediaDirection.INACTIVE);
+
+        // If a track of a given media type is added for the first time, replace the track on the first sender. This
+        // will be called when the user joins the call muted but unmutes during the call.
```
paweldomas, line 365: how do you know it's okay to just pick the first one in the multi track scenario? were we not supposed to generate dedicated mlines for each local track and re-use them as tracks are added/remove/replaced? I thought that the recvonly mlines will not be used for sending local tracks, but dedicate 'sendonly' mlines will be created?

modules/sdp/SDP.js
```diff
@@ -48,6 +51,39 @@ SDP.prototype.removeTcpCandidates = false;
  */
 SDP.prototype.removeUdpCandidates = false;
 
+/**
+ * Adds a new m-line for receiving source of the given media type.
```
paweldomas, line 55: Isn't it for sending local tracks? In which case this should say "for sending source" of given media type?

modules/sdp/SDP.js
```diff
@@ -48,6 +51,39 @@ SDP.prototype.removeTcpCandidates = false;
...
+ *
+ * @param {Mediatype} mediaType media type of the new source that is being added.
+ */
+SDP.prototype.addMlineForNewLocalSource = function(mediaType) {
+    const mid = this.media.length;
```
paweldomas, line 60: As mentioned in another comment, I think if we put source names as `mid` for the sender mlines that would make it easier to find transceivers and recycle the mlines.

modules/xmpp/JingleSessionPC.js
```diff
@@ -2030,6 +2031,49 @@ export default class JingleSessionPC extends JingleSession {
...
+    /**
+     * Adds a new track to the peerconnection. This method needs to be called only when a secondary JitsiLocalTrack is
+     * being added to the peerconnection for the first time.
+     *
+     * @param {JitsiLocalTrack} localTrack track to be added to the peer connection.
+     * @returns {Promise<void>} that resolves when the track is successfully added to the peerconnection, rejected
+     * otherwise.
+     */
+    addTrack(localTrack) {
+        if (!this.usesUnifiedPlan || localTrack.type !== MediaType.VIDEO) {
```
paweldomas, line 2043: should this also throw when the feature flag is not enabled (if called by mistake by other code)?

modules/xmpp/JingleSessionPC.js
```diff
@@ -2030,6 +2031,49 @@ export default class JingleSessionPC extends JingleSession {
...
+            return Promise.reject(new Error('Multiple tracks of a given media type are not supported'));
+        }
+
+        const workFunction = finishedCallback => {
+            const remoteSdp = new SDP(this.peerconnection.peerconnection.remoteDescription.sdp);
+
+            // Add a new transceiver by adding a new mline in the remote description.
```
paweldomas, line 2050: Interesting, so the mlines are added by injecting into remote sdp. Will `addTransceiver` not work here?

types/auto/modules/sdp/SDP.d.ts
```diff
@@ -29,6 +29,12 @@ export default class SDP {
      * @type {boolean}
      */
     removeUdpCandidates: boolean;
+    /**
+     * Adds a new m-line for receiving source of the given media type.
+     *
+     * @param {Mediatype} mediaType media type of the new source that is being added.
+     */
+    addMlineForNewLocalSource(mediaType: any): void;
```
paweldomas, line 37: is there no type definition for `mediaType`?

types/auto/modules/xmpp/JingleSessionPC.d.ts
```diff
@@ -454,6 +454,15 @@ export default class JingleSessionPC extends JingleSession {
...
     private _initiatorRenegotiate;
+    /**
+     * Adds a new track to the peerconnection. This method needs to be called only when a secondary JitsiLocalTrack is
+     * being added to the peerconnection for the first time.
+     *
+     * @param {JitsiLocalTrack} localTrack track to be added to the peer connection.
+     * @returns {Promise<void>} that resolves when the track is successfully added to the peerconnection, rejected
+     * otherwise.
+     */
+    addTrack(localTrack: any): Promise<void>;
```
paweldomas, line 465: why `any`? is there no type definition for `JitsiLocalTrack`?

paweldomas (commented):
Left few questions/comments. I'm trying to understand what's the final approach going to look like. Now it seems slightly different to what we were talking about.

modules/RTC/TPCUtils.js
jallamsetty1: No we don't want to hardcode mids to the source names since we won't be able recycle them for other sources. The mid order cannot be changed at all, we can only append extra mids to the current ones. The idea is so to use the first transceiver (created by the jvb ssrcs in the remote offer) for the first local source and then create a new transceiver when a new local source is added.

modules/RTC/TPCUtils.js
jallamsetty1: When we add a transceiver by adding a new m-line in the remote description without a ssrc attached, the browser will create a transceiver with the direction as RECVONLY since the local track hasn't been added yet and the current direction will be INACTIVE since there is no media flowing yet. Current direction is the differentiator here. After the track is added and reneg is done, the transceiver direction will change to sendonly.

modules/sdp/SDP.js
jallamsetty1: Yeah but technically when you add an m-line in the remote description, you are technically adding a receiver for receiving a source, be it local or remote. I can change the wording if its confusing.

modules/sdp/SDP.js
jallamsetty1: No I don't think this is a good idea. The same mids show up in both the local and remote description and these mids do not make sense when you are trying to generate new m-lines for new remote sources. They wouldn't follow a pattern and recycling them will not be possible.

modules/xmpp/JingleSessionPC.js
jallamsetty1: It should only be called from JitsiConference.addTrack and calling it from elsewhere will result in unexpected behavior. I don't mind adding the check here.

modules/xmpp/JingleSessionPC.js
jallamsetty1: Even with addTransceiver, the remote description would still have to be munged before setting the remote description after the transceiver is added. I played around with addTransceiver a bit but it resulted in the browser renegotiation loop triggered by 'negotiationneeded' event fired by the browser.

types/auto/modules/sdp/SDP.d.ts
jallamsetty1: These are auto generated types based on the SDP JS file. This is the inferred type since type is not specified in the JS file.

types/auto/modules/xmpp/JingleSessionPC.d.ts
jallamsetty1: same reason as above.

modules/RTC/TPCUtils.js
paweldomas: How is it a problem to create dedicated mline for each local track? It actually makes it easier to recycle the lines and to do track replacement (because you act on specific mline tied to the track). The dedicated mlines can be recycled and the only waste is that they would be inactive during the time there's no local track sent on that slot. To me it seems more brittle approach and error prone trying to rely on the ordering, but if that's easier then it's fine I guess.

JitsiConference.js
jallamsetty1: I will be revisiting all the presence related stuff in a another PR. The only time this will be undefined is when there are no local video tracks and that point I don't know what source name will be sent in the presence.

modules/RTC/JitsiLocalTrack.js
jallamsetty1: I don't understand why sources with the same source names cannot be shared across multiple peerconnections ?

modules/sdp/SDP.js
paweldomas: They would be dedicated to local tracks and can be recycled for that purpose, but again if you think it's easier this way then that's ok. Still I think it will be more error prone. The `if` conditions that rely on ordering, finding media type and inactive or recvonly is hard to understand and follow without tons of comments.

modules/xmpp/JingleSessionPC.js
paweldomas: yes, but it will make it easier to detect when the unexpected call occurs, thanks!

types/auto/modules/xmpp/JingleSessionPC.d.ts
paweldomas: then why not declare the type in the method signature? is it a rabbit hole? I think it's good to try adding more strict typing step by step (as long as it doesn't require lots of extra changes which would grow the PR).

jallamsetty1: Type annotations are allowed in the TS files and the original file where the methods is defined is a JS file. Am I missing something ?

JitsiConference.js
paweldomas: Ok got it. If there's no local track then the track should be removed from the presence.

modules/RTC/JitsiLocalTrack.js
paweldomas: That's not what I meant. The thing is the `JitsiLocalTrack` when created does not belong to any conference and I thought that one can participate in multiple `JitsiConference` instances, but after checking that is not the case, because there's `conference` field which allows only one. I thought that maybe Spot was doing that for proxying between the peerconnections, but I think it uses raw MediaStreamTrack to do that.

types/auto/modules/xmpp/JingleSessionPC.d.ts
paweldomas: ah okay I wasn't paying attention to that

modules/sdp/SDP.js
jallamsetty1: The idea is to use the first available sender when you are adding/replacing it on the peerconnection. I still don't get what you mean by dedicated m-lines. Having one fixed per JitsiLocalTrack means there will be many unused ones since every device change for a given source is a new JitsiLocalTrack and will be a new m-line.

paweldomas (approved):

zsinba:
There is no documentation on how to use multistreaming support.
Does that mean I can use the built-in camera and another external camera on one computer, and the remote user can see both of my videos?
How to invoke it now

zsinba:
Do you have an estimated release date? @paweldomas

zsinba:
Has this been released?

paweldomas:
@zsinba the estimated release date is around end of March

zsinba:
> @zsinba the estimated release date is around end of March
I mentioned this feature at the end of 2019.
I'm really looking forward to this。

Thanks。

zsinba:
HowAbouIt

zsinba:
> > @zsinba the estimated release date is around end of March
> > I mentioned this feature at the end of 2019.
> > I'm really looking forward to this。
> 
> Thanks。

Hello，  Can We  Have a try now?

zsinba:
Hello  all. how about it now?

zsinba:
I tried it out and it was great.
It combines desktop streams and local camera streams into one screenshot to push out.
More than that, if I have a computer with two screens (Windows display extension, one master screen, one secondary screen), is there any way I can share both screens with my remote friend ? I want to give two screen to remote.

paweldomas:
> I tried it out and it was great. It combines desktop streams and local camera streams into one screenshot to push out. 

Uhm, I'm confused about this part. It is actually not supposed to combine local camera and desktop into one stream anymore.

> More than that, if I have a computer with two screens (Windows display extension, one master screen, one secondary screen), is there any way I can share both screens with my remote friend ? I want to give two screen to remote.

You could eventually do that, but you'd have to modify jitsi-meet yourself.

zsinba:
I may not have made myself clear.


I downloaded and ran the latest version of Jitsi of docker; I could share My desktop and open my camera at the same time. They will be pushed to the remote end in the same stream. as follows:

12dae86efef8fc24146ffe1b38a2ae6

My question is:
My Windows 11  (OS)  has two screens, is it possible for me to share both screens and use both streams

image


Thank you very much for your constant reply. Thank you for your answer too.

zsinba:
> ou could eventually do that, but you'd have to modify jitsi-meet yourself.

I consulted about this about two years ago. The response: No. You can't push two streams out at the same time.

paweldomas:
> I consulted about this about two years ago. The response: No. You can't push two streams out at the same time.

Right, but this changes with the multi stream mode supported on the lib-jitsi-meet/JVB level. You should be able to do that, but you need to modify jitsi-meet.

About your screenshots, that is not the new presenter mode, that's how it works currently and it's a single stream. Maybe it's not available in the docker release yet.

PR #1874, opened 2022-02-06
feat(ScreenObtainer): add support for tab capture on Electron

DanielMcAssey:
With https://github.com/electron/electron/pull/31204 being merged into Electron and released with Electron 17.x.x, Electron can now capture individual WebContents (Tabs). This change allows the standard flow to support tab capture

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -164,7 +164,7 @@ const ScreenObtainer = {
                             audio: audioConstraints,
                             video: {
                                 mandatory: {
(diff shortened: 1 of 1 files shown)


saghul:
But how to "tabs" manifest on Electron? Like Chrome tabs?

DanielMcAssey:
Essentially yeah, so you can have a BrowserWindow with different WebContents, in addition capturing the WebContents also removes the Window Frame

saghul:
I see. We need to make sure this does not break existing apps before landing it though.

PR #1875, opened 2022-02-08
JitsiTrackErrors to typescript enum

garyhuntddn:
@saghul I'd like to spend some time gradually working through the files and converting them to TypeScript where they are zero risk.

If you're happy for me to work through this then a quick conversation might be good to start off with.

My thoughts are:

1. Go through and convert the easy files first - these will allow us to then build upon them in the more complex files
2. This will start with the files that contains lists of exported strings (the constants) into enums so that we can then correctly type methods and properties that use them
3. I'd then move onto the files that are already ES6 classes - as these should be easy conversions
4. Finally I'd tackle some of those non ES6 files - but those are going to be risks and I'd likely start by adding a bunch more tests

I could either do these one at a time - thereby reducing the risk of clashing with any other code changes that are happening - if we hit a merge conflict due to edits of the original .js file then we'd abandon my change completely. Or I could do a block of changes at a time, but I feel this would be harder for you to then merge. Please let me know your thoughts.

Oh - and if you'd prefer I didn't do this at all then I'll take no offence :-) I personally find code that is strongly typed much easier to maintain but you're doing lots more on this than I'll ever be able to.

This example change to JitsiTrackErrors shows the sort of thing I think we could do - e.g. use of a typescript enum whilst maintaining backward compatibility by performing the existing exports. Eventually we'd likely get rid of those individual exports and just use the enum.


saghul:

saghul:
> @saghul I'd like to spend some time gradually working through the files and converting them to TypeScript where they are zero risk.
> 
> 
> 
> If you're happy for me to work through this then a quick conversation might be good to start off with.
> 
> 
> 
> My thoughts are:
> 
> 
> 
> 1. Go through and convert the easy files first - these will allow us to then build upon them in the more complex files
> 
> 2. This will start with the files that contains lists of exported strings (the constants) into enums so that we can then correctly type methods and properties that use them
> 
> 3. I'd then move onto the files that are already ES6 classes - as these should be easy conversions
> 
> 4. Finally I'd tackle some of those non ES6 files - but those are going to be risks and I'd likely start by adding a bunch more tests
> 
> 
> 
> I could either do these one at a time - thereby reducing the risk of clashing with any other code changes that are happening - if we hit a merge conflict due to edits of the original .js file then we'd abandon my change completely. Or I could do a block of changes at a time, but I feel this would be harder for you to then merge. Please let me know your thoughts.
> 
> 
> 
> Oh - and if you'd prefer I didn't do this at all then I'll take no offence :-) I personally find code that is strongly typed much easier to maintain but you're doing lots more on this than I'll ever be able to.
> 
> 
> 
> This example change to JitsiTrackErrors shows the sort of thing I think we could do - e.g. use of a typescript enum whilst maintaining backward compatibility by performing the existing exports. Eventually we'd likely get rid of those individual exports and just use the enum.

This sounds like a fantastic plan! Thanks a lot for getting started!

Generally speaking, the smaller the PRs the easier they are to review and merge, while also avoiding conflicts.

So I'd say let's start with your plan, one file at a time, creating one PR for each.

saghul (approved):

PR #1878, opened 2022-02-09
fix: Cleans up rooms after failure to join.

damencho:


JitsiConferenceEventManager.js
```diff
@@ -231,6 +231,10 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
     this.chatRoomForwarder.forward(XMPPEvents.RESERVATION_ERROR,
         JitsiConferenceEvents.CONFERENCE_FAILED,
         JitsiConferenceErrors.RESERVATION_ERROR);
+    chatRoom.addListener(XMPPEvents.RESERVATION_ERROR,
+        () => {
+            conference.connection.emuc.doLeave(conference.room.roomjid);
```
shawnchin, line 236: `conference.connection.emuc` was undefined when I tested this locally. 

Seems to work if I replace this with a rather clumsy looking `conference.connection.xmpp.connection.emuc.doLeave(conference.room.roomjid)`

saghul:
Damencho can you elaborate on what this fixes pl?

shawnchin:
@saghul  Apologies if I'm being nosy and speaking out of place, but I believe this PR is in response to the issue I raised here -- https://community.jitsi.org/t/ux-when-reservation-fails/110095?u=shawn

In short, room data cached on emuc was not being cleared up on room creation failure e.g on reservation or max user error. As a result, user can no longer retry the join and has to reload to recover.

damencho:
I have edited the commit message with an explanation. And dropped the ts changes and open another PR for that.

jallamsetty1 (approved):

PR #1891, opened 2022-02-20
Tests to monitor regressions in the TypeScript conversion

garyhuntddn:
These are the brittle tests across the files that contain simple constant style exports.

The aim here is to ensure that when the TypeScript versions are merged that we have a baseline and that we can ensure no regressions.

Note that the tests are setup look for additional members and to fail if they appear - this would indicate a change that needs the test updating (hence they are brittle). You may also spot that I've done these with knowledge of the TypeScript versions that I have tucked away.

Please note the inclusion of @types/jasmine to keep TypeScript / VSCode happy and the updates to both tsconfig.json and karma.conf.js to include the .ts tests

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,6 +40,7 @@
     "@babel/preset-env": "7.16.0",
     "@babel/preset-typescript": "7.16.7",
     "@jitsi/eslint-config": "4.0.0",
+    "@types/jasmine": "^3.10.3",
     "@types/sdp-transform": "2.4.5",
     "babel-loader": "8.2.3",
     "core-js": "3.19.1",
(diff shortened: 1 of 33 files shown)


saghul (approved):

package.json
```diff
@@ -40,6 +40,7 @@
     "@babel/preset-env": "7.16.0",
     "@babel/preset-typescript": "7.16.7",
     "@jitsi/eslint-config": "4.0.0",
+    "@types/jasmine": "^3.10.3",
```
saghul, line 43: Nit: can you please remove the caret? We use specific versions all across.

saghul (changes requested):

package.json
garyhuntddn: Change made, and I agree, sensible plan to pin everything

garyhuntddn:
I've found a regression from the previous PR due to be not updating the auto-generated types in the types/auto folder - that's now fixed. If you want me I can cherry-pick that commit into a separate PR - but as I believe this one is ready to go that might be the quickest option

saghul:
This is great!

PR #1893, opened 2022-02-21
Compile the typescript prior to running the karma tests to ensure that the auto-generated declarations are up-to-date

garyhuntddn:
I won't be offended if you consider this to be a bit lazy - but as I'm doing lots of work within karma it makes sense for me to auto-update the typescript definition files when I run the tests - as I'm effectively driving the TypeScript conversion via TDD.

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,7 +63,7 @@
     "lint": "eslint .",
     "lint-fix": "eslint . --fix",
     "prepack": "npm run build",
-    "test": "karma start karma.conf.js",
+    "test": "tsc && karma start karma.conf.js",
     "test-watch": "karma start karma.conf.js --no-single-run",
     "validate": "npm ls",
     "watch": "webpack --config webpack.config.js --watch --mode development"


saghul:
I suppose this is so you don't forget? If that's the case, the CI will tell you, because the working tree won't be clean: https://github.com/jitsi/lib-jitsi-meet/blob/35b3d17a456d1f17956e166293e4567130683122/.github/workflows/ci.yml#L20

garyhuntddn:
Yep - I've created an issue to explain my reasoning - I've now been through the github build logs a few too many times across 3 PRs ;-) Once I'm used to the build process you have it'll become second nature - but I'd hate to think that I'd broken the build!

saghul:
Not a fan of the change, sorry. Note that running `tsc` alone doesn't give you the certainty the build is not broken, running `npm install` will generate both bundles and give you that, auto-generating the types in the process...

PR #1894, opened 2022-02-21
Gh/ts/patch 002 jitsi track events

garyhuntddn:


.gitattributes
```diff
@@ -1,4 +1,3 @@
 lib-jitsi-meet.min.js -text -diff
 lib-jitsi-meet.js -text -diff
 lib-jitsi-meet.js.map -text -diff
-text eol=lf
```
saghul, line 4: Don't we want to keep this?

saghul (approved):
LGTM with a small question.

.gitattributes
garyhuntddn: Let me check the next PR and see if my branch includes that same change too - at one point I'd accidently committed that change to my fork's master - but I thought I'd reset it correctly.... maybe not

garyhuntddn: The next one has worked correctly - I must have screwed up reverting master - so I'll abandon this PR and cherry pick the right commits into a new one

PR #1912, opened 2022-02-21
Gh/ts/patch 031 interfaces

garyhuntddn:


saghul:
What improvement does this give us right now?

garyhuntddn:
These ones need a good explanation as they are a step towards the other branches I want to implement.

Effectively by declaring the interfaces on the major components in the project we can then correctly identify and restrict the type in methods that use them. They are a stepping stone to using the real type but I don't believe that will be possible until we've converted lots more of the code.

e.g. a function that takes the JitsiConference object somewhere in one of the modules

there might be an existing function that does something like
```
/**
*  @param conf {JitsiConference}
*/
function ( conf ) {
   const x = conf.options.x;
}
```

Converting the JitsiConference class first is going to be very hard - so my aim will be to convert as many other modules first. I can do a good job of that by converting the above code to:

```
function( conf: IJitsiConference ) {
   const x = conf.options.x;
}
```

This will only work if I've defined:

```
interface IJitsiConference {
   options: IJitsiConferenceOptions;
}

interface IJitsiConferenceOptions {
   x: string;
}
```

Does that explain what I'm trying to achieve? The nice side effect is that we can then observe the dependencies using tools that track usage (e.g. the F12 key in VS code)

I'm very open to suggestions if you have any about alternative methods.

The eventual goal will then be a simple replace of IJitsiConference with JitsiConference once that type is converted - at present we know that for example JitsiConference isn't an ES6 module and contains a bunch of notes about things that are public/private that shouldn't be etc. It's also a file that's likely to be actively development against in other branches so the likelihood of merge clashes is high and we also have little test coverage so the risk is high.

saghul:
I see, makes sense!

garyhuntddn:
Happy for me to take this one out of draft?

I think the process will work well - but I think I'll first put something akin to the explanation above into the top of the file.

Anything you think I should add to that to make it clear to future committers?

PR #1915, opened 2022-02-21
Gh/ts/patch 009 xmpp events

garyhuntddn:


service/xmpp/XMPPEvents.ts
```diff
@@ -0,0 +1,484 @@
...
+export const SUBJECT_CHANGED = XMPPEvents.SUBJECT_CHANGED;
+export const SUSPEND_DETECTED = XMPPEvents.SUSPEND_DETECTED;
+export const TRANSCRIPTION_STATUS_CHANGED = XMPPEvents.TRANSCRIPTION_STATUS_CHANGED;
+export const TRANSPORT_INFO = XMPPEvents.TRANSPORT_INFO;
+export const VIDEO_SIP_GW_AVAILABILITY_CHANGED = XMPPEvents.VIDEO_SIP_GW_AVAILABILITY_CHANGED;
+export const VIDEO_SIP_GW_SESSION_STATE_CHANGED = XMPPEvents.VIDEO_SIP_GW_SESSION_STATE_CHANGED;
+export const ICE_CONNECTION_STATE_CHANGED = XMPPEvents.ICE_CONNECTION_STATE_CHANGED;
+export const JSON_MESSAGE_RECEIVED = XMPPEvents.JSON_MESSAGE_RECEIVED;
+
+// TODO: this was a pre-ES6 module using module.exports = XMPPEvents which doesn't translate well
```
saghul, line 481: How about we do this in this very PR? I can trigger a full e2e test to make sure we haven't broken anything.

garyhuntddn: We could do.

My aim in this round was to ensure 100% backward compatibility without breaking any consumers of these files - you know a great deal more than I do about whether these are used by others or whether they are effectively "internal" code.

My personal preference would be to do these in two stages - get the TS merged in - and then for me to work through the TODO's I've added with PRs that cause breaking changes. These then get evaluated in their own right.

Flip of a coin stuff isn't it? There are advantages to both approaches.

saghul: Up until very recently the only way this library was consumed was through the webpack bundle. Now there is a tsc build, which mobile uses, but we export a single object with everything, and AFAICT we can keep all of that working just fine and change the internals. That's fair game.

garyhuntddn: Yeah - that's the sort of knowledge that you have that I haven't the slightest about :-)

So is the decision to make the changes in this PR?

saghul: Yeah let's do it!

PS: Very happy with the work you're doing, thank you! ❤️

garyhuntddn: OK - that's done - but boy it now affects a lot of files - my concern with a branch like this is it becomes much less transparent what the change is due to it affecting so many places

saghul: Let's see what the tests have to say :-)

saghul:

saghul (approved):

PR #1921, opened 2022-02-21
Gh/ts/patch 019 service rtc media direction

garyhuntddn:


saghul:
Converted to draft since there are some merge shenanigans.

PR #1922, opened 2022-02-21
Gh/ts/patch 020 service rtc codec mime type

garyhuntddn:


saghul:

JitsiConference.js
```diff
@@ -1612,6 +1612,14 @@ JitsiConference.prototype.getLastN = function() {
     return this.receiveVideoController.getLastN();
 };
 
+/**
+ * Obtains the forwarded sources list in this conference.
+ * @return {Array<string>|null}
+ */
+JitsiConference.prototype.getForwardedSources = function() {
```
saghul, line 1619: Unrelated change?

modules/RTC/BridgeChannel.js
```diff
@@ -359,11 +359,25 @@ export default class BridgeChannel {
                 break;
             }
             case 'LastNEndpointsChangeEvent': {
-                // The new/latest list of last-n endpoint IDs (i.e. endpoints for which the bridge is sending video).
-                const lastNEndpoints = obj.lastNEndpoints;
+                if (!FeatureFlags.isSourceNameSignalingEnabled()) {
```
saghul, line 362: Where does this come from? Bad merge perhaps?

saghul (changes requested):
I think there was some weird merge  going on here...

modules/RTC/BridgeChannel.js
garyhuntddn: Yeah - that'll be a bad merge - I was plagued with them yesterday with these changes that are affecting multiple files - I think going forward we should do them in multiple steps

PR #1926, opened 2022-02-21
fix(JingleSessionPC) fix "echo" on mobile

saghul:
This patch removes aggressive layer suspension on RN. The way it was
implemented it would start sending media to the JVB even while in P2P.

I tried to look for a way around it, but I could not find a "is media
active in this Jingle session PC?" flag, so I chose the shortest path to
the fix and yanked the (mis)feature altogether. This will start to work
once we have transceivers.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -1517,9 +1517,10 @@ export default class JingleSessionPC extends JingleSession {
             // RN doesn't support RTCRtpSenders yet, aggresive layer suspension on RN is implemented
             // by changing the media direction in the SDP. This is applicable to jvb sessions only.
             if (!this.isP2P && browser.isReactNative() && typeof maxFrameHeight !== 'undefined') {
-                const videoActive = maxFrameHeight > 0;
+                const audioActive = this.peerconnection.audioTransferActive;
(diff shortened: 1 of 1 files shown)


saghul:
@jallamsetty1 I opened this as a PR so we can discuss a bit. If you know of a way not to re-enable the media session in this case, I'll be happy to adjust the PR!

jallamsetty1:
```
            if (!this.isP2P && browser.isReactNative() && typeof maxFrameHeight !== 'undefined') {
                const audioActive = this.peerconnection.audioTransferActive;
                const videoActive = this.peerconnection.videoTransferActive && maxFrameHeight > 0;

                return this.setMediaTransferActive(audioActive, videoActive);
            }
```
Can we do this instead ? This should handle the layer suspension as well. WDYT ?

saghul:
I'll give it a try!

saghul:
I'll do some more tests tomorrow, but the first attempt wasn't promising :-(

Because we check before calling the function, it's possible we see it as active while the enqueued function is going to change it to inactive, so we end up queuing an inactive / active succession and end up with duplicated audio.

saghul:
Scratch that, I had tested a stale apk. It does work!

jallamsetty1 (approved):

PR #1928, opened 2022-02-21
introduce declaration files for @jitsi/js-utils and @jitsi/logger

garyhuntddn:
These are only the bits I've needed so far - I'm sure there will be plenty more types to be added

### DIFF
--- a/modules/js-utils.d.ts
+++ b/modules/js-utils.d.ts
@@ -1,0 +1,15 @@
+declare module '@jitsi/js-utils' {
+    class BrowserDetection {
+        getName: () => string;
+        getVersion: () => string;
+        isFirefox: () => boolean;
+        isChrome: () => boolean;
+        isElectron: () => boolean;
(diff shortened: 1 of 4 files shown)


saghul:
How about we add those to the respective projects? We can cut a new release is necessary, in fact they are automated on every PR merge.

garyhuntddn:
That would be better wouldn't it - we can auto-generate like we do with the auto ones in this lib

Are we able to publish NPMs? e.g. @types/jitsi/js-utils

garyhuntddn:
To use the same technique would we need to move to webpack etc? we'll need to introduce tsc and a tsconfig.json

garyhuntddn:
Just a quick update on this one - we're very likely going to add the types directly within the NPM packages themselves - so this PR will most likely be abandoned.

saghul:
Yeah, I'd say we do that.

PR #1935, opened 2022-02-27
clean-up the scripts where typescript conversion is complete

garyhuntddn:
Now that a number of the files are typescript we can update the tests accordingly

PR #1941, opened 2022-02-28
ref(ts) remove todo on already converted imports

NewEraCracker:


saghul:
@garyhuntddn any objections?

garyhuntddn:
I think I've got a very similar PR in the queue - go ahead with whichever comes first - probably this one as you're in front of it :-)

My plan is to do a full review of the .spec.ts files once all the merges are in place.

saghul (approved):

saghul:
Ah, I think it overlaps with https://github.com/jitsi/lib-jitsi-meet/pull/1935, which is more complete. I'll merge that first.

saghul:
Can you rebase, just in case, @NewEraCracker ?

NewEraCracker:
@saghul done.

saghul (approved):

PR #1949, opened 2022-03-04
typescript version of StreamEventTypes (and conversion to named export)

garyhuntddn:


garyhuntddn:
I've converted this one, but anticipate that we might end up removing it as it doesn't appear to be in use.

When I convert the other code I'll look for places that the constants are used as literals instead of via the enum

saghul:
Indeed we are no longer using these, can you remove them instead?

garyhuntddn:
Yep if you know that to be the case then they can go

saghul:
Yep, let's kill them!

garyhuntddn:
That's revised to now be a simple delete of the three files affected.

One thing to be aware of - running the build didn't remove the .d.ts file from the auto types folders - so if we're not careful we may end up with things being left in there that are orphaned.

I'll take a look in a separate PR at a "clean" process that empties that folder prior to the build.

saghul (approved):

PR #1952, opened 2022-03-04
use path.join in webpack-shared.config

garyhuntddn:
In theory it is good practice to use path.join in webpack config files rather than doing string concatenation - it also happens to make this line cross-environment compatible


saghul:

saghul (approved):

PR #1954, opened 2022-03-08
tsc clean

garyhuntddn:
Actually appears to be a very simple fix

### DIFF
--- a/package.json
+++ b/package.json
@@ -60,7 +60,7 @@
     "webpack-cli": "4.9.0"
   },
   "scripts": {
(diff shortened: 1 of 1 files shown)


garyhuntddn:
I've tested this by adding

```export const TEST = '123';```

to JitsiParticpant.js

This resulted in a build producing a /types/auto/JitsiParticpant.d.ts.

When I then removed the contents of JitsiParticpant.js the /types/auto/JitsiParticpant.d.ts was removed.

saghul (approved):

PR #1955, opened 2022-03-09
fix(multi-stream-support) Support muting of desktop track.

jallamsetty1:


modules/RTC/JitsiLocalTrack.js
```diff
@@ -442,11 +444,11 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
                 }
 
-                this.containers.map(
-                    cont => RTCUtils.attachMediaStream(cont, this.stream));
+                this.containers.map(cont => RTCUtils.attachMediaStream(cont, this.stream));
 
                 return this._addStreamToConferenceAsUnmute();
-            });
+            })
+            .catch(error => Promise.reject(error));
```
paweldomas, line 451: hmm why is this call necessary? isn't it equivalent to not catching?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -368,7 +370,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
         const logMuteInfo = () => logger.info(`Mute ${this}: ${muted}`);
 
         if (this.isAudioTrack()
-                || this.videoType === VideoType.DESKTOP
+                || (this.videoType === VideoType.DESKTOP && !FeatureFlags.isMultiStreamSupportEnabled())
```
paweldomas, line 373: can you explain how muting a desktop track will work? does it remove the actual stream from the conference? why is it not enough to just use the "enabled" flag like for audio? is the AV moderation the main use case for being able to mute a desktop track?

modules/xmpp/JingleSessionPC.js
```diff
@@ -2031,7 +2031,14 @@ export default class JingleSessionPC extends JingleSession {
 
             // Add a new transceiver by adding a new mline in the remote description.
             remoteSdp.addMlineForNewLocalSource(MediaType.VIDEO);
-            this._renegotiate(remoteSdp.raw)
+
+            // Always initiate a responder renegotiate since the new m-line is added to remote SDP.
```
paweldomas, line 2033: why does it execute the responder flow? what if it's the p2p offerer?

jallamsetty1: Even if the ep is the p2p offerer, the new m-line is always added to the remote SDP. Therefore the reneg has to start with a sRD call so that the a new m-line is added to the local SDP returned by createAnswer call.

modules/RTC/JitsiLocalTrack.js
jallamsetty1: Oh, this was unintentional, I had added some debugging info here when I was testing error handling in the JM PR and I removed since it wasn't necessary to log it here, forgot to remove this line. I will remove this.

modules/RTC/JitsiLocalTrack.js
paweldomas: Can you add a comment which explains why mute in multi stream works different than in regular mode? Also mention that this behavior is used in jitsi-meet to avoid extra source-remove/source-add?

modules/RTC/TPCUtils.js
```diff
@@ -354,6 +354,11 @@ export class TPCUtils {
     replaceTrack(oldTrack, newTrack) {
         const mediaType = newTrack?.getType() ?? oldTrack?.getType();
         const track = newTrack?.getTrack() ?? null;
+        const isNewSecondaryLocalSource = FeatureFlags.isMultiStreamSupportEnabled()
```
paweldomas, line 357: this variable name suggest that only two streams are supported, but that's not the case is it?

modules/RTC/TPCUtils.js
```diff
@@ -363,20 +368,17 @@ export class TPCUtils {
...
+        } else if (isNewSecondaryLocalSource) {
             transceiver = this.pc.peerconnection.getTransceivers().find(
                 t => t.receiver.track.kind === mediaType
                 && t.direction === MediaDirection.RECVONLY
                 && t.currentDirection === MediaDirection.INACTIVE);
 
-        // For unmute operations, find the transceiver based on the track index in the source name if present, otherwise
-        // it is assumed to be the first local track that was added to the peerconnection.
+        // For mute/unmute operations, find the transceiver based on the track index in the source name if present,
+        // otherwise it is assumed to be the first local track that was added to the peerconnection.
```
paweldomas, line 378: is it possible to always use the index instead of having a special case for "the first local track that was added to the peerconnection"?

modules/sdp/LocalSdpMunger.js
```diff
@@ -364,13 +365,21 @@ export default class LocalSdpMunger {
...
         for (const source of sources) {
             const nameExists = mediaSection.ssrcs.find(ssrc => ssrc.id === source && ssrc.attribute === 'name');
+            const msid = mediaSection.ssrcs.find(ssrc => ssrc.id === source && ssrc.attribute === 'msid')?.value;
+            let sourceName = null;
 
+            if (msid) {
+                const msidInfo = msid.split('-');
+
+                msidInfo.splice(3);
+                sourceName = msidInfo.join('-');
```
paweldomas, line 375: why does source name come from the msid?

modules/xmpp/JingleSessionPC.js
paweldomas: so it's allowed to use the two flows interchangeably? interesting...

jallamsetty1: It doesn't matter whether you do sRD->cA->sLD or cO->sLD->sRD once the call has established. The initiator or responder flows matter only for the initial call establishment where one of the peers has to start by creating an offer.

modules/RTC/TPCUtils.js
jallamsetty1: This is needed for the clients that don't support source name signaling and send only one track per media type.

modules/RTC/TPCUtils.js
jallamsetty1: changed the variable name to `isNewLocalSource`

modules/sdp/LocalSdpMunger.js
jallamsetty1: I am trying to extract the trackIndex to be used for source name generation since we don't have a reference to the track object here. I modified it slightly now.

paweldomas (approved):

PR #1956, opened 2022-03-12
feat(build) remove export from build script

NewEraCracker:
this moves the determination of the git commit to webpack-shared-config

LIB_JITSI_MEET_COMMIT_HASH is still kept for backward compatibility

related to #1951 and #1952

Issue #1951, opened 2022-03-04
Building on windows - issue with webpack string-replace-loader


garyhuntddn:
## Description
I've got a personal branch where I'm attempting to use cross-env to replace the npm run script with one that is cross-environment.
The first bit has been easy - I know cross-env and it's easy to use.
The problem I have though is that the use of "string-replace-loader" in "webpack-shared-config.js" is failing and I can't see why. Anybody know an answer and want to help?

## Current behavior
"npm run build" fails on Window due to the use of the "export" command.
If we eliminate that step and run webpack directly using
```
SET LIB_JITSI_MEET_COMMIT_HASH="0000000"
node ./node_modules/webpack/bin/webpack.js
```
then the build succeeds - but the "{#COMMIT_HASH#}" variable is not substituted.

## Expected Behavior
It would be nice if "npm run build" would work on Windows, but that is outside of the scope of this issue.
If we run webpack directly as above then the "{#COMMIT_HASH#}" substitution should have been made.

## Possible Solution
I'll be proposing an update to use "cross-env" instead of the "export" in the package.json file but there's no point in doing that until the substitution works.

## Steps to reproduce
See above

# Environment details
Windows 10

saghul:
We don't support building in Windows.

I'd be hesitant of even taking patches because it would send the wrong message, we don't build nor test on Windows so we really can't claim to support it.

If the patch is not too involved and we can test it with GH actions maybe we can reconsider.

garyhuntddn:
I don't disagree with the policy but the project is actually really close to working with a tiny change - it's just the "string-replace-loader" that's failing and I can't understand why :-) I'm 99% sure that the webpack part of things will work happily on Windows but something in the config for lib-jitsi-meet but be causing it to not apply

garyhuntddn:
Ha ha - I've cracked it... it's the darn slash ;-)

```
test: `${__dirname}/JitsiMeetJS.js`
```

needs to be

```
test: `${__dirname}\\JitsiMeetJS.js`
```

on Windows (what a pile of pooh it is using the wrong kind of slash!)

saghul:
Nice! Wasn't there a constant in the path module?

garyhuntddn:
From a bit of reading around the subject it's good practice to use path.join and path.resolve in webpack config files - so that's what I've done in https://github.com/jitsi/lib-jitsi-meet/pull/1952.

Will let you decide whether to accept the PR or not.

[...]

PR #1957, opened 2022-03-13
fix(build) patch transpile exclude regex for windows

NewEraCracker:
on windows replaces slashes with backslashes and then escapes them for the regex to be valid

related to #1951 and #1956

### DIFF
--- a/webpack-shared-config.js
+++ b/webpack-shared-config.js
@@ -29,12 +29,11 @@ module.exports = (minimize, analyzeBundle) => {
             }, {
                 // Transpile ES2015 (aka ES6) to ES5.
 
-                exclude: [
+                include: [
                     new RegExp(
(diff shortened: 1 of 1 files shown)


saghul:
Wouldn't using path.join in the Regex building avoid the problem?

NewEraCracker:
I don't envision a good fix using path.join here as we need to care for the negated @jitsi/js-utils condition and still need to escape backslashes for the regex to be valid.

saghul:
@horymury can you PTAL?

saghul:
Actually, on a closer inspection I think we no longer need that: https://github.com/jitsi/lib-jitsi-meet/pull/1958

garyhuntddn:
I'm pretty sure it won't be needed once my conversion of js-utils and jitsi-logger make some more progress

PR #1958, opened 2022-03-14
chore(build) remove unneeded code

saghul:
There is no need to treat js-utils differently, since it's now declared
as a module in package.json

Ref: https://github.com/jitsi/lib-jitsi-meet/pull/1957

### DIFF
--- a/webpack-shared-config.js
+++ b/webpack-shared-config.js
@@ -29,9 +29,6 @@ module.exports = (minimize, analyzeBundle) => {
             }, {
                 // Transpile ES2015 (aka ES6) to ES5.
 
-                exclude: [
-                    new RegExp(`${__dirname}/node_modules/(?!@jitsi/js-utils)`)
-                ],
                 loader: 'babel-loader',
                 options: {
                     presets: [


horymury (approved):

damencho (approved):

NewEraCracker:
Please note this change causes the entire tree to be transpiled including the node_modules folder. Before, all node_modules except @jitsi/js-utils would be ignored from the transpile.

However for lib-jitsi-meet the effect of this change is minimal, there are no babel warnings and build time only increases by a few seconds.

garyhuntddn:
It's much nicer not having that weird double-negative in the excludes

saghul:
> Please note this change causes the entire tree to be transpiled including the node_modules folder. Before, all node_modules except @jitsi/js-utils would be ignored from the transpile.
> 
> 
> 
> However for lib-jitsi-meet the effect of this change is minimal, there are no babel warnings and build time only increases by a few seconds.

That was a mistake if you look at the webpack file history.

Since the main output is the webpack bundle we do need the transpilation.

PR #1959, opened 2022-03-14
Conversion of Resolutions to typescript (and named export)

garyhuntddn:


saghul (dismissed):

saghul:

NewEraCracker:
@garyhuntddn Currently as you have declared the IDE gives this suggestion.
image

However you could change it so to alias each individual member to Resolution giving a suggestion that enumerates the object keys
image

My two cents.

garyhuntddn:
Weird - my reply by email whilst I was on hols didn't get to you.

@NewEraCracker - I agree with what you are saying and that it's better to have the intellisense there.

I'd originally tried something similar as I want enum semantics but with a Resolution rather than string value (something that typescript doesn't support as far as I am aware)

I've just pushed an update which attempts to satisfy both requirements (ensuring that all member properties are of type Resolution and that the intellisense works well) - see what you think.

If anybody knows better type manipulation that can do this is a better way then please make a suggestion - I'm sure there is a better way but I can't seem to find it

garyhuntddn:
Cracked what I was trying to do - we now have a type ```ResolutionsMap``` which means that any member of ```resolutions``` must be of type ```Resolution``` without the need to use ```as Resolution``` which I consider a bit dangerous (and not very DRY)

NewEraCracker (approved):

saghul:

garyhuntddn:
Yeah - I wanted enum semantics and also to constrain that each element was a Resolution. But enums in typescript can only be strings or numbers not objects :-(

I agree that the intellisense is better off without the type constraint.

The current usage across the other modules is to pass in a string but I wander if I can improve that with the use of keyof? Worth a try isn’t it.

The other complication is that the entries are strings containing numbers which was limiting what I could pull off.

Let me see if I can do a bit of type voodoo and export a constant that is both a map of string and also the known set of keys and therefore get the best of both worlds.

PR #1961, opened 2022-03-15
End-to-end ping updates

bgrozev:
- Do not send analytics periodically.
- Stop sending e2eping after 5 responses.
- Only send e2eping in one direction in a pair of endpoints.
- Fix syntax, lint errors.
- squash: Cleanup logs, stop pings if we don't get responses.
- Use a timeout instead of interval.
- feat: Space out e2e pings based on conference size.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -407,7 +407,7 @@ JitsiConference.prototype._init = function(options = {}) {
     this.room.addListener(XMPPEvents.SOURCE_ADD_ERROR, this._removeLocalSourceOnReject);
     this.room.addListener(XMPPEvents.SOURCE_REMOVE, this._updateRoomPresence);
 
-    if (config.e2eping.enabled === true) {
+    if (config.e2eping && config.e2eping.enabled === true) {
         this.e2eping = new E2ePing(
             this,
             config,
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -407,17 +407,21 @@ JitsiConference.prototype._init = function(options = {}) {
     this.room.addListener(XMPPEvents.SOURCE_ADD_ERROR, this._removeLocalSourceOnReject);
     this.room.addListener(XMPPEvents.SOURCE_REMOVE, this._updateRoomPresence);
 
-    this.e2eping = new E2ePing(
-        this,
-        config,
-        (message, to) => {
-            try {
-                this.sendMessage(
-                    message, to, true /* sendThroughVideobridge */);
-            } catch (error) {
-                logger.warn('Failed to send E2E ping request or response.', error && error.msg);
-            }
-        });
+    if (config.e2eping.enabled === true) {
+        this.e2eping = new E2ePing(
+            this,
+            config,
+            (message, to) => {
+                try {
+                    this.sendMessage(
+                        message, to, true /* sendThroughVideobridge */);
+                } catch (error) {
+                    logger.warn(
+                        'Failed to send E2E ping request or response.',
+                         error && error.msg);
+                }
+            });
+    }
 
     if (!this.rtc) {
         this.rtc = new RTC(this, options);
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -407,18 +407,15 @@ JitsiConference.prototype._init = function(options = {}) {
     this.room.addListener(XMPPEvents.SOURCE_ADD_ERROR, this._removeLocalSourceOnReject);
     this.room.addListener(XMPPEvents.SOURCE_REMOVE, this._updateRoomPresence);
 
-    if (config.e2eping && config.e2eping.enabled === true) {
+    if (config.e2eping?.enabled) {
         this.e2eping = new E2ePing(
             this,
             config,
             (message, to) => {
                 try {
-                    this.sendMessage(
-                        message, to, true /* sendThroughVideobridge */);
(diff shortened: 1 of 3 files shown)


modules/e2eping/e2eping.js
```diff
@@ -104,45 +107,35 @@ class ParticipantWrapper {
...
                     rtt = Math.min(rtt, request.rtt);
                 }
             }
         }
 
-        if (rtt < Infinity) {
-            this.sendAnalytics(rtt);
+        if (numRequestsWithResponses >= e2eping.numRequests) {
+            console.log(`Measured RTT=${rtt} ms to participant ${this.id} (in
+                region ${this.participant.getProperty('region')}`));
```
JonathanLennox: This is the remote participant's connected region, rather than the region of the bridge they're connected to?

modules/e2eping/e2eping.js
```diff
@@ -266,7 +266,13 @@ export default class E2ePing {
             delete this.participants[id];
         }
 
-        this.participants[id] = new ParticipantWrapper(participant, this);
+        // We don't need to send e2eping in both directions for a pair of
+        // endpoints. Force only one direction with just string comparison of
+        // the IDs.
+        if (conference.myUserId() > id) {
+            console.log(`Starting e2eping for participant ${id}`);
+            this.participants[id] = new ParticipantWrapper(participant, this);
```
JonathanLennox, line 326: It's not new here, but I don't feel like the name `ParticipantWrapper` is very descriptive of the fact that it's used for e2e pings.

modules/e2eping/e2eping.js
```diff
@@ -126,8 +129,13 @@ class ParticipantWrapper {
...
         if (numRequestsWithResponses >= this.e2eping.numRequests) {
-            console.log(`Measured RTT=${rtt} ms to participant ${this.id} (in
-                region ${this.participant.getProperty('region')}`);
+            logger.info(`Measured RTT=${rtt} ms to participant ${this.id} (in `
+                + `region ${this.participant.getProperty('region')}`);
+            this.clearIntervals();
+        } else if (totalNumRequests > 2 * this.e2eping.numRequests) {
+            logger.info(`Stopping e2eping for ${this.id} because we've sent `
+                + `${totalNumRequests} with only ${numRequestsWithResponses} `
+                + `responses.`)
```
JonathanLennox: The log message doesn't match the condition?

JonathanLennox (commented):
I don't know js or ljm well enough to review that part of it, but the scaling logic makes sense to me.

modules/e2eping/e2eping.js
bgrozev: Correct. How to log this data to rtcstats is still TBD

modules/e2eping/e2eping.js
bgrozev: How so? It should log something like:
```
Stopping e2eping for ${this.id} because we've sent 11 with only 0 responses.
```

bgrozev: These are not the same. The purpose of this check is to stop sending requests if we're not getting responses for some reason. 

this.e2eping.numRequests is essentially a constant (5) for the number of successful requests we want before we log the RTT measurement (I think it needs a better name, any suggestions?), so I didn't add it to the log message because it's not very interesting. `numRequestsWithResponses` is the number of responses we've got, which is interesting to log.

JitsiConference.js
```diff
@@ -407,17 +407,21 @@ JitsiConference.prototype._init = function(options = {}) {
...
-        config,
-        (message, to) => {
-            try {
-                this.sendMessage(
-                    message, to, true /* sendThroughVideobridge */);
-            } catch (error) {
-                logger.warn('Failed to send E2E ping request or response.', error && error.msg);
-            }
-        });
+    if (config.e2eping && config.e2eping.enabled === true) {
```
jallamsetty1, line 410: Suggestion: You can re-write the if condition to `config.e2eping?.enabled`. The equality check is redundant there.

JitsiConference.js
```diff
@@ -407,17 +407,21 @@ JitsiConference.prototype._init = function(options = {}) {
...
+        this.e2eping = new E2ePing(
+            this,
+            config,
+            (message, to) => {
+                try {
+                    this.sendMessage(
+                        message, to, true /* sendThroughVideobridge */);
+                } catch (error) {
+                    logger.warn(
```
jallamsetty1, line 419: nit: eslint max-len is set to 120, there is no need to break into a new line here and there are few other instances in this PR. We are trying to adjust any code we touch to conform to 120 max-len limit.

jallamsetty1 (approved):

modules/e2eping/e2eping.js
gpolitis: @nils-ohlmeier or @andrei-gavrilescu can correct me if I'm wrong, but I think this either needs to be bubbled up to jitsi-meet (and add a method similar to [this](https://github.com/jitsi/jitsi-meet/blob/85c505a29d5b6d39cabb26c26a1fb054b3ce098e/react/features/rtcstats/RTCStats.js#L94) one to send the rtt back to the rtcstats server) or add the capability of sending stats entries from lib-jitsi-meet. Not sure which approach is better suited here. /cc @jallamsetty1 @hristoterezov in case you guys have any thoughts to share

jallamsetty1 (approved):

Issue #1649, opened 2021-07-01
Conference onRemoteTrackAdded videoType is the previous remoteTrack's videoType


08BRYLL23:
*This Issue tracker is only for reporting bugs and tracking code related issues.*

Before posting, please make sure you check community.jitsi.org to see if the same or similar bugs have already been discussed. General questions, installation help, and feature requests can also be posted to community.jitsi.org.

## Description
---  When I add a local track of type video/desktop, it seems that when I check on onRemoteTrackAdded (function for **JitsiMeetJS.events.conference.TRACK_ADDED**), it seems the videoType is undefined. If I try to create another local track, it seems to get the previous one's videoType. (One thing to take note of is that the previous value only happens when I try to retrieve the videoType but when I check in the logs the videoType seems to be accurate)

## Current behavior
---onRemoteTrackAdded the videoType of the remote track is the previous's videoType

## Expected Behavior
---onRemoteTrackAdded the videoType of the remote track should be the videoType that was added

## Possible Solution
---

## Steps to reproduce
---
![Screen Shot 2021-07-01 at 11 50 54 AM](https://user-images.githubusercontent.com/40222666/124087526-227a7400-da63-11eb-9266-38be16c879e8.png)

# Environment details
---

jadhavsaurabh:
@08BRYLL23 @damencho Any updates on this?

damencho:
Do you reproduce with latest master? If yes, give us a simple example to reproduce it, thanks.

jadhavsaurabh:
@damencho 

https://meet.jit.si/libs/lib-jitsi-meet.min.js?v=5666.2368 I am using this minified lib-jitsi-meet
````
let room = await connection.initJitsiConference('name', {
  enableLayerSuspension: false,
  p2p: {
              enabled: false,
              enableUnifiedOnChrome: true,
              preferredCodec: 'VP9',
              disableH264: true,
  },
  e2eping: { pingInterval: -1 }
});
````

and I am listening to events like this
```
room &&
      room.on(JitsiMeetJS.events.conference.TRACK_ADDED, (track) => {
        console.log("isTrackDesktopTrack?", track, track.videoType);
        !track.isLocal() && !remoteTrackMatchesLocal(track) && onRemoteTrackAdd(track);
      });
```
Expected behaviour would be:
`isTrackDesktopTrack? {TrackObject}, "desktop"`

But I am getting
`isTrackDesktopTrack? {TrackObject}, undefined`

[...]

PR #1963, opened 2022-03-17
Fix connection stats

jallamsetty1:
fix(TPC) Update the local SSRC cache when a track is replace with another.
Fixes an issue where the local stats are not emitted in Unified plan after a track replace.

fix(multi-stream) Add video tracks sequentially.
If there are more than one video track at pc creation time, add them sequentially to avoid renegotiation loop.

fix(TPC) Send presence before signaling for SS tracks.
This is needed for the track to be identified as a desktop track on the remote peer when the JitsiRemoteTrack is created.
Fixes https://github.com/jitsi/lib-jitsi-meet/issues/1649.

fix(multi-stream) Do not emit ParticipantConnectionStatus updates when source-name signaling is enabled.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -984,10 +984,20 @@ export default class JingleSessionPC extends JingleSession {
         }
         const workFunction = finishedCallback => {
             const addTracks = [];
+            const audioTracks = localTracks.filter(track => track.getType() === MediaType.AUDIO);
+            const videoTracks = localTracks.filter(track => track.getType() === MediaType.VIDEO);
+            let tracks = localTracks;
 
-            for (const localTrack of localTracks) {
-                addTracks.push(this.peerconnection.addTrack(localTrack, this.isInitiator));
+            // Add only 1 video track at a time. Adding 2 or more video tracks to the peerconnection at the same time
+            // makes the browser go into a renegotiation loop by firing 'negotiationneeded' event after every
+            // renegotiation.
+            if (FeatureFlags.isMultiStreamSupportEnabled() && videoTracks.length > 1) {
+                tracks = [ ...audioTracks, videoTracks[0] ];
             }
+            for (const track of tracks) {
+                addTracks.push(this.peerconnection.addTrack(track, this.isInitiator));
+            }
+            videoTracks.length && videoTracks.splice(0, 1);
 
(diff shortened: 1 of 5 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -2015,20 +2036,25 @@ export default class JingleSessionPC extends JingleSession {
...
      *
-     * @param {JitsiLocalTrack} localTrack track to be added to the peer connection.
+     * @param {Array<JitsiLocalTrack>} localTracks - Tracks to be added to the peer connection.
      * @returns {Promise<void>} that resolves when the track is successfully added to the peerconnection, rejected
      * otherwise.
      */
-    addTrack(localTrack) {
-        if (!FeatureFlags.isMultiStreamSupportEnabled() || localTrack.type !== MediaType.VIDEO) {
-            return Promise.reject(new Error('Multiple tracks of a given media type are not supported'));
+    addTracks(localTracks = null) {
```
paweldomas, line 2043: we need to add an adapter and keep the `addTrack` method (mark as deprecated) as it's an API breaking change, something that will do `addTracks([ track])` internally and fail for the multi stream I guess? throw an error that says use addTracks in multi stream, WDYT?

jallamsetty1: This is the new method that we added only for multi-stream support at JingleSessionPC level. This wouldn't be considered a breaking change.

modules/xmpp/JingleSessionPC.js
```diff
@@ -2015,20 +2036,25 @@ export default class JingleSessionPC extends JingleSession {
...
+        if (!FeatureFlags.isMultiStreamSupportEnabled()
+            || !localTracks?.length
+            || localTracks.find(track => track.getType() !== MediaType.VIDEO)) {
+            return Promise.reject(new Error('Multiple tracks of the given media type are not supported'));
```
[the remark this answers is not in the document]
jallamsetty1: Yes, since multi-stream is supported only for video as of now.

modules/xmpp/JingleSessionPC.js
paweldomas: Oh got it, I thought it was JitsiConference

paweldomas (approved):

PR #1965, opened 2022-03-23
fix(multi-stream) Handle source-name based presence updates.

jallamsetty1:
Handle presence updates for multiple sources per remote endpoint.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -205,15 +205,17 @@ export default function TraceablePeerConnection(
 
     // SignalingLayer listeners
     this._peerVideoTypeChanged = this._peerVideoTypeChanged.bind(this);
-    this.signalingLayer.on(
-        SignalingEvents.PEER_VIDEO_TYPE_CHANGED,
-        this._peerVideoTypeChanged);
+    this.signalingLayer.on(SignalingEvents.PEER_VIDEO_TYPE_CHANGED, this._peerVideoTypeChanged);
 
     this._peerMutedChanged = this._peerMutedChanged.bind(this);
-    this.signalingLayer.on(
-        SignalingEvents.PEER_MUTED_CHANGED,
-        this._peerMutedChanged);
+    this.signalingLayer.on(SignalingEvents.PEER_MUTED_CHANGED, this._peerMutedChanged);
     this.options = options;
+
+    // Setup SignalingLayer listeners for source-name based events.
+    this.signalingLayer.on(SignalingEvents.SOURCE_MUTED_CHANGED,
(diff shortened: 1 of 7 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -572,6 +569,33 @@ TraceablePeerConnection.prototype._peerMutedChanged = function(
...
+/**
+ * Handles remote source mute, unmute and videoType changed events.
+ *
+ * @param {string} sourceName - The name of the remote source.
+ * @param {boolean} isMuted - The new mute state.
+ * @param {string} videoType - The new videoType value.
+ */
+TraceablePeerConnection.prototype._sourceMutedOrVideoTypeChanged = function(sourceName, isMuted, videoType) {
+    // Ignore presence updates for local sources.
+    const localSource = this.getLocalTracks().find(t => t.getSourceName() === sourceName);
```
paweldomas, line 581: what's the point in doing local source look up if the method returns anyway when it's true?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -572,6 +569,33 @@ TraceablePeerConnection.prototype._peerMutedChanged = function(
...
 };
 
+/**
+ * Handles remote source mute, unmute and videoType changed events.
+ *
+ * @param {string} sourceName - The name of the remote source.
+ * @param {boolean} isMuted - The new mute state.
+ * @param {string} videoType - The new videoType value.
+ */
+TraceablePeerConnection.prototype._sourceMutedOrVideoTypeChanged = function(sourceName, isMuted, videoType) {
```
paweldomas, line 579: can you split this into 2: source muted and source video type changed methods?

modules/RTC/TraceablePeerConnection.js
jallamsetty1: because we want to ignore presence updates for local sources. Do you want me to rename `localSource` to `isLocalSource` ?

modules/RTC/TraceablePeerConnection.js
jallamsetty1: I had it that way before but its code repetition and there is only 1 line that is different.

modules/RTC/TraceablePeerConnection.js
paweldomas: why don't break the circuit much earlier at the time when the events are going out of the presence handler? btw. this code ignores them anyway because no remote track will be found

modules/RTC/TraceablePeerConnection.js
paweldomas: The code repetition is only for the track lookup which calls for a utility method eventually.

paweldomas (approved):

PR #1966, opened 2022-03-23
fix: Set affiliation.

damencho:
XEP-0045:
Affiliations are granted, revoked, and maintained based on the user’s bare JID, not the nick as with roles.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1721,7 +1721,7 @@ JitsiConference.prototype.grantOwner = function(id) {
     if (!participant) {
         return;
     }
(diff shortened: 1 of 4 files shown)


JitsiConference.js
```diff
@@ -1737,11 +1737,10 @@ JitsiConference.prototype.revokeOwner = function(id) {
     if (isMyself) {
         this.room.setAffiliation(this.room.myroomjid, role);
```
bgrozev, line 1738: I think this also needs an update

bgrozev (approved):

saghul:
Oooh, this will help simplify some BR code!

PR #1970, opened 2022-03-25
Fix Parse Connection Failed Message

muratersin:
"lastIndex" must be set as 0 for each time because of when "_parse ConnectionFailedMessage" triggered twice message will be undefined.

[For more information](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp/lastIndex)


saghul (approved):

saghul:
Nice find!

saghul:

PR #1971, opened 2022-03-29
Fix p2p multistream

jallamsetty1:
fix(multi-stream) add new m-line when remote adds a second video source in p2p.
A new m-line needs to be added in the remote description when a new secondary video source is signaled by the peer.

fix(TPC) Suppress lower layers only for screenshare tracks.
With multi-stream support enabled, an extra check for videoType for local track is needed to dtermine if the lower layers need to be suppressed.

fix(TPC) Mark the direction inactive on inactive media connection.
If a new secondary video source is added on the jvb connection while p2p is active, make sure the direction is set to inactive on the jvb connection, so that no media is being sent out on the jvb connection.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -1697,6 +1697,8 @@ export default class JingleSessionPC extends JingleSession {
                 });
             });
 
+            let midFound = false;
+
             /* eslint-enable no-invalid-this */
             currentRemoteSdp.media.forEach((media, i2) => {
                 if (!SDPUtil.findLine(media, `a=mid:${name}`)) {
@@ -1706,7 +1708,14 @@ export default class JingleSessionPC extends JingleSession {
                     addSsrcInfo[i2] = '';
                 }
                 addSsrcInfo[i2] += lines;
+                midFound = true;
             });
+
+            // In p2p unified mode with multi-stream enabled, the new sources will have content name that doesn't exist
+            // in the current remote description. Add a new m-line for this newly signaled source.
+            if (!midFound && this.isP2P && FeatureFlags.isSourceNameSignalingEnabled()) {
+                addSsrcInfo[name] = lines;
+            }
         });
 
         return addSsrcInfo;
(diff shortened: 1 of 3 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -2640,12 +2658,12 @@ export default class JingleSessionPC extends JingleSession {
         // context a common object for one run of ssrc update (source-add and source-remove) so we can match them if we
         // need to
         const ctx = {};
+        const removedSsrcInfo = getSignaledSourceInfo(sdpDiffer);
 
-        if (removedAnySSRCs) {
-            const sourceInfo = getSignaledSourceInfo(sdpDiffer);
-
+        if (removedAnySSRCs && removedSsrcInfo.ssrcs.length) {
```
paweldomas, line 2663: why would `removedAnySSRCs` be true while `removedSsrcInfo.ssrcs.length` is `0`?

modules/xmpp/JingleSessionPC.js
```diff
@@ -2670,19 +2688,18 @@ export default class JingleSessionPC extends JingleSession {
             );
 
         const containsNewSSRCs = sdpDiffer.toJingle(add);
+        const addedSsrcInfo = getSignaledSourceInfo(sdpDiffer);
 
-        if (containsNewSSRCs) {
-            const sourceInfo = getSignaledSourceInfo(sdpDiffer);
-
+        if (containsNewSSRCs && addedSsrcInfo.ssrcs.length) {
```
paweldomas, line 2693: same question as in the `removedAnySSRCs` case, why would it contain new sources while actually there's 0 SSRCs?

modules/xmpp/JingleSessionPC.js
jallamsetty1: This would happen in p2p when a remote peer adds a second video source. A recv-only m-line will be created as a result in the local SDP without any SSRC.

modules/xmpp/JingleSessionPC.js
jallamsetty1: Same as above

modules/xmpp/JingleSessionPC.js
paweldomas: you're writing about mline being added while the if condition is about any sources being removed

paweldomas: I don't understand how an mline being added/removed affects the two conditions specifically

jallamsetty1: Sorry was AFK and had typed the original response on my mobile. This comment is meant for source-add case. When the remote peer adds a new source, a new m-line in the remote description is created and therefore a recv-only m-line without any SSRC is added to the local SDP as a result. Therefore on renegotiation, this code detects that there is a new m-line and SDPDiffer.toJIngle returns true because of the new m-line even though there is no new SSRC. I didn't change the existing variables and therefore it may be confusing here. This happens only in the case of P2P. For JVB case, the interop.toPlanB() in the getLocalDescription getter takes care of ignoring the recv-only m-lines.  
I added this extra check for the source-remove just in case to make sure source-remove with no SSRCs is ever signaled to the peer.

paweldomas: maybe the condition then should be based only on `removedSsrcInfo.ssrcs.length`? it is confusing to have two sources of truth

PR #1973, opened 2022-03-30
Plumbing for e2e pings.

gpolitis:


JitsiConferenceEvents.ts
```diff
@@ -251,6 +251,11 @@ export enum JitsiConferenceEvents {
      */
     PARTICIPANT_PROPERTY_CHANGED = 'conference.participant_property_changed',
 
+    /**
+     * Indicates that the end-to-end RTT between the local participant and a remote participant has been measured.
+     */
+    PARTICIPANT_E2ERTT_RECEIVED = 'conference.participant_e2ertt_received',
```
bgrozev, line 257: There's already an event defined [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/service/e2eping/E2ePingEvents.ts). It's not used anywhere since my latest changed, but I left it in case it needed to be re-used.

I don't have a preference as to which one is used, but can you please remove whichever one ends up unused?

nils-ohlmeier (approved):
LGTM

JitsiConferenceEvents.ts
gpolitis: Thanks, Boris 👍

bgrozev (approved):

PR #1975, opened 2022-03-31
fix(browserCapabilities) bump minimum chrome version to 88

NewEraCracker:
This is the target chrome version in webpack babel

https://github.com/jitsi/lib-jitsi-meet/blob/3b38ab37f2bd2777d9ac56ee006840d412aac83e/webpack-shared-config.js#L47

### DIFF
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -4,7 +4,7 @@ import { getLogger } from '@jitsi/logger';
 const logger = getLogger(__filename);
 
 /* Minimum required Chrome / Chromium version. This applies also to derivatives. */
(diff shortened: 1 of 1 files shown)


saghul:
I'd say let's bump it to 90. Then once ~110 hits we can bump it to 96 since it will be an LTS release: https://support.google.com/chrome/a/answer/11333726

NewEraCracker:
I am more inclined to change this instead to 88. Our custom application still gets some mobile users on that version.

saghul:
> I am more inclined to change this instead to 88. Our custom application still gets some mobile users on that version.

Sounds good.

saghul (approved):

saghul:

jallamsetty1:
Closing this since the plan is to bump the version to 96.

PR #1980, opened 2022-04-05
fix(SignalingLayer) add sourceName argument to getPeerMediaInfo

NewEraCracker:
Related to 6d8060a4c6a09f40c6a582de851f9751eb00a41e

### DIFF
--- a/service/RTC/SignalingLayer.js
+++ b/service/RTC/SignalingLayer.js
@@ -86,6 +86,7 @@ export default class SignalingLayer extends Listenable {
      * @param {string} owner the MUC jid of the participant for whom
      * {@link PeerMediaInfo} will be obtained.
      * @param {MediaType} mediaType the type of the media for which presence
+     * @param {SourceName} sourceName - The name of the source for which the info is to be obtained.
      * info will be obtained.
      * @return {PeerMediaInfo|null} presenceInfo an object with media presence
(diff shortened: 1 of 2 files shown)


service/RTC/SignalingLayer.js
```diff
@@ -86,14 +86,15 @@ export default class SignalingLayer extends Listenable {
...
+     * @param {SourceName} sourceName - The name of the source for which the info is to be obtained.
      * info will be obtained.
      * @return {PeerMediaInfo|null} presenceInfo an object with media presence
      * info or <tt>null</tt> either if there is no presence available for given
      * JID or if the media type given is invalid.
      *
      * @deprecated This method is to be replaced with getPeerSourceInfo.
      */
-    getPeerMediaInfo(owner, mediaType) { // eslint-disable-line no-unused-vars
+    getPeerMediaInfo(owner, mediaType, sourceName) { // eslint-disable-line no-unused-vars
```
paweldomas, line 97: We don't want to do that, this is a legacy method used for the clients who are not using source names yet. Eventually it will be removed and replaced with `getPeerSourceInfo`

paweldomas: Oh I see what's going on in the other commit... oh well

jallamsetty1 (approved):

jallamsetty1:
Jenkins test this please

PR #1983, opened 2022-04-12
fix(video-quality) Update frame heights in content-modify for p2p.

jallamsetty1:
When the user changes the preferred video quality settings from the UI, update the frame height values in content-modify for p2p connection when source-name signaling is enabled.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1123,7 +1123,12 @@ JitsiConference.prototype.addTrack = function(track) {
         return Promise.reject(new Error(`Cannot add second ${mediaType} track to the conference`));
     }
 
-    return this.replaceTrack(null, track);
+    return this.replaceTrack(null, track)
+        .then(() => {
+            if (track.getVideoType() === VideoType.DESKTOP && FeatureFlags.isMultiStreamSupportEnabled()) {
+                this._updateRoomPresence(this.getActiveMediaSession());
+            }
+        });
 };
 
 /**
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -1125,6 +1125,9 @@ JitsiConference.prototype.addTrack = function(track) {
 
     return this.replaceTrack(null, track)
         .then(() => {
+            // Presence needs to be sent here for desktop track since we need the presence to reach the remote peer
+            // before signaling so that a fake participant tile is created for screenshare. Otherwise, presence will
+            // only be sent after a session-accept or source-add is ack'ed.
             if (track.getVideoType() === VideoType.DESKTOP && FeatureFlags.isMultiStreamSupportEnabled()) {
                 this._updateRoomPresence(this.getActiveMediaSession());
             }
--- a/modules/qualitycontrol/ReceiveVideoController.js
+++ b/modules/qualitycontrol/ReceiveVideoController.js
@@ -2,6 +2,7 @@ import { getLogger } from '@jitsi/logger';
 import isEqual from 'lodash.isequal';
 
 import * as JitsiConferenceEvents from '../../JitsiConferenceEvents';
+import { MediaType } from '../../service/RTC/MediaType';
 import FeatureFlags from '../flags/FeatureFlags';
 
 const logger = getLogger(__filename);
@@ -179,6 +180,14 @@ export default class ReceiveVideoController {
         // The number representing the maximum video height the local client should receive from the bridge.
         this._maxFrameHeight = MAX_HEIGHT_ONSTAGE;
 
+        /**
+         * The map that holds the max frame height requested for each remote source when source-name signaling is
+         * enabled.
+         *
+         * @type Map<string, number>
+         */
+        this._sourceReceiverConstraints = new Map();
+
         // Enable new receiver constraints by default unless it is explicitly disabled through config.js.
         const useNewReceiverConstraints = config?.useNewBandwidthAllocationStrategy ?? true;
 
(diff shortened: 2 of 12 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -2524,32 +2581,30 @@ export default class JingleSessionPC extends JingleSession {
...
+        const newMaxFrameHeight = JingleSessionPC.parseMaxFrameHeight(jingleContents);
+        const sourceMaxFrameHeights = JingleSessionPC.parseSourceMaxFrameHeight(jingleContents);
 
         // frame height is optional in our content-modify protocol
         if (newMaxFrameHeight) {
             logger.info(`${this} received remote max frame height: ${newMaxFrameHeight}`);
             this.remoteRecvMaxFrameHeight = newMaxFrameHeight;
-            this.eventEmitter.emit(
-                MediaSessionEvents.REMOTE_VIDEO_CONSTRAINTS_CHANGED, this);
+            this.eventEmitter.emit(MediaSessionEvents.REMOTE_VIDEO_CONSTRAINTS_CHANGED, this);
```
paweldomas, line 2592: can we have two separate events? `MediaSessionEvents.REMOTE_VIDEO_CONSTRAINTS_CHANGED` for the old flow and `MediaSessionEvents.REMOTE_SOURCE_CONSTRAINTS_CHANGED` for the new flow?
[the code which is processing the event] is hard to follow (hard to tell why would there be no constraints)

[the code which is processing the event]:https://github.com/jitsi/lib-jitsi-meet/pull/1983/files#diff-88c6b4c65414c3f9776f316273316a331f84e1755c848039295038175ac8a9b1R78

modules/qualitycontrol/SendVideoController.js
```diff
@@ -74,8 +74,12 @@ export default class SendVideoController {
     _onMediaSessionStarted(mediaSession) {
         mediaSession.addListener(
             MediaSessionEvents.REMOTE_VIDEO_CONSTRAINTS_CHANGED,
-            session => {
-                if (session === this._conference.getActiveMediaSession()) {
+            (session, videoConstraints) => {
+                if (session === this._conference.getActiveMediaSession() && videoConstraints) {
```
paweldomas, line 78: I think the old logic used to care about "is session the active session". Shouldn't it be the following?
```
if (session === this._conference.getActiveMediaSession()) {
    if (videoConstraints) {
     // source name logic
    } else {
     // legacy logic
    }
}
```

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -330,7 +360,31 @@ export default class ReceiveVideoController {
...
 
-            if (p2pSession) {
+            if (!p2pSession) {
+                return;
+            }
+
+            if (FeatureFlags.isSourceNameSignalingEnabled()) {
+                const sourceReceiverConstraints = constraints.constraints;
+
+                // Add the new ones from the passed constraints to the cached map.
```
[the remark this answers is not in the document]
jallamsetty1: 1. It's not necessary but that is how I chose to do, I can change it.
2. `_getSourceReceiverConstraints(maxFrameHeight) `generates the default constraints when the application doesn't do different layouts and just relies on the[ legacy API ](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3798) to set the receive height.  `this._sourceReceiverConstraints` on the other hand is generated based on the [new constraints](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3786)  provided by the application. I can rename `_getSourceReceiverConstraints` to `_getDefaultSourceReceiverConstraints` if that makes it clear and maybe add an extra comment. Will that work ?

modules/qualitycontrol/SendVideoController.js
jallamsetty1: I thought that the isActive media session chheck wasn't necessary since content-modify is sent only on p2p conn and it is always the active session if it exists. However, at the beginning of the call, p2p is not the active one until p2p ICE connection get established. I brought the check back.

modules/qualitycontrol/ReceiveVideoController.js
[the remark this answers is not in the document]
jallamsetty1: I had introduced _getSourceReceiverConstraints(maxFrameHeight) to generate the default constraints when the application doesn't use the [newer constraints ](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3786) and instead uses the [legacy](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3798) API.

JitsiConference.js
```diff
@@ -1123,7 +1123,12 @@ JitsiConference.prototype.addTrack = function(track) {
         return Promise.reject(new Error(`Cannot add second ${mediaType} track to the conference`));
     }
 
-    return this.replaceTrack(null, track);
+    return this.replaceTrack(null, track)
+        .then(() => {
+            if (track.getVideoType() === VideoType.DESKTOP && FeatureFlags.isMultiStreamSupportEnabled()) {
```
paweldomas, line 1128: can you add a comment which explains that it's needed for desktop, because camera is the default video type? but then I'm wondering if the presence does not need to be updated anyway to send the "unmuted" status.

paweldomas (commented):
one last question/comment

JitsiConference.js
jallamsetty1: The videoType is what matters here more than the unmuted status. If the signaling reaches the peer before the presence does, then Jitsi-meet won't create a fake participant tile since the default videoType is camera.

jallamsetty1: I will add a comment

paweldomas: hmm okay... I hope at some point we'll be able to remove the video type from presence, freeze it (disallow video type changes) and send it over the signaling only. Then this could go away

paweldomas (approved):

PR #1984, opened 2022-04-12
JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED expose jitsiTrack as listener parameter

pangrr:
So that we don't have to maintain `jitsiTrack` in the scope where listeners are defined. 

This change is a dependency for https://github.com/jitsi/jitsi-meet/pull/11377

### DIFF
--- a/modules/connectivity/TrackStreamingStatus.ts
+++ b/modules/connectivity/TrackStreamingStatus.ts
@@ -337,7 +337,7 @@ export class TrackStreamingStatusImpl {
                 }));
 
             // It's common for the event listeners to access the JitsiRemoteTrack. Thus pass it as a parameter here.
-            this.track.emit(JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED, newStatus, this.track);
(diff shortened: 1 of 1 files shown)


modules/connectivity/TrackStreamingStatus.ts
```diff
@@ -336,7 +336,8 @@ export class TrackStreamingStatusImpl {
                     status: newStatus
                 }));
 
-            this.track.emit(JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED, newStatus);
+            // It's common for the event listeners to access the JitsiRemoteTrack. Thus pass it as a parameter here.
+            this.track.emit(JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED, newStatus, this.track);
```
paweldomas, line 340: lets have the track argument come first and then the status

pangrr: changed

pangrr:
I think this PR is not needed if UI components keep their own trackStreamingStatus locally.

paweldomas (approved):

PR #1987, opened 2022-04-15
fix(multi-stream) Send the initial session-accept and then add the secondary video tracks when a session is re-created

jallamsetty1:
When a media session is re-created with multiple video sources in the conference as result of peer leaving and joining the call, send out the session-accept with first video source and then add the secondary source and signal it. Otherwise, ice candidates can reach the peer before the seesion-accept because of timing and not be processed (since a remote sdp won't be set) and result in a broken connection. This is needed only for the responder case. In the initiator case, all the tracks are added to the pc and then a session-initiate is sent to the peer. This doesn't seem to trigger the renegotiate loop.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -354,12 +354,13 @@ export class TPCUtils {
      */
     replaceTrack(oldTrack, newTrack) {
         const mediaType = newTrack?.getType() ?? oldTrack?.getType();
+        const localTracks = this.pc.getLocalTracks(mediaType);
         const track = newTrack?.getTrack() ?? null;
         const isNewLocalSource = FeatureFlags.isMultiStreamSupportEnabled()
-            && this.pc.getLocalTracks(mediaType)?.length
+            && localTracks?.length
             && !oldTrack
             && newTrack
-            && !newTrack.conference;
+            && !localTracks.find(t => t === newTrack);
         let transceiver;
 
         // If old track exists, replace the track on the corresponding sender.
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -1007,12 +1007,27 @@ export default class JingleSessionPC extends JingleSession {
                 // modify sendSessionAccept method to do that
                 this.sendSessionAccept(() => {
                     success();
-
                     this.room.eventEmitter.emit(XMPPEvents.SESSION_ACCEPT, this);
+
+                    // The first video track is added to the peerconnection and signaled as part of the session-accept.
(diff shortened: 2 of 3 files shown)


paweldomas (dismissed):

modules/xmpp/JingleSessionPC.js
```diff
@@ -1007,12 +1007,27 @@ export default class JingleSessionPC extends JingleSession {
...
+                    // This will happen when someone shares a secondary source to a two people call, the other user
+                    // leaves and joins the call again, a new peerconnection is created for p2p connection. At this
+                    // point, there are 2 video tracks which need to be signaled to the remote peer.
+                    const videoTracks = localTracks.filter(track => track.getType() === MediaType.VIDEO);
+
+                    videoTracks.length && videoTracks.splice(0, 1);
+                    if (FeatureFlags.isMultiStreamSupportEnabled() && videoTracks.length) {
+                        // Clear the conference field on the secondary local tracks if it was set previously, this is
+                        // needed for correctly identifying the RTCRtpSender for track replacement.
+                        videoTracks.forEach(track => {
```
paweldomas, line 1023: is this needed only for the p2p case?

jallamsetty1: This would be needed for jvb case as well. A user shares a secondary video stream, then everyone else leaves the call and all the peerconnections on the ep are closed after the idle timeout. When someone new joins the call, then new p2p and jvb connections are created. The conference field needs to be set to null for all the secondary tracks.

modules/RTC/TPCUtils.js
```diff
@@ -354,12 +354,13 @@ export class TPCUtils {
...
         const mediaType = newTrack?.getType() ?? oldTrack?.getType();
+        const localTracks = this.pc.getLocalTracks(mediaType);
         const track = newTrack?.getTrack() ?? null;
         const isNewLocalSource = FeatureFlags.isMultiStreamSupportEnabled()
-            && this.pc.getLocalTracks(mediaType)?.length
+            && localTracks?.length
             && !oldTrack
             && newTrack
-            && !newTrack.conference;
+            && !localTracks.find(t => t === newTrack);
```
paweldomas, line 363: why would `replaceTrack` allow to replace a track (as the replace target/ new track argument) that is already part of the peerconnection?

jallamsetty1: This would happen in case of unmute operation when the track needs to be added back to the pc.

paweldomas: oh okay, can you put a comment?

paweldomas (approved):

PR #1994, opened 2022-04-25
fix(multi-stream) Inject multiple SIM groups on Firefox when multiple sources are added.

jallamsetty1:
Firefox doesn't generate a `a=ssrc-group: SIM` line in its local SDP for every local video source. When multiple video sources are added to the peerconnection, a SIM group for every 3 FID groups need to be injected so that they can signaled to Jicofo.


paweldomas (dismissed):

jallamsetty1:
Jenkins test this please

paweldomas (approved):

PR #1996, opened 2022-04-27
fix missing stats for screen share tiles by using source name as key

pangrr:
pair with https://github.com/jitsi/jitsi-meet/pull/11475

This fixes missing stats for screen share tiles in jitsi-meet when source name signaling is on.

### DIFF
--- a/modules/statistics/RTPStatsCollector.js
+++ b/modules/statistics/RTPStatsCollector.js
@@ -3,6 +3,7 @@ import { getLogger } from '@jitsi/logger';
 import { MediaType } from '../../service/RTC/MediaType';
 import * as StatisticsEvents from '../../service/statistics/Events';
 import browser from '../browser';
+import FeatureFlags from '../flags/FeatureFlags';
 
 const GlobalOnErrorHandler = require('../util/GlobalOnErrorHandler');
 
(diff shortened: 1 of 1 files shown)


paweldomas:
@jallamsetty1 I'm wondering what are your thoughts about this? to me it seems like it should have been possible to still take the stats on SSRC basis without extra source name mapping

jallamsetty1:
> @jallamsetty1 I'm wondering what are your thoughts about this? to me it seems like it should have been possible to still take the stats on SSRC basis without extra source name mapping

I would want to use the source name mapping whenever source-name signaling is enabled irrespective of the videoType of the track. We need a follow up JM PR that does this for existing video tiles as well. With ssrc-rewriting feature that is currently underway, the SSRCs of a track will change, source-name of the track is the only property that will not change for a given track.

paweldomas:
> > @jallamsetty1 I'm wondering what are your thoughts about this? to me it seems like it should have been possible to still take the stats on SSRC basis without extra source name mapping
> 
> I would want to use the source name mapping whenever source-name signaling is enabled irrespective of the videoType of the track. We need a follow up JM PR that does this for existing video tiles as well. With ssrc-rewriting feature that is currently underway, the SSRCs of a track will change, source-name of the track is the only property that will not change for a given track.

Sounds good! I will the review of this part up to you.

modules/statistics/RTPStatsCollector.js
```diff
@@ -309,6 +310,38 @@ StatsCollector.prototype._processAndEmitReport = function() {
...
+                if (sourceName) {
+                    const resolution = ssrcStats.resolution;
+
+                    if (resolution.width // eslint-disable-line max-depth
+                        && resolution.height
+                        && resolution.width !== -1
+                        && resolution.height !== -1) {
+                        const sourceResolutions = resolutions[sourceName] || {};
+
+                        sourceResolutions[ssrc] = resolution;
```
jallamsetty1, line 331: Can `sourceResolutions` be a map <sourceName, resolution> instead ? Same with `sourceFramerates` as well. SSRCs for a given track can change because of an upcoming feature. This can also be made in a follow-up PR since this current implementation is good for the release.

pangrr: They can be maps <sourceName, resolution> but jitsi-meet also need to change to adapt to the new structure. e.g. https://github.com/jitsi/jitsi-meet/blob/master/react/features/connection-stats/components/ConnectionStatsTable.js#L656
I think a separate PR for this structural change is a good idea.

pangrr:
Found a bug where stats are n/a for local screen share thumbnail. Logging track.getSourceName() after line https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/statistics/RTPStatsCollector.js#L301 shows no source name for local screen share track. Do you know the cause?

jallamsetty1:
> Found a bug where stats are n/a for local screen share thumbnail. Logging track.getSourceName() after line https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/statistics/RTPStatsCollector.js#L301 shows no source name for local screen share track. Do you know the cause?

maybe TPC's `localSSRCs` map is not getting updated when a second video track is added. I can debug whats happening there but that shouldn't block this PR.

jallamsetty1:
Jenkins test this please.

pangrr:
> @pangrr I am still seeing resolution stats for both camera and desktop tracks on the video tile with your changes. Am I missing something ? Screen Shot 2022-04-27 at 2 14 29 PM

@jallamsetty1 It seems you are looking at remote participant's stats which I can also see. It's the local participant's screen share missing stats. Can you verify?

jallamsetty1:
@pangrr I thought the idea was to split the resolution and fps stats per source and show them on the respective tiles. I would expect to see the camera resolution/fps on the video tile and the screenshare resolution/fps on the fake SS tile.

pangrr:
@jallamsetty1 Ah I see what you mean. Working on it...

pangrr:
I kind of changed my mind to remove ssrc under the flag. I am working on a jitsi-meet PR to pair with this.

jallamsetty1:
Jenkins pls add to whitelist.

jallamsetty1:
Jenkins test this please.

jallamsetty1 (approved):

PR #1999, opened 2022-04-28
ref(test) remove unused import

NewEraCracker:
Removes unused import by previous commit.


saghul (approved):

NewEraCracker:
I have updated this PR to the latest Jitsi version.

If you check before my change: `const { CodecMimeType } = exported as any;` CodecMimeType is undefined, but because of the *if statement* the test does not fail.

Also PR #1922 is now obsolete.

saghul (approved):

PR #2004, opened 2022-05-06
fix(BridgeChannel): close before opening a new one

paweldomas:
This happens in the ICE restart scenario when new JingleSession is created.


paweldomas:
What's wrong with the github action?

damencho:
You miss commiting some TS files.

damencho:
Check the git diff, just before the failure.

PR #2011, opened 2022-05-23
fix(e2ee) increases the number of unencrypted bytes for key and non-key frames to accomodate the VP9 descriptor

tmoldovan8x8:


saghul (commented):
Let's do this per media type please, and only do the header.

PR #2015, opened 2022-05-24
fix(RTC): Add function declaration to JitsiTrack.

jallamsetty1:
RN clients throw a function undefined error otherwise.

### DIFF
--- a/modules/RTC/JitsiTrack.js
+++ b/modules/RTC/JitsiTrack.js
@@ -408,6 +408,16 @@ export default class JitsiTrack extends EventEmitter {
     }
 
     /**
+     * Returns the streaming status of the track.
+     *
+     * @returns {string} the streaming status <tt>TrackStreamingStatus</tt> of the track. Returns null if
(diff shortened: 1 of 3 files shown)


damencho (approved):

modules/RTC/JitsiTrack.js
```diff
@@ -407,6 +407,16 @@ export default class JitsiTrack extends EventEmitter {
         return null;
     }
 
+    /**
+     * Returns the streaming status of the track.
+     *
+     * @returns {string} the streaming status <tt>TrackStreamingStatus</tt> of the track. Returns null if
+-    * trackStreamingStatusImpl hasn't been initialized.
+     */
+    getTrackStreamingStatus() {
```
paweldomas, line 416: why is it needed on a local track? it's not supposed to be available for local tracks

jallamsetty1: Sorry was too quick to merge this one. I found the bug in my JM PR, this change wasn't needed, will be reverting this one.

PR #2022, opened 2022-05-31
Improve documentation that jwt works on websocket

derTobsch:
see https://community.jitsi.org/t/lib-jitsi-meet-jwt-support-with-websockets/79806


saghul:
Cheers!

PR #2024, opened 2022-06-02
E2EE fixes

tmoldovan8x8:


modules/e2ee/ManagedKeyHandler.js
```diff
@@ -41,10 +41,14 @@ export class ManagedKeyHandler extends KeyHandler {
             this._onParticipantPropertyChanged.bind(this));
         this.conference.on(
             JitsiConferenceEvents.USER_JOINED,
-            this._onParticipantJoined.bind(this));
+            async () => {
```
saghul, line 44: Do you need to do this? Making `_onParticipantJoined` async and using await inside should be enough no?

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -41,10 +41,14 @@ export class ManagedKeyHandler extends KeyHandler {
...
+                await this._onParticipantJoined();
+            });
         this.conference.on(
             JitsiConferenceEvents.USER_LEFT,
-            this._onParticipantLeft.bind(this));
+            async () => {
```
saghul, line 49: Ditto.

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -102,21 +106,21 @@ export class ManagedKeyHandler extends KeyHandler {
...
      * Rotates the current key when a participant leaves the conference.
      * @private
      */
-    _onParticipantLeft(id) {
+    async _onParticipantLeft(id) {
         this.e2eeCtx.cleanup(id);
 
         if (this.enabled) {
-            this._rotateKey();
+            await this._rotateKey();
```
saghul, line 123: Not sure I understand the reasoning behind this whole change. Is it to avoid 2 participant joined / left operations from running concurrently? If so shouldn't we perhaps use some kind of operations queue?

saghul (commented):
❤️  the small and focused commits! Left some comments, PTAL!

modules/e2ee/ManagedKeyHandler.js
tmoldovan8x8: For rotating I have not seen the problem described for ratcheting, but I figured some race condition can occur here, as well. Indeed an operation queue would make more sense, but I think we can add this in the state management system we discussed.

saghul (approved):
🚀

tmoldovan8x8:

PR #2042, opened 2022-06-24
fix(FF): Don't add recvonly SSRC in session-accept

hristoterezov:


modules/sdp/LocalSdpMunger.js
```diff
@@ -245,8 +246,28 @@ export default class LocalSdpMunger {
...
+        //
+        // For all other use cases (when the user has had media but then the user has stopped it) we want to keep the
+        // receive only SSRCs in the SDP. Otherwise source-remove will be triggered and the next time the user add a
+        // track we will reuse the SSRCs and send source-add with the same SSRCs. This is problematic because of issues
+        // on Chrome and FF (https://bugzilla.mozilla.org/show_bug.cgi?id=1768729) when removing and then adding the
+        // same SSRC in the remote sdp the remote track is not rendered.
+        if (browser.isFirefox()
+            && (mediaDirection === MediaDirection.RECVONLY || mediaDirection === MediaDirection.INACTIVE)
+            && (
+                (mediaType === MediaType.VIDEO && !this.tpc._hadVideoAdded)
```
jallamsetty1, line 263: Aren't these vars supposed to be `_hasHadVideoTrack` and `_hasHadAudioTrack` instead of `_hadVideoAdded` and `_hadAudioAdded` ?

hristoterezov: Yes! Good catch!

PR #2047, opened 2022-06-27
remove unused config parameters

TheGor1lla:


saghul:
If anything, we should probably clarify the comment above saying that it's not necessary because we proxy the connections.

TheGor1lla:
We are currently trying to get token authentication to work with out setup.  
Following the readme "Also check if client to server encryption is enforced. If not then token authentication won't work:" this part reads that the encryption must be set to true.  
If  I correctly read what @saghul said  this part is irrelevant? If so shouldnt this part of the doc be removed?

damencho:
That setting is irrelevant to tokens, I think. So I don't see why not removing it ...

saghul:
> That setting is irrelevant to tokens, I think. So I don't see why not removing it ...

Agreed.

TheGor1lla:
Checked that it tokens work with either true/false to the mentioned parameter and removed the specific lines from the documentation

PR #2049, opened 2022-06-30
Improve 'JSON Web Token (JWT) authentication Prosody plugin' documentation

derTobsch:
improve the current `JWT token authentication Prosody plugin` documentation with all information we gathered over the last weeks while implementing token authentication with a shared secret.

### DIFF
--- a/doc/tokens.md
+++ b/doc/tokens.md
@@ -1,126 +1,232 @@
-JWT token authentication Prosody plugin
-==================
+# JSON Web Token (JWT) authentication Prosody plugin
 
-This plugin implements a Prosody authentication provider that verifies a client connection based on a JWT token described in [RFC7519].
-It allows use of an external form of authentication with lib-jitsi-meet. Once your user authenticates you need to
-generate the JWT token as described in the RFC and pass it to your client app. Once it connects with a valid token it is considered authenticated by the jitsi-meet system.
+This plugin implements a **Prosody authentication provider** that verifies a client connection based on a JWT
+described in [RFC7519]. It allows use of an external form of authentication with _lib-jitsi-meet_.
+Once your user authenticates you need to generate the JWT as described in the RFC and pass it to your client app.
+Once it connects with a valid token, it is considered authenticated by the jitsi-meet system.
 
-During configuration you will need to provide the *application ID* that identifies the client and a *secret* shared by both server and JWT token generator. Like described in the RFC, the secret is used to compute a HMAC hash value which allows authentication of the generated token. There are many existing libraries which can be used to implement token generation. More info can be found here: [http://jwt.io/#libraries-io]
+During configuration, you can choose between two JWT validation methods:
 
-JWT token authentication works with BOSH and websocket connections.
+* Shared Secret Validation or
+* Public Key Validation 
 
-[RFC7519]: https://tools.ietf.org/html/rfc7519
-[http://jwt.io/#libraries-io]: http://jwt.io/#libraries-io
+In both cases you will need to provide the _application ID_ that identifies the client.
 
-### Token structure
+For the **Shared Secret Validation** an _application secret_ is shared by both the server (Prosody) and the JWT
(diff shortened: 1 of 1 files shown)


doc/tokens.md
```diff
@@ -1,148 +1,251 @@
...
 ```
+
 ### Token verification
 
-JWT token is currently checked in 2 places:
-- when a user connects to Prosody through BOSH. The token value is passed as the 'token' query paramater of the BOSH URL. User uses XMPP anonymous authentication method.
-- when a MUC room is being created/joined Prosody compares the 'room' claim with the actual name of the room. In addition, the 'sub' claim is compare to either the tenant (for TENANT/ROOM URLs) or the base domain (for /ROOM URLs).  This prevents a stolen token being abused by unathorized users to allocate new conference rooms in the system. Admin users are not required to provide a valid token which is used for example by Jicofo.
+JWT is currently checked in 2 places:
+
+- when a user connects to Prosody through BOSH or websocket. The token value is passed as the 'token' query parameter of
```
saghul, line 190: WebSocket

doc/tokens.md
```diff
@@ -1,148 +1,251 @@
...
+At the current level of integration every user that joins the conference has to provide the token and not just the one
+who creates the room. It should be possible to change that by using a second anonymous domain, but that hasn't been
+tested yet.
 
 ### Installing the token plugin
 
-Token authentication can be integrated automatically using a Debian package install. Once you have jitsi-meet installed
-just install 'jitsi-meet-tokens' on top of it. In order to have it configured automatically at least version 779 of
+Token authentication can be integrated automatically using a Debian package. Once you have `jitsi-meet` installed
+just install `jitsi-meet-tokens` on top of it. In order to have it configured automatically at least version 779 of
```
saghul, line 226: I believe this package will be going away soon, right @damencho ?

saghul (commented):
Nice improvements! Left some comments.

saghul (approved):
LGTM, but I'd love an extra pair of eyes here.

derTobsch:
Hey @saghul,
is there anything I can help to merge this?

doc/tokens.md
```diff
@@ -1,148 +1,228 @@
...
+
+#### Public Key Validation
+
+##### Headers
+
+`kid` must be the name of the public key to validate the token against.
+
+```json
+{
+    "kid": "jitsi/custom_key_name",
```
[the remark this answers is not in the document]
damencho: Yep, confirm anything here.

doc/tokens.md
damencho: Well we have talked about it, but nothing is planned ... I think we need to figure out and move it forward when doing the authentication refactor.

doc/tokens.md
```diff
@@ -1,148 +1,251 @@
...
+
+**Sub with tenant** If the 'sub' contains the tenant. You need to define the tenant in the `muc` when creating a new `JitsiConnection` in the options 
+within the `hosts` section, see
+
+```javascript
+    const connection = new JitsiMeetJS.JitsiConnection(
+    '${applicationId}',
+    '${token}',
+    {
+        serviceUrl: 'wss://muc.server.net/xmpp-websocket',
```
damencho, line 57: In other places `muc.server.net` is referenced as the muc component. So here the websocket needs to have the address `server.net`.
You also need to add the tenant here to the address. So it would be `'wss://server.net/tenant/xmpp-websocket'`
And it uses this nginx rule to set some correct params: https://github.com/jitsi/jitsi-meet/blob/fcb1845e2a6d7778dacd29f5c75e3d48c52b9c47/doc/debian/jitsi-meet/jitsi-meet.example#L158

damencho (changes requested):

damencho:
@derTobsch just a small change, other than that it looks good and tidy, I think :) Thank you for your contribution.

doc/tokens.md
derTobsch: done.

PR #2050, opened 2022-07-01
ref(sdp/LocalSdpMunger) add back else and comments from previous version

NewEraCracker:
Adds back some code that was removed by d742dff746a076cc5247e65ad403ffc0ff4762e2 and re-added in a different fashion later.

Still testing the implications of this in our environment. But if someone has comments about this changes, I'd appreciate.

### DIFF
--- a/modules/sdp/LocalSdpMunger.js
+++ b/modules/sdp/LocalSdpMunger.js
@@ -246,6 +246,9 @@ export default class LocalSdpMunger {
             return;
         }
 
+        // If the msid attribute is missing, then remove the ssrc from the transformed description so that a
+        // source-remove is signaled to Jicofo. This happens when the direction of the transceiver (or m-line)
+        // is set to 'inactive' or 'recvonly' on Firefox, Chrome (unified) and Safari.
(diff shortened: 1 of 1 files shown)


jallamsetty1:
@NewEraCracker  What are you attempting to fix in this PR ?

NewEraCracker:
You can see the diff without whitespace changes here: https://github.com/jitsi/lib-jitsi-meet/pull/2050/files?w=1

`mediaSection.ssrcs` is set to `undefined` in certain situations[1] but below it is used inside the `for` block[2]

[1] https://github.com/jitsi/lib-jitsi-meet/blob/28aab9fcf88ba136fcb367122b5db0a5b0c11070/modules/sdp/LocalSdpMunger.js#L267
[2] https://github.com/jitsi/lib-jitsi-meet/blob/28aab9fcf88ba136fcb367122b5db0a5b0c11070/modules/sdp/LocalSdpMunger.js#L276

The `else` was there in the previous version of code likely to protect from this corner case.

Just something that caught my eye.

My two cents.

Issue #2048, opened 2022-06-29
Multistream: When sharing desktop, camera blur causes camera video to freeze 


DanielMcAssey:
*This Issue tracker is only for reporting bugs and tracking code related issues.*

Before posting, please make sure you check community.jitsi.org to see if the same or similar bugs have already been discussed. General questions, installation help, and feature requests can also be posted to community.jitsi.org.

## Description
---
This was tested on https://beta.meet.jit.si/ to test multi-stream.

When sharing a camera with a blurred background, when you start then stop desktop sharing, then turn off camera blur, it causes the camera feed to freeze on the remote side and wont recover, even if the camera is turned off and on again.

## Current behavior
---
When sharing a desktop with a camera with a blurred background, starting then stopping the desktop sharing, then turning off the blur causes the camera stream to freeze on the remote side

## Expected Behavior
---
When sharing a desktop stream with a camera with a blurred background, starting then stopping the desktop share should keep the camera stream and not freeze

## Possible Solution
---
Not sure, but I believe it has to do with setEffect, when setEffect is called, it seems to create a 3rd video track that appears on the remote side, so 3 sources in total appear, {{participantId}}-v0 (Camera), {{participantId}}-v1 (Desktop), {{participantId}}-v2 (Camera stream with effect?).

## Steps to reproduce
---
0. 2 participants join the same room with camera enabled with both with blurred background
1. Participant A shares their screen
2. After a bit of time, Participant A stops sharing their screen
3. Participant A turns off blur
4. Participant B sees a frozen camera frame

# Environment details
---
https://beta.meet.jit.si/

saghul:
Thanks for the report! Ping @WillLiang918 ?

DanielMcAssey:
Thank you! That sorted it

PR #2054, opened 2022-07-05
fix(RTC) Use mute/unmute track operation for effects.

jallamsetty1:
Since the track is only temporarily removed and added back, it can be treated like mute & unmute operation. Fixes https://github.com/jitsi/lib-jitsi-meet/issues/2048. Remove a TODO thats no longer needed, track replace operation don't force renegotiations anymore.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -227,7 +227,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // TPC and JingleSessionPC which would contain the queue and would notify the signaling layer when local SSRCs
         // are changed. This would help to separate XMPP from the RTC module.
         return new Promise((resolve, reject) => {
-            this.conference._addLocalTrackAsUnmute(this)
+            this.conference._addLocalTrackToPc(this)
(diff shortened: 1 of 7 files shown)


modules/RTC/JitsiLocalTrack.js
```diff
@@ -863,15 +863,14 @@ export default class JitsiLocalTrack extends JitsiTrack {
 
         this._setEffectInProgress = true;
 
-        // TODO: Create new JingleSessionPC method for replacing a stream in JitsiLocalTrack without offer answer.
-        return conference.removeTrack(this)
+        return conference._removeLocalTrackAsMute(this)
```
paweldomas, line 866: This is not ideal to mix mute operation into the effects flow. What does need to happen for the effect operation? Can we add a new flow for the effect?

jallamsetty1: Like discussed, I am going to rename the remove/add methods to something generic since both the mute/unmute operation and effect have the exact same flow, i.e., remove or add the track from/to RTCPeerConnection but not from/to the TraceablePeerConnection.

paweldomas (approved):

PR #2075, opened 2022-07-22
feat(ts) add hand-crafted types for red

NewEraCracker:


saghul:
Can you drop the auto ones? Looks like they landed: https://github.com/jitsi/lib-jitsi-meet/blob/master/types/auto/modules/red/red.d.ts

saghul (approved):

PR #2078, opened 2022-07-25
Fix renegotiations for p2p

jallamsetty1:
ref(JingleSessionPC): Remove unncessary reneg cycles for p2p.

When an answer is received from the p2p peer, only set remote description instead of calling the full renegotiation cycle. Also, always initiate a sRD->cA->sLD cycle when adding remote sources irrespective of whether the local endpoint is an initiator or responder for the p2p connection. This lets us remove the hacks added for Chromium (unified-plan) bugs related to video not rendering in p2p cases.

feat(red) Add missing type def for RFC2198Encoder.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -978,25 +978,19 @@ export default class JingleSessionPC extends JingleSession {
         return this.peerconnection.getConfiguredVideoCodec();
     }
 
-    /* eslint-disable max-params */
     /**
-     * Accepts incoming Jingle 'session-initiate' and should send
-     * 'session-accept' in result.
-     * @param jingleOffer jQuery selector pointing to the jingle element of
-     * the offer IQ
-     * @param success callback called when we accept incoming session
-     * successfully and receive RESULT packet to 'session-accept' sent.
-     * @param failure function(error) called if for any reason we fail to accept
-     * the incoming offer. 'error' argument can be used to log some details
-     * about the error.
-     * @param {Array<JitsiLocalTrack>} [localTracks] the optional list of
-     * the local tracks that will be added, before the offer/answer cycle
-     * executes. We allow the localTracks to optionally be passed in so that
-     * the addition of the local tracks and the processing of the initial offer
-     * can all be done atomically. We want to make sure that any other
-     * operations which originate in the XMPP Jingle messages related with
-     * this session to be executed with an assumption that the initial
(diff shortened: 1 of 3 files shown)


types/auto/modules/red/red.d.ts
```diff
@@ -0,0 +1,31 @@
+/**
+ * An encoder for RFC 2198 redundancy using WebRTC Insertable Streams.
+ */
+export class RFC2198Encoder {
```
saghul, line 4: Why is this part of this PR?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1877,21 +1868,15 @@ export default class JingleSessionPC extends JingleSession {
...
             const remoteDescription = new RTCSessionDescription({
                 type: 'offer',
                 sdp: newRemoteSdp.raw
             });
-            const promise = isAdd && this.usesUnifiedPlan && this.isP2P && browser.isChromiumBased()
-                ? this._responderRenegotiate(remoteDescription)
-                : this._renegotiate(newRemoteSdp.raw);
 
-            promise.then(() => {
+            // Always initiate a sRD->cA->sLD cycle when a remote source is added or removed irrespective of whether
```
saghul, line 1876: Is this safe to add right now? Note that mobile 2 mobile will still use p2p in plan B mode.

saghul (approved):
LGTM with a question.

modules/xmpp/JingleSessionPC.js
jallamsetty1: Yes it, it should have always been implemented this way. Without this change, on p2p offerer, we call createOffer, setLocalDescription and then send out the offer to the peer. When the response is received, we again run the same cycle of cO->sLD->sRD. So we are calling createOffer (on receiving answer) when the peerconnection state is "have-local-offer" which is not correct. Plan-b is more forgiving and we don't see the rendering issues there but unified-plan is more strict. I don't see issues doing it this way for plan-b p2p.

types/auto/modules/red/red.d.ts
jallamsetty1: These were generated when I ran the build, didn't want gh to complain that the repo is not clean. This is going to be a separate commit anyways, so didn't see any harm in adding it to this PR.

saghul: 👍

PR #2079, opened 2022-07-26
fix(safari-ios) Fix audio being lost on safari ios after phone call

horymury:


modules/browser/BrowserCapabilities.js
```diff
@@ -200,11 +200,7 @@ export default class BrowserCapabilities extends BrowserDetection {
      */
     supportsReceiverStats() {
         return typeof window.RTCRtpReceiver !== 'undefined'
-            && Object.keys(RTCRtpReceiver.prototype).indexOf('getSynchronizationSources') > -1
-
-            // Disable this on Safari because it is reporting 0.000001 as the audio levels for all
-            // remote audio tracks.
-            && !this.isWebKitBased();
```
saghul, line 207: Why is this change necessary?

JitsiMeetJS.js
```diff
@@ -340,6 +340,17 @@ export default _mergeNamespaceAndModule({
                                 () => {
                                     Statistics.stopLocalStats(mStream);
                                 });
+                            track.addEventListener(
```
saghul, line 343: Maybe this could be embedded in statistics since `startLocalStats`, since it gets a reference to the stream? Maybe we could  pass the track instead. Then I think you'd avoid having to create new public APIs which are a bit weird on their own.

modules/browser/BrowserCapabilities.js
horymury: undid the change.

JitsiMeetJS.js
horymury: done, also moved the existing eventlistener inside `startLocalStats`.

modules/statistics/statistics.js
```diff
@@ -238,10 +239,42 @@ Statistics.prototype.startRemoteStats = function(peerconnection) {
...
+    track.addEventListener(
+        JitsiTrackEvents.NO_DATA_FROM_SOURCE,
+
+        /**
+         * Closes AudioContext on no audio data, and enables it on data received again.
+         *
+         * @param {boolean} value - Whether we receive audio data or not.
+         */
+        async value => {
+
```
saghul, line 258: style: unnecessary nnewline

modules/statistics/statistics.js
```diff
@@ -238,10 +239,42 @@ Statistics.prototype.startRemoteStats = function(peerconnection) {
...
+            if (value) {
+                for (let i = 0; i < Statistics.localStats.length; i++) {
```
saghul, line 260: for..of since we are moving code around please.

modules/statistics/LocalStatsCollector.js
```diff
@@ -142,5 +143,45 @@ LocalStatsCollector.prototype.stop = function() {
...
+    // XXX Not all browsers define a suspend method on AudioContext. As the
+    // invocation is at the (ES6 module) global execution level, it breaks the
+    // loading of the lib-jitsi-meet library in such browsers and, consequently,
+    // the loading of the very Web app that uses the lib-jitsi-meet library. For
+    // example, Google Chrome 40 on Android does not define the method but we
+    // still want to be able to load the lib-jitsi-meet library there and
+    // display a page which notifies the user that the Web app is not supported
+    // there.
+    if (context.suspend) {
+        context.suspend();
```
saghul, line 180: For symmetry, and to avoid surprises, we should `await` here. Also, AFAICT suspend has been supported long enough that we don't need this check. See: https://caniuse.com/mdn-api_audiocontext_suspend

modules/statistics/LocalStatsCollector.js
```diff
@@ -142,5 +143,45 @@ LocalStatsCollector.prototype.stop = function() {
...
+    }
+};
+
+/**
+ * Initialize the audio context on startup.
+ */
+LocalStatsCollector.connectAudioContext();
```
saghul, line 187: Since this would now return a promise (see above) we should store it in some `initContext` variable start awaits on, to make sure it's initialized when we start.

modules/statistics/statistics.js
```diff
@@ -238,10 +239,42 @@ Statistics.prototype.startRemoteStats = function(peerconnection) {
...
+                    Statistics.localStats[i].stop();
+                }
+
+                await LocalStats.disconnectAudioContext();
+            } else {
+                LocalStats.connectAudioContext();
+                for (let i = 0; i < Statistics.localStats.length; i++) {
```
saghul, line 267: ditto, for..of

saghul (changes requested):
Much cleaner now! Left some more comments, PTAL.

modules/statistics/LocalStatsCollector.js
horymury: I had it with await initially, but it seem to never resolve the promise for some reason  as the for after connectAudioContext() call was not executing

saghul: Can you check if there any reported bugs? If so, please put a comment.

modules/statistics/statistics.js
```diff
@@ -238,10 +239,41 @@ Statistics.prototype.startRemoteStats = function(peerconnection) {
...
 
-Statistics.startLocalStats = function(stream, callback) {
+Statistics.startLocalStats = function(track, callback) {
+    track.addEventListener(
+        JitsiTrackEvents.LOCAL_TRACK_STOPPED,
+        () => {
+            Statistics.stopLocalStats(track);
+        });
+
+    track.addEventListener(
```
hristoterezov, line 249: I understand that you need to kill the context and that's why we need to stop all localStats instances but the part I don't like is that there is only one localStats instance related to the track that triggers the no data from source event. Anyway I don't think there is anything you can do about it... and also I'm not sure if we will actually have a scenario with multiple local audio tracks.  

Could you please add a comment in the code with explanation what we are actually fixing here? 

Also since this is an issue only with mobile Safari could you please add the listener only for mobile Safari?

modules/statistics/LocalStatsCollector.js
```diff
@@ -142,5 +143,43 @@ LocalStatsCollector.prototype.stop = function() {
...
+
+    // XXX Not all browsers define a suspend method on AudioContext. As the
+    // invocation is at the (ES6 module) global execution level, it breaks the
+    // loading of the lib-jitsi-meet library in such browsers and, consequently,
+    // the loading of the very Web app that uses the lib-jitsi-meet library. For
+    // example, Google Chrome 40 on Android does not define the method but we
+    // still want to be able to load the lib-jitsi-meet library there and
+    // display a page which notifies the user that the Web app is not supported
+    // there.
+    context.suspend();
```
hristoterezov, line 179: Why we don't check if context.suspend is available?

The comment above is related to this check, so if we don't need the check anymore maybe we should remove the comment as well?

modules/statistics/statistics.js
```diff
@@ -238,10 +239,41 @@ Statistics.prototype.startRemoteStats = function(peerconnection) {
 
 Statistics.localStats = [];
 
-Statistics.startLocalStats = function(stream, callback) {
+Statistics.startLocalStats = function(track, callback) {
+    track.addEventListener(
```
hristoterezov, line 243: Do we need to add the listeners if audio levels are disabled?

modules/statistics/LocalStatsCollector.js
horymury: Ok so the AudioContext is in `suspended` state right after the instance is created, that probably explains why the await context.suspend() did not resolve. For testing I added a state change listener on the context, and at some point it did go to `running` state (probably when `oncomplete` event triggers)
 
So there are 2 choices here:
1. Call `await context.suspend()` when state first changes to `running`  ( or on the `oncomplete` event handler- have to check if that event can be used)
2. Leave it as it is without awaiting- as from my test it does do suspend - as the context does not get to `running` state anymore -  and add the `JitsiTrackEvents.NO_DATA_FROM_SOURCE` listener only for safari ios - this will basically keep the current logic for all other browser, as existing code does not await for suspend  call either.


@saghul  WDYT?

modules/statistics/LocalStatsCollector.js
horymury: Discussed this with @saghul  and he pointed out that as of now all major browsers support `suspend` so no check is needed.

modules/statistics/statistics.js
horymury: yep, because the AudioContext is always created.

modules/statistics/LocalStatsCollector.js
```diff
@@ -142,5 +143,43 @@ LocalStatsCollector.prototype.stop = function() {
...
+ */
+LocalStatsCollector.connectAudioContext = function() {
+    if (!LocalStatsCollector.isLocalStatsSupported()) {
+        return;
+    }
+
+    logger.info('Connecting audio context');
+    context = new AudioContext();
+
+    // XXX Not all browsers define a suspend method on AudioContext. As the
```
saghul, line 171: can you drop this comment then?

modules/statistics/LocalStatsCollector.js
saghul: Discussed that the promise won't resolve until resume is called, so we're dumping it.

modules/statistics/statistics.js
saghul: > Also since this is an issue only with mobile Safari could you please add the listener only for mobile Safari?

Good point, we should do that. For any iOS browser actually.

Chrome on Android doesn't seem to emit the muted events for local tracks so it doesn't apply there...

modules/statistics/LocalStatsCollector.js
saghul: LEt's go with option 2 then.

modules/statistics/statistics.js
horymury: Did for any iOS browser

modules/statistics/LocalStatsCollector.js
horymury: removed the whole comment,

modules/statistics/LocalStatsCollector.js
horymury: I did go with option 2 in the end, as I was not able to make option 1 work no matter what I tried. Since I added the condition to add the listener for audio muted only on iOS browsers, the behavior for the rest of OS's is just as before.

saghul (approved):
LGTM, great work @horymury !

hristoterezov (approved):

PR #2082, opened 2022-07-28
feat(JitsiConference) add ability to set the leave reason

saghul:
Ref: https://xmpp.org/extensions/xep-0045.html#exit example 83.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -627,9 +627,10 @@ JitsiConference.prototype.isP2PTestModeEnabled = function() {
 
 /**
  * Leaves the conference.
+ * @param reason {string|undefined} The reason for leaving the conference.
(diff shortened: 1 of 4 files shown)


abora8x8 (dismissed):

DanielMcAssey:
Should this also update the event that's fired when a participant leaves and include the reason?

https://github.com/jitsi/lib-jitsi-meet/blob/2fc3c454d0aee95f4593c9d217436ae7ac923000/modules/xmpp/ChatRoom.js#L1057

saghul:
> Should this also update the event that's fired when a participant leaves and include the reason?

Good point, probably yeah, I'll take a look.

damencho (approved):

PR #2084, opened 2022-08-01
fix(moderation): media not working

hristoterezov:
When the participant has joined after A/V moderation has been enabled
they are auto muted with the startmuted node in the session-initiate
packet. We were not marking those as muted by the focus and therefore on
unmute we were not sending the <muted>false</muted> IQ to jicofo.
The result was that the bridge wasn't forwarding the media to the
remote participants.

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -692,6 +692,14 @@ JitsiConferenceEventManager.prototype.setupXMPPListeners = function() {
             conference.startAudioMuted = audioMuted;
             conference.startVideoMuted = videoMuted;
 
+            if (audioMuted) {
+                conference.isMutedByFocus = true;
+            }
+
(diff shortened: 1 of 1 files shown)


damencho (approved):

saghul:
Do we need to have something meaningful in the actor field?

damencho:
> Do we need to have something meaningful in the actor field?

No, jicofo checks the from field of the iq.

saghul:
> > Do we need to have something meaningful in the actor field?
> 
> No, jicofo checks the from field of the iq.

LJM does check for it when firing some mute events, I suppose we don't need it?

damencho:
This is only for information purposes when someone is muting you.
This is jicofo muting the bridge channel. In the same way, as moderators can mute someone,  you can unmute yourself after being muted by jicofo.

PR #2090, opened 2022-08-09
fix(package.json) Replace 'module' field with 'export' in package.json

michalsnik:
This PR updates package.json and replaces currently used 'module' field with a recommended by NodeJS 'export' field. [According to documentation](https://nodejs.org/dist/latest-v16.x/docs/api/packages.html#dual-commonjses-module-packages) the 'module' field is ignored by NodeJS and only used by bundlers. After changing it however everything seems to be building fine.

### The reasoning behind this change:
- Currently when installing the package via npm using a link to a released archive, for example: https://github.com/jitsi/lib-jitsi-meet/releases/download/v1486.0.0+73bc6ba2/lib-jitsi-meet.tgz importing the lib throws a build error as it can't locate the module, given 'module' field being ignored by NodeJS and there is no index.js in the root directory.

With this change it's possible to import the package without any hiccups:
```js
import JitsiMeetJS from 'lib-jitsi-meet';
```

I noticed that in jitsi-meet app you're importing the package like that already, but I'm not entirely sure why it works for you, perhaps you got some extra configuration that resolves modules by looking at that 'module' field 🤔

### DIFF
--- a/types/auto/modules/RTC/TraceablePeerConnection.d.ts
+++ b/types/auto/modules/RTC/TraceablePeerConnection.d.ts
@@ -23,7 +23,7 @@
  *
  * @constructor
  */
-export default function TraceablePeerConnection(rtc: RTC, id: number, signalingLayer: any, pcConfig: object, constraints: object, isP2P: boolean, options: {
+export default function TraceablePeerConnection(rtc: any, id: number, signalingLayer: any, pcConfig: object, constraints: object, isP2P: boolean, options: {
     disableSimulcast: boolean;
     disableRtx: boolean;
     disabledCodec: string;
@@ -57,7 +57,7 @@ export default class TraceablePeerConnection {
      *
      * @constructor
      */
-    constructor(rtc: RTC, id: number, signalingLayer: any, pcConfig: object, constraints: object, isP2P: boolean, options: {
+    constructor(rtc: any, id: number, signalingLayer: any, pcConfig: object, constraints: object, isP2P: boolean, options: {
         disableSimulcast: boolean;
         disableRtx: boolean;
         disabledCodec: string;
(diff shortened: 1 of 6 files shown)


saghul:
Have you verified the mobile app also builds ok?

michalsnik:
I just run the app on iOS and build via metro went just fine.

What's interesting though, it shouldn't build due to Metro not supporting "exports" fields yet (https://github.com/facebook/metro/issues/670), however conducting more experiments I noticed that metro actually takes the package from the "browser" field, not "modules". Even in log messages it seems like they treat "browser" as "main" field (if we specify wrong package in "browser" field - the error from metro tells that field "main" points to a non-existent package 🤷)

Btw. when you work on the lib, how do you link it to the mobile app locally? Metro doesn't seem to follow symlinks so can't do typical `npm link ...`, or am I missing something?

michalsnik:
Replacing 'browser' with 'react-native' and pointing to ESM module works fine, though it throws one warning about circular dependency. Do you want me to fix it as part of this PR too?

I didn't get that warning using current version of the lib, so I'm thinking that RN currently uses the UMD version of the lib, not the ESM.

Simulator Screen Shot - iPod touch (7th generation) - 2022-08-09 at 23 26 44

saghul:
Hum, I was convinced we were using the ESM build, I may be wrong.

Yeah we should get the cycle fixed if that's the only one.

types/auto/modules/RTC/TraceablePeerConnection.d.ts
```diff
@@ -23,7 +23,7 @@
  *
  * @constructor
  */
-export default function TraceablePeerConnection(rtc: RTC, id: number, signalingLayer: any, pcConfig: object, constraints: object, isP2P: boolean, options: {
+export default function TraceablePeerConnection(rtc: any, id: number, signalingLayer: any, pcConfig: object, constraints: object, isP2P: boolean, options: {
```
michalsnik, line 26: Removing RTC import from `TraceablePeerConnection.js` results in missing type here.. not sure if we can do something about that right now, thoughts?

michalsnik:
I fixed the circular dependency in latest commit @saghul

PR #2091, opened 2022-08-09
fix(electron):Respect desktopSharingSources option

hristoterezov:


jallamsetty1 (dismissed):

jallamsetty1:
> Do we need it? Why not hardcode it to screen and window?

Because electron client passes  different options (only screen) for remote control and that is not being honored right now.

saghul (approved):

PR #2092, opened 2022-08-10
Use conventional syntax for depending on git repos as deps

PsychoLlama:
I am adding this as a dependency to a project managed by Yarn v2, which is a bit more persnickety regarding dependency syntax. It can't infer how to resolve the package.

```diff
-https://git@github.com
+git+https://github.com
```

This change should be functionally identical to the current behavior, but now it plays nicely with Yarn v1, v2, and npm.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -11,7 +11,7 @@
       "dependencies": {
         "@jitsi/js-utils": "2.0.0",
         "@jitsi/logger": "2.0.0",
-        "@jitsi/sdp-interop": "https://git@github.com/jitsi/sdp-interop#3d49eb4aa26863a3f8d32d7581cdb4321244266b",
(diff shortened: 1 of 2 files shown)


NewEraCracker:
This is not possible as this change was implemented as a workaround for known npm bug.

See:
https://github.com/npm/cli/issues/2610#issuecomment-914610832
and https://github.com/jitsi/lib-jitsi-meet/commit/21337c75ae228cf33ed7d655a07f98fdb3a4baad

saghul:
But npm 8.52 solved it, so we should be fine now I think.

PsychoLlama:
I should have the CLA signed hopefully next week. I'm just waiting on internal permission.

PsychoLlama:
ICLA signed ✅

PsychoLlama:
^^^ Rebased to master and manually reinstalled with Yarn, npm v6.14.17, npm v7.7.6, and npm v8.19.2. Unless there's some funky caching, it seems even npm v7 is tolerant of the syntax.

saghul (approved):

PR #2102, opened 2022-08-29
Fix of case when findLine returns false

NewEraCracker:
Replaces #1442

See discussion at: https://github.com/jitsi/lib-jitsi-meet/pull/1442/files


saghul (approved):

saghul:

PR #2104, opened 2022-09-02
fix(build) add missed jquery slim imports

NewEraCracker:
Looks like these have slipped away. Most likely due to the header being slightly different `/* global $ */` vs `/* globals $ */`


damencho (approved):

damencho:
Jenkins test this please

saghul (approved):

saghul:
Thank you!

PR #2105, opened 2022-09-05
Fix ProxyConnectionPC, use a custom signaling layer

saghul:
The existing SignalingLayerImpl is too intertwined with ChatRoom, and it has become too hard to use for ProxyConnectionPC.

Thus I ventured into creating a custom implementation that has the absolute bare minimum for that connection to work. WDYT @jallamsetty1 ?

While I was there I noticed the interface was outdated since the new functionality for multi-stream requires some implementation of certain methods, so I added them there.

Generally speaking, I'm not happy at all with this, and if we get to work on Spot I'd love to try and replace this with something that doesn't need to involve LJM internals so much. Oh well.

### DIFF
--- a/modules/proxyconnection/CustomSignalingLayer.js
+++ b/modules/proxyconnection/CustomSignalingLayer.js
@@ -1,0 +1,105 @@
+import { getLogger } from '@jitsi/logger';
+
+import SignalingLayer from '../../service/RTC/SignalingLayer';
+
+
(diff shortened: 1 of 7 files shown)


saghul:
FWIW: I jerry-rigged a Spot setup pointing to alpha through the localhost proxy used with `make dev` and I got it tested, so it works!

jallamsetty1 (approved):

PR #2113, opened 2022-09-19
fix(e2ee) discards frames that cannot be decrypted

tmoldovan8x8:


modules/e2ee/Context.js
```diff
@@ -253,11 +249,9 @@ export class Context {
             newUint8.set(new Uint8Array(plainText), frameHeader.byteLength);
 
             encodedFrame.data = newData;
-        } catch (error) {
-            if (this._sharedKey) {
```
saghul, line 257: Why was this check for shared key performed here?

saghul (dismissed):
LGTM with a question.

modules/e2ee/Context.js
tmoldovan8x8: I think cause the rotation should be handled externally in sharedKey mode.

saghul: Then shouldn't we just return in case of a shared key? We don't want to ratchet in that case.

tmoldovan8x8: Yeah, true that. We should not return the encoded frame, but still return from the method.

saghul (approved):

PR #2114, opened 2022-09-19
Prepare to export LJM types

saghul:


modules/connectivity/NetworkInfo.js
```diff
@@ -26,7 +26,6 @@ export class NetworkInfo extends Listenable {
 
     /**
      * Updates the network info state.
-     * @param {boolean} isOnline - {@code true} if internet is online or {@code false} otherwise.
```
robertpin, line 29: Why was this removed?

saghul: It's incorrectly defined as a Boolean not an object containing a Boolean. TS guesses it right.

robertpin (approved):

PR #2115, opened 2022-09-19
fix(TPC): Ignore stopped transceivers on FF during new track addition.

jallamsetty1:
Firefox lists stopped transceivers when getTranceivers is called, ignore these when picking a trancceiver for a screenshare track. Fixes an issue when starting screenshare fails if there are stopped tranceivers in the peerconnection (i.e., if some remote users with sources have left the call).


saghul:
We currently do list stopped transceivers in RN and didn't implement the stopped property because it's deprecated: https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpTransceiver/stopped

Do we use it elsewhere?

jallamsetty1:
Nope, we don't use it anywhere else. When ssrc-rewriting gets enabled, this will not be an issue anymore since the remote tracks will never get removed (on leave), they will get re-mapped.

PR #2123, opened 2022-09-29
Don't pollute the global scope in the ESM build

saghul:


package.json
```diff
@@ -62,10 +62,13 @@
     "webpack-cli": "4.9.0"
   },
   "scripts": {
-    "build": "export LIB_JITSI_MEET_COMMIT_HASH=$(git rev-parse --short HEAD 2>/dev/null) && webpack && tsc --build --clean && tsc",
+    "build": "npm run build:webpack && npm run build:tsc",
+    "build:webpack": "LIB_JITSI_MEET_COMMIT_HASH=$(git rev-parse --short HEAD 2>/dev/null) webpack",
```
hristoterezov, line 66: Is this called somewhere?

hristoterezov (approved):
Left a question. But LGTM!

package.json
saghul: By npm run build. The idea of separating them is to be able to add extra steps to each or to build just UMD or ESM if you need for testing.

hristoterezov: Oh yes! I'm just blind :) Please discard this comment!

Issue #2127, opened 2022-10-05
recording icon not leaving on mute


maxirozay:
## Description
The browser recording icon next to the site title stays after muting the audio.

## Current behavior
Turn off your microphone and the recording icon stays like you were still being recorded.

## Expected Behavior
Turn off your microphone and the icon leaves so you know that you're not being recorded.

## Possible Solution
In [JitsiLocalTrack.js](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/JitsiLocalTrack.js#L376) if the track is audio it only set enabled to false `this.track.enabled = !muted;` else it calls `_removeStreamFromConferenceAsMute`.

I tried to change the _setMuted function from `if (this.isAudioTrack()` to `if (this._streamEffect` so if there' an effect it would still use the effect `setMuted` function else it would mute the audio the same way it does for video. It works but then when the unmute part of the function handle only video so I think it could be easily fixed but I don't know if there's a good reason for not really turning the audio off.

## Steps to reproduce
Go to a room on meet.jit.si and turn off your microphone and camera. Look at the recording icon next to the tab title.

# Environment details
All

saghul:
Hey there!

When you mute the mic is not technically closed. No audio is sent to the other side (it's disabled) but it's not closed, for a number of reasons:

- Opening it takes time, so unmuting would be a much more costly operation
- We cannot give user hints such as "are you trying to speak?"

That indicator is technically correct because the mic is still open, even though no data is being transmitted.

maxirozay:
Thank you, I understand I'll close this issue but before I have another question/remark related to this issue.

If my video has an effect I have to call `await video.setEffect()` before `dispose()`. If I don't call and wait `setEffect` the browser icon stays.

The [dispose](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/JitsiLocalTrack.js#L600) function call setEffect but maybe it should wait for `setEffect` to finish before calling `stopStream` no ?

alimtunc:
Hello, I've got the same issue today. When you mute the track, browser should said that we don't send data anymore, no? Like jitsi client is working today. Muting mic or disabling video will turn off the camera is recording indicator. But today with jitsi-low-level, the camera indicator is turned on every time.

I've got also an issue when leaving the room with a muted track keeping the camera indicator on.. I think there some troubleshoot with this today. Am I the only one getting these issues?

maxirozay:
[...]

PR #2130, opened 2022-10-07
fix(RTC) fix stop track with effect (#2127)

maxirozay:
When a track has an effect, `setEffect` is finished after `stopStream` which leave the original track not stopped.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.js
+++ b/modules/RTC/JitsiLocalTrack.js
@@ -592,19 +592,16 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * @extends JitsiTrack#dispose
      * @returns {Promise}
      */
-    dispose() {
-        let promise = Promise.resolve();
+    async dispose() {
 
         // Remove the effect instead of stopping it so that the original stream is restored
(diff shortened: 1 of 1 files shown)


maxirozay:
I tried to run the test but got the following error
```
karma-webpack does not currently support customized filenames via
webpack output.filename, if this is something you need consider opening an issue.
defaulting [name].js to [name].js.
Webpack bundling...
07 10 2022 12:10:10.097:ERROR [karma-server]: Error during file loading or preprocessing
Error: error:0308010C:digital envelope routines::unsupported
```

NewEraCracker:
Looks like the CI system broke.

Anyway I would recommend that you refactor this to ignore promise failures as that seems to have been the original intention of the previous commit touching these lines: https://github.com/jitsi/lib-jitsi-meet/commit/ac88ed98b59824ff4fe2022a8c54ba4f50c740e7

I'd add some catch blocks with `logger.error` or `logger.warn` to be safe.

Just my two cents.

saghul (approved):
LGTM

saghul:

PR #2133, opened 2022-10-17
Fix codec negotiation and other multi-stream issues

jallamsetty1:
Please look at individual commits for description.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2626,6 +2626,11 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
     }
 
     if (FeatureFlags.isSourceNameSignalingEnabled()) {
+        const sourceName = localVideoTrack.getSourceName();
+
+        if (this._senderMaxHeights.get(sourceName) === frameHeight) {
+            return Promise.resolve();
+        }
         this._senderMaxHeights.set(localVideoTrack.getSourceName(), frameHeight);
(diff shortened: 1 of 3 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -2240,15 +2240,25 @@ TraceablePeerConnection.prototype._adjustRemoteMediaDirection = function(remoteD
...
-            ? MediaDirection.SENDRECV
-            : hasLocalSource
-                ? MediaDirection.RECVONLY
-                : hasRemoteSource ? MediaDirection.SENDONLY : MediaDirection.INACTIVE;
+        const media = transformer.selectMedia(mediaType);
+        const localSources = this.getLocalTracks(mediaType).length;
+        const remoteSources = this.getRemoteTracks(null, mediaType).length;
+
+        media.forEach((mLine, idx) => {
+            if (localSources && localSources === remoteSources) {
```
hristoterezov, line 2248: I am lost what we are doing here with this change. Could you please explain? Maybe adding a comment will be good?

jallamsetty1: Before, we were setting the direction only on the first audio and video m-line. Now, we are setting on each of the m-lines based on how many local and remote sources are present.

hristoterezov (commented):
Generally LGTM! I don't understand part of the PR and left a question! Also added a proposal for a small change.

hristoterezov (approved):

PR #2135, opened 2022-10-18
Remove support for callstats.

bgrozev:
- fix: Always set recroding status=on when a jibri is present
- Remove support for callstats.


JitsiConnection.js
```diff
@@ -43,12 +43,6 @@ export default function JitsiConnection(appID, token, options) {
                     ANALYTICS_CONNECTION_DISCONNECTED,
                     { message: msg });
             }
-            Statistics.sendLog(
```
saghul, line 46: Weird this was duplicated?!

saghul (commented):
Left some quick comments!

PR #2138, opened 2022-10-20
feat(ScreenObtainer) Allow seamless switching of tab capture.

jallamsetty1:
Allow users to seamlessly switch which tab they are sharing without having to stop the current share and select a new tab again. This is supported on Chrome 107 onwards.
A new button "Share this tab instead" will be shown when the user switches to a tab that is not being shared.

Screen Shot 2022-10-20 at 2 42 27 PM

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -221,6 +221,11 @@ const ScreenObtainer = {
             && this.options?.testing?.setScreenSharingResolutionConstraints;
         let video = {};
 
+        // Allow users to seamlessly switch which tab they are sharing without having to select the tab again.
+        if (browser.isChromiumBased() && browser.isVersionGreaterThan(106)) {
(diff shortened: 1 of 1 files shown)


damencho (approved):

saghul:
How are we creating the local recording track? Because it shouldn't have this...

jallamsetty1:
> How are we creating the local recording track? Because it shouldn't have this...

You mean how are we capturing the tab ? Using getDisplayMedia. Sorry, I don't understand your question.

saghul:
Are we using gDM directly or via this API? If it's the latter that would unfortunate but I think we do call gDM directly.

jallamsetty1:
We used gDM directly but the explainer for this feature says
`Chrome will try to guess what’s best for your application. This guess may change between milestones. `
I am explicitly setting the surfaceSwitching constraint to 'include' instead of letting Chrome decide.

PR #2140, opened 2022-10-27
fix(chore) typos and jsdoc

NewEraCracker:
Some things I've noticed that could be better.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -3110,7 +3110,7 @@ TraceablePeerConnection.prototype._processLocalSSRCsMap = function(ssrcMap) {
 
 /**
  * Track the SSRCs seen so far.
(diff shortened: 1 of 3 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1059,7 +1059,7 @@ TraceablePeerConnection.prototype._createRemoteTrack = function(
             + 'deleting the existing track');
         const existingTrack = Array.from(userTracksByMediaType)[0];
 
-        // The existing track needs to be removed here. This happens on Safari sometimes when an SSRC is removed from
+        // The existing track needs to be removed here. This happens on Safari sometimes when a SSRC is removed from
```
saghul, line 1062: Just to be pedantic, I think "an SSRC" is correct, because even though it's a consonant, it sounds like a vowel: "an esesarcee"

saghul (approved):

PR #2146, opened 2022-11-02
Add e2ee sas

tmoldovan8x8:


modules/e2ee/E2EEncryption.js
```diff
@@ -71,4 +71,15 @@ export class E2EEncryption {
     setEncryptionKey(keyInfo) {
         this._keyHandler.setKey(keyInfo);
     }
+
+    startVerification(participant) {
```
saghul, line 75: Let's encapsulate all verification functionality in an object like so:

```
    get sasVerification() {
        this._keyHandler.sasVerification;
    }
```

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +544,172 @@ export class OlmAdapter extends Listenable {
...
+            if (olmData.session) {
+                if (olmData.sas) {
+                    logger.warn(`SAS already created for participant ${pId}`);
+
+                    return;
+                }
+
+                const { uuid } = msg.data;
+
+                // Send ACCEPT.
```
saghul, line 557: needs to include the commitment

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +544,172 @@ export class OlmAdapter extends Listenable {
...
+                this._sendError(participant, 'No session found while processing sas-init');
+            }
+            break;
+        }
+        case OLM_MESSAGE_TYPES.SAS_ACCEPT: {
+            if (olmData.session) {
+                const { commitment, uuid } = msg.data;
+                olmData.sasCommitment = commitment;
+
+                olmData.sas = new Olm.SAS()
```
saghul, line 581: no need to create this, or don't create it early in start.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +544,172 @@ export class OlmAdapter extends Listenable {
...
+                const pub_key = olmData.sas.get_pubkey();
+
+                // Send KEY.
+                const ack = {
+                    [JITSI_MEET_MUC_TYPE]: OLM_MESSAGE_TYPE,
+                    olm: {
+                        type: OLM_MESSAGE_TYPES.SAS_KEY,
+                        data: {
+                            key: pub_key,
+                            isInitializer: true,
```
saghul, line 591: instead of encoding this here each participant should be aware of their own state

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +544,172 @@ export class OlmAdapter extends Listenable {
...
+                logger.debug(`Received sas accept message from ${pId} but we have no session for them!`);
+
+                this._sendError(participant, 'No session found while processing sas-accept');
+            }
+            break;
+        }
+        case OLM_MESSAGE_TYPES.SAS_KEY: {
+            if (olmData.session) {
+                if (!olmData.sas) {
+                    olmData.sas = new Olm.SAS();
```
saghul, line 608: there  should always be a sas created at this point or something is completely wrong

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +544,172 @@ export class OlmAdapter extends Listenable {
...
+
+                const { key: theirKey, isInitializer, uuid } = msg.data;
+
+                if (olmData.sasCommitment) {
+                    const olmUtil = new Olm.Utility();
+                    const commitment = olmUtil.sha256(theirKey + olmData.startContent);
+                    olmUtil.free();
+
+                    if (olmData.sasCommitment !== commitment) {
+                        throw new Error('OlmAdapter commitments mismatched');
```
saghul, line 625: emit event instead?

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +544,172 @@ export class OlmAdapter extends Listenable {
...
+                    }
+                }
+
+                olmData.sas.set_their_key(theirKey);
+
+                const pub_key = olmData.sas.get_pubkey();
```
saghul, line 631: we need to avoid sending another key from Alice if the already received Bob's key.

modules/e2ee/OlmAdapter.js
```diff
@@ -636,6 +876,46 @@ export class OlmAdapter extends Listenable {
...
         return d;
     }
+    
+    /**
+     * Builds and sends the SAS MAC message to the given participant.
+     */
+    _sendSasMac(participant) {
+        const pId = participant.getId();
+        const olmData = this._getParticipantOlmData(participant);
+        const uuid = uuidv4();
```
saghul, line 886: shouldn't this be the UUID that started the whole verification process?

modules/e2ee/OlmAdapter.js
```diff
@@ -636,6 +876,46 @@ export class OlmAdapter extends Listenable {
...
+    
+        // Calculate and send MAC with the keys to be verified.
+        const mac = {};
+        const keyList = [];
+        const baseInfo = `${OLM_KEY_VERIFICATION_MAC_INFO}${this.myId}${pId}${uuid}`;
+
+        const deviceKeyId = `ed25519:${uuid}`;
```
saghul, line 893: this should be my participant ID

modules/e2ee/OlmAdapter.js
```diff
@@ -636,6 +876,46 @@ export class OlmAdapter extends Listenable {
...
+
+        mac[deviceKeyId] = olmData.sas.calculate_mac(
+            this._key,
```
saghul, line 896: This needs to be our own ed25519 key (the public part).

saghul: Let's also rename "this._key" to avoid confusion.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +544,172 @@ export class OlmAdapter extends Listenable {
...
+
+                if (keysMac !== keys) {
+                    logger.error('SAS verification error: keys MAC mismatch');
+                    this.eventEmitter.emit(OlmAdapterEvents.PARTICIPANT_SAS_COMPLETED, pId, false);
+                    return;
+                }
+
+                for (const [ keyInfo, computedMac ] of Object.entries(mac)) {
+                    const keyType = keyInfo.split(':')[0];
+                    const pubKey = olmData.lastKey;
```
saghul, line 684: This should be the target participant's ed25519 key which you need to learn it from presence.

modules/e2ee/E2EEncryption.js
```diff
@@ -71,4 +71,25 @@ export class E2EEncryption {
...
+    }
+
+    /**
+     * Marks the channel as verified
+     *
+     * @param {boolean} isVerified - whether the verification was succesfull.
+     * @param {Participant} - participant to be verified.
+     * @returns {void}
+     */
+    markChannelVerified(isVerified, participant) {
```
saghul, line 92: See the previous comment.

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -167,6 +184,28 @@ export class ManagedKeyHandler extends KeyHandler {
...
+    }
+
+    /**
+     * Handles the SAS completed event.
+     *
+     * @param {string} pId - The participant ID.
+     * @param {boolean} success - Wheter the verification was succesfull.
+     * @private
+     */
+    _onParticipantSasCompleted(pId, success) {
```
saghul, line 205: The event should just say the participant is verified with success or not, but SAS completed sounds too low level here.

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -167,6 +184,28 @@ export class ManagedKeyHandler extends KeyHandler {
...
 
+    /**
+     * Handles the SAS ready event.
+     *
+     * @param {string} pId - The participant ID.
+     * @param {Uint8Array} sas - The bytes from sas.generate_bytes..
+     * @private
+     */
+    _onParticipantSasReady(pId, sas) {
+        this.conference.eventEmitter.emit(JitsiConferenceEvents.E2EE_SAS_READY, pId, sas);
```
saghul, line 195: Let's call this `E2EE_VERIFICATION_READY`

modules/e2ee/OlmAdapter.js
```diff
@@ -222,29 +260,69 @@ export class OlmAdapter extends Listenable {
...
+        const pId = participant.getId();
+        const olmData = this._getParticipantOlmData(participant);
 
+        if (!olmData.session) {
+            logger.warn(`Tried to start verification with participant ${pId} but we have no session`);
+
+            return;
+        }
+
+        olmData.sas = new Olm.SAS();
```
saghul, line 291: The `sas`, `uuid` and `startContent` belong to the process of "SAS verification" so let;s group them in a `sasVerification` object.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +548,175 @@ export class OlmAdapter extends Listenable {
...
         }
+        case OLM_MESSAGE_TYPES.SAS_START: {
+            if (olmData.session) {
+                if (olmData.sas) {
+                    logger.warn(`SAS already created for participant ${pId}`);
+
+                    return;
+                }
+
+                olmData.sas = new Olm.SAS();
```
saghul, line 559: Same thing, let's group them.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +548,175 @@ export class OlmAdapter extends Listenable {
...
+
+                this._sendError(participant, 'No session found while processing sas-init');
+            }
+            break;
+        }
+        case OLM_MESSAGE_TYPES.SAS_ACCEPT: {
+            if (olmData.session) {
+                const { commitment, uuid } = msg.data;
+
+                olmData.sasCommitment = commitment;
```
saghul, line 595: Group.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +548,175 @@ export class OlmAdapter extends Listenable {
...
+
+                const { key: theirKey, isInitializer, uuid } = msg.data;
+
+                if (olmData.sasCommitment) {
+                    const olmUtil = new Olm.Utility();
+                    const commitment = olmUtil.sha256(theirKey + olmData.startContent);
+
+                    olmUtil.free();
+
+                    if (olmData.sasCommitment !== commitment) {
```
saghul, line 636: If this happens we should destroy the SAS verification state and return.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +548,175 @@ export class OlmAdapter extends Listenable {
...
+                        this._sendError(participant, 'OlmAdapter commitments mismatched');
+                    }
+                }
+
+                olmData.sas.set_their_key(theirKey);
```
saghul, line 641: Can we keep a state variable to not rely on the above check tio stop sending keys indefinitely.

modules/e2ee/OlmAdapter.js
```diff
@@ -636,6 +887,46 @@ export class OlmAdapter extends Listenable {
 
         return d;
     }
+
+    /**
+     * Builds and sends the SAS MAC message to the given participant.
+     */
+    _sendSasMac(participant) {
+        const pId = participant.getId();
+        const olmData = this._getParticipantOlmData(participant);
```
saghul, line 896: we should check if this is ready no?

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +548,175 @@ export class OlmAdapter extends Listenable {
...
+
+                for (const [ keyInfo, computedMac ] of Object.entries(mac)) {
+                    const ourComputedMac = olmData.sas.calculate_mac(
+                        olmData.ed25519,
+                        baseInfo + keyInfo
+                    );
+
+                    if (computedMac !== ourComputedMac) {
+                        logger.error('SAS verification error: MAC mismatch');
+                        this.eventEmitter.emit(OlmAdapterEvents.PARTICIPANT_SAS_COMPLETED, pId, false);
```
saghul, line 705: we should probably include a  "reason" field in case of an error

saghul (changes requested):

saghul:
One more thing: pl add comments with references to the matrix spec, for reference.

JitsiConference.js
```diff
@@ -4031,7 +4020,7 @@ JitsiConference.prototype.markParticipantVerified = function(participantId) {
         return;
     }
 
-    this._e2eEncryption.markVerified(participant);
+    this._e2eEncryption.markParticipantVerified(isVerified, participant);
```
saghul, line 4024: You need to define this parameter. Also, make it the second one.

modules/e2ee/E2EEncryption.js
```diff
@@ -89,7 +89,7 @@ export class E2EEncryption {
      * @param {Participant} - participant to be verified.
      * @returns {void}
      */
-    markChannelVerified(isVerified, participant) {
-        this._keyHandler.sasVerification?.markChannelVerified(isVerified, participant);
+    markParticipantVerified(isVerified, participant) {
```
saghul, line 92: Please reverse this parameters, we usually put them that way, like when sending messages.

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -202,8 +202,8 @@ export class ManagedKeyHandler extends KeyHandler {
      * @param {boolean} success - Wheter the verification was succesfull.
      * @private
      */
-    _onParticipantSasCompleted(pId, success) {
-        this.conference.eventEmitter.emit(JitsiConferenceEvents.E2EE_SAS_COMPLETED, pId, success);
+    _onParticipantVerificationCompleted(pId, success) {
```
saghul: We should add the failure reason...

saghul (changes requested):

JitsiConference.js
saghul: Ignore the "define" part, I reviewed a partial state it seems...

modules/e2ee/OlmAdapter.js
```diff
@@ -207,6 +226,27 @@ export class OlmAdapter extends Listenable {
         }
     }
 
+    /**
+     * Sends sacMac if channel verification waas successful.
+     *
+     */
+    markParticipantVerified(isVerified, participant) {
+        if (isVerified) {
```
saghul, line 234: Else, we need to destroy everything we have regarding this SAS, don't we? Also we need to support trying again I think.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +572,199 @@ export class OlmAdapter extends Listenable {
...
+                if (olmData.sasVerification?.sas) {
+                    logger.warn(`SAS already created for participant ${pId}`);
+
+                    return;
+                }
+
+                const { uuid } = msg.data;
+
+                olmData.sasVerification.sas = new Olm.SAS();
+                olmData.sasVerification.uuid = uuid;
```
saghul, line 586: Should we set isInitiator to false here?

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +572,199 @@ export class OlmAdapter extends Listenable {
...
+
+                /* The first phase of the verification process, the Key agreement phase
+                   https://spec.matrix.org/latest/client-server-api/#short-authentication-string-sas-verification
+                */
+                const ack = {
+                    [JITSI_MEET_MUC_TYPE]: OLM_MESSAGE_TYPE,
+                    olm: {
+                        type: OLM_MESSAGE_TYPES.SAS_ACCEPT,
+                        data: {
+                            uuid,
```
saghul, line 602: Can we rename this uuid to `transaction_id` everywhere in the SAS part? It better matches the Matrix spec.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +572,199 @@ export class OlmAdapter extends Listenable {
...
+                }
+
+                const { uuid } = msg.data;
+
+                olmData.sasVerification.sas = new Olm.SAS();
+                olmData.sasVerification.uuid = uuid;
+
+                const pubKey = olmData.sasVerification.sas.get_pubkey();
+                const olmUtil = new Olm.Utility();
+                const commitment = olmUtil.sha256(pubKey + msg.data);
```
saghul, line 590: When computing the commitment we need to JSON encode the data so it's a string. I suggest currently the commitment looks like `'lalala[object Object]'`

saghul: Since we need to do it twice, maybe it would be a good idea to create a `_computeCommitment(pubKey: string, data: object) ` helper that takes care of this.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +572,199 @@ export class OlmAdapter extends Listenable {
...
+                const sasBytes = sas.generate_bytes(info, OLM_SAS_NUM_BYTES);
+                const generatedSas = generateSas(sasBytes);
+
+                this.eventEmitter.emit(OlmAdapterEvents.PARTICIPANT_SAS_READY, pId, generatedSas);
+
+                if (keySent) {
+                    return;
+                }
+
+                const ack = {
```
saghul, line 690: Can you pl give this auxiliary variables some more meaningful name?

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +572,199 @@ export class OlmAdapter extends Listenable {
...
+                    baseInfo + OLM_KEY_VERIFICATION_MAC_KEY_IDS
+                );
+
+                if (keysMac !== keys) {
+                    logger.error('SAS verification error: keys MAC mismatch');
+                    this.eventEmitter.emit(
+                        OlmAdapterEvents.PARTICIPANT_VERIFICATION_COMPLETED,
+                        pId,
+                        false,
+                        'keys MAC mismatch');
```
saghul, line 736: let's create constants for this please, so the app can then render nice(r) messages.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +572,199 @@ export class OlmAdapter extends Listenable {
...
+                        baseInfo + keyInfo
+                    );
+
+                    if (computedMac !== ourComputedMac) {
+                        logger.error('SAS verification error: MAC mismatch');
+                        this.eventEmitter.emit(
+                            OlmAdapterEvents.PARTICIPANT_VERIFICATION_COMPLETED,
+                            pId,
+                            false,
+                            'MAC mismatch');
```
saghul, line 753: Ditto

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +572,199 @@ export class OlmAdapter extends Listenable {
...
+
+                        return;
+                    }
+                }
+
+                logger.info(`SAS MAC verified for participant ${pId}`);
+                this.eventEmitter.emit(OlmAdapterEvents.PARTICIPANT_VERIFICATION_COMPLETED, pId, true);
+            } else {
+                logger.debug(`Received sas mac message from ${pId} but we have no session for them!`);
```
saghul, line 762: Reverse the check, so the majority of the function has one level on indentation less

modules/e2ee/OlmAdapter.js
```diff
@@ -533,6 +829,12 @@ export class OlmAdapter extends Listenable {
                     this._sendMessage(data, participantId);
                 }
             }
+            break;
+        case 'e2ee.idKey.ed25519':
+            olmData.sasVerification = olmData.sasVerification ?? {};
+
+            olmData.sasVerification.ed25519 = newValue;
```
saghul, line 836: Here we should probably emit some "verification possible / available / supported" event.

saghul (changes requested):
Discard the rpevious review, it was weird. I made a complete round now.

modules/e2ee/OlmAdapter.js
```diff
@@ -966,19 +972,31 @@ export class OlmAdapter extends Listenable {
...
-        this._sendMessage(data, pId);
+        this._sendMessage(macMessage, pId);
+    }
+
+    /**
+     * Computes the commitment.
+     */
+    _computeCommitment(pubKey, data) {
+        const olmUtil = new Olm.Utility();
+        const commitment = olmUtil.sha256(pubKey + safeJsonParse(data));
```
saghul, line 995: It should be JSON.stringify rather than safeJsonParse.

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -36,6 +36,18 @@ export class ManagedKeyHandler extends KeyHandler {
             OlmAdapter.events.PARTICIPANT_KEY_UPDATED,
             this._onParticipantKeyUpdated.bind(this));
 
+        this._olmAdapter.on(
+            OlmAdapter.events.PARTICIPANT_SAS_READY,
+            this._onParticipantSasReady.bind(this));
+
+        this._olmAdapter.on(
+                OlmAdapter.events.PARTICIPANT_SAS_KEY_AVAILABLE,
```
saghul, line 44: nit: indentation seems one level too deep here

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -36,6 +36,18 @@ export class ManagedKeyHandler extends KeyHandler {
...
+                this._onParticipantSasKeyAvailable.bind(this));
+
+        this._olmAdapter.on(
+                OlmAdapter.events.PARTICIPANT_VERIFICATION_COMPLETED,
```
saghul, line 48: ditto

modules/e2ee/ManagedKeyHandler.js
```diff
@@ -167,6 +188,39 @@ export class ManagedKeyHandler extends KeyHandler {
...
+        this.conference.eventEmitter.emit(JitsiConferenceEvents.E2EE_VERIFICATION_READY, pId, sas);
+    }
+
+    /**
+     * Handles the sas available event.
+     *
+     * @param {string} pId - The participant ID.
+     * @private
+     */
+    _onParticipantSasKeyAvailable(pId) {
```
saghul, line 208: nit: the name reads weird because it says "sas key", but it's in fact the fingerprint key. So let's just say "sas available"

modules/e2ee/OlmAdapter.js
```diff
@@ -19,17 +21,30 @@ const OLM_MESSAGE_TYPES = {
...
+const OLM_KEY_VERIFICATION_MAC_INFO = 'Jitsi-KEY_VERIFICATION_MAC';
+const OLM_KEY_VERIFICATION_MAC_KEY_IDS = 'Jitsi-KEY_IDS';
+
 const kOlmData = Symbol('OlmData');
 
 const OlmAdapterEvents = {
-    OLM_ID_KEY_READY: 'olm.id_key_ready',
     PARTICIPANT_E2EE_CHANNEL_READY: 'olm.participant_e2ee_channel_ready',
-    PARTICIPANT_KEY_UPDATED: 'olm.partitipant_key_updated'
+    PARTICIPANT_SAS_KEY_AVAILABLE: 'olm.participant_sas_key_available',
```
saghul, line 39: ditto, let's just say SAS available

modules/e2ee/OlmAdapter.js
```diff
@@ -19,17 +21,30 @@ const OLM_MESSAGE_TYPES = {
...
+    PARTICIPANT_SAS_READY: 'olm.participant_sas_ready',
+    PARTICIPANT_KEY_UPDATED: 'olm.partitipant_key_updated',
+    PARTICIPANT_VERIFICATION_COMPLETED: 'olm.participant_verification_completed'
 };
 
+const keysMacMismatchedError = 'keys MAC mismatch';
```
saghul, line 45: these need to be exporsded somewhere for the app to use and be something like "e2ee.sas.keys-mac-mismatch" so it can be translated easily

modules/e2ee/OlmAdapter.js
```diff
@@ -19,17 +21,30 @@ const OLM_MESSAGE_TYPES = {
...
+const macMismatchedError = 'MAC mismatch';
```
saghul, line 46: ditto

modules/e2ee/OlmAdapter.js
```diff
@@ -207,6 +230,27 @@ export class OlmAdapter extends Listenable {
...
+            const { sas, sasMacSent } = olmData.sasVerification;
+
+            if (sas && sas.is_their_key_set() && !sasMacSent) {
+                this._sendSasMac(participant);
+
+                // Mark the MAC as sent so we don't send it multiple times.
+                olmData.sasVerification.sasMacSent = true;
+            }
+        } else {
+            olmData.sasVerification = undefined;
```
saghul, line 250: When would we create this object again? Also, we should likely log an error / warning, no?

modules/e2ee/OlmAdapter.js
```diff
@@ -222,29 +266,93 @@ export class OlmAdapter extends Listenable {
...
+        const pId = participant.getId();
+        const olmData = this._getParticipantOlmData(participant);
+
+        if (!olmData.session) {
+            logger.warn(`Tried to start verification with participant ${pId} but we have no session`);
+
+            return;
+        }
+
+        olmData.sasVerification = {
```
saghul, line 318: Ah here is where you create it, great! Now we should check if it already exists, not to re-create it again...

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +578,201 @@ export class OlmAdapter extends Listenable {
             }
             break;
         }
+        case OLM_MESSAGE_TYPES.SAS_START: {
+            if (olmData.session) {
```
saghul, line 582: return early here pl, to avoid having one extra level of indentation

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +578,201 @@ export class OlmAdapter extends Listenable {
...
+                this._sendMessage(acceptMessage, pId);
+            } else {
+                logger.debug(`Received sas init message from ${pId} but we have no session for them!`);
+
+                this._sendError(participant, 'No session found while processing sas-init');
+            }
+            break;
+        }
+        case OLM_MESSAGE_TYPES.SAS_ACCEPT: {
+            if (olmData.session) {
```
saghul, line 625: ditto

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +578,201 @@ export class OlmAdapter extends Listenable {
...
+                const { commitment, transactionId } = msg.data;
+
+                olmData.sasVerification.sasCommitment = commitment;
```
saghul, line 628: We should check if we already received the commitment and bail

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +578,201 @@ export class OlmAdapter extends Listenable {
...
+                olmData.sasVerification.keySent = true;
+            } else {
+                logger.debug(`Received sas accept message from ${pId} but we have no session for them!`);
+
+                this._sendError(participant, 'No session found while processing sas-accept');
+            }
+            break;
+        }
+        case OLM_MESSAGE_TYPES.SAS_KEY: {
+            if (olmData.session) {
```
saghul, line 655: ditto

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +578,201 @@ export class OlmAdapter extends Listenable {
...
+                    return;
+                }
+
+                const { key: theirKey, transactionId } = msg.data;
+
+                if (sasCommitment) {
+                    const commitment = this._computeCommitment(theirKey, startContent);
+
+                    if (sasCommitment !== commitment) {
+                        this._sendError(participant, 'OlmAdapter commitments mismatched');
```
saghul, line 670: rather than a generic error we should send the sas completed event with an error indicator

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +578,201 @@ export class OlmAdapter extends Listenable {
...
+
+                        return;
+                    }
+                }
+
+                sas.set_their_key(theirKey);
+
+                const pubKey = sas.get_pubkey();
+
+                const myInfo = `${this.myId}|${pubKey}`;
```
saghul, line 681: `this.myId()`

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +578,201 @@ export class OlmAdapter extends Listenable {
...
+        case OLM_MESSAGE_TYPES.SAS_MAC: {
+            if (!olmData.session) {
+                logger.debug(`Received sas mac message from ${pId} but we have no session for them!`);
+
+                this._sendError(participant, 'No session found while processing sas-mac');
+
+                return;
+            }
+
+            const { keys, mac, transactionId } = msg.data;
```
saghul, line 725: shouldn't we also check if the sas object has both keys set at this point?

modules/e2ee/OlmAdapter.js
saghul: Nevermind, it's a getter...

saghul (changes requested):
Veeeery close! Let's get this in tomorrow!

modules/e2ee/OlmAdapter.js
tmoldovan8x8: Actually, if the user startsVerification multiple times it should be fine to recreate this object, no?

modules/e2ee/OlmAdapter.js
tmoldovan8x8: Actually, aren't this just an implementation detail that should not concern the app. The important thing would be that the verification has failed, not the commitment, mac or any other internal detail?

saghul: Yes, but it's good to provide some details. Not sure if we'll want to surface them in the UI in the end, we'll see, but since this is the low level library, having it there makes sense IMHO.

modules/e2ee/OlmAdapter.js
saghul: But it will all be messed up because there are 2 transaction ids floating around. I'd say the process cannot be started unless the previous one has ended. Just like when creating the session.

modules/e2ee/OlmAdapter.js
```diff
@@ -207,6 +228,27 @@ export class OlmAdapter extends Listenable {
...
+            const { sas, sasMacSent } = olmData.sasVerification;
+
+            if (sas && sas.is_their_key_set() && !sasMacSent) {
+                this._sendSasMac(participant);
+
+                // Mark the MAC as sent so we don't send it multiple times.
+                olmData.sasVerification.sasMacSent = true;
+            }
+        } else {
+            olmData.sasVerification = undefined;
```
saghul, line 248: we need to send a notification here that the participant is completed and not verified.

modules/e2ee/OlmAdapter.js
```diff
@@ -207,6 +228,27 @@ export class OlmAdapter extends Listenable {
...
+     * Sends sacMac if channel verification waas successful.
+     *
+     */
+    markParticipantVerified(participant, isVerified) {
+        const olmData = this._getParticipantOlmData(participant);
+
+        if (isVerified) {
+            const { sas, sasMacSent } = olmData.sasVerification;
+
+            if (sas && sas.is_their_key_set() && !sasMacSent) {
```
saghul, line 241: Log a warning at least in the else?

modules/e2ee/OlmAdapter.js
```diff
@@ -207,6 +228,27 @@ export class OlmAdapter extends Listenable {
...
 
+    /**
+     * Sends sacMac if channel verification waas successful.
+     *
+     */
+    markParticipantVerified(participant, isVerified) {
+        const olmData = this._getParticipantOlmData(participant);
+
+        if (isVerified) {
+            const { sas, sasMacSent } = olmData.sasVerification;
```
saghul, line 239: check if this object exists before processing and error out otherwise

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +582,232 @@ export class OlmAdapter extends Listenable {
...
+                logger.debug(`Received sas accept message from ${pId} but we have no session for them!`);
+
+                this._sendError(participant, 'No session found while processing sas-accept');
+
+                return;
+            }
+
+            const { commitment, transactionId } = msg.data;
+
+            if (olmData.sasVerification.sasCommitment) {
```
saghul, line 641: This will fail if we get an out of the blue message like this one.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +582,232 @@ export class OlmAdapter extends Listenable {
...
+        case OLM_MESSAGE_TYPES.SAS_KEY: {
+            if (!olmData.session) {
+                logger.debug(`Received sas key message from ${pId} but we have no session for them!`);
+
+                this._sendError(participant, 'No session found while processing sas-key');
+
+                return;
+            }
+
+            const { isInitiator, sas, sasCommitment, startContent, keySent } = olmData.sasVerification;
```
saghul, line 679: ditto, check that `sasVerification` exists.

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +582,232 @@ export class OlmAdapter extends Listenable {
...
+
+            const { keys, mac, transactionId } = msg.data;
+
+            if (!mac || !keys) {
+                logger.warn('Invalid SAS MAC message');
+
+                return;
+            }
+
+            const sas = olmData.sasVerification.sas;
```
saghul, line 756: same

saghul (changes requested):

modules/e2ee/OlmAdapter.js
```diff
@@ -470,6 +603,266 @@ export class OlmAdapter extends Listenable {
...
+                    false,
+                    E2EEErrors.E2EE_SAS_INVALID_SAS_VERIFICATION);
+
+                return;
+            }
+
+            if (olmData.sasVerification.sasCommitment) {
+                logger.debug(`Already received sas commitment message from ${pId}!`);
+
+                this._sendError(participant, 'Already received sas commitment message from ${pId}!');
```
saghul, line 682: Should we send an error here or just bail with the above log? I know we are a tad inconsistent but this would imply a programming error in the client I think...

saghul (approved):
LGTM! What's with the tests? Maybe this needs a rebase?

saghul (approved):

PR #2147, opened 2022-11-02
Remove support for legacy endpoint based signaling.

jallamsetty1:


JitsiConference.js
```diff
@@ -1705,20 +1643,6 @@ JitsiConference.prototype.setLastN = function(lastN) {
...
- * Checks if the participant given by participantId is currently included in
- * the last N.
- * @param {string} participantId the identifier of the participant we would
- * like to check.
- * @return {boolean} true if the participant with id is in the last N set or
- * if there's no last N set, false otherwise.
- * @deprecated this method should never be used to figure out the UI, but
- * {@link ParticipantConnectionStatus} should be used instead.
- */
-JitsiConference.prototype.isInLastN = function(participantId) {
```
saghul, line 1718: Why is this function removed? Can we provide a reasonable replacement?

JitsiConferenceEvents.ts
```diff
@@ -223,22 +223,6 @@ export enum JitsiConferenceEvents {
...
-     * - active - the connection is active
-     * - inactive - the connection is inactive, was intentionally interrupted by
-     * the bridge
-     * - interrupted - a network problem occurred
-     * - restoring - the connection was inactive and is restoring now
-     *
-     * The current status value can be obtained by calling
-     * JitsiParticipant.getConnectionStatus().
-     */
-    PARTICIPANT_CONN_STATUS_CHANGED = 'conference.participant_conn_status_changed',
```
saghul, line 240: Is there a replacement for this?

modules/RTC/BridgeChannel.js
```diff
@@ -391,21 +375,19 @@ export default class BridgeChannel {
...
-                            }
-                        );
-                    } else {
-                        logger.error(`Invalid SenderSourceConstraints: ${JSON.stringify(obj)}`);
-                    }
+                const { sourceName, maxHeight } = obj;
+
+                if (typeof sourceName === 'string' && typeof maxHeight === 'number') {
+                    // eslint-disable-next-line object-property-newline
+                    logger.info(`SenderSourceConstraints: ${JSON.stringify({ sourceName, maxHeight })}`);
```
saghul, line 382: Since you are building the same object literal down there to emit the event it might be worth it to create a n auxiliary object to log and emit the event with.

modules/qualitycontrol/SendVideoController.js
```diff
@@ -95,27 +83,22 @@ export default class SendVideoController {
...
-                    logger.debug(`Sender constraints for source:${sourceName} changed to maxHeight:${maxHeight}`);
-                    this._propagateSendMaxFrameHeight(sourceName);
-                }
+        const { maxHeight, sourceName } = videoConstraints;
+        const localVideoTracks = this._conference.getLocalVideoTracks() ?? [];
+
+        for (const track of localVideoTracks) {
+            // Propagate the sender constraint only if it has changed.
+            if (track.getSourceName() === sourceName
+                && (!this._sourceSenderConstraints.has(sourceName)
```
saghul, line 92: Do you really need to check with `has`? It will return `undefined` otherwise, won't it?

modules/xmpp/xmpp.js
```diff
@@ -255,14 +254,11 @@ export default class XMPP extends Listenable {
...
         // Advertise source-name signaling when the endpoint supports it.
-        if (FeatureFlags.isSourceNameSignalingEnabled()) {
-            logger.info('Source-name signaling is enabled');
-            this.caps.addFeature('http://jitsi.org/source-name');
-        }
-        if (FeatureFlags.isReceiveMultipleVideoStreamsSupported()) {
-            logger.info('Receiving multiple video streams is enabled');
-            this.caps.addFeature('http://jitsi.org/receive-multiple-video-streams');
-        }
+        logger.info('Source-name signaling is enabled');
```
saghul, line 257: Maybe it now deserves to be downgraded to debug level?

modules/xmpp/xmpp.js
```diff
@@ -255,14 +254,11 @@ export default class XMPP extends Listenable {
...
+        this.caps.addFeature('http://jitsi.org/source-name');
+
+        logger.info('Receiving multiple video streams is enabled');
```
saghul, line 260: ditto

saghul (commented):
Love me some deleted code! Left some comments (minor) PTAL!

JitsiConference.js
jallamsetty1: Jvb doesn't send the `LastNEndpointsChangeEvent` message anymore when the endpoint signals support for source-names. The equivalent function for the new signaling format is [getForwardedSources](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L1670).

saghul: 👍

JitsiConferenceEvents.ts
jallamsetty1: In the new signaling mode, connectivity is determined per source (remote track) basis and JitsiTrackEvents.TRACK_STREAMING_STATUS_CHANGED is fired.

modules/RTC/BridgeChannel.js
jallamsetty1: Well I just removed the check, was trying to keep refactors to a minimum. I don't see any harm in making the changes you requested.

saghul (approved):

PR #2149, opened 2022-11-04
fix: Do not send trickle candidates to jicofo

bgrozev:
The candidates are not necessary when we use jvb.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -518,7 +518,9 @@ export default class JingleSessionPC extends JingleSession {
                     });
                 this._gatheringReported = true;
             }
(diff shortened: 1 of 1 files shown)


damencho (approved):

saghul:
Are they not useful for network switching?

bgrozev:
> Are they not useful for network switching?

Assuming JVB is properly deployed and doesn't require hole-punching, we don't need to signal ANY of the client's candidates to JVB (it will discover them as peer reflexive candidates when it receives binding requests). So this change doesn't prevent trickle from being used, it just doesn't signal them.

The context for this change is reducing signaling for large conferences, I don't know if trickle will play a role for network switching.

saghul:
Thanks for explaining!

PR #2155, opened 2022-11-11
fix: allow disabling and setting preferred codec

WillLiang918:
### Bug Description
If both `disabledCodec` and `preferredCodec` are set, browsers (Chrome) that support VP9 are sending VP8 video streams initially and then VP9 after another user joins. The new user that joins sends VP8 until another user joins, so on and so forth. 

### Setup/configs
```
config.p2p = { enabled: false }
config.videoQuality.disabledCodec = 'H264'
config.videoQuality.preferredCodec = 'VP9'
```

### STR
- Join a conference from 2 different Chrome tabs
- chrome://webrtc-internals/ shows inbound and outbound streams as VP8 for both participants (VP9 is expected)
- Have a 3rd users join the conference from another Chrome tab
- chrome://webrtc-internals/ shows first two participants sending VP9, newly join participant sending VP8

### Observation
- In `CodecSelection.js` there are 3 events that triggers a change in preferred codec, `JitsiConferenceEvents.USER_JOINED`, `JitsiConferenceEvents.USER_LEFT`, and `JitsiConferenceEvents._MEDIA_SESSION_STARTED`. `JitsiConferenceEvents._MEDIA_SESSION_STARTED` is the only event that takes into consideration the `videoQuality.disabledCodec` config.
- However, if both `disabledCodec` and `preferredCodec` is provided, it seems we will only execute logic to disable codec. A subsequent user left or join event is required to set the preferred codec.

### Fix
Remove the `this.codecPreference.enable` field used to determine to disable or set preferred codec. Introduce `this.codecPreferenc.disabledCodecMimeType` explicitly for disabling codec to allow disabling and setting preferred codec at the same time. However, if `disabledCodec === preferredCodec` we will only disable.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -1648,7 +1648,11 @@ TraceablePeerConnection.prototype._mungeCodecOrder = function(description) {
     }
 
     for (const mLine of mLines) {
-        if (this.codecPreference.enable) {
+        if (this.codecPreference.disabledCodecMimeType) {
+            SDPUtil.stripCodec(mLine, this.codecPreference.disabledCodecMimeType);
+        }
+
+        if (this.codecPreference.mimeType !== this.codecPreference.disabledCodecMimeType) {
             SDPUtil.preferCodec(mLine, this.codecPreference.mimeType);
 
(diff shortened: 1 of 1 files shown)


jallamsetty1:
Thanks for the PR @WillLiang918!

jallamsetty1 (approved):

Issue #2037, opened 2022-06-20
Other side not seeing camera track removal


NewEraCracker:
## Description
---
We have a custom application using lib-jitsi-meet. Currently P2P mode is disabled on our configuration.

## Current behavior
---

When a new call is made and camera is activated, if during call, participant disables camera, the other participant will see a frozen picture (there is no track removal event).

## Expected Behavior
---

Track is removed.

## Possible Solution
---

IDK. But for testing we've reverted f3a2f61e3fd6ffef04dc5a7bf8534d9c0866a529 and d742dff746a076cc5247e65ad403ffc0ff4762e2 on our build. And this works again.

## Steps to reproduce
---

Make a new call with camera on.
Turn it off while there is a 2nd participant.
2nd participant sees frozen picture.

# Environment details
---

Latest lib-jitsi-meet v1456.0.0+261736ba
Chrome 102.0.5005.115

jitsi-meet-prosody/stable,now 1.0.6155-1 all [installed,upgradable to: 1.0.6260-1]
jitsi-meet-turnserver/stable,now 1.0.6155-1 all [installed,upgradable to: 1.0.6260-1]
jitsi-meet-web-config/stable,now 1.0.6155-1 all [installed,upgradable to: 1.0.6260-1]
jitsi-videobridge2/stable,now 2.1-681-g3544ed05-1 all [installed,upgradable to: 2.2-9-g8cded16e-1]
jicofo/stable,now 1.0-877-1 all [installed,upgradable to: 1.0-900-1]
prosody-0.11/unknown,now 1nightly143-1~focal amd64 [installed]

---

Should jitsi packages be updated to latest versions in our systems as an alternative fix?

Thanks.

DanielMcAssey:
Getting this too, we also run a custom application and the remote side does not get the TRACK_REMOVED event, they only receive an onmute RTC event and no further.

This is will multi-stream enabled though

NewEraCracker:
In our application we now hook the TRACK_MUTE_CHANGED and run the remove logic on mute (to hide camera) and add logic on unmute (to attach camera again to html).

So far works good. We'll be testing during next week with multiple devices to see if that does the trick.

If this new behaviour is intended going forward, please let me know and this issue can be closed.

DanielMcAssey:
@NewEraCracker not sure if its because I am using the unstable docker images, but I dont get a TRACK_MUTE_CHANGED event either, I see the RTC onmute event, but no further propagation. Will test further

[...]

PR #2161, opened 2022-11-26
fix(JitsiConference) send mute sourceInfo before removing an unmuted track

NewEraCracker:
Workarounds around issue #2037 to have other participants receive the mute event and therefore avoid frozen camera bug.

There may be a better way to fix this, so I leave to the experts to review and figure out what is causing the `sourceInfo` to not be sent on track removal without this trick (or workaround at application level).

The good news is this is far less invasive then my previous workaround in the original issue. My two cents.

Kind regards and thanks in advance.


saghul:
The approach SGTM, but I defer to the expert: @jallamsetty1

jallamsetty1 (approved):

PR #2163, opened 2022-12-01
fix(startMutedFocus):remove tracks not added to PC

hristoterezov:
Currenly those tracks are stuck in the conference and can't be removed
because of the check on replaceTrack for transciever of the old track.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -295,6 +295,11 @@ export default function JitsiConference(options) {
      * again by Jicofo.
      */
     this._videoSenderLimitReached = undefined;
+
+    /**
+     * Bind functions.
+     */
+    this._shouldUseTrackOnInit = this._shouldUseTrackOnInit.bind(this);
 }
 
 // FIXME convert JitsiConference to ES6 - ASAP !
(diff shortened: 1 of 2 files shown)


saghul (approved):

JitsiConference.js
```diff
@@ -2205,6 +2205,19 @@ JitsiConference.prototype.onIncomingCall = function(jingleSession, jingleOffer,
...
 
+/**
+ * Removes all tracks that don't match the criteria to be added to the peerconnection on init.
+ *
+ * @returns {void}
+ */
+JitsiConference.prototype._removeUnusedTracksOnInit = function() {
+    this.getLocalTracks().forEach(track => {
+        if (!this._shouldUseTrackOnInit(track)) {
+            this.removeTrack(track);
```
jallamsetty1, line 2216: Why are we trying to remove the track from the pc if it wasn't added in the first place ? This effectively results in replaceTrack(null, null) being called since the old track wasn't added to the conference. 

Can we call `onLocalTrackRemoved` or fire `JitsiConferenceEvents.TRACK_REMOVED` directly ?

hristoterezov: You say we don't need any of those:
```
if (oldTrackBelongsToConference && !oldTrack.isMuted() && !newTrack) {
                oldTrack._sendMuteStatus(true);
            }
            oldTrackBelongsToConference && this.onLocalTrackRemoved(oldTrack);
            newTrack && this._setupNewTrack(newTrack);

            // Send 'VideoTypeMessage' on the bridge channel when a video track is added/removed.
            if ((oldTrackBelongsToConference && oldTrack?.isVideoTrack()) || newTrack?.isVideoTrack()) {
                this._sendBridgeVideoTypeMessage(newTrack);
            }
            this._updateRoomPresence(this.getActiveMediaSession());
            if (newTrack !== null && (this.isMutedByFocus || this.isVideoMutedByFocus)) {
                this._fireMuteChangeEvent(newTrack);
            }
```

jallamsetty1: Well, none of those get executed since it fails `oldTrackBelongsToConference` check. The tracks were never added to the conference. So if your intention is to remove the tracks from RTC and fire the TRACK_REMOVED event, you should call `onLocalTrackRemoved` directly.

jallamsetty1: Scratch that ... the old track is added to the conference but not to the pc.

jallamsetty1 (approved):

saghul:
Is this still necessary?

PR #2171, opened 2022-12-06
fix: hosts is not defined in options, point to XMPP options

DanielMcAssey:
Fails to join a conference because, when joining a conference it cant find the hosts parameter in the ChatRoom options, because its not looking in the correct options object.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -190,12 +190,12 @@ export default class ChatRoom extends Listenable {
 
             // there is no point of sending conference iq when in visitor mode
             const preJoin
-                = this.options.disableFocus || this.options.hosts.visitorFocus
+                = this.options.disableFocus || this.xmpp.options.hosts.visitorFocus
                     ? Promise.resolve()
                     : this.moderator.allocateConferenceFocus();
 
-            if (this.options.hosts.visitorFocus) {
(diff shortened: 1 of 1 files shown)


DanielMcAssey:
Just want to confirm with @damencho that this is correct.

Because when I update to lib-jitsi-meet after commit 271cce2f623c8c4c824d3ed3d89b4ebbc6ce2ae6 it breaks joining a conference, because it cant find the hosts object inside options.

modules/xmpp/ChatRoom.js
```diff
@@ -190,12 +190,12 @@ export default class ChatRoom extends Listenable {
 
             // there is no point of sending conference iq when in visitor mode
             const preJoin
-                = this.options.disableFocus || this.options.hosts.visitorFocus
+                = this.options.disableFocus || this.xmpp.options.hosts.visitorFocus
```
damencho, line 193: I don't think that is correct. 
Can you explain what is the problem you face? 
There is issue with that code, but the fix was just merged https://github.com/jitsi/lib-jitsi-meet/commit/8acdeb1ddc30e665b3de995ace893630abda45c7

damencho: I'm guessing you were hitting the same ...

DanielMcAssey: Ah that's the exact issue I was experiencing! Thank you

PR #2178, opened 2022-12-12
Use insertable streams to drop all media when in lite mode.

JonathanLennox:
(Rather than moving streams to inactive in SDP.)

### DIFF
--- a/modules/litemode/LiteModeContext.js
+++ b/modules/litemode/LiteModeContext.js
@@ -42,7 +42,7 @@ export class LiteModeContext {
         const receiver = tpc.findReceiverForTrack(track.track);
 
         if (!receiver) {
-            logger.warn(`Could set up lite mode for ${track}: receiver not found in: ${tpc}`);
+            logger.warn(`Could not set up lite mode for ${track}: receiver not found in: ${tpc}`);
 
             return;
         }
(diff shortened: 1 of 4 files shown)


modules/litemode/LiteModeContext.js
```diff
@@ -0,0 +1,78 @@
...
+
+        const receiverStreams = receiver.createEncodedStreams();
+
+        this._setupLiteMode(receiverStreams.readable, receiverStreams.writable);
+    }
+
+    /**
+     * Setup lite mode for receiver streams
+     */
+    _setupLiteMode(readableStream, writableStream) {
```
saghul, line 63: I'd inline this function, and the next even in the above one, so it does the whole setup.

saghul (approved):
Left a suggestion, but LGTM! Question: we want to do this for all received tracks right (audio and video) ?

modules/litemode/LiteModeContext.js
JonathanLennox: Ok.

JonathanLennox:
> Left a suggestion, but LGTM! Question: we want to do this for all received tracks right (audio and video) ?

Yes, we want to turn off both audio and video decoding, which I believe is what this PR should do?

saghul (approved):

PR #2179, opened 2022-12-13
fix(BridgeChannel): prevent multiple websockets on retries

sunduev:
Multiple WebSockets could lead to client being unable to send any bridge messages due to lost `this._channel` reference.

Scenario:
1. `WebSocket #1` (`WS1`) is closed somehow by server -> start backoff retrying.
3. Spawn `WS2` in 1 sec -> `WS2` connecting -> store `WS2` in `this._channel`.
4. Spawn `WS3` in 2 sec -> `WS3` connecting -> store `WS3` in `this._channel`.
5. `WS2` connects -> stop retrying. But we have `WS3` in `this._channel`.
6. `WS3` connects moments after `WS2` -> at same moment `WS2` is closed by server (same url with `WS3`) -> `WS2` `onclose` handler resets `this._channel` to `null` and should start retries again, but `WS3` `onopen` handler stops them.
7. `WS3` survives and receives messages, but `this._channel` is `null` -> `isOpen` returns `false` -> `_send` throws error

### DIFF
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -100,6 +100,16 @@ export default class BridgeChannel {
         let timeoutS = 1;
 
         const reload = () => {
+            const isConnecting = this._channel && (this._channel.readyState === 'connecting'
+                    || this._channel.readyState === WebSocket.CONNECTING);
+
+            // Should not spawn new websockets while one is already trying to connect.
+            if (isConnecting) {
+                this._retryTimeout = setTimeout(reload, timeoutS * 1000);
+
+                return;
+            }
+
             if (this.isOpen()) {
                 return;
             }
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -422,10 +422,10 @@ export default class BridgeChannel {
         };
 
         channel.onclose = event => {
-            logger.info(`Channel closed by ${this._closedFromClient ? 'client' : 'server'}`);
+            logger.debug(`Channel closed by ${this._closedFromClient ? 'client' : 'server'}`);
 
             if (channel !== this._channel) {
-                logger.info('Skip close handler, channel instance is not equal to stored one');
+                logger.debug('Skip close handler, channel instance is not equal to stored one');
 
                 return;
             }
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -420,11 +420,13 @@ export default class BridgeChannel {
         };
 
         channel.onclose = event => {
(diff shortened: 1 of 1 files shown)


paweldomas:
It probably also makes sense to ignore `onclose` events from any websocket instance other than the current one, right?

sunduev:
> It probably also makes sense to ignore `onclose` events from any websocket instance other than the current one, right?

It makes sense and looks like it should not cause any side effects.
Adding it in second commit, thanks!

modules/RTC/BridgeChannel.js
```diff
@@ -100,6 +100,16 @@ export default class BridgeChannel {
         let timeoutS = 1;
 
         const reload = () => {
+            const isConnecting = this._channel && (this._channel.readyState === 'connecting'
+                    || this._channel.readyState === WebSocket.CONNECTING);
+
+            // Should not spawn new websockets while one is already trying to connect.
+            if (isConnecting) {
+                this._retryTimeout = setTimeout(reload, timeoutS * 1000);
```
paweldomas, line 108: I'm wondering if scheduling another timeout is really necessary? If there's a websocket that's "connecting" what will happen if the connection attempt fails? Will there be onerror or other handler triggered which would enable the timer loop again?

sunduev: Yeah, seems to be necessary.
There is flag `_areRetriesEnabled` that blocks scheduling of new retry loop until any channel is "reconnected".

damencho:
Everything is good @paweldomas

modules/RTC/BridgeChannel.js
paweldomas: can you add a comment here which explains why we still need to schedule a new retry timeout? LGTM otherwise

sunduev: done!

paweldomas (approved):

modules/RTC/BridgeChannel.js
paweldomas: thank you!

modules/RTC/BridgeChannel.js
```diff
@@ -412,6 +424,12 @@ export default class BridgeChannel {
         channel.onclose = event => {
             logger.info(`Channel closed by ${this._closedFromClient ? 'client' : 'server'}`);
 
+            if (channel !== this._channel) {
+                logger.info('Skip close handler, channel instance is not equal to stored one');
```
saghul, line 428: Can you please make this debug instead of info?

sunduev: done!

saghul (approved):

PR #2184, opened 2022-12-14
fix(SignalingLayer) Update SSRC owners on leave.

jallamsetty1:
Update the SSRC owners in the following cases:
1. When a remote endpoint leaves the call.
2. When a source-remove is received.
3. When a source is remapped (with ssrc-rewriting enabled). 

Create the remote track even if presence is not yet received. The ssrc owner check prevents the client from creating a dummy track when the call switches over from p2p to jvb after the last remote endpoint leaves the call.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -958,7 +958,9 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
                 + 'track creation failed!'));
 
         return;
-    } else if (!ownerEndpointId) {
+    }
+
+    if (!ownerEndpointId) {
         GlobalOnErrorHandler.callErrorHandler(
             new Error(`No SSRC owner known for remote stream[ssrc=${trackSsrc},id=${streamId},type=${mediaType}]`
             + 'track creation failed!'));
@@ -969,15 +971,17 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
     const sourceName = this.signalingLayer.getTrackSourceName(trackSsrc);
     const peerMediaInfo = this.signalingLayer.getPeerMediaInfo(ownerEndpointId, mediaType, sourceName);
 
-    if (!peerMediaInfo) {
-        GlobalOnErrorHandler.callErrorHandler(
-            new Error(`${this}: no sourceInfo available for ${ownerEndpointId}:${sourceName} track creation failed!`));
+    // Assume default presence state for remote source. Presence can be received after source signaling. This shouldn't
+    // prevent the endpoint from creating a remote track for the source.
+    let muted = true;
+    let videoType = VideoType.CAMERA;
 
-        return;
+    if (peerMediaInfo) {
+        muted = peerMediaInfo.muted;
+        videoType = peerMediaInfo.videoType; // can be undefined
+    } else {
+        logger.info(`${this}: no source-info available for ${ownerEndpointId}:${sourceName}, assuming default state`);
     }
-
-    const muted = peerMediaInfo.muted;
-    const videoType = peerMediaInfo.videoType; // can be undefined
 
     this._createRemoteTrack(ownerEndpointId, stream, track, mediaType, videoType, trackSsrc, muted, sourceName);
 };
@@ -1082,9 +1086,8 @@ TraceablePeerConnection.prototype._remoteTrackRemoved = function(stream, track) 
     const streamId = stream.id;
     const trackId = track?.id;
 
+    // Ignore stream removed events for JVB "mixed" sources (used for RTCP termination).
     if (!RTC.isUserStreamById(streamId)) {
-        logger.info(`${this} ignored remote 'stream removed' event for non-user stream[id=${streamId}]`);
-
         return;
     }
 
@@ -1101,8 +1104,7 @@ TraceablePeerConnection.prototype._remoteTrackRemoved = function(stream, track) 
     }
 
     const toBeRemoved = this.getRemoteTracks().find(
-        remoteTrack => remoteTrack.getStreamId() === streamId
-        && remoteTrack.getTrackId() === trackId);
+        remoteTrack => remoteTrack.getStreamId() === streamId && remoteTrack.getTrackId() === trackId);
 
     if (!toBeRemoved) {
         GlobalOnErrorHandler.callErrorHandler(new Error(`${this} remote track removal failed - track not found`));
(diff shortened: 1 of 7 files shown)


jallamsetty1:
Jenkins test this please.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -969,16 +971,18 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
     const sourceName = this.signalingLayer.getTrackSourceName(trackSsrc);
     const peerMediaInfo = this.signalingLayer.getPeerMediaInfo(ownerEndpointId, mediaType, sourceName);
 
-    if (!peerMediaInfo) {
-        GlobalOnErrorHandler.callErrorHandler(
```
hristoterezov, line 973: Why don't we use the `GlobalOnErrorHandler` anymore?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -969,16 +971,18 @@ TraceablePeerConnection.prototype._remoteTrackAdded = function(stream, track, tr
...
-            new Error(`${this}: no sourceInfo available for ${ownerEndpointId}:${sourceName} track creation failed!`));
+    // Assume default presence state for remote source. Presence can be received after source signaling. This shouldn't
+    // prevent the endpoint from creating a remote track for the source.
+    let muted = true;
+    let videoType = VideoType.CAMERA;
 
-        return;
```
hristoterezov, line 976: Could you please add more details why we were not creating the track before and why do we want to create it now?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1101,16 +1104,14 @@ TraceablePeerConnection.prototype._remoteTrackRemoved = function(stream, track)
...
-        && remoteTrack.getTrackId() === trackId);
+        remoteTrack => remoteTrack.getStreamId() === streamId && remoteTrack.getTrackId() === trackId);
 
     if (!toBeRemoved) {
         GlobalOnErrorHandler.callErrorHandler(new Error(`${this} remote track removal failed - track not found`));
 
         return;
     }
 
-    logger.info(`${this} remote track removed stream[id=${streamId},trackId=${trackId}]`);
```
hristoterezov, line 1113: Why we are removing this log? I've found it quite useful for debugging in some cases.

modules/proxyconnection/CustomSignalingLayer.js
```diff
@@ -58,6 +48,34 @@ export default class CustomSignalingLayer extends SignalingLayer {
         return this.ssrcOwners.get(ssrc);
     }
 
+    /**
+     * @inheritDoc
+     */
+    getTrackSourceName(ssrc) { // eslint-disable-line no-unused-vars
+        return undefined;
```
hristoterezov, line 55: Since `setTrackSourceName` is an empty function maybe we should do this an empty function as well.

I don't have any preferences whether it will be empty or `return undefined`. Lets be consistent.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1855,6 +1855,9 @@ export default class JingleSessionPC extends JingleSession {
                 if (track) {
                     logger.debug(`Existing SSRC ${s.ssrc}: new owner ${s.owner}. name=${s.source}`);
 
+                    // Update the SSRC owner.
+                    this._signalingLayer.setSSRCOwner(s.ssrc, s.owner);
```
hristoterezov, line 1859: Can we rename the method? I think this one stopped to be a getter a long time ago :)

Another option would be to somehow split it in 2 functions - 1 getter and 1 that actually modifies the state.

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -357,6 +354,25 @@ export default class SignalingLayerImpl extends SignalingLayer {
...
+        if (typeof ssrc !== 'number') {
+            throw new TypeError(`SSRC(${ssrc}) must be a number`);
+        }
+
+        // Now signaling layer instance is shared between different JingleSessionPC instances, so although very unlikely
+        // an SSRC conflict could potentially occur. Log a message to make debugging easier.
+        const existingName = this._sourceNames.get(ssrc);
+
+        if (existingName && existingName !== sourceName) {
+            logger.error(`SSRC(${ssrc}) sourceName re-assigned from ${existingName} to ${sourceName}`);
```
hristoterezov, line 370: Shouldn't we add the check for ssrc rewriting support?

hristoterezov (changes requested):
Left some questions and comments!

modules/RTC/TraceablePeerConnection.js
jallamsetty1: If there is no presence, we don't stop track creation anymore. We go ahead and create the remote track assuming default presence.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Before ssrc-rewriting, source signaling and presence were received on the same channel, i.e., on the strophe ws. So there is a little chance of an endpoint receiving those messages out of order. However, with ssrc-rewriting, the sender sends source signaling and presence on the same strophe ws but on the receiver, source info is received on the bridge ws and presence on the strophe ws. There is a higher probability for these messages to be received out of order.
Till now client always assumed that presence is received before source-signaling but a remote track can be created assuming default presence.

modules/proxyconnection/CustomSignalingLayer.js
jallamsetty1: Actually, it is consistent with the rest of the class. All the other methods return `undefined` for getters and since `setTrackSourceName` is a setter, it does nothing.

modules/xmpp/SignalingLayerImpl.js
jallamsetty1: Yep, it needs to be an error when ssrc-rewriting is not enabled, a debug message otherwise.

modules/proxyconnection/CustomSignalingLayer.js
jallamsetty1: BTW, I just did an alpha sort of the methods, this is how it is was before :)

modules/xmpp/JingleSessionPC.js
hristoterezov: I meant the method where this line is executed - `getNewSources`

hristoterezov: Not very relevant to the PR but here you add more set operations...

hristoterezov (approved):

PR #2185, opened 2022-12-15
fix(BridgeChannel) skip close notification if the code is 1001

saghul:
That's a graceful close by the bridge.

### DIFF
--- a/modules/RTC/BridgeChannel.js
+++ b/modules/RTC/BridgeChannel.js
@@ -430,27 +430,30 @@ export default class BridgeChannel {
                 return;
             }
 
-            if (this._mode === 'websocket') {
-                if (!this._closedFromClient) {
-                    this._retryWebSocketConnection(event);
-                }
-            }
+            // When the JVB closes the connection gracefully due to the participant being alone in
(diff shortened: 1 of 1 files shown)


damencho (dismissed):

modules/RTC/BridgeChannel.js
```diff
@@ -418,7 +418,10 @@ export default class BridgeChannel {
                 }
             }
 
-            if (!this._closedFromClient) {
+            // When the JVB closes the connection gracefully due to the participant being alone in
+            // the meeting it uses code 1001, so treat that as a graceful close and don't say
+            // anything.
+            if (!this._closedFromClient && event.code !== 1001) {
```
paweldomas, line 424: hmm and it's really okay not to emit `RTCEvents.DATA_CHANNEL_CLOSED` and not adjust `this._connected = false;` ?

paweldomas (dismissed):

modules/RTC/BridgeChannel.js
saghul: Good point Pawel! Updated, can you PTAL?

PR #2190, opened 2022-12-21
fix(codec-selection): Codec selection fixes.

jallamsetty1:
1. Checks peer's preferred codec in p2p case. Mobile and web have different preferred codecs.
2. Log an error message when the preferred codec is not offered by JVB.
3. Clean up code related to deprecated config.js settings 'preferH264' and 'disableH264'.
4. Refactor the codec selection logic so that correct codec is picked.

### DIFF
--- a/modules/RTC/CodecSelection.js
+++ b/modules/RTC/CodecSelection.js
@@ -32,24 +32,29 @@ export class CodecSelection {
     constructor(conference, options) {
         this.conference = conference;
         this.options = options;
+        this.enforcePreferredCodec = options.enforcePreferredCodec;
 
-        // VP8 cannot be disabled and it will be the default codec when no preference is set.
-        this.disabledCodec = options.disabledCodec === CodecMimeType.VP8
-            ? undefined
-            : this._getCodecMimeType(options.disabledCodec);
+        // VP8 cannot be disabled since it the default codec.
+        this.p2pDisabledCodec = options.p2pDisabledCodec !== CodecMimeType.VP8
+            && this._isCodecSupported(options.p2pDisabledCodec)
+            && options.p2pDisabledCodec;
+        this.jvbDisabledCodec = options.jvbDisabledCodec !== CodecMimeType.VP8
+            && this._isCodecSupported(options.jvbDisabledCodec)
+            && options.jvbDisabledCodec;
 
-        // Check if the codec values passed are valid.
-        const jvbCodec = this._getCodecMimeType(options.jvbCodec);
-        const p2pCodec = this._getCodecMimeType(options.p2pCodec);
+        // Determine the preferred codecs.
+        this.p2pPreferredCodec = this._isCodecSupported(options.p2pPreferredCodec)
+            && options.p2pPreferredCodec !== options.p2pDisabledCodec
+            ? options.p2pPreferredCodec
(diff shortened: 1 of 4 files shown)


modules/RTC/CodecSelection.js
```diff
@@ -32,24 +32,29 @@ export class CodecSelection {
...
-            P2P: ${this.p2pPreferredCodec}`);
-
-        if (this.jvbPreferredCodec === CodecMimeType.VP9 && !browser.supportsVP9()) {
-            this.jvbPreferredCodec = CodecMimeType.VP8;
-        }
+        this.enforcePreferredCodec = options.enforcePreferredCodec;
+
+        // VP8 cannot be disabled since it the default codec.
+        this.p2pDisabledCodec = options.p2pDisabledCodec !== CodecMimeType.VP8
+            && this._isCodecSupported(options.p2pDisabledCodec)
```
saghul, line 39: Why do we need to support the codec in order to disable it?

modules/RTC/CodecSelection.js
```diff
@@ -85,8 +75,16 @@ export class CodecSelection {
      * @private
      */
     _isCodecSupported(preferredCodec) {
+        if (!preferredCodec) {
+            return false;
+        }
+
+        if (preferredCodec === CodecMimeType.VP9 && !this.enforcePreferredCodec && !browser.supportsVP9()) {
```
saghul, line 82: Why do we check the enforcement flag for this codec specifically?

modules/RTC/CodecSelection.js
```diff
@@ -98,50 +96,43 @@ export class CodecSelection {
...
 
-                if (peerCodec
-                    && peerCodec !== currentCodec
-                    && (peerCodec !== CodecMimeType.VP9 || browser.supportsVP9())) {
-                    selectedCodec = peerCodec;
-                }
+                return peerMediaInfo?.codecType;
+            });
+
+            const nonPreferredCodecs = remoteCodecs.filter(codec => codec !== selectedCodec);
```
saghul, line 123: Should we also filter out any disabled codecs?

saghul (approved):
LGTM with some questions.

modules/RTC/CodecSelection.js
jallamsetty1: This is because our client doesn't support VP9 on Firefox/Safari though the browser supports it. There are folks in the community that use this flag to enforce VP9 selection on those browsers.

modules/RTC/CodecSelection.js
jallamsetty1: I am filtering them out in the next step but it makes sense to filter them out here. Will make the change.

modules/RTC/CodecSelection.js
jallamsetty1: I am checking it here to avoid unnecessary renegotiations. `SDPUtil.stripCodec` and` SDPUtil.preferCodec` methods don't return a value indicating if the order was actually changed and calling setVideoCodecs triggers a renegotiation automatically.

saghul (approved):

jallamsetty1:
Merging because the failed Lobby test has been flaky lately.

PR #2192, opened 2022-12-28
Feat ssrc rewriting

jallamsetty1:
In this feature:

- There will be a fixed set of remote tracks and therefore a fixed set of SSRCs.

- The bridge will re-map those SSRCs to a remote sender based on what is being requested and then the owner (source) of a remote track will change.

- The bridge does the remapping based on the receiver constraints received from the client.

- However, the receiver constraints currently are derived from the tracks in redux but with ssrc-rewriting, the tracks won't be created unless it is signaled. So it is a chicken and egg problem. Therefore, we will be constructing the receiver constraints based on the remote presence since all the source information is already signaled in presence.

There is jitsi-meet PR which does the receiver constraints part - https://github.com/jitsi/jitsi-meet/pull/12408/files#diff-ce724de6e7f229a65b2483d045183950192efdb68e35f9c5587a139609a49de6L371

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -53,7 +53,8 @@ import BridgeVideoType from './service/RTC/BridgeVideoType';
 import CodecMimeType from './service/RTC/CodecMimeType';
 import { MediaType } from './service/RTC/MediaType';
 import RTCEvents from './service/RTC/RTCEvents';
-import { getSourceNameForJitsiTrack } from './service/RTC/SignalingLayer';
+import { SignalingEvents } from './service/RTC/SignalingEvents';
+import { getMediaTypeFromSourceName, getSourceNameForJitsiTrack } from './service/RTC/SignalingLayer';
 import { VideoType } from './service/RTC/VideoType';
 import {
     ACTION_JINGLE_RESTART,
@@ -394,6 +395,17 @@ JitsiConference.prototype._init = function(options = {}) {
     );
 
     this._signalingLayer.setChatRoom(this.room);
+    this._signalingLayer.on(
+        SignalingEvents.SOURCE_UPDATED,
+        (sourceName, endpointId, muted, videoType) => {
+            const participant = this.participants.get(endpointId);
+            const mediaType = getMediaTypeFromSourceName(sourceName);
+
+            if (participant) {
+                participant._setSources(mediaType, muted, sourceName, videoType);
+                this.eventEmitter.emit(JitsiConferenceEvents.PARTICIPANT_SOURCE_UPDATED, participant);
+            }
+        });
 
     // Connection interrupted/restored listeners
     this._onIceConnectionInterrupted
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -156,7 +156,12 @@ export default function JitsiConference(options) {
     this.eventEmitter = new EventEmitter();
     this.options = options;
     this.eventManager = new JitsiConferenceEventManager(this);
-    this.participants = {};
+
+    /**
+     * List of all the participants in the conference.
+     * @type {Map<string, JitsiParticipant>};
+     */
+    this.participants = new Map();
 
     /**
      * The signaling layer instance.
@@ -1199,11 +1204,11 @@ JitsiConference.prototype._fireMuteChangeEvent = function(track) {
     if (this.mutedByFocusActor && track.isAudioTrack()) {
         const actorId = Strophe.getResourceFromJid(this.mutedByFocusActor);
 
-        actorParticipant = this.participants[actorId];
+        actorParticipant = this.participants.get(actorId);
     } else if (this.mutedVideoByFocusActor && track.isVideoTrack()) {
         const actorId = Strophe.getResourceFromJid(this.mutedVideoByFocusActor);
 
-        actorParticipant = this.participants[actorId];
+        actorParticipant = this.participants.get(actorId);
     }
 
     // Send the video type message to the bridge if the track is not removed/added to the pc as part of
@@ -1632,11 +1637,10 @@ JitsiConference.prototype.setLastN = function(lastN) {
 };
 
 /**
- * @return Array<JitsiParticipant> an array of all participants in this
- * conference.
+ * @return Array<JitsiParticipant> an array of all participants in this conference.
  */
 JitsiConference.prototype.getParticipants = function() {
-    return Object.values(this.participants);
+    return Array.from(this.participants.values());
 };
 
 /**
@@ -1662,7 +1666,7 @@ JitsiConference.prototype.getParticipantCount = function(countHidden = false) {
  * @param id the id of the participant.
  */
 JitsiConference.prototype.getParticipantById = function(id) {
-    return this.participants[id];
+    return this.participants.get(id);
 };
 
 /**
@@ -1810,7 +1814,7 @@ JitsiConference.prototype.onMemberJoined = function(
         participant._tracks.push(track);
     }
 
-    this.participants[id] = participant;
+    this.participants.set(id, participant);
     this.eventEmitter.emit(
         JitsiConferenceEvents.USER_JOINED,
         id,
@@ -1923,10 +1927,10 @@ JitsiConference.prototype.onMemberLeft = function(jid, reason) {
         });
     }
 
-    const participant = this.participants[id];
+    const participant = this.participants.get(id);
 
     if (participant) {
-        delete this.participants[id];
+        this.participants.delete(id);
         this.eventEmitter.emit(JitsiConferenceEvents.USER_LEFT, id, participant, reason);
     }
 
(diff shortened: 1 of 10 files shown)


JitsiConference.js
```diff
@@ -156,7 +157,12 @@ export default function JitsiConference(options) {
     this.eventEmitter = new EventEmitter();
     this.options = options;
     this.eventManager = new JitsiConferenceEventManager(this);
-    this.participants = {};
+
+    /**
+     * List of all the participants in the conference.
+     * @type {Map<string, JitsiParticipant>};
+     */
+    this.participants = new Map();
```
hristoterezov, line 165: Out of curiosity why did you change it to `Map`?

JitsiParticipant.js
```diff
@@ -294,6 +308,32 @@ export default class JitsiParticipant {
...
 
+    /**
+     * Sets source info.
+     * @param {MediaType} mediaType The media type, 'audio' or 'video'.
+     * @param {boolean} muted The new muted state.
+     * @param {string} sourceName The name of the source.
+     * @param {string} videoType The video type of the source.
+     * @returns {void}
+     */
+    setSources(mediaType, muted, sourceName, videoType) {
```
hristoterezov, line 319: Should we rename it to `_setSources` in order to indicate that this method is not part of the public API of LJM? WDYT?

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -180,12 +180,14 @@ export default class SignalingLayerImpl extends SignalingLayer {
                 = this._remoteSourceState[endpointId] || (this._remoteSourceState[endpointId] = {});
 
             for (const sourceName of Object.keys(sourceInfoJSON)) {
+                let presenceChanged = false;
```
hristoterezov, line 183: Isn't `sourcesChanged` more suitable name?

jallamsetty1:
> I'm wondering why do we decided to send this information trough presence and not trough the bridge channels?
> 
> Could you please give example of what do we expect to receive trough presence?
> 
> I was thinking that this feature will include:
> 
> * having a fixed set of sources per receiver.
> * reusing those sources for different participants.
> 
> Is this different feature? If it is different could you please add some details.

This information(mute state, video type) has always been sent through presence. All I am doing now is saving the source information (mute state, videoType and source name) as a member of JitsiParticipant whenever presence from remote participant is received.

In this feature:

- There will be a fixed set of remote tracks and therefore a fixed set of SSRCs.
- The bridge will re-map those SSRCs to a remote senders based on what is being requested and then the owner (source) of a remote track will change.
- The bridge does the remapping based on the receiver constraints received from the client. 
- However, the receiver constraints currently are derived from the tracks in redux but with ssrc-rewriting, the tracks won't be created unless it is signaled. So it is a chicken and egg problem. Therefore, we will be constructing the receiver constraints based on the remote presence since all the source information is already signaled in presence.
There is jitsi-meet PR which does this - https://github.com/jitsi/jitsi-meet/pull/12408/files#diff-ce724de6e7f229a65b2483d045183950192efdb68e35f9c5587a139609a49de6L371

JitsiParticipant.js
jallamsetty1: Done

modules/xmpp/SignalingLayerImpl.js
jallamsetty1: Done

JitsiConference.js
jallamsetty1: A Map made more sense for the participant list, also it is going to make the transition to TS easier. I am trying to make small incremental changes whenever I am touching a class so that its easy to make the switch later.

hristoterezov: I don't see how it could affect the transition to TS but I see how a map can be more clear as a type.

JitsiConference.js
```diff
@@ -389,6 +395,17 @@ JitsiConference.prototype._init = function(options = {}) {
     );
 
     this._signalingLayer.setChatRoom(this.room);
+    this._signalingLayer.on(
+        SignalingEvents.SOURCE_UPDATED,
+        (sourceName, endpointId, muted, videoType) => {
```
saghul, line 400: Why do we get the video type here, but for the media type we need to query it?

JitsiParticipant.js
```diff
@@ -43,6 +43,7 @@ export default class JitsiParticipant {
         this._isReplacing = isReplacing;
         this._isReplaced = isReplaced;
         this._features = new Set();
+        this._sources = new Map();
```
saghul, line 46: It would be good to have a comment here mentioning how the data is mapped.

JitsiParticipant.js
```diff
@@ -61,6 +62,32 @@ export default class JitsiParticipant {
...
+
+        if (sourceByMediaType?.size) {
+            sourceByMediaType.set(sourceName, sourceInfo);
+
+            return;
+        }
+
+        sourceByMediaType = new Map();
+        sourceByMediaType.set(sourceName, sourceInfo);
+        this._sources.set(mediaType, sourceByMediaType);
```
saghul, line 88: Am I reading this correctly, that we'll have this Map with media type (so audio, video) pointing to another map, with the source infos, indexes by source name?

Wouldn't this make it simpler?

```
this._sources = {
    audio: new Map(),
    video: new Map()
};
```

Since the keys for the outermost map are always constant, are they not?

JitsiParticipant.js
```diff
@@ -130,6 +157,19 @@ export default class JitsiParticipant {
         return this._role;
     }
 
+    /**
+     * Returns the sources for a given media type.
+     * @param {string} mediaType The type of media, 'audio' or 'video'.
+     * @returns {Map<string, Object>}
+     */
+    getSources(mediaType) {
+        if (!mediaType) {
```
saghul, line 166: When would be need this? It feels weird to have 2 types of returns...

JitsiParticipant.js
```diff
@@ -130,6 +157,19 @@ export default class JitsiParticipant {
         return this._role;
     }
 
+    /**
+     * Returns the sources for a given media type.
+     * @param {string} mediaType The type of media, 'audio' or 'video'.
+     * @returns {Map<string, Object>}
```
saghul, line 163: If the parameter is missing, the return type is different.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1793,115 +1794,88 @@ export default class JingleSessionPC extends JingleSession {
...
-                    }
+                    // Update the track with all the relevant info.
+                    track.setSourceName(source);
+                    track.setOwner(owner);
+                    if (mediaType === MediaType.VIDEO) {
+                        const type = videoType === 'CAMERA' ? VideoType.CAMERA : VideoType.DESKTOP;
 
-                    track.setSourceName(s.source);
-                    track.setOwner(s.owner);
+                        track._setVideoType(type);
```
saghul, line 1826: why not assign videoType directly?

modules/xmpp/SignalingLayerImpl.js
```diff
@@ -142,12 +142,14 @@ export default class SignalingLayerImpl extends SignalingLayer {
                 = this._remoteSourceState[endpointId] || (this._remoteSourceState[endpointId] = {});
 
             for (const sourceName of Object.keys(sourceInfoJSON)) {
+                let sourceChanged = false;
```
saghul, line 145: General comment on this file. Can you pl make sure Spot screen sharing won't break with these changes?

saghul (commented):
Left some comments, nothing major! PTAL.

JitsiConference.js
jallamsetty1: The media type can be inferred from the source-name using `getMediaTypeFromSourceName` but not the video type.

JitsiParticipant.js
jallamsetty1: JM calls both `getSources()` and `getSources('video')`.  I added the first case as an after thought. I will return the map here and the caller of the function can do further processing to get the video sources.

modules/xmpp/JingleSessionPC.js
jallamsetty1: It is a matter of preference I guess. I try to avoid it if I need to update it outside of the class and there is a function available.

JitsiParticipant.js
jallamsetty1: Yes, it does make it simpler but then also creates the need for declaring a new type for this Object here (when we transition to TS) and in [JM](https://github.com/jitsi/jitsi-meet/blob/master/react/features/base/participants/types.ts#L54).  The current approach doesn't sound too complicated for me, we do it the same way for remote tracks in TPC but do let me know if you have a strong preference.

JitsiConference.js
saghul: 👍

JitsiParticipant.js
saghul: No strong preference, and since it's an internal thing it can be changed anytime if we have a change of heart.

saghul (approved):

PR #2198, opened 2023-01-05
Update JitsiTrack

jallamsetty1:
Alpha sort the methods.
Add setSourceName/getSourceName methods to the JitsiTrack class. They are already implemented in JitsiLocalTrack and JitsiRemoteTrack.

### DIFF
--- a/modules/RTC/JitsiRemoteTrack.js
+++ b/modules/RTC/JitsiRemoteTrack.js
@@ -265,12 +265,19 @@ export default class JitsiRemoteTrack extends JitsiTrack {
      * Update the properties when the track is remapped to another source.
      *
      * @param {string} owner The endpoint ID of the new owner.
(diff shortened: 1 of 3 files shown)


hristoterezov (changes requested):
It looks like we don't actually have `setSourceName` in JitsiRemoteTrack. We have  `setNewSource` which will set the owner also (which IMHO is a little confusing). Maybe we can split this `setNewSource` into `setSourceName` and `setOwner`? WDYT?

jallamsetty1:
> It looks like we don't actually have `setSourceName` in JitsiRemoteTrack. We have `setNewSource` which will set the owner also (which IMHO is a little confusing). Maybe we can split this `setNewSource` into `setSourceName` and `setOwner`? WDYT?

Done

hristoterezov (approved):

PR #2207, opened 2023-01-20
fix: Bring back code setting externalAuthEnabled.

bgrozev:


damencho (approved):

saghul:
Why is this necessary? Can you please amend the commit message with an explanation, for posterity?

damencho:
That was deleted and broke Shibboleth, I think.

PR #2211, opened 2023-01-29
feat(Jingle+SDP): parse extmap-allow-mixed

paweldomas:


JonathanLennox:
Should this also have the reverse translation, i.e. mapping the attribute from SDP to (nonstandard) Jingle?

paweldomas:
Do we expect to receive it back in Jicofo from the client? I though it goes one way Jicofo -> client (even if expressed in both local and remote SDP some operations are client local).

JonathanLennox:
I guess the question is whether there're any cases where the JVB needs to know whether the client really supports it.  Right now we just assume that if you've negotiated any header extensions that need it (bigger than 16 bytes or IDs greater than 16) you support it, but I'm worried that that might be fragile, in which case we might actually want the JVB to know what the client has negotiated?

paweldomas:
Makes sense, I will try to adjust

paweldomas:
Closing, because another approach has landed in https://github.com/jitsi/lib-jitsi-meet/pull/2212

PR #2213, opened 2023-02-07
fix: added better types

vasav-prashar:
@hristoterezov @saghul @bgrozev can you please review this.

### DIFF
--- a/service/authentication/AuthenticationEvents.spec.ts
+++ b/service/authentication/AuthenticationEvents.spec.ts
@@ -7,7 +7,10 @@ describe( "/service/authentication/AuthenticationEvents members", () => {
         IDENTITY_UPDATED,
         AuthenticationEvents,
         ...others
-    } = exported as any; // TODO: remove cast after typescript conversion
+    } = exported as {
+        IDENTITY_UPDATED: string,
+        AuthenticationEvents: { IDENTITY_UPDATED: string },
+    } // TODO: remove cast after typescript conversion
 
     it( "known members", () => {
         expect( IDENTITY_UPDATED ).toBe( 'authentication.identity_updated' );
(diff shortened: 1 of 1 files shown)


Parth0105 (approved):

Parth0105 (dismissed):

service/authentication/AuthenticationEvents.spec.ts
```diff
@@ -7,7 +7,10 @@ describe( "/service/authentication/AuthenticationEvents members", () => {
         IDENTITY_UPDATED,
         AuthenticationEvents,
         ...others
-    } = exported as any; // TODO: remove cast after typescript conversion
```
Parth0105, line 10: @hristoterezov  @saghul  @bgrozev can you please review this.

saghul:
Sorry, what are you fixing here? "other" is still unaccounted for.

Parth0105:
> Sorry, what are you fixing here? "other" is still unaccounted for.

@saghul

Parth0105:
> The whole point of that test is to make sure we don't miss any other exports. Maybe I'm missing something...

```
{
        IDENTITY_UPDATED: string,
        AuthenticationEvents: { IDENTITY_UPDATED: string },
} & { [key: string]: unknown }
``` i guess this would be a better type then taking into account any unneeded exports.?

File: service/authentication/AuthenticationEvents.spec.ts [hunk 2]
```diff
@@ -7,7 +7,10 @@ describe( "/service/authentication/AuthenticationEvents members", () => {
...
+    } = exported as {
+        IDENTITY_UPDATED: string,
+        AuthenticationEvents: { IDENTITY_UPDATED: string },
+    } // TODO: remove cast after typescript conversion
```
NewEraCracker, line 13:
```suggestion
    } & { [key: string]: unknown }
```

NewEraCracker (changes requested):
PTAL

service/authentication/AuthenticationEvents.spec.ts
vasav-prashar: Done @NewEraCracker

NewEraCracker (approved):

PR #2216, opened 2023-02-09
fix(SDP): search for extmap at session level

nils-ohlmeier:
According to the RFC's extmap attributes are allowed at the media and the session level. This makes us look for them at the session level as well in case something starts doing that one day.

### DIFF
--- a/modules/sdp/SDP.js
+++ b/modules/sdp/SDP.js
@@ -319,7 +319,7 @@ SDP.prototype.toJingle = function(elem, thecreator) {
             this.rtcpFbToJingle(i, elem, '*');
 
             // XEP-0294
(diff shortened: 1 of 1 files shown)


JonathanLennox (approved):
LGTM, but I'd feel more comfortable with an actual front-end person reviewing it too.

I also have no idea why the Breakout Room test failed on Firefox; I assume it's unrelated?

fippo:
TIL: libwebrtc supports parsing extmap at session level [here](https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/pc/webrtc_sdp.cc;l=2294;bpv=1;bpt=1?q=webrtc_sdp.cc)

fippo (approved):
The findLine makes me cry ;-)

saghul (approved):

PR #2220, opened 2023-02-21
fix(TPCUtils) filter stopped transceivers

saghul:
When tracks are removed from a PC the transceiver will transition to a stopped state.

In the JVB PC this means we want to filter it out when setting the connection to inactive, since changing the direction of a stopped transceiver throws an exception.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -468,6 +468,10 @@ export class TPCUtils {
 
         logger.info(`${this.pc} ${active ? 'Enabling' : 'Suspending'} ${mediaType} media transfer.`);
         transceivers.forEach(transceiver => {
+            if (transceiver.stopped) {
+                return;
+            }
+
             if (active) {
                 const localTrackMids = Array.from(this.pc._localTrackTransceiverMids);
 


modules/RTC/TPCUtils.js
```diff
@@ -468,6 +468,10 @@ export class TPCUtils {
 
         logger.info(`${this.pc} ${active ? 'Enabling' : 'Suspending'} ${mediaType} media transfer.`);
         transceivers.forEach(transceiver => {
+            if (transceiver.stopped) {
```
jallamsetty1, line 471: Doesn't native-webrtc filter out stopped transceivers from the `RTCPeerConnection::getTransceivers()` output ? Browsers do that and our code assumes that. How complicated is it to fix this in native-webrtc ? We do not want `RTCPeerConnection::getTransceivers()` to return stopped transceivers because it messes up our logic for adding/replacing video streams. That is why we reject the m-lines after remote sources are removed for a jvb call here https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/xmpp/JingleSessionPC.js#L2057 which results in transceivers getting stopped (does native-webrtc implement this ?)
So we either fix native-webrtc or filter out stopped transceivers after every getTransceivers call.

saghul: I'll dig a bit into this but if I do a simple test and call transceiver.stop() then a call to getTransceivers() returns it just fine.

IIRC we added a similar stopped check for Firefox, I'll check what the spec says and see what we can do here.

FWIW this is the o my error of this sort I've ran into.

jallamsetty1: The spec is clear about `getSenders()` and `getReceivers()` in that those methods should not return the sender/receiver associated with a transceiver that is stopped. - https://www.w3.org/TR/webrtc/#dom-peerconnection-getsenders
Spec doesn't say anything about the getTransceivers() method though.

saghul: It actually does: https://w3c.github.io/webrtc-pc/#set-the-session-description (6, 12): 

```
If description is of type "[answer](https://w3c.github.io/webrtc-pc/#dom-rtcsdptype-answer)", then run the following steps:

For each transceiver in the connection's [set of transceivers](https://w3c.github.io/webrtc-pc/#transceivers-set) run the following steps:

If transceiver is [stopped](https://w3c.github.io/webrtc-pc/#dfn-stopped), [associated](https://w3c.github.io/webrtc-pc/#dfn-associated) with an m= section and the associated m= section is rejected in connection.[[[CurrentLocalDescription]]](https://w3c.github.io/webrtc-pc/#dfn-currentlocaldescription) or connection.[[[CurrentRemoteDescription]]](https://w3c.github.io/webrtc-pc/#dfn-currentremotedescription), remove the transceiver from the connection's [set of transceivers](https://w3c.github.io/webrtc-pc/#transceivers-set).
```

So I'm fixing this in RN WebRTC: https://github.com/react-native-webrtc/react-native-webrtc/pull/1330

Thank you formaking me look into this!

jallamsetty1: Thanks for pointing me to it. Does react-native-webrtc stop the transceiver when the m-line is rejected ? I checked the sRD call here https://github.com/react-native-webrtc/react-native-webrtc/blob/master/src/RTCPeerConnection.ts#L160 and also in WebRTCModule.peerConnectionSetRemoteDescription and couldn't find it.

saghul: That happens in libwebrtc itself. Then we get the transceiver updates which are applied in _updateTransceivers. It is right there that we filter them if they have been stopped.

PR #2222, opened 2023-02-24
calirfy how to generate the keys for public validation

enoch85:


saghul (approved):
LGTM @damencho can you confirm?

enoch85:
Maybe I should add that I made this yesterday and it's working as intended. So; tested and works! :)

enoch85:
OK to merge?

PR #2225, opened 2023-03-02
fix(ScreenObtainer): make electron desktop resolution constraints configurable

DanielMcAssey:
This allows the client app to decide the maximum resolution that needs to be captured using electron.

We have found machines that try desktop capture at 4k to be sluggish, and would like to limit it to 720p.

Follows the same constraints configuration format as Camera capture, with fallback to previous functionality if no constraints configured.


saghul (approved):

saghul:
LGTM, @jallamsetty1 WDYT?

jallamsetty1 (approved):

PR #2226, opened 2023-03-07
fix(TPC) Disable media instead of changing dir for p2p->jvb switch.

jallamsetty1:
Resume or suspend the media on the jvb peerconnection by changing the RTCRtpEncodingParamters.active state instead of changing the direction on the transceiver. This avoids the needs to start a O/A renegotiation cycle for these operations. The media direction will be changed only for p2p lastn=0 case since video needs to be disabled on both the sender and the peer for p2p lastn=0 case.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -61,10 +61,10 @@ export class TPCUtils {
 
         return localTrack.isVideoTrack()
             ? [ {
-                active: true,
+                active: this.pc.videoTransferActive,
                 maxBitrate: this.videoBitrates.high
             } ]
-            : [ { active: true } ];
+            : [ { active: this.pc.audioTransferActive } ];
     }
 
     /**
@@ -88,19 +88,19 @@ export class TPCUtils {
 
         return [
             {
-                active: true,
+                active: this.pc.videoTransferActive,
                 maxBitrate: browser.isFirefox() ? maxVideoBitrate : this.encodingBitrates.low,
                 rid: SIM_LAYER_1_RID,
                 scaleResolutionDownBy: browser.isFirefox() ? HD_SCALE_FACTOR : LD_SCALE_FACTOR
             },
             {
-                active: true,
+                active: this.pc.videoTransferActive,
                 maxBitrate: this.encodingBitrates.standard,
                 rid: SIM_LAYER_2_RID,
                 scaleResolutionDownBy: SD_SCALE_FACTOR
             },
             {
-                active: true,
+                active: this.pc.videoTransferActive,
                 maxBitrate: browser.isFirefox() ? this.encodingBitrates.low : maxVideoBitrate,
                 rid: SIM_LAYER_3_RID,
                 scaleResolutionDownBy: browser.isFirefox() ? LD_SCALE_FACTOR : HD_SCALE_FACTOR
@@ -353,12 +353,13 @@ export class TPCUtils {
     }
 
     /**
-     * Returns the max resolution encoded by the client for a given local video track.
+     * Returns the max resolution that the client is configured to encode for a given local video track. The actual
+     * send resolution might be downscaled based on cpu and bandwidth constraints.
      *
      * @param {JitsiLocalTrack} localVideoTrack - The local video track.
      * @returns {number} The max encoded resolution for the given video track.
      */
-    getEncodedResolution(localVideoTrack) {
+    getConfiguredEncodeResolution(localVideoTrack) {
         const localTrack = localVideoTrack.getTrack();
         const { height } = localTrack.getSettings();
         const videoSender = this.pc.findSenderForTrack(localVideoTrack.getTrack());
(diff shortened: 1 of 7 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -420,16 +449,31 @@ export class TPCUtils {
...
+
+            if (parameters?.encodings?.length) {
+                for (const encoding of parameters.encodings) {
+                    encoding.active = enable;
+                }
+            }
+            promises.push(sender.setParameters(parameters));
+        }
+
+        return Promise.all(promises);
```
saghul, line 476: Is that what we want? This will resolve with the first failing promise too. I'd say we want `allSettled`.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2588,13 +2550,16 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
...
-
     const sourceName = localVideoTrack.getSourceName();
 
-    if (this._senderMaxHeights.get(sourceName) === frameHeight) {
+    this._senderMaxHeights.set(sourceName, frameHeight);
+
+    // Ignore sender constraints for the following cases -
+    // 1. If the media on the peerconnection is suspended (jvb conn when p2p is currently active).
+    // 2. If the client is already sending video of the requested resolution.
+    if (!this.videoTransferActive || this.tpcUtils.getEncodedResolution(localVideoTrack) === frameHeight) {
```
saghul, line 2560: Does this generate any problem if we set our consraint to a certain value but end up encoding different value, for whatever reason?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2668,6 +2633,8 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
                     .find(layer => layer.scaleResolutionDownBy === scaleFactor)?.maxBitrate ?? bitrate;
             }
             parameters.encodings[0].maxBitrate = bitrate;
+        } else {
+            parameters.encodings[0].maxBitrate = undefined;
```
saghul, line 2637: would 0 work here? In RN the native APIs only support a number or null.

modules/xmpp/JingleSessionPC.js
```diff
@@ -2277,6 +2277,20 @@ export default class JingleSessionPC extends JingleSession {
         });
     }
 
+    /**
+     * Resumes or suspends media transfer over the underlying peer connection.
+     *
+     * @param {boolean} enable
+     * @returns {Promise}
+     */
+    resumeOrSuspendMedia(enable) {
```
saghul, line 2286: Ditto on the function name.

modules/xmpp/JingleSessionPC.js
```diff
@@ -2277,6 +2277,20 @@ export default class JingleSessionPC extends JingleSession {
...
+        return this.peerconnection.tpcUtils.resumeOrSuspendMedia(enable)
+            .then(() => {
```
saghul, line 2288: What if it fails?

saghul (commented):
I did a quick pass and left some comments.

modules/RTC/TPCUtils.js
```diff
@@ -420,16 +449,31 @@ export class TPCUtils {
...
-    setAudioTransferActive(active) {
-        this.setMediaTransferActive(MediaType.AUDIO, active);
+     * Resumes or suspends media on the peerconnection by setting the active state on RTCRtpEncodingParameters
+     * associates with all the senders that have a track attached to it.
+     *
+     * @param {boolean} enable - whether media needs to be enabled or suspended.
+     * @returns {Promise} - A promise that is resolved with an array of results when the change is succesful on all the
+     * senders, rejected otherwise.
+     */
+    resumeOrSuspendMedia(enable) {
```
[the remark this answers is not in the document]
jallamsetty1: We already have `setMediaTransferActive`, didn't want to confuse it with that one. I could rename the other one to `setVideoTransferActive` since its being used for exclusively for turning off incoming and outgoing video for p2p for lastn=0 and name this `setMediaTransferActive`. Does that work ?

modules/RTC/TraceablePeerConnection.js
jallamsetty1: I don't see how that can happen. The encode resolution depends on how many simulcast streams are enabled and this the only method that configures the encodings.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: We want the maxBitrate to be not limited here for VP9-p2p case so 0 will not work.

modules/xmpp/JingleSessionPC.js
jallamsetty1: It logs an error at the JitsiConference level - [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3308)  or [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiConference.js#L3437)

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Also, when the actual encode resolution changes because of cpu or bandwidth constraints, the encodings `scaleResolutionDownBy` value doesn't change. So this method only checks if the correct encodings are enabled based on the `frameHeight` received in the sender constraints.

jallamsetty1: I can rename the method to getConfiguredEncodeResolution if that makes it more descriptive.

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
saghul: Sounds good! How about going a bit further and call the P2P one setP2pVideoTransferActive ?

modules/RTC/TraceablePeerConnection.js
saghul: 👍

modules/RTC/TraceablePeerConnection.js
saghul: Got it. We might need to adjust RN I think.

modules/RTC/TPCUtils.js
jallamsetty1: Right , but we do want to resolve with the first failing promise and log an error in the calling function. There is not point in waiting for others to resolve since operation is unsuccessful on one of the senders.

jallamsetty1:
Jenkins test this please.

saghul (dismissed):
LGTM!

modules/RTC/TPCUtils.js
jallamsetty1: Good point! It also means that something has changed in the browser recently and that we need to push all setParameters operations to the jingle session queue. I will try to look up what the spec says and see if some changes have been made in Chrome in that area.

PR #2228, opened 2023-03-10
fix(RTCRtpSender) Race condition on resetting transaction id.

mykhailo-ua:
Made updating video sender parameters in sequential manner in order to solve race condition when transaction id (which is generated during getParameters) can be reset at the end of videoSender setParameters https://chromium.googlesource.com/external/webrtc/+/master/pc/rtp_sender.cc#340 and cause exception if another update was already in fly and before its check for transaction id presence.

The issue is not consistent, but I was able to reproduce it when e.g. in chrome browser for some time you are starting/stopping screen-share with choosing different options like full screen, browser tab or desktop application. 

Also after the change  I run npm run test/lint
Screenshot 2023-03-10 at 11 40 28

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -422,6 +422,8 @@ export default function TraceablePeerConnection(
         }, 1000);
     }
 
+    this._lastVideoSenderUpdatePromise = Promise.resolve();
+
     logger.info(`Create new ${this}`);
 }
 
@@ -2571,6 +2573,27 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
     if (!localVideoTrack || localVideoTrack.isMuted()) {
         return Promise.resolve();
     }
+
+    // Updating video sender parameters as wrapped-up promise to have ability
+    // to chain promises sequentially later and avoid chrome problem with transaction id
+    // which is reset at the end of video sender setParameters method and can affect
+    // next on-fly update
+    const nextPromise = this._lastVideoSenderUpdatePromise
+        .finally(() => this._updateVideoSenderParameters(frameHeight, localVideoTrack));
+
+    this._lastVideoSenderUpdatePromise = nextPromise;
+
+    return nextPromise;
+};
+
+/**
+ * Configures the video stream with resolution / degradation / maximum bitrates
+ *
+ * @param {number} frameHeight - The max frame height to be imposed on the outgoing video stream.
+ * @param {JitsiLocalTrack} - The local track for which the sender constraints have to be applied.
+ * @returns {Promise} promise that will be resolved when the operation is successful and rejected otherwise.
+ */
+TraceablePeerConnection.prototype._updateVideoSenderParameters = function(frameHeight, localVideoTrack) {
     const videoSender = this.findSenderForTrack(localVideoTrack.getTrack());
 
     if (!videoSender) {


mykhailo-ua:
@jallamsetty1 could you please take a look?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -422,6 +422,8 @@ export default function TraceablePeerConnection(
         }, 1000);
     }
 
+    this._lastVideoSenderUpdatePromise = null;
```
saghul, line 425: You could initialize this to Promise.resolve(), then you can just chain the promises allways.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2571,6 +2573,27 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
     if (!localVideoTrack || localVideoTrack.isMuted()) {
         return Promise.resolve();
     }
+
+    const prevPromise = this._lastVideoSenderUpdatePromise;
```
saghul, line 2577: Why did you cut this function here instead of also checking if there is a sender and returning early?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2571,6 +2573,27 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
...
+    const nextPromise = Promise.allSettled([ prevPromise ])
```
saghul, line 2578: You no longer need to extract prevPromise, you can just use the instance variable.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2571,6 +2573,27 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
...
+ * Updating video sender parameters as wrapped-up promise to have ability
+ * to chain promises sequentially later and avoid problem with transaction id
+ * which is reset at the end of video sender setParameters method and can affect
+ * next on-fly update
+ *
+ * @param {number} frameHeight - The max frame height to be imposed on the outgoing video stream.
+ * @param {JitsiLocalTrack} - The local track for which the sender constraints have to be applied.
+ * @returns {Promise} promise that will be resolved when the operation is successful and rejected otherwise.
+ */
+TraceablePeerConnection.prototype.updateVideoSenderParameters = function(frameHeight, localVideoTrack) {
```
saghul, line 2596: I'd suspect this should be private.

saghul (commented):
Left some more comments.

modules/RTC/TraceablePeerConnection.js
mykhailo-ua: Right, corrected that one.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2571,6 +2573,27 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
     if (!localVideoTrack || localVideoTrack.isMuted()) {
         return Promise.resolve();
     }
+
+    const prevPromise = this._lastVideoSenderUpdatePromise;
+    const nextPromise = Promise.allSettled([ prevPromise ])
+        .then(() => this.updateVideoSenderParameters(frameHeight, localVideoTrack));
```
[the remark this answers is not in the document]
mykhailo-ua: Yeah, finally looks better here, changed that.

modules/RTC/TraceablePeerConnection.js
mykhailo-ua: added underscore prefix here for making it private.

modules/RTC/TraceablePeerConnection.js
mykhailo-ua: I did not want to split logical part of taking videoSender and getting params, also in case of later split I would need to either check it once again for videoSender presence (as it might be delayed) or add additional videoSender parameter to _updateVideoSenderParameters. 
Do you think it would be better to move videoSender check in front of _updateVideoSenderParameters and pass it as additional parameter?

saghul (dismissed):
LGTM, but I defer to Jaya for her final thoughts.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2571,6 +2573,26 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
...
+    const nextPromise = this._lastVideoSenderUpdatePromise
+        .finally(() => this._updateVideoSenderParameters(frameHeight, localVideoTrack));
+
+    this._lastVideoSenderUpdatePromise = nextPromise;
+
+    return nextPromise;
+};
+
+/**
+ * Updating video sender parameters as wrapped-up promise to have ability
```
jallamsetty1, line 2586: I would prefer to move this text as a comment up to line 2577 because it is very specific to Chrome's implementation of get/setParameters. `_updateVideoSenderParameters` configures the video stream and the JSDoc for this method should reflect that.

jallamsetty1 (commented):
LGTM, left a minor comment.

jallamsetty1:
Jenkins test this please.

jallamsetty1 (approved):

jallamsetty1:
@mykhailo-ua Looks like the issue is not fixed. I just noticed the same error on a build with the fix.

mykhailo-ua:
> @mykhailo-ua Looks like the issue is not fixed. I just noticed the same error on a build with the fix.

Oh.. Will try to investigate it further then.

jallamsetty1:
@paweldomas We still see this issue on our deployments and in fact one of the torture tests has started failing in the past week with the following error

2023-03-21T13:23:44.097Z [JitsiConference.js] replaceTrack failed: Error: Failed to execute 'setParameters' on 'RTCRtpSender': Failed to set parameters since getParameters() has never been called on this sender
2023-03-21T13:23:44.097Z [JitsiMeetJS.ts] <Object.getGlobalOnErrorHandler>:  UnhandledError: Failed to execute 'setParameters' on 'RTCRtpSender': Failed to set parameters since getParameters() has never been ca

paweldomas:
I suspect it may be because there's another place where setParameters is called:
https://github.com/jitsi/lib-jitsi-meet/blob/5155e550213369726b72e7401c0ab28e9b7b5532/modules/RTC/TraceablePeerConnection.js#L2062

We're not running on the latest master and this code is different, so maybe that's why we stopped seeing it. Please note it's rather involving setup for us to test vanilla jitsi and we're unlikely to follow up with additional fix at this time.

I'd suggest to add logging around the transaction IDs, how and when promises with those are resolved then the test should reveal what happens there.

jallamsetty1:
Actually the line that you referred to gets executed only on Firefox but we recently added another call to setParameters for p2p<->jvb switch. I have a PR open for fixing that.

PR #2233, opened 2023-03-16
fix(codec-selection) Apply codec preferences to initial offer/answer.

jallamsetty1:
This fixes an issue where p2p clients (with different codec preferences) fail to decode video because the negotiated codecs are removed from the supported codecs list after the media session is established. The codec preferences will be applied when the first offer/answer is created.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2254,6 +2254,11 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(jingleSession, jingl
             this._signalingLayer,
             {
                 ...this.options.config,
+                codecSettings: {
+                    mediaType: MediaType.VIDEO,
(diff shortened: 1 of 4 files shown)


JitsiConference.js
```diff
@@ -2254,6 +2254,11 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(jingleSession, jingl
             this._signalingLayer,
             {
                 ...this.options.config,
+                codecSettings: {
+                    mediaType: MediaType.VIDEO,
+                    preferred: this.codecSelection.jvbPreferredCodec,
```
saghul, line 2259: How can we guarantee this is not undefined?

saghul (approved):
LGTM with a question.

PR #2245, opened 2023-03-25
feat(visitors): Visitors chat.

damencho:
For messages coming from visitors, which are not in the current room we need to supply a display name.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -1159,8 +1159,18 @@ export default class ChatRoom extends Listenable {
                 this.eventEmitter.emit(XMPPEvents.PRIVATE_MESSAGE_RECEIVED,
                         from, txt, this.myroomjid, stamp);
             } else if (type === 'groupchat') {
+                const nickEl = $(msg).find('>nick');
+                let nick;
+
+                if (nickEl.length > 0) {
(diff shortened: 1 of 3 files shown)


modules/xmpp/ChatRoom.js
```diff
@@ -1159,8 +1159,13 @@ export default class ChatRoom extends Listenable {
                 this.eventEmitter.emit(XMPPEvents.PRIVATE_MESSAGE_RECEIVED,
                         from, txt, this.myroomjid, stamp);
             } else if (type === 'groupchat') {
+                const nick = $(msg).find('>nick').text();
+
+                // we will fire explicitly that this is a guest(isGuest:true) to the conference
+                // informing that this is probably a message from a guest to the conference (visitor)
+                // a message with explicit name set
                 this.eventEmitter.emit(XMPPEvents.MESSAGE_RECEIVED,
-                        from, txt, this.myroomjid, stamp);
```
bgrozev, line 1163: Doesn't this mark *all* group messages as coming from a visitor?

nit: IMO we should just use "visitor" and avoid introducing "guest"

damencho: yes 🤦

bgrozev (approved):

PR #2250, opened 2023-03-30
Cleanup

jallamsetty1:


JitsiConference.js
```diff
@@ -475,7 +468,7 @@ JitsiConference.prototype._init = function(options = {}) {
             callStatsID: config.callStatsID,
             callStatsSecret: config.callStatsSecret,
             callStatsApplicationLogsDisabled: config.callStatsApplicationLogsDisabled,
-            enableCallStats,
+            enableCallStats: true,
```
damencho, line 471: Can we drop this now, when it is always true? It was used to selectively enable/disable integration even when configured correctly. If we don't need it let it 🔥

damencho (changes requested):

JitsiConference.js
jallamsetty1: Done

damencho (approved):

PR #2253, opened 2023-04-03
feat(bandwidth) send bandwidth constraints to the bridge

mihhu:


saghul (dismissed):
LGTM with a suggestion.

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -7,6 +7,25 @@ import { MediaType } from '../../service/RTC/MediaType';
 const logger = getLogger(__filename);
 const MAX_HEIGHT = 2160;
 const LASTN_UNLIMITED = -1;
+const ASSUMED_BANDWIDTH_BPS = {
+    MAX_VALUE: 10000000,
```
saghul, line 11: Is this the max? I thought it's now capped by the JVB setting, so then we should probably not limit it here?

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -7,6 +7,25 @@ import { MediaType } from '../../service/RTC/MediaType';
...
+};
+
+/**
+ * Sanitizes the assumed bandwidth bps value to be numeric and within its interval.
+ * @param {number} value - the count of bps.
+ * @returns {number} the sanitized assumed bandwidth bps value.
+ */
+const sanitizeAssumedBandwidthBps = value => {
+    if (isNaN(value) || value > ASSUMED_BANDWIDTH_BPS.MAX_VALUE) {
+        return ASSUMED_BANDWIDTH_BPS.MIN_VALUE;
```
saghul, line 22: I guess you meant the MAX_VALUE here?

saghul (commented):
@bgrozev can you PTAL too?

modules/qualitycontrol/ReceiveVideoController.js
mihhu: No no, the negative values are ignored, so it's intended as a reset mechanism, to ignore this setting.

modules/qualitycontrol/ReceiveVideoController.js
mihhu: Yes, it should be the max we allow, I'll remove it from here.

saghul (approved):
LGTM. @bgrozev can you PTAL too?

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -133,8 +170,16 @@ export default class ReceiveVideoController {
          */
         this._sourceReceiverConstraints = new Map();
 
+        /**
+         * The number of bps requested from the bridge.
+         */
+        this._assumedBandwidthBps = sanitizeAssumedBandwidthBps(config?.assumedBandwidthBps);
```
saghul, line 176: Let's default to -1 and not use config, I don't think it makes much sense to put a config like that, wholesale...

saghul (changes requested):
Remembered the discussion about config.js, left an extra comment.

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -185,6 +220,18 @@ export default class ReceiveVideoController {
...
 
+    /**
+     * Sets the assumed bandwidth bps the local participant should receive from remote participants.
+     *
+     * @param {number|undefined} assumedBandwidthBps - the new value.
+     * @returns {void}
+     */
+    setAssumedBandwidthBps(assumedBandwidthBps) {
+        if (this._receiverVideoConstraints.updateAssumedBandwidthBps(assumedBandwidthBps)) {
+            this._rtc.setReceiverVideoConstraints(this._receiverVideoConstraints.constraints);
```
bgrozev, line 231: Note that the different fields in the ReceiverConstraints message can be updated independently. So you could send a message with only `assumedBandwidthBps` set instead of sending the full constraints. I don't understand the flow enough to see if it makes sense though

saghul: AFAICT this follows the same pattern as for lastN for example.

saghul (approved):

PR #2259, opened 2023-04-14
fix(statistics) Use RTCRtpReceiver:getSynchronizationSources for audioLevels always

jallamsetty1:
Track based stats have been removed recently and RTCRtpReceiver:getSynchronizationSources has been available in the browsers for a while now.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -531,7 +531,7 @@ JitsiConference.prototype._init = function(options = {}) {
     }
 
     // Generates events based on no audio input detector.
-    if (config.enableNoAudioDetection) {
+    if (config.enableNoAudioDetection && !config.disableAudioLevels) {
         this._noAudioSignalDetection = new NoAudioSignalDetection(this);
         this._noAudioSignalDetection.on(DetectionEvents.NO_AUDIO_INPUT, () => {
             this.eventEmitter.emit(JitsiConferenceEvents.NO_AUDIO_INPUT);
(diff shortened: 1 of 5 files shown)


saghul:
This is not supported on RN since WebRTC doesn't implement it, it's implemented on the Chromium side. Will this be a problem?

jallamsetty1:
> This is not supported on RN since WebRTC doesn't implement it, it's implemented on the Chromium side. Will this be a problem?

Actually this will fix the case for the newer releases of Chromium where the support for track statistics was dropped. Since RN WebRTC doesn't support this API, we were falling back to the legacy impl which wouldn't yield any results for newer versions of Chromium. RTCRtpReceiver:getSynchronizationSources has been supported since Chrome 73 so this change should fix this issue.

saghul (approved):

saghul:
Sorry, you lost me in the seconf half :-) I think we should be fine because we do disable audio levels on RN, I just wanted to make sure we wouldn't call that API on RN because it's not implemented.

jallamsetty1:
> Sorry, you lost me in the seconf half :-) I think we should be fine because we do disable audio levels on RN, I just wanted to make sure we wouldn't call that API on RN because it's not implemented.

🤦‍♀️ now I get what you are saying. I can bring back the check and do nothing if the check fails. This should atleast help us get rid of the old getStats based impl. WDYT ?

JitsiConference.js
```diff
@@ -531,7 +531,7 @@ JitsiConference.prototype._init = function(options = {}) {
     }
 
     // Generates events based on no audio input detector.
-    if (config.enableNoAudioDetection) {
+    if (config.enableNoAudioDetection && !RTC.options.disableAudioLevels) {
```
saghul, line 534: Shouldn't this also be `!config.disableAudioLevels` ?

saghul (approved):

PR #2265, opened 2023-04-24
fix(stats) Obtain resolution/fps from 'outbound-rtp' stats.

jallamsetty1:
'Track' based stats were dropped in Chrome 112, therefore send resolution and fps for the simulcast case needs to be calculated based on the 'outbound-rtp' streams that are currently active.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -681,6 +681,33 @@ TraceablePeerConnection.prototype.getAudioLevels = function(speakerList = []) {
 };
 
 /**
+ * Checks if the browser is currently doing true simulcast where in three different media streams are being sent to the
+ * bridge. Currently this happens only when VP8 is the selected codec.
+ * @returns {boolean}
+ */
+TraceablePeerConnection.prototype.doesTrueSimulcast = function() {
+    return this.isSimulcastOn() && this.getConfiguredVideoCodec() === CodecMimeType.VP8;
+};
+
+/**
+ * Returns the SSRCs associated with a given local video track.
+ *
+ * @param {JitsiLocalTrack} localTrack
+ * @returns
+ */
+TraceablePeerConnection.prototype.getLocalSSRCs = function(localTrack) {
+    const ssrcs = [];
+
+    if (!localTrack || !localTrack.isVideoTrack()) {
+        return ssrcs;
+    }
+
+    const ssrcGroup = this.isSimulcastOn() ? 'SIM' : 'FID';
+
+    return this.localSSRCs.get(localTrack.rtcId)?.groups?.find(group => group.semantics === ssrcGroup)?.ssrcs || ssrcs;
+};
+
+/**
  * Obtains local tracks for given {@link MediaType}. If the <tt>mediaType</tt>
  * argument is omitted the list of all local tracks will be returned.
  * @param {MediaType} [mediaType]
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -695,7 +695,7 @@ TraceablePeerConnection.prototype.doesTrueSimulcast = function() {
  * @param {JitsiLocalTrack} localTrack
  * @returns
  */
-TraceablePeerConnection.prototype.getLocalSSRCs = function(localTrack) {
+TraceablePeerConnection.prototype.getLocalVideoSSRCs = function(localTrack) {
     const ssrcs = [];
 
     if (!localTrack || !localTrack.isVideoTrack()) {
@@ -1263,6 +1263,18 @@ TraceablePeerConnection.prototype._extractSSRCMap = function(desc) {
                     }
                     groupsMap.get(primarySSRC).push(group);
                 }
+            }
+
+            const simGroup = mLine.ssrcGroups.find(group => group.semantics === 'SIM');
+
+            // Add a SIM group if its missing in the description (happens on Firefox).
+            if (!simGroup) {
+                const groupSsrcs = mLine.ssrcGroups.map(group => group.ssrcs[0]);
+
+                groupsMap.get(groupSsrcs[0]).push({
+                    semantics: 'SIM',
+                    ssrcs: groupSsrcs
+                });
             }
         }
 
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -220,6 +220,17 @@ export default class BrowserCapabilities extends BrowserDetection {
     }
 
     /**
+     * Returns true if the browser supports track based statistics for the local video track. Otherwise,
+     * track resolution and framerate will be calculated based on the 'outbound-rtp' statistics.
+     * @returns {boolean}
+     */
+    supportsTrackBasedStats() {
+        return (this.isChromiumBased() && this.isVersionLessThan(112))
+            || this.isWebKitBased()
+            || this.isFirefox();
+    }
+
+    /**
      * Returns true if VP9 is supported by the client on the browser. VP9 is currently disabled on Firefox and Safari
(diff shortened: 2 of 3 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -680,6 +680,33 @@ TraceablePeerConnection.prototype.getAudioLevels = function(speakerList = []) {
...
+    return this.isSimulcastOn() && this.getConfiguredVideoCodec() === CodecMimeType.VP8;
+};
+
+/**
+ * Returns the SSRCs associated with a given local video track.
+ *
+ * @param {JitsiLocalTrack} localTrack
+ * @returns
+ */
+TraceablePeerConnection.prototype.getLocalSSRCs = function(localTrack) {
```
saghul, line 698: Can we call this `getLocalVideoSSRCs` instead please?

modules/statistics/RTPStatsCollector.js
```diff
@@ -275,61 +275,86 @@ StatsCollector.prototype._processAndEmitReport = function() {
         bitrateDownload += ssrcStats.bitrate.download;
         bitrateUpload += ssrcStats.bitrate.upload;
 
+        ssrcStats.resetBitrate();
```
saghul, line 278: Why didn't we need this before?

modules/browser/BrowserCapabilities.js
```diff
@@ -219,6 +219,17 @@ export default class BrowserCapabilities extends BrowserDetection {
         return !this.isFirefox();
     }
 
+    /**
+     * Returns true if the browser supports track based statistics for the local video track. Otherwise,
+     * track resolution and framerate will be calculated based on the 'outbound-rtp' statistics.
+     * @returns {boolean}
+     */
+    supportsTrackBasedStats() {
```
saghul, line 227: Currently we have WebRTC M111 on React Native, what stats should we be using?

modules/statistics/RTPStatsCollector.js
```diff
@@ -275,61 +275,86 @@ StatsCollector.prototype._processAndEmitReport = function() {
...
+        // However for the SVC case, there will be only 1 "outbound-rtp" stream which will have the correct
+        // send resolution width and height.
+        if (track.isLocal() && !browser.supportsTrackBasedStats() && this.peerconnection.doesTrueSimulcast()) {
+            const localSsrcs = this.peerconnection.getLocalSSRCs(track);
+
+            for (const localSsrc of localSsrcs) {
+                const ssrcResolution = this.ssrc2stats.get(localSsrc)?.resolution;
+
+                // The code processes resolution stats only for 'outbound-rtp' streams that are currently active.
+                if (ssrcResolution?.height && ssrcResolution?.width) {
```
saghul, line 342: Why don't we sanitize these?

modules/statistics/RTPStatsCollector.js
```diff
@@ -552,29 +577,34 @@ StatsCollector.prototype.processStatsReport = function() {
...
-                    height: now.frameHeight,
-                    width: now.frameWidth
-                };
-                const frameRate = now.framesPerSecond;
-
-                if (resolution.height && resolution.width) {
-                    ssrcStats.setResolution(resolution);
-                }
+            const resolution = {
+                height: now.frameHeight,
```
saghul, line 581: Why do we assign these without checking for the type first, as before?

modules/statistics/RTPStatsCollector.js
```diff
@@ -552,29 +577,34 @@ StatsCollector.prototype.processStatsReport = function() {
...
 
-                if (before) {
-                    ssrcStats.addBitrate({
-                        'download': this._calculateBitrate(now, before, 'bytesReceived'),
-                        'upload': 0
-                    });
-                }
+            // Reset the stats if the current stream is suspended when 'outbound-rtp' stats are used for resolution
+            // and framerate.
+            } else if (!browser.supportsTrackBasedStats()) {
```
saghul, line 598: If i'm reading this right you could end up here if `resolution.height` is 0 above, even if the type is `inbound-rtp`.

saghul (commented):
Left some comments!

modules/browser/BrowserCapabilities.js
fippo: inbound-rtp based stats contain everything you need for a while (in Chromium-based browsers) already.

fippo: Should have it since 2020 at least: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/api/stats/rtcstats_objects.h;l=500;drc=b3e59696585d222a53c1c59292eb92a66e604643

modules/browser/BrowserCapabilities.js
```diff
@@ -219,6 +219,17 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+        return (this.isChromiumBased() && this.isVersionLessThan(112))
+            || this.isWebKitBased()
+            || this.isFirefox();
```
fippo, line 230: I don't see track stats in Firefox? Sadly it does not have framesPerSecond in outbound-rtp but it does have framesSent at least which allows calculating

modules/RTC/TraceablePeerConnection.js
```diff
@@ -680,6 +680,33 @@ TraceablePeerConnection.prototype.getAudioLevels = function(speakerList = []) {
     return audioLevels;
 };
 
+/**
+ * Checks if the browser is currently doing true simulcast where in three different media streams are being sent to the
+ * bridge. Currently this happens only when VP8 is the selected codec.
+ * @returns {boolean}
+ */
+TraceablePeerConnection.prototype.doesTrueSimulcast = function() {
```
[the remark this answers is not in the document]
jallamsetty1: I am afraid there is no way to use feature detection here since it will be depend on our implementation. If we choose `L3T3` scalability mode, VP9/AV1 will continue to do SVC and will do true simulcast if we choose `L1T3` as a scalability mode and set `scaleResolutionDownBy` values on the encodings. I can update this method when we implement simulcast for VP9/AV1.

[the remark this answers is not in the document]
saghul: Sounds good!

modules/statistics/RTPStatsCollector.js
jallamsetty1: We needed this before also, but it was called at the end. I switched to using continue instead of using 5 nested if loops, therefore I need to call this before the first continue.

modules/browser/BrowserCapabilities.js
saghul: Nice!

jallamsetty1: Yes, `outbound-rtp `have been available for quite a while now so it shouldn't affect the current React Native build. The check here is mostly for older Chrome versions (72) that do not generate `outbound-rtp` stats and for Safari that still doesn't support it. @fippo I checked the current stable for Firefox and see the stats, but will check again and update the check here.

modules/statistics/RTPStatsCollector.js
jallamsetty1: Yup, l will fix that.

saghul: IMHO it makes the code less readable because you need to know what you just said for it to be correct. I'd rather the code be explicit.

modules/statistics/RTPStatsCollector.js
saghul: 👍

modules/browser/BrowserCapabilities.js
jallamsetty1: They are working on Safari too so I removed the check for that as well.

modules/statistics/RTPStatsCollector.js
jallamsetty1: Ok, I addressed this in the most recent commit.

saghul (approved):
Looks good now!

PR #2274, opened 2023-05-04
fix(JingleSession) start modification queue after session is established

jallamsetty1:
Pull initial offer/answer tasks out of the modification queue and execute them right away. Only track and codec related operations that necessitate a renegotiation cycle need to be pushed to the modification queue. The queue execution is paused until the session is established. This avoids track operations being executed before the session is established. This fixes an issue seen in p2p connections where sources of the initiator are not signaled to the remote since the tracks are added while the initiator is waiting for a session-accept from the peer.

### DIFF
--- a/modules/util/AsyncQueue.js
+++ b/modules/util/AsyncQueue.js
@@ -35,6 +35,13 @@ export default class AsyncQueue {
     }
 
     /**
+     * Pauses the execution of the tasks on the queue.
+     */
+    pause() {
+        this._queue.pause();
+    }
+
+    /**
      * The 'task' function will be given a callback it MUST call with either:
      *  1) No arguments if it was successful or
      *  2) An error argument if there was an error
@@ -58,6 +65,13 @@ export default class AsyncQueue {
             return;
         }
         this._queue.push(task, callback);
+    }
+
+    /**
+     * Resumes the execution of the tasks on the queue.
+     */
+    resume() {
+        this._queue.resume();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


jallamsetty1:
> Isn't the asymmetry of operations going to complicate things in the long run?

What asymmetry of operations are you referring to ? All the other tasks that get queued up need a LD/RD to be available which wouldn't be case if they initial offer/answer are also pushed to the queue.

saghul:
> > Isn't the asymmetry of operations going to complicate things in the long run?
> 
> What asymmetry of operations are you referring to ? All the other tasks that get queued up need a LD/RD to be available which wouldn't be case if the initial offer/answer are also pushed to the queue.

Before:

>          * The queue used to serialize operations done on the peerconnection.

After:

>         * The queue used to serialize operations done on the peerconnection after the session is established.
>         * The queue is paused until the first offer/answer cycle is complete. Only track or codec related
>         * operations which necessitate a renegotiation cycle need to be pushed to the modification queue.
>         * These tasks will be executed after the session has been established.

The former is clearly easier to understand.

For anyone not working with this code daily "all operations involving negotiation go on the queue" is much easier to understand that "all except the first, for reasons".

My 2 cents.

jallamsetty1:
> For anyone not working with this code daily "all operations involving negotiation go on the queue" is much easier to understand that "all except the first, for reasons".

Actually it is supposed to be "all operations involving renegotiation". There is no renegotiation if the session itself doesn't exist. We have been lucky that it worked until now without any issues but like the issue we are running into now with slow gUM, we cannot queue up tasks between the initial offer and answer. 
I don't think there is any other way around it.

saghul (approved):

PR #2276, opened 2023-05-08
fix(SignalingLayer) prevent Jicofo «Source does not exist» error

NewEraCracker:
Use `0` as track index if `undefined`, otherwise there will be `aundefined`, `vundefined` tracks in corner cases.

We've been getting this error lately on some percentage of the clients !? (I've snipped all identifiable information away for privacy):
```
Jicofo 2023-05-08 15:08:39.148 INFO: [1308] [remoteJid=*SNIP*@conference.*SNIP*/*SNIP* sid=*SNIP*] JingleSession.doProcessIq#145: Returning error: request=<iq xmlns='jabber:client' to='focus@auth.*SNIP*/focus' from='*SNIP*@conference.*SNIP*/*SNIP*' id='*SNIP*' type='set'><jingle xmlns='urn:xmpp:jingle:1' action='source-remove' initiator='*SNIP*@conference.*SNIP*/focus' sid='*SNIP*'><content name='audio'><description xmlns='urn:xmpp:jingle:apps:rtp:1' media='audio'><source xmlns='urn:xmpp:jingle:apps:rtp:ssma:0' name='*SNIP*-aundefined' ssrc='*SNIP*'/></description></content></jingle></iq>, error=<error xmlns='jabber:client' type='modify'><bad-request xmlns='urn:ietf:params:xml:ns:xmpp-stanzas'/><text xmlns='urn:ietf:params:xml:ns:xmpp-stanzas' xml:lang='en'>Source does not exist or is not owned by endpoint null.</text></error>
```

### DIFF
--- a/service/RTC/SignalingLayer.js
+++ b/service/RTC/SignalingLayer.js
@@ -27,7 +27,7 @@ import { MediaType } from '../../service/RTC/MediaType';
 export function getSourceNameForJitsiTrack(endpointId, mediaType, trackIdx) {
     const firstLetterOfMediaType = mediaType.substring(0, 1);
 
-    return `${endpointId}-${firstLetterOfMediaType}${trackIdx}`;
+    return `${endpointId}-${firstLetterOfMediaType}${trackIdx || 0}`;
 }
 
 /**


saghul (approved):
LGTM @jallamsetty1 ? I see a couple of places where it might end up undefined. Not sure it's worth to look into those...

jallamsetty1:
@NewEraCracker, how does your client end up sending a source-remove ? Replacing a track with `null` on the pc shouldn't generate a source-remove, this is something we try to avoid.

NewEraCracker:
I'm not entirely sure, either.

However, our application logic is set to mute tracks before removing them, with explicit removeTrack, and this action is also taken before leaving the conference to ensure proper handling of track removal.

As I was looking today on that part of the code (leaving the conference) today, I've touched it up to just straight leave the conference and clean the tracks later.

We'll see...

jallamsetty1:
> I'm not entirely sure, either.
> 
> However, our application logic is set to mute tracks before removing them, with explicit removeTrack, and this action is also taken before leaving the conference to ensure proper handling of track removal.
> 
> As I was looking on that part of the code (leaving the conference) today, I've touched it up to just straight leave the conference and clean the tracks later.
> 
> We'll see...

Please share your findings so that I can understand why source-remove is getting triggered. This fix won't work for all the cases since the source-remove generated will always have the first track even if the source being removed could be a secondary track like screenshare.

NewEraCracker:
Hi,

With this change, in the past, I was occasionally seeing `a0` or `v0` instead of the before change situation `aundefined` and `vundefined`.

However, while running a recent tag (v1635.0.0+152fdb21) patched with this change, I have not seen any errors for the past two days (no longer seeing ANY `Source does not exist` errors).

Thanks.

PS: For the moment in our app, we are only enabling P2P on Chromium as cross P2P Chromium-Firefox has some frozen picture bug (this happened in the past and led to the whole remove-will-only-mute workaround) when the Firefox client transitions from P2P to the Bridge on a new participant join.

damencho:
There were fixes added for p2p on FF and it is now enabled on meet.jit.si https://github.com/jitsi/lib-jitsi-meet/pull/2279

jallamsetty1:
Closing this since its no longer needed.

PR #2279, opened 2023-05-10
feat(signaling) Only include 'msid' in the source description.

jallamsetty1:
'cname', 'label' and 'mslabel' attributes that are included in the source description of the signaling messages are ignored by Jicofo. Only 'msid' is used for source validation. Therefore, drop those attributes and make sure 'msid' is always included and in the format `'<endpoint_id>-<mediaType>-<trackIndex>-<tpcId>`'.

fix(p2p) Enable p2p for Firefox.
Set the senders attribute correctly in the outgoing jingle message. Ignore the media direction from the session, only look for the direction in the media desccription. This fixes an issue where Firefox rejects an answer because the direction is not set correctly.


JitsiConference.js
```diff
@@ -324,6 +324,9 @@ export default function JitsiConference(options) {
      * again by Jicofo.
      */
     this._videoSenderLimitReached = undefined;
+
+    this._firefoxP2pEnabled = browser.isVersionGreaterThan(109)
```
saghul, line 328: Why not check if the browser is Firefox here too, thus simplifying further checks?

saghul (commented):
Drive by comment

JitsiConference.js
jallamsetty1: Nope, it won't work because of the rejection logic at https://github.com/jitsi/lib-jitsi-meet/pull/2279/files#diff-b2f6195536d4c09eac93cd5482c9258cae0dc21129fca78a125fe545594eca30R3465 and https://github.com/jitsi/lib-jitsi-meet/pull/2279/files#diff-b2f6195536d4c09eac93cd5482c9258cae0dc21129fca78a125fe545594eca30R2172. The client will end up disabling p2p for the other browsers as well.

damencho (approved):

JitsiConference.js
saghul: Ack

saghul (approved):

PR #2280, opened 2023-05-11
fix(JitsiConference) Do not apply the startMutedPolicy locally on the moderator.

jallamsetty1:
Fixes a corner case where the local tracks are not added to the jvb pc if the session-initiate from Jicofo is received after the policy is set. This doesn't reproduce manually and happens only in torture on certain deployments because its timing specific.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2675,7 +2675,10 @@ JitsiConference.prototype.setStartMutedPolicy = function(policy) {
 
         return;
     }
-    this.startMutedPolicy = policy;
+
+    // Do not apply the startMutedPolicy locally on the moderator, the moderator should join with available local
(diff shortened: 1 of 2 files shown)


damencho:
Just a note for future reference. The StartMutedTest was failing in 2 different ways. 
The test is p1 enters room as moderator and enabled start audio and video muted and then p2 joins (p2p is enabled).
The first failure was p1 was not sending any media although is unmuted.
The second failure is that p1 is not updating its presence although it is sending media and for p2 it looks like muted in the screenshot.
The two failures depend on timing(when the jvb invite arrives.) and who is the p2p initiator.

damencho (dismissed):

damencho (approved):

PR #2295, opened 2023-06-16
fix(log) Log presence updates sent to prosody.

jallamsetty1:


saghul:
Please don't log whole stanzas. It hurts mobile because they are huge nested objects.

jallamsetty1:
@saghul changed it to logging it only when the mute state changes since we didn't want to log display names.

saghul (approved):

PR #2296, opened 2023-06-16
New Codec selection logic

jallamsetty1:
Use a configurable preferred codecs list to select the codecs.

Allow asymmetric codecs to be configured on the endpoints. This means that Firefox and Safari which have bugs with VP9 encode will now encode VP8 but decode VP9 coming in from Chromium endpoints.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -51,7 +51,7 @@ import {
     JITSI_MEET_MUC_TYPE
 } from './modules/xmpp/xmpp';
 import BridgeVideoType from './service/RTC/BridgeVideoType';
-import CodecMimeType from './service/RTC/CodecMimeType';
+import { CodecMimeType } from './service/RTC/CodecMimeType';
 import { MediaType } from './service/RTC/MediaType';
 import RTCEvents from './service/RTC/RTCEvents';
 import { SignalingEvents } from './service/RTC/SignalingEvents';
@@ -381,11 +381,16 @@ JitsiConference.prototype._init = function(options = {}) {
 
     // Get the codec preference settings from config.js.
     const codecSettings = {
-        jvbDisabledCodec: _getCodecMimeType(config.videoQuality?.disabledCodec),
-        p2pDisabledCodec: _getCodecMimeType(config.p2p?.disabledCodec),
-        enforcePreferredCodec: config.videoQuality?.enforcePreferredCodec,
-        jvbPreferredCodec: _getCodecMimeType(config.videoQuality?.preferredCodec),
-        p2pPreferredCodec: _getCodecMimeType(config.p2p?.preferredCodec)
+        jvb: {
+            preferenceOrder: config.videoQuality?.codecPreferenceOrder,
+            disabledCodec: _getCodecMimeType(config.videoQuality?.disabledCodec),
+            preferredCodec: _getCodecMimeType(config.videoQuality?.preferredCodec)
+        },
+        p2p: {
+            preferenceOrder: config.p2p?.codecPreferenceOrder,
+            disabledCodec: _getCodecMimeType(config.p2p?.disabledCodec),
+            preferredCodec: _getCodecMimeType(config.p2p?.preferredCodec)
+        }
     };
 
     this.codecSelection = new CodecSelection(this, codecSettings);
@@ -563,8 +568,8 @@ JitsiConference.prototype._init = function(options = {}) {
             'region', config.deploymentInfo.userRegion);
     }
 
-    // Publish the codec type to presence.
-    this.setLocalParticipantProperty('codecType', this.codecSelection.getPreferredCodec());
+    // Publish the codec preference to presence.
+    this.setLocalParticipantProperty('codecList', this.codecSelection.getCodecPreferenceList('jvb'));
 
     // Set transcription language presence extension.
     // In case the language config is undefined or has the default value that the transcriber uses
@@ -1016,7 +1021,6 @@ JitsiConference.prototype.sendCommand = function(name, values) {
     } else {
         logger.warn('Not sending a command, room not initialized.');
     }
-
 };
 
 /**
@@ -2260,8 +2264,7 @@ JitsiConference.prototype._acceptJvbIncomingCall = function(jingleSession, jingl
                 ...this.options.config,
                 codecSettings: {
                     mediaType: MediaType.VIDEO,
-                    preferred: this.codecSelection.jvbPreferredCodec,
-                    disabled: this.codecSelection.jvbDisabledCodec
+                    codecList: this.codecSelection.getCodecPreferenceList('jvb')
                 },
                 enableInsertableStreams: this.isE2EEEnabled() || FeatureFlags.isRunInLiteModeEnabled()
             });
@@ -3047,8 +3050,7 @@ JitsiConference.prototype._acceptP2PIncomingCall = function(jingleSession, jingl
             ...this.options.config,
             codecSettings: {
                 mediaType: MediaType.VIDEO,
-                preferred: this.codecSelection.p2pPreferredCodec,
-                disabled: this.codecSelection.p2pDisabledCodec
+                codecList: this.codecSelection.getCodecPreferenceList('p2p')
             },
             enableInsertableStreams: this.isE2EEEnabled() || FeatureFlags.isRunInLiteModeEnabled()
         });
@@ -3418,8 +3420,7 @@ JitsiConference.prototype._startP2PSession = function(remoteJid) {
             ...this.options.config,
             codecSettings: {
                 mediaType: MediaType.VIDEO,
-                preferred: this.codecSelection.p2pPreferredCodec,
-                disabled: this.codecSelection.p2pDisabledCodec
+                codecList: this.codecSelection.getCodecPreferenceList('p2p')
             },
             enableInsertableStreams: this.isE2EEEnabled() || FeatureFlags.isRunInLiteModeEnabled()
         });
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -51,7 +51,7 @@ import {
     JITSI_MEET_MUC_TYPE
 } from './modules/xmpp/xmpp';
 import BridgeVideoType from './service/RTC/BridgeVideoType';
-import { CodecMimeType } from './service/RTC/CodecMimeType';
+import CodecMimeType from './service/RTC/CodecMimeType';
 import { MediaType } from './service/RTC/MediaType';
 import RTCEvents from './service/RTC/RTCEvents';
 import { SignalingEvents } from './service/RTC/SignalingEvents';
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -382,12 +382,16 @@ JitsiConference.prototype._init = function(options = {}) {
     // Get the codec preference settings from config.js.
     const codecSettings = {
         jvb: {
-            preferenceOrder: config.videoQuality?.codecPreferenceOrder,
+            preferenceOrder: browser.isMobileDevice() && config.videoQuality?.mobileCodecPreferenceOrder
+                ? config.videoQuality.mobileCodecPreferenceOrder
+                : config.videoQuality?.codecPreferenceOrder,
             disabledCodec: _getCodecMimeType(config.videoQuality?.disabledCodec),
             preferredCodec: _getCodecMimeType(config.videoQuality?.preferredCodec)
         },
         p2p: {
-            preferenceOrder: config.p2p?.codecPreferenceOrder,
+            preferenceOrder: browser.isMobileDevice() && config.p2p?.mobileCodecPreferenceOrder
+                ? config.p2p.mobileCodecPreferenceOrder
+                : config.p2p?.codecPreferenceOrder,
             disabledCodec: _getCodecMimeType(config.p2p?.disabledCodec),
             preferredCodec: _getCodecMimeType(config.p2p?.preferredCodec)
         }
--- a/modules/RTC/CodecSelection.js
+++ b/modules/RTC/CodecSelection.js
@@ -156,8 +156,7 @@ export class CodecSelection {
         }
 
         // Reconfigure the codecs on the media session.
-        if (currentCodecOrder.length !== selectedCodecOrder.length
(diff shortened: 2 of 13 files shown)


JitsiConference.js
```diff
@@ -381,11 +381,20 @@ JitsiConference.prototype._init = function(options = {}) {
 
     // Get the codec preference settings from config.js.
     const codecSettings = {
-        jvbDisabledCodec: _getCodecMimeType(config.videoQuality?.disabledCodec),
-        p2pDisabledCodec: _getCodecMimeType(config.p2p?.disabledCodec),
-        enforcePreferredCodec: config.videoQuality?.enforcePreferredCodec,
-        jvbPreferredCodec: _getCodecMimeType(config.videoQuality?.preferredCodec),
-        p2pPreferredCodec: _getCodecMimeType(config.p2p?.preferredCodec)
+        jvb: {
+            preferenceOrder: browser.isMobileDevice() && config.videoQuality?.mobileCodecPreferenceOrder
```
saghul, line 385: Isn't this a more generic version of `preferredCodec` ?

JitsiConference.js
```diff
@@ -381,11 +381,20 @@ JitsiConference.prototype._init = function(options = {}) {
...
+                ? config.videoQuality.mobileCodecPreferenceOrder
+                : config.videoQuality?.codecPreferenceOrder,
+            disabledCodec: _getCodecMimeType(config.videoQuality?.disabledCodec),
```
saghul, line 388: While we are changing things, why not make this also a list of disabled codecs?

JitsiConference.js
```diff
@@ -563,8 +572,8 @@ JitsiConference.prototype._init = function(options = {}) {
             'region', config.deploymentInfo.userRegion);
     }
 
-    // Publish the codec type to presence.
-    this.setLocalParticipantProperty('codecType', this.codecSelection.getPreferredCodec());
```
saghul, line 567: How is backwards compatibility going to work here?

modules/RTC/CodecSelection.js
```diff
@@ -21,40 +21,67 @@ export class CodecSelection {
...
+                // Select all codecs that are supported by the browser.
+                selectedOrder = preferenceOrder
+                    .filter(codec => supportedCodecs.has(codec.toLowerCase()));
+
+                if (!browser.supportsVP9()) {
+                    const index = selectedOrder.findIndex(codec => codec.toLowerCase() === CodecMimeType.VP9);
+
+                    if (index !== -1) {
+                        selectedOrder.splice(index, 1);
+                        selectedOrder.push(CodecMimeType.VP9);
```
saghul, line 48: Maybe a comment here, indicating that we are putting VP9 last so at least decoding works would help future readers.

JitsiConference.js
saghul: As an example, someone might want to disable VP9 and AV1 to use VP8 only.

modules/RTC/CodecSelection.js
```diff
@@ -68,36 +95,20 @@ export class CodecSelection {
...
-        if (browser.isFirefox()) {
-            return true;
-        }
-
-        return window.RTCRtpReceiver
-            && window.RTCRtpReceiver.getCapabilities
-            && window.RTCRtpReceiver.getCapabilities('video').codecs
-            .some(codec => codec.mimeType.toLowerCase() === `video/${preferredCodec}`);
+    _getSupportedVideoCodecs() {
+        return VideoCodecMimeTypes.filter(codec => window.RTCRtpReceiver
```
saghul, line 103: Not sure if more readable?

```
(window.RTCRtpReceiver?.getCapabilities?.('video')?.codecs ?? []).some(...)
```

modules/RTC/CodecSelection.spec.js
```diff
@@ -0,0 +1,214 @@
...
+     * @param {MockParticipant} endpoint
+     */
+    removeParticipant(endpoint) {
+        this.participants = this.participants.filter(p => p !== endpoint);
+        this._signalingLayer.setPeerMediaInfo(false, endpoint.getId());
+        this.eventEmitter.emit(JitsiConferenceEvents.USER_LEFT);
+    }
+}
+
+describe('Codec Selection', () => {
```
saghul, line 82: ❤️

modules/RTC/CodecSelection.spec.js
```diff
@@ -0,0 +1,214 @@
...
+
+            expect(jingleSession.setVideoCodecs).toHaveBeenCalledWith([ 'vp8' ]);
+
+            // Make p1 leave the call
+            conference.removeParticipant(participant1);
+            expect(jingleSession.setVideoCodecs).toHaveBeenCalledWith([ 'vp9', 'vp8' ]);
+        });
+    });
+
+    describe('when deprecated configs are used in config.js', () => {
```
saghul, line 165: Can you add a test which tests the opposite prepferences? Web prefers vp9 and disabled h264 and mobile the opposite. It should result in vp8.

modules/RTC/MockClasses.js
```diff
@@ -1,4 +1,110 @@
+import transform from 'sdp-transform';
+
 /* eslint-disable no-empty-function */
+/* eslint-disable no-useless-constructor */
```
saghul, line 4: Is the empty constructor necessary?

modules/RTC/MockClasses.js
```diff
@@ -1,4 +1,110 @@
...
+            'a=fmtp:126 0-15\r\n',
+            'a=fmtp:111 minptime=10;useinbandfec=1\r\n',
+            'a=rtcp:9 IN IP4 0.0.0.0\r\n',
+            'a=setup:active\r\n',
+            'a=mid:0\r\n',
+            'a=msid:26D16D51-503A-420B-8274-3DD1174E498F 8205D1FC-50B4-407C-87D5-9C45F1B779F0\r\n',
+            'a=sendrecv\r\n',
+            'a=ice-ufrag:tOQd\r\n',
+            'a=ice-pwd:3sAozs7hw6+2O6DBp2pt9fvY\r\n',
+            /* eslint-disable-next-line max-len */
```
saghul, line 41: Maybe disable it for the whole file?

modules/RTC/MockClasses.js
```diff
@@ -117,4 +241,46 @@ export class MockRTC {
...
+                codecType
+            };
+        } else {
+            this._remoteSourceState[endpointId] = undefined;
+        }
+    }
+}
+
 /* eslint-enable no-empty-function */
+/* eslint-enable no-useless-constructor */
```
saghul, line 286: No need to do this, there are per file.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1688,19 +1688,15 @@ TraceablePeerConnection.prototype._mungeCodecOrder = function(description) {
...
+        for (const codec of this.codecSettings.codecList.slice().reverse()) {
+            SDPUtil.preferCodec(mLine, codec);
 
             // Strip the high profile H264 codecs on mobile clients for p2p connection. High profile codecs give better
             // quality at the expense of higher load which we do not want on mobile clients. Jicofo offers only the
             // baseline code for the jvb connection and therefore this is not needed for jvb connection.
             // TODO - add check for mobile browsers once js-utils provides that check.
-            if (this.codecSettings.preferred === CodecMimeType.H264 && browser.isReactNative() && this.isP2P) {
-                SDPUtil.stripCodec(mLine, this.codecSettings.preferred, true /* high profile */);
+            if (codec === CodecMimeType.H264 && browser.isReactNative() && this.isP2P) {
```
saghul, line 1698: Why not remove the TODO while we're at it, and change the check to isMobileDevice?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1907,6 +1903,24 @@ TraceablePeerConnection.prototype.getConfiguredVideoCodec = function() {
     return defaultCodec;
 };
 
+/**
+ * Returns the codecs in the current order of preference as configured on the peerconnection.
+ *
+ * @returns {Array}
+ */
+TraceablePeerConnection.prototype.getConfiguredVideoCodecs = function() {
+    const sdp = this.peerconnection.localDescription?.sdp;
```
saghul, line 1912: Maybe shield against calling this when `this.peerconnection` is null ?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1907,6 +1903,24 @@ TraceablePeerConnection.prototype.getConfiguredVideoCodec = function() {
...
+
+    if (!sdp) {
+        return [];
+    }
+    const parsedSdp = transform.parse(sdp);
+    const mLine = parsedSdp.media.find(m => m.type === MediaType.VIDEO);
```
saghul, line 1918: Can there be a case in which there is no video m line?

modules/browser/BrowserCapabilities.js
```diff
@@ -37,6 +37,17 @@ export default class BrowserCapabilities extends BrowserDetection {
...
 
+    /**
+     * Checks if the client is running on an Android browser.
+     *
+     * @returns {boolean}
+     */
+    isAndroidBrowser() {
+        const { userAgent } = navigator;
+
+        return this.isIosBrowser() || (!this.isReactNative() && userAgent.match(/Android/i));
```
saghul, line 48: Why does this return true if it's an iOS browser?

service/RTC/CodecMimeType.js
```diff
@@ -34,5 +34,11 @@ const CodecMimeType = {
     VP9: 'vp9'
 
 };
+const AudioCodecMimeTypes = [ CodecMimeType.OPUS, CodecMimeType.ULPFEC ];
+const VideoCodecMimeTypes = [ CodecMimeType.VP9, CodecMimeType.VP8, CodecMimeType.H264 ];
 
-module.exports = CodecMimeType;
+module.exports = {
```
saghul, line 40: While you're here, maybe use ES6 exports?

saghul (commented):
Did a round!

JitsiConference.js
jallamsetty1: The code assumes that older clients that publish only one codec in presence supports only that particular codec - https://github.com/jitsi/lib-jitsi-meet/pull/2296/files#diff-3ed599db7a8e6368e4f4b60ff8eda596e4a49fbd800dbe7128342ff3fc2735cbR131
Since the older clients don't publish the disabled codecs in presence, its safer to assume that the preferred codec is the only supported one.

JitsiConference.js
saghul: Aha!

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Yes, didn't realize that! will do

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Nope, `offerToReceiveVideo` guarantees that they will always be a video m-line.

modules/browser/BrowserCapabilities.js
jallamsetty1: Oops, I moved things around and forgot to remove this one, will delete it.

modules/RTC/MockClasses.js
jallamsetty1: Doesn't seem to be, removed it.

jallamsetty1:
Failure seems to be a bridge issue where the bridge stops sending media to p2.

JitsiConference.js
```diff
@@ -381,11 +381,20 @@ JitsiConference.prototype._init = function(options = {}) {
 
     // Get the codec preference settings from config.js.
     const codecSettings = {
-        jvbDisabledCodec: _getCodecMimeType(config.videoQuality?.disabledCodec),
-        p2pDisabledCodec: _getCodecMimeType(config.p2p?.disabledCodec),
-        enforcePreferredCodec: config.videoQuality?.enforcePreferredCodec,
-        jvbPreferredCodec: _getCodecMimeType(config.videoQuality?.preferredCodec),
-        p2pPreferredCodec: _getCodecMimeType(config.p2p?.preferredCodec)
+        jvb: {
```
saghul, line 384: Question: wouldn't it make sense to compute the new setting here even if the deprecated ones are used and then only use the new one internally?

saghul (approved):
LGTM with a question. Can you pl test mobile before landing?

JitsiConference.js
jallamsetty1: I plan to remove handling of deprecated settings after few releases so its going to look much cleaner then. I don't think its good idea to move all the validation logic here, IMHO it needs to be handled in the CodecSelection constructor itself.

modules/RTC/CodecSelection.js
```diff
@@ -2,12 +2,16 @@
...
-import { CodecMimeType, VideoCodecMimeTypes } from '../../service/RTC/CodecMimeType';
+import CodecMimeType from '../../service/RTC/CodecMimeType';
 import { MediaType } from '../../service/RTC/MediaType';
 import browser from '../browser';
 
 const logger = getLogger(__filename);
 
+// Default video codec preferences on mobile and desktop endpoints.
+const DESKTOP_VIDEO_CODEC_ORDER = [ CodecMimeType.VP9, CodecMimeType.VP8, CodecMimeType.H264 ];
+const MOBILE_VIDEO_CODEC_ORDER = [ CodecMimeType.VP8, CodecMimeType.VP9, CodecMimeType.H264 ];
```
saghul, line 13: On P2P we want H.264 first though...

saghul (changes requested):
One minor thing (RN would like to prefer H.264 for P2P) otherwise LGTM, very excited for this to land!

modules/RTC/CodecSelection.js
jallamsetty1: Do we need to leave to out VP9 for p2p though ? If its at the end of the list, it will be negotiated but never used.

saghul: I'd say that's ok

saghul (approved):

PR #2297, opened 2023-06-19
conference leave can block under certain circumstances

luocongqiu:


luocongqiu:
@jitsi-jenkins yes I sign a icla

JitsiConference.js
```diff
@@ -712,7 +712,8 @@ JitsiConference.prototype.leave = async function(reason) {
     let leaveError;
 
     try {
-        await room.leave(reason);
+        await room.leave(reason)
+            .catch(() => {}); // eslint-disable-line no-empty-function
```
saghul, line 716: Why do you need this? It will break the logic below because onMemberLeft won't be called in that case.

modules/xmpp/ChatRoom.js
```diff
@@ -1865,6 +1865,7 @@ export default class ChatRoom extends Listenable {
                 // let's just clean
                 this.connection.emuc.doLeave(this.roomjid);
                 this.clean();
+                resolve();
```
saghul, line 1868: This looks correct, good catch!

saghul (changes requested):
Please undo the first change, the second one is ok.

saghul:

saghul (approved):

saghul:

saghul:
Rebased in https://github.com/jitsi/lib-jitsi-meet/pull/2349

PR #2299, opened 2023-06-25
Adding the ability to modify content hint

ProdigyView:
Adds the ability to change a tracks contentHint.  For more information on contentHint, reads here: https://webrtc.github.io/samples/src/content/capture/video-contenthint/


saghul:
Maybe it should be called "desktopContentHint", since we only apply it to tracks for desktop sharing?

PR #2302, opened 2023-06-27
fix(RTCRtpSender) Race condition on resetting transaction id.

mykhailo-ua:
Adjusted wrapping method to have RTCRtpSender setting parameters in sequential manner in order to solve race condition when transaction id can be reset at the end of videoSender setParameters and cause exception if another update was already in fly and before its check for transaction id presence.

The issue "Failed to execute 'setParameters' on 'RTCRtpSender': Failed to set parameters since getParameters() has never been called" happened because the actual setting params were not called sequentially, the function was executed and promise was chained what broke the logic - [code_link](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/TraceablePeerConnection.js#L2654)

Similar change was made in past - [github_link](https://github.com/jitsi/lib-jitsi-meet/pull/2228)

Also after the change I run npm run test/lint
Screenshot 2023-06-27 at 10 18 11

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -472,13 +472,12 @@ export class TPCUtils {
             return Promise.resolve();
         }
         parameters.encodings = this._getStreamEncodings(track);
-        const promise = transceiver.sender.setParameters(parameters);
 
         if (mediaType === MediaType.VIDEO) {
-            return this.pc._updateVideoSenderParameters(promise);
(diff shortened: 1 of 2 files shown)


mykhailo-ua:
@jallamsetty1 could you please take a look?

jallamsetty1 (approved):

jallamsetty1:
Jenkins test this please.

Issue #2298, opened 2023-06-21
Screening Sharing Not Getting Above 5 FPS


ProdigyView:
I am re-opening the issue presented here:

https://github.com/jitsi/lib-jitsi-meet/issues/1089

But also the issue isn't just with Jitsi but a bug with Chrome as presented here:

https://bugs.chromium.org/p/webrtc/issues/detail?id=13016

The fix in the above was to set contentHint to motion. In Jitsi, I accomplished this will the following:

```
for (const track of tracks) {
                    
                    if (track.type === MediaType.VIDEO
                        && track.videoType === 'desktop') {
                        this.setVideoTrackContentHints(track.track, 'motion');
                    }
                }
```

The section of the code that was changed was this: https://github.com/jitsi/lib-jitsi-meet/blob/9db908a3658c05674f48e8dd356ceafe8bb5fa88/JitsiMeetJS.ts#LL336C16-L336C16

But it seems to be still be capped at 5 fps. Any thoughts?

DanielMcAssey:
For desktop share the content hint should stay as detail, if set to motion it degrades the quality significantly that its hard to read text on desktop shares.

saghul:
Ping @jallamsetty1

jallamsetty1:
I agree with @ProdigyView that we should set the contentHint to 'motion' when the captured fps > 5. I am working on fixing this. For low fps screenshare, we set the contentHint to 'detail' and send only the highest spatial resolution and for high fps screenshare we need to set the contentHInt to 'motion' and send all the spatial-temporal layers so that the jvb can pick the best spatial-temporal layer based on the receivers BWE.

PR #2304, opened 2023-06-28
Fix screenshare issues

jallamsetty1:
- fix(JingleSession) Reconfigure the stream encodings after p2p->jvb switch.

Fixes blurry screenshare in some cases. There is a higher probability of the higher layers getting suspended when all the stream encodings are enabled for low fps SS. Make sure only the highest spatial layer is sent for low fps SS after p2p->jvb switch. Make sure all the stream encodings are enabled for high fps SS after p2p->jvb switch.

- fix(ScreenObtainer): Apply contentHint on captured desktop track correctly.

contentHint needs to be set to 'motion' for high fps SS and 'detail' otherwise. Fixes https://github.com/jitsi/lib-jitsi-meet/issues/2298.

### DIFF
--- a/JitsiMeetJS.ts
+++ b/JitsiMeetJS.ts
@@ -334,15 +334,6 @@ export default {
                     }
                 }
 
-                // set the contentHint to "detail" for desktop tracks
-                // eslint-disable-next-line prefer-const
-                for (const track of tracks) {
-                    if (track.type === MediaType.VIDEO
-                        && track.videoType === 'desktop') {
-                        this.setVideoTrackContentHints(track.track, 'detail');
-                    }
-                }
-
                 return tracks;
             })
             .catch(error => {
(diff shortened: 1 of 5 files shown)


JitsiMeetJS.ts
```diff
@@ -531,24 +522,6 @@ export default {
...
     },
 
-    /**
-     * Set the contentHint on the transmitted stream track to indicate
-     * charaterstics in the video stream, which informs PeerConnection
-     * on how to encode the track (to prefer motion or individual frame detail)
-     * @param {MediaStreamTrack} track - the track that is transmitted
-     * @param {String} hint - contentHint value that needs to be set on the track
-     */
-    setVideoTrackContentHints(track, hint) {
```
saghul, line 541: Doesn't this break the public API?

saghul (approved):
Approving but I think we should retain the public API function and turn it into a noop and log a warning.

JitsiMeetJS.ts
jallamsetty1: Actually this is not documented as a public API, this should have been a private function when it was added.

PR #2305, opened 2023-06-29
Move rtcstats to lib

nils-ohlmeier:
Folks who are only using lib-jitsi-meet would like to still be able to utilize rtcstats.
Thus this PR is moving rtcstats into lib-jitsi-meet.
There is going to be another PR for jitsi-meet which allows to configure which of two version to utilize.


JitsiMeetJS.ts
```diff
@@ -166,6 +167,8 @@ export default {
             Statistics.sendLog(JSON.stringify(logObject));
         }
 
+        RtcStats.init(options);
```
saghul, line 170: Any reason not to do this inside Statistics.init?

JitsiMeetJS.ts
```diff
@@ -194,6 +197,14 @@ export default {
         Logger.setLogLevel(level);
     },
 
+    getRtcStatsTrace() {
```
saghul, line 200: Can we group the API in an object so we can use it like so: `JitsiMeetJS.rtcstats.getTrace()` ?

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
+import Logger from '@jitsi/logger';
```
saghul, line 1: Please make this TypeScript since it's new code.

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+/**
+ * Filter out RTCPeerConnection that are created by callstats.io.
+ *
+ * @param {*} config - Config object sent to the PC c'tor.
+ * @returns {boolean}
+ */
+function connectionFilter(config) {
+    if (config?.iceServers[0] && config.iceServers[0].urls) {
+        for (const iceUrl of config.iceServers[0].urls) {
+            if (iceUrl.indexOf('callstats.io') >= 0) {
```
saghul, line 16: you can use `string.includes(string2)` instead.

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+    meetingFqn = null;
+    setMeetingF;
+    meetingFqnSet = new Promise(r => {
+        this.setMeetingF = r;
+    });
+
+    /**
+     * Initializes the underlying rtcstats node moduke.
+     */
+    init(options) {
```
saghul, line 44: There needs to be a way to "de-init" since mobile needs that.

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+        const { rtcstatsUseLibJitsi } = options.analytics;
```
saghul, line 45: Why would we want this option? Can't we migrate in a single shot?

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+        return this.traceConfigured.then(t => t);
+    }
+
+    /**
+     * Set the meetingFqn value needed for initialzing rtcstats.
+     *
+     * @param {String} - the meeting fqn.
+     * @returns {void}
+     */
+    setMeetingFqn(config) {
```
saghul, line 99: Can you call this `options`, not to shadow the global `config` ?

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * RTCStats client can notify the App of any PeerConnection related events.
+     *
+     * @param {Object} event - Rtcstats event.
+     * @returns {void}
+     */
+    handleRtcStatsEvent(event) {
+        logger.info(`RtcStats received event ${JSON.stringify(event)}`);
```
saghul, line 117: Won't this be too chatty? Shouldn't it at least be debug?

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+    /**
+     * Handle WS close.
+     *
+     * @param {Object} closeEvent - Event sent by ws onclose.
+     * @returns {void}
+     */
+    handleTraceWSClose(closeEvent) {
+        logger.warn('RtcStats trace WS closed', closeEvent);
+        if (typeof this.closeCallback === 'function') {
+            this.closeCallback(closeEvent);
```
saghul, line 132: Use `this.closeCallback.?()` instead, so there is no need to check if it's defined.

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Handle WS close.
+     *
+     * @param {Object} closeEvent - Event sent by ws onclose.
+     * @returns {void}
+     */
+    handleTraceWSClose(closeEvent) {
+        logger.warn('RtcStats trace WS closed', closeEvent);
```
saghul, line 130: Reconnects happen inside rtcstats, right? If so, shouldn't there be a way to notify users about when we re-connected?

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+}
+
+/**
+ * Class the controls the rtcstats node module.
+ */
+class RtcStats {
+    initialized = false;
+    trace = null;
+    eventCallback = null;
```
saghul, line 29: I think it would be more ergonomic if we used event emitter for these events. It will also be easier to extend with more events in the future.

saghul (changes requested):

modules/RtcStats/RtcStats.js
nils-ohlmeier: See https://github.com/jitsi/jitsi-meet/pull/13508#issuecomment-1615110001
I think it is easier to be able to simply turn the feature off via a config flag, rather than having to roll back a software release (plus once a software release got rolled back because of an issue, the fix for the issues remains a blocker for any new releases, unless you backout the whole PR before doing a new release).

modules/RtcStats/RtcStats.js
nils-ohlmeier: Is there a specific place this de-init needs to be called from?

modules/RtcStats/RtcStats.js
nils-ohlmeier: Interesting question.
It looks like the client side reconnection PR got merged https://github.com/jitsi/rtcstats/pull/28 and it just internally buffers all messages which didn't get send to the server successfully. But it is a good question why we now only have a notification for close, but not for reconnect. If the rtcstats library handles everything internally even the disconnect should not get exposed. 
@andrei-gavrilescu do you have any thoughts on how to proceed here? Should we removed the trace WS close callback since everything get handled internally? Or should we expose the reconnect event as well (but then we would need to make it clear that still everything in between got buffer)?

modules/RtcStats/RtcStats.js
saghul: Init will be called a second time, so as long as that works we are good!

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+    }
+
+    /**
+     * Set the meetingFqn value needed for initialzing rtcstats.
+     *
+     * @param {String} - the meeting fqn.
+     * @returns {void}
+     */
+    setMeetingFqn(config) {
+        const { meetingName, eventCallback, onCloseCallback } = config;
```
andrei-gavrilescu, line 100: Couldn't we initialize rtcstats simply when we have all the option available, so we don't go through the promise dance. i.e. instead of the jitsi-meet calling setMeetingFqn, it would call init.
Technically jitsi-meet calls init when the LIB_WILL_INIT event occurs in jitsi-meet, which is fired from lib-jitsi-meet afaik, so we could completely decouple the init process from jitsi-meet, but this works for a first iteration.

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+
+    /**
+     * Initializes the underlying rtcstats node moduke.
+     */
+    init(options) {
+        const { rtcstatsUseLibJitsi } = options.analytics;
+
+        if (rtcstatsUseLibJitsi) {
+            if (!this.initialized) {
+                this.meetingFqnSet.then(() => {
```
andrei-gavrilescu, line 49: I'm assuming that this will somehow reset when adding the de-init step which @saghul mentioned, otherwise the promise will be called a second time on mobile (when a user joins another meeting) but by that time it will already have been resolved.

modules/RtcStats/RtcStats.js
```diff
@@ -0,0 +1,138 @@
...
+
+                    const rtcstatsOptions = {
+                        connectionFilter,
+                        pollInterval: rtcstatsPollInterval,
+                        useLegacy: rtcstatsUseLegacy,
+                        sendSdp: rtcstatsSendSdp,
+                        eventCallback: this.handleRtcStatsEvent.bind(this)
+                    };
+
+                    this.trace = traceInit(traceOptions);
```
andrei-gavrilescu, line 70: When reconnecting on mobile we need to reset the trace and reinitialize, which happens here, however we don't want to call rtcstatsInit otherwise it will proxy the global webrtc and gum objects again.
In jitsi-meet this is guarded by this check [here](https://github.com/jitsi/jitsi-meet/blob/a3c9630a19e1f78cea7112ecf10117e5d45020c4/react/features/rtcstats/RTCStats.ts#L138C16-L138C21)

modules/RtcStats/RtcStats.js
nils-ohlmeier: Oh yes even on the web this gets initialized several times. I haven't looked into the why, but it works.

modules/RtcStats/RtcStats.js
nils-ohlmeier: From my testing on mobile the worries around re-init are already been taken care of by George code, which re-inits all of the rtcstats code. But I'll double check for the room name to make sure.

modules/RtcStats/RtcStats.js
saghul: I think this might require some more thought, let's punt on it.

modules/RtcStats/RtcStats.js
andrei-gavrilescu: The scenario which could happen (I think). Is the following.

`if (changed) {
            this.reset();

            if (newOptions.meetingFqn && newOptions.endpoint) {
                this.init(newOptions);
            } else {
                logger.warn('RTCStats is enabled but is waiting for meetingFqn and endpoint.');
            }
        }
    }
   `
   This would trigger
   `
       init(options: InitOptions) {

        const { useLibRtcStats, meetingFqn } = options;

        if (useLibRtcStats) {
            logger.info('setting up libRtcStats');

            const rtcOptions = {
                meetingName: meetingFqn,
                eventCallback: this.handleRtcstatsEvent.bind(this),
                onCloseCallback: this.handleTraceWSClose.bind(this)
            };

            JitsiMeetJS.setRtcStatsMeetingFqn(rtcOptions);

            JitsiMeetJS.getRtcStatsTrace().then((traceF: any) => {
                this.setTrace(traceF);
            });

        }
   `
So on mobile when a user joins another room, we would reset the connection (delete the trace) but `JitsiMeetJS.setRtcStatsMeetingFqn(rtcOptions);`  would be called again and by this point the promise was already resolved, additional resolve calls would have no effect, unless we somehow reset the promise.
From what I understand ljm init will be called a second time but the RtcStats instance inside is a singleton so the class and hence the promise doesn't get re-initialized, @saghul is this right?

andrei-gavrilescu: The scenario which could happen (I think). Is the following.
```
if (changed) {
            this.reset();

            if (newOptions.meetingFqn && newOptions.endpoint) {
                this.init(newOptions);
            } else {
                logger.warn('RTCStats is enabled but is waiting for meetingFqn and endpoint.');
            }
        }
    }
```
This would trigger:
```
     init(options: InitOptions) {

        const { useLibRtcStats, meetingFqn } = options;

        if (useLibRtcStats) {
            logger.info('setting up libRtcStats');

            const rtcOptions = {
                meetingName: meetingFqn,
                eventCallback: this.handleRtcstatsEvent.bind(this),
                onCloseCallback: this.handleTraceWSClose.bind(this)
            };

            JitsiMeetJS.setRtcStatsMeetingFqn(rtcOptions);

            JitsiMeetJS.getRtcStatsTrace().then((traceF: any) => {
                this.setTrace(traceF);
            });

        }
 ```
So on mobile when a user joins another room, we would reset the connection (delete the trace) but JitsiMeetJS.setRtcStatsMeetingFqn(rtcOptions);  would be called again and by this point the promise was already resolved, additional resolve calls would have no effect, unless we somehow reset the promise.
From what I understand ljm init will be called a second time but the RtcStats instance inside is a singleton so the class and hence the promise doesn't get re-initialized, @saghul is this right?

saghul: Sounds about right! Which is wrong :-P

modules/RtcStats/RtcStats.js
nils-ohlmeier: Sorry I ran out of time to make more progress on this.

saghul:
Superseded by https://github.com/jitsi/lib-jitsi-meet/pull/2335

PR #2307, opened 2023-07-04
fix(JingleSession) actually call browser.supportsCodecPreferences

NewEraCracker:
supportsCodecPreferences is a function, and must be called. Am I right?

Thanks.

PS: needs testing as AFAIK _mungeCodecOrder was never getting called.


saghul (approved):
🤦  I missed this during review, thanks for being in the lookout!

PR #2309, opened 2023-07-06
fix(TPCUtils): Fix suspending media on jvb.

jallamsetty1:
Fixes a regression introduced in https://github.com/jitsi/lib-jitsi-meet/commit/17ade96a70971b7429b6beac73b919885e665f16. [JitsiConference.js] Failed to suspend media transfer over the JVB connection: Error: Failed to change encodings on the RTCRtpSendersInvalidStateError: getParameters() needs to be called before setParameters().

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -109,6 +109,25 @@ export class TPCUtils {
     }
 
     /**
+     * Updates the sender parameters in the stream encodings.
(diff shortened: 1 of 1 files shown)


saghul (approved):

jallamsetty1:
The test failure is a genuine failure that I am able to reproduce manually, still looking into it so pls don't merge this.

PR #2311, opened 2023-07-10
watchRTC as a feature for Jitsi meet

subhamcyara:
Following PR contains changes required to add watchRTC as a module within lib jitsi meet

This involves adding

- watchRTC SDK package (npm i @testrtc/watchrtc-sdk).
- watchRTC SDK is initialized within Statistics.init to proxy WebRTC functions such as GUM and RTCPeerConnection.
- watchRTC SDK then starts collecting required stats when the conference begins. Here we check if the user has not configured room name or user/peer name, if not, then use relevant data available from Jitsi.
- watchRTC SDK does not require to function for react native. 
- checks implemented to run watchRTC SDK when
-- watchRTC is enabled
-- rtcstats must be disabled
-- analytics should be enabled
-- third party requests are not disabled

### DIFF
--- a/modules/watchRTC/WatchRTC.ts
+++ b/modules/watchRTC/WatchRTC.ts
@@ -1,0 +1,82 @@
+
+import Logger from '@jitsi/logger';
+import watchRTC from '@testrtc/watchrtc-sdk';
+
+import { isAnalyticsEnabled, isRtcstatsEnabled, isWatchRTCEnabled } from './functions';
+import { IWatchRTCConfiguration } from './interfaces';
+
+const logger = Logger.getLogger(__filename);
+
+/**
+ * Class that controls the watchRTC flow, because it overwrites and proxies global function it should only be
+ * initialized once.
+ */
+class WatchRTCHandler {
+    options?: IWatchRTCConfiguration;
+    /**
+     * Initialize watchRTC, it overwrites GUM and PeerConnection global functions and adds a proxy over them
+     * used to capture stats.
+     *
+     * @param {Object} options - Init options.
+     * @returns {void}
+     */
+    init(options: any): void {
+
+        if (isWatchRTCEnabled(options)) {
+
+            if (!isAnalyticsEnabled(options)) {
+                logger.error('Cannot initialize WatchRTC when analytics or third party requests are disabled.');
+                return;
+            }
+
+            if (isRtcstatsEnabled(options)) {
+                logger.error('Cannot initialize WatchRTC when RTCStats is enabled.');
+                return;
+            }
+
+            // watchRTC "proxies" WebRTC functions such as GUM and RTCPeerConnection by rewriting the global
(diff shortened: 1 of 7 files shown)


modules/watchRTC/WatchRTC.ts
```diff
@@ -0,0 +1,82 @@
...
+                logger.error('Cannot initialize WatchRTC when analytics or third party requests are disabled.');
+                return;
+            }
+
+            if (isRtcstatsEnabled(options)) {
+                logger.error('Cannot initialize WatchRTC when RTCStats is enabled.');
+                return;
+            }
+
+            // watchRTC "proxies" WebRTC functions such as GUM and RTCPeerConnection by rewriting the global
```
saghul, line 37: This is no longer the case. We used to store `window.PeerConnection` but we no longer do. Or did we miss a spot?

package.json
```diff
@@ -20,6 +20,7 @@
     "@jitsi/logger": "2.0.0",
     "@jitsi/sdp-interop": "git+https://github.com/jitsi/sdp-interop#3d49eb4aa26863a3f8d32d7581cdb4321244266b",
     "@jitsi/sdp-simulcast": "0.4.0",
+    "@testrtc/watchrtc-sdk": "^1.36.3",
```
saghul, line 23: Please use the exact version, aka, leave ^ out.

saghul:
Please see the test failure, we are going to need to bump the library size.

modules/watchRTC/WatchRTC.ts
subhamcyara: I am not sure I understand. Is this about the verbose comment written and behaviour of the lib jitsi meet written within the comment ?!

saghul: It's about the comment.

subhamcyara:
@saghul , I have tried to address all review points. Please let me know your feedback.
Thanks

saghul:
The test failures seem legit. I think we should gate starting of watchRTC if it's not enabled or no room name was provided.

subhamcyara:
I believe the failure should resolve now. For context, there are checks to see if watchRTC is enabled or not. The idea is to use the roomName and/or peerName from the watchRTC config and if not supplied, we can use what jitsi provides. If we get nothing from jitsi too, watchRTC SDK would handle it.

saghul (approved):
LGTM. @andrei-gavrilescu can you PTAL too?

andrei-gavrilescu:
Looks good 👍

saghul:
@subhamcyara Can you please rebase? Looks like there is some small conflict. Once solved it's good to merge.

saghul (approved):

subhamcyara:
Hi, thanks for the review here. I believe we might also need to take a look at https://github.com/jitsi/jitsi-meet/pull/13527 for changes related to JM.

saghul:

PR #2312, opened 2023-07-10
feat: Handles join error about required display name.

damencho:


JitsiConferenceEventManager.js
```diff
@@ -199,6 +199,11 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
         JitsiConferenceEvents.CONFERENCE_FAILED,
         JitsiConferenceErrors.CONNECTION_ERROR);
 
+    this.chatRoomForwarder.forward(XMPPEvents.DISPLAY_NAME_REQUIRED,
+        JitsiConferenceEvents.CONFERENCE_FAILED,
+        JitsiConferenceErrors.DISPLAY_NAME_REQUIRED);
+    chatRoom.addListener(XMPPEvents.DISPLAY_NAME_REQUIRED, () => conference.leave());
```
saghul, line 205: Won't this trigger CONFERENCE_ENDED ?

damencho: We haven't joined yet, so no. We do the same in several other occasions like max-occupabta reach and reservation error. This is to clean internals otherwise if you try again it will fail creating the conference as such conference already exists.

hristoterezov (approved):

JitsiConferenceEventManager.js
damencho: And I need to drop that and few more ... as this is already done internally in jitsi-meet. 
https://github.com/jitsi/jitsi-meet/blob/364e63da144e2f9569d324cb90db4ea9c4691be7/react/features/base/conference/middleware.any.ts#L247

hristoterezov (approved):

PR #2313, opened 2023-07-11
fix(AsyncQueue) use async-es

saghul:
It's the same as async, but with ESM which is more bundle friendly.


saghul:
It saves us ~20KB of the final bundle size!

damencho (approved):

PR #2316, opened 2023-07-13
Update jwt docs key generation.

damencho:


saghul:
Can we move this to the handbook please? And just leave a link there in this file.

saghul (approved):

damencho:
Yes we can do that.

PR #2319, opened 2023-07-13
fix(screenshare): Ignore mute events on remote desktop tracks.

jallamsetty1:
Workaround for Chromium bug https://bugs.chromium.org/p/chromium/issues/detail?id=1099280 which results in client switching to showing avatar instead of static content after 10 secs.


saghul (approved):

jallamsetty1:
Test failure is a known unrelated failure.

PR #2322, opened 2023-07-19
fix(CodecSelection) Switch codec to VP8 when E2EE is enabled.

jallamsetty1:
E2EE is not supported by the application when VP9 is the selected codec. Therefore, switch to using VP8 for video encode/decode.

### DIFF
--- a/modules/RTC/CodecSelection.js
+++ b/modules/RTC/CodecSelection.js
@@ -70,12 +70,17 @@ export class CodecSelection {
             // Push VP9 to the end of the list so that the client continues to decode VP9 even if its not
             // preferable to encode VP9 (because of browser bugs on the encoding side or added complexity on mobile
             // devices).
-            if (!browser.supportsVP9()) {
+            if (!browser.supportsVP9() || this.conference.isE2EEEnabled()) {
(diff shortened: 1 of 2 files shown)


damencho (approved):

modules/RTC/CodecSelection.js
```diff
@@ -124,7 +143,7 @@ export class CodecSelection {
         }
         const currentCodecOrder = session.peerconnection.getConfiguredVideoCodecs();
         const localPreferredCodecOrder = session === this.conference.jvbJingleSession
-            ? this.codecPreferenceOrder.jvb
+            ? this._maybeFilterJvbCodecs()
             : this.codecPreferenceOrder.p2p;
```
saghul, line 147: Why don't we apply the same filtering logic for P2P?

jallamsetty1: P2P is disabled whenever E2EE is enabled. Should we even enable E2EE for P2P?

saghul: Oh I forgot about that... we should, disabling it was a stop-gap measure.

I can't recall what the problem was though.

PR #2327, opened 2023-07-26
ref(JingleSession) Fixes media handling and renegotiations for p2p.

jallamsetty1:
1. For renegotiations triggered by the application, it is ok to execute sRD->cA-sLD cycle on the p2p initiator instead of executing cO->sLD->sRD cycle. Chrome sometimes creates a third m-line for p2p when createOffer is called which is unexpected and therefore it breaks the client's renegotiation cycle. This has been done in other places already, i.e., when remote sources are signaled, etc. so making it uniform across all renegotiations.

2. fix(JingleSession) Modify encoding to stop/start outgoing media on p2p.
This is much faster than changing direction on the transceiver and doesn't need a renegotiation for the suspending/resuming media. Also fixes an issue with audio-only mode in p2p where media doesn't resume when audio-only mode is disabled if it was enabled while the jvb connection was active.

3. fix(TPC): Make sure all RTCRtpSender:setParameters calls are sequential.
Recommit https://github.com/jitsi/lib-jitsi-meet/commit/17ade96a70971b7429b6beac73b919885e665f16 with the jvb leak fix

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -498,14 +498,17 @@ export class TPCUtils {
      * Resumes or suspends media on the peerconnection by setting the active state on RTCRtpEncodingParameters
      * associated with all the senders that have a track attached to it.
      *
-     * @param {boolean} enable - whether media needs to be enabled or suspended.
+     * @param {boolean} enable - whether outgoing media needs to be enabled or disabled.
+     * @param {string} mediaType - media type, 'audio' or 'video', if neither is passed, all outgoing media will either
+     * be enabled or disabled.
      * @returns {Promise} - A promise that is resolved when the change is succesful on all the senders, rejected
      * otherwise.
      */
-    setMediaTransferActive(enable) {
+    setMediaTransferActive(enable, mediaType) {
         logger.info(`${this.pc} ${enable ? 'Resuming' : 'Suspending'} media transfer.`);
 
-        const senders = this.pc.peerconnection.getSenders().filter(s => Boolean(s.track));
+        const senders = this.pc.peerconnection.getSenders()
+            .filter(s => Boolean(s.track) && (!mediaType || s.track.kind === mediaType));
         const promises = [];
 
         for (const sender of senders) {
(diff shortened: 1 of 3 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -108,6 +108,25 @@ export class TPCUtils {
...
     }
 
+    /**
+     * Updates the sender parameters in the stream encodings.
+     *
+     * @param {RTCRtpSender} sender - the sender associated with a MediaStreamTrack.
+     * @param {boolean} enable - whether the streams needs to be enabled or disabled.
+     * @returns {Promise} - A promise that resolves when the operation is successful, rejected otherwise.
+     */
+    _updateSenderParameters(sender, enable) {
```
hristoterezov, line 118: Maybe we should add `encodings` in the name of the method for clarity. WDYT?

modules/RTC/TPCUtils.js
```diff
@@ -485,43 +504,39 @@ export class TPCUtils {
...
-            const setActivePromise = sender.setParameters(parameters);
-
             if (sender.track.kind === MediaType.VIDEO) {
-                promises.push(this.pc._updateVideoSenderParameters(setActivePromise));
+                promises.push(this.pc._updateVideoSenderParameters(this._updateSenderParameters(sender, enable)));
             } else {
-                promises.push(setActivePromise);
+                const parameters = sender.getParameters();
+
+                parameters?.encodings?.length && (parameters.encodings[0].active = enable);
```
hristoterezov, line 538: Why we don't use `this._updateSenderParameters` here?

modules/RTC/TPCUtils.js
```diff
@@ -485,43 +504,39 @@ export class TPCUtils {
...
-            if (parameters?.encodings?.length) {
-                for (const encoding of parameters.encodings) {
-                    encoding.active = enable;
-                }
-            }
-            const setActivePromise = sender.setParameters(parameters);
-
             if (sender.track.kind === MediaType.VIDEO) {
-                promises.push(this.pc._updateVideoSenderParameters(setActivePromise));
+                promises.push(this.pc._updateVideoSenderParameters(this._updateSenderParameters(sender, enable)));
```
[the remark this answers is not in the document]
jallamsetty1: `this._updateSenderParameters()` is a method right? What am I missing?

modules/RTC/TPCUtils.js
jallamsetty1: Yes we can, I will update it.

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
jallamsetty1: ah I see, I am making the same mistake again 🤦‍♀️

hristoterezov (approved):

PR #2328, opened 2023-07-31
feat(compute-pressure) monitor cpu pressure

mihhu:


mihhu:

CLA signed

saghul:
Interesting choice, adding this to LJM instead of doing it in JM. Any reason for it? Generally speaking I _think_ this might fit better on the latter.

While being in a conference is the main activity, joining / leaving breakout rooms should likely not trigger attaching / detaching from the observer, I see this more like a global thing.

mihhu:
> Interesting choice, adding this to LJM instead of doing it in JM. Any reason for it? Generally speaking I _think_ this might fit better on the latter.

I thought of making it available for those who use only LJM too. 

> While being in a conference is the main activity, joining / leaving breakout rooms should likely not trigger attaching / detaching from the observer, I see this more like a global thing.

So should I drop this PR and move the observer in JM?

saghul:
> I thought of making it available for those who use only LJM too.

It's not a bad idea, and the line is blurry indeed, but since we currently don't do anything other that fire an event IMHO it's better suited for JM.

> So should I drop this PR and move the observer in JM?

I think so yeah.

PR #2329, opened 2023-08-01
fix(codec-selection) Pick only 1 VP9 codec for p2p negotiation.

jallamsetty1:
More than one VP9 payload type is produced in the offer, more experimental ones if the local endpoint has no local video source. Munge the SDP to remove all codecs besides the one with profile-id=0 so that only one VP9 codec is negotiated with p2p peer.


saghul (approved):
LGTM! Any chance the codec stripping could be tested ?

jallamsetty1:
> LGTM! Any chance the codec stripping could be tested ?

Like discussed, the changes are in TPC and it doesn't seem to be possible without duplicating the logic in the mock classes.

PR #2335, opened 2023-08-10
feat(rtcstats): move rtcstats to ljm

andrei-gavrilescu:
PR aims to move rtctats reporting to ljm.
The solution is meant to work out of the box, a client using ljm will start sending rtcstats compatible statitics by simply enabling it in the provided config.


JitsiMeetJS.ts
```diff
@@ -194,6 +196,24 @@ export default {
...
+       * @param {string} statsType - The type of stats to send.
+       * @param {Object} data - The stats data to send.
+       */
+      sendStatsEntry(statsType, data) {       
+        RTCStats.sendStatsEntry(statsType, null, data);
+      },
+      /**
+       * Events generated by rtcstats, such as PeerConnections state, and websocket connection state.
+       */
+      events: RTCStats.events
```
saghul, line 214: All event types are defined here: https://github.com/jitsi/lib-jitsi-meet/blob/77e6803f1915174ba4bce2cb36f942cce55d6005/JitsiMeetJS.ts#L112

So let's have those there.

Also, they should be constants, not an event emitter.

I'd say a sample usage should look like this:

```
JitsiMeetJS.rtcstats.on(JitsiMeetJS.events.rtcstats.RTC_STATS_PC_EVENT, ev => ...);
```

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+import EventEmitter from 'events';
+
+import { 
+    CONFERENCE_JOINED, 
+    CONFERENCE_LEFT, 
+    CONFERENCE_UNIQUE_ID_SET 
+} from '../../JitsiConferenceEvents';
+import JitsiConference from '../../JitsiConference';
+
+const RTC_STATS_WC_DISCONNECTED = 'rtcstats_ws_disconnected';
```
saghul, line 12: These need to be exported so they can be attached in the JitsiMeetJS.events.rtcstats object.

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+const RTC_STATS_PC_EVENT = 'rtstats_pc_event';
+
+/**
+ * Filter out RTCPeerConnection that are created by callstats.io.
+ *
+ * @param {*} config - Config object sent to the PC c'tor.
+ * @returns {boolean}
+ */
+function connectionFilter(config) {
+    if (config?.iceServers[0] && config.iceServers[0].urls) {
```
saghul, line 22: Is it enough to only took at the first ice server?

I think you can simplify the check with:

```
        for (const iceUrl of (config?.iceServers ?? [])[0]?.urls ?? []) {
```

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+
+    /**
+     * RTCStats "proxies" WebRTC functions such as GUM and RTCPeerConnection by rewriting the global objects.
+     * The proxies will then send data to the rtcstats server via the trace object.
+     * The initialization procedure must be called once when lib-jitsi-meet is loaded.
+     * 
+     * @param initConfig initial config for rtcstats.
+     * @returns {void}
+     */
+    init(initConfig: any) {
```
saghul, line 48: Let's type the options please.

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+     * The proxies will then send data to the rtcstats server via the trace object.
+     * The initialization procedure must be called once when lib-jitsi-meet is loaded.
+     * 
+     * @param initConfig initial config for rtcstats.
+     * @returns {void}
+     */
+    init(initConfig: any) {        
+        const {
+            analytics: {
+                rtcstatsUseLegacy: useLegacy = false,
```
saghul, line 51: Does this make sense still?

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+                rtcstatsEnabled = false
+            } = {}
+        } = initConfig;
+
+        // If rtcstats is not enabled or already initialized, do nothing.
+        // Calling rtcsatsInit multiple times will cause the global objects to be rewritten multiple times,
+        // with unforseseen consequences.
+        if (!rtcstatsEnabled || this._initialized) return;
+        
+        rtcstatsInit(
```
saghul, line 63: Can this fail?

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+                rtcstatsPollInterval: pollInterval= 10000,
+                rtcstatsSendSdp: sendSdp = false,
+                rtcstatsEnabled = false
+            } = {}
+        } = initConfig;
+
+        // If rtcstats is not enabled or already initialized, do nothing.
+        // Calling rtcsatsInit multiple times will cause the global objects to be rewritten multiple times,
+        // with unforseseen consequences.
+        if (!rtcstatsEnabled || this._initialized) return;
```
saghul, line 61: On mobile one could join a meeting on a server with rtcstats enabled and then join another with it disabled. AFAIS the second meeting would still send stats because it would retain the enabled flag from the first one. We need to address that.

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+     * new conference's config. On a normal conference flow this wouldn't be necessary, as the whole page is
+     * reloaded, but in the case of breakout rooms or react native the js context doesn't reload, hence the
+     * RTCStats singleton and its config persists between conferneces.
+     * 
+     * @param conference - JitsiConference instance that's about to start.
+     * @returns {void}
+     */
+    start(conference: JitsiConference) {
+        // If rtcstats proxy module is not initialized, do nothing.
+        if (!this._initialized) return;
```
saghul, line 86: When would this happen?

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,189 @@
...
+     * Resets the trace module by closing the websocket and deleting the object.
+     * After reset, the rtcstats proxy module that tries to send data via `sendStatsEntry`, will no longer
+     * send any data, until the trace module is initialized again. This comes in handy on react-native
+     * where ljm doesn't get reloaded, so we need to switch the trace module between conferences.
+     *
+     * @returns {void}
+     */
+    reset() {
+        this.trace?.close();
+        this.trace && delete this.trace;
```
saghul, line 174: Why not just set it to null?

modules/statistics/statistics.js
```diff
@@ -157,16 +160,17 @@ Statistics.init = function(options) {
  * @param {StatisticsOptions} options - The options to use creating the
  * Statistics.
  */
-export default function Statistics(xmpp, options) {
+export default function Statistics(conference, config) {
```
saghul, line 163: Update the jsdoc or keep using options pl.

modules/statistics/statistics.js
```diff
@@ -199,10 +203,13 @@ export default function Statistics(xmpp, options) {
 
     Statistics.instances.add(this);
 
+    conference && RTCStats.start(this.conference);
```
saghul, line 206: There is no way conference is undefined, is there?

webpack.config.js
```diff
@@ -18,7 +18,11 @@ module.exports = (_env, argv) => {
                 library: 'JitsiMeetJS',
                 libraryTarget: 'umd',
                 path: path.join(process.cwd(), 'dist', 'umd')
-            })
+            }),
+            performance: {
```
saghul, line 22: This is already set on the shared file, any reason to set it here too?

saghul (changes requested):
Sorry it took this long to review!

webpack.config.js
andrei-gavrilescu: Didn't check the purpose of the config from the previous PR. I suppose not, I'll remove it.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: Ideally it wouldn't, it's just meant as some basic invariant checking. But if you prefer a fail fast approach I can remove it.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: When the second meeting is joined, the start method is called again, which gets the new config, there's a comment about that case [here](https://github.com/jitsi/lib-jitsi-meet/blob/4ae62fffbdc9f5d30cce0b1c64161b5457895874/modules/RTCStats/RTCStats.ts#L108).
If this happens trace will be set to null and stats will be ignored, however the proxy objects will still be set as we don't have a way to "unload" the rtcstats module yet.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: It could, would you rather I swallow and log the exception or let it fail?

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: Seems like chrome removed legacy support as of 117, but rtcstats still has it as an option, I'll remove the option here and make a task for removing support from rtcstats client.

saghul: Hold on :-) What does it really imply? Note we support Chrome >= 72.

modules/RTCStats/RTCStats.ts
saghul: I guess it's fine since the outcome is the same.

modules/RTCStats/RTCStats.ts
saghul: Gotcha!

modules/RTCStats/RTCStats.ts
saghul: Let's add a log line at least, since it implies something is bad.

modules/statistics/statistics.js
andrei-gavrilescu: I didn't know the full intent of the Statistics class (just group some logic that's tightly coupled to the JitsiConference class or more generic), technically it has functionality that can work without a conference. But as it's currently used I suppose conference will be there. I'll remove it.

modules/statistics/statistics.js
andrei-gavrilescu: updated

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: Added.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: Welp, removing it wouldn't break older Chrome users, as standard stats have been available for a while, so we would simply be forcing clients to use standard.
But this requires some work on the rtcstats client as previously mentioned, I'll add a note to make sure that standard stats are available from Chrome >= 72.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: I typed it, but not sure I did it right, I've added an interface file for it. Let me know if that is the way for ljm.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: makes sense, changed.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: I've added separate events and bubbled them up, using the template from other services. Adding the changes to jitsi-meet asap.

andrei-gavrilescu:
> Sorry it took this long to review!

No worries, thanks for being so thorough.

JitsiMeetJS.ts
andrei-gavrilescu: I've changed event propagation a bit, to match other examples in ljm, let me know if that approach is ok.

andrei-gavrilescu:

damencho:
I'm in a middle of updating the grid, it will be back in 10-15 mins and I will make sure this runs when ready, sorry.

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,194 @@
...
+                rtcstatsUseLegacy: useLegacy = false
+            } = {}
+        } = confConfig;
+
+        // The new conference config might have rtcstats disabled, so we need to check again.
+        if (!rtcstatsEnabled) return;
+
+        // When the conference is joined, we need to initialize the trace module with the new conference's config.
+        // The trace module will then connect to the rtcstats server and send the identity data.
+        conference.on(CONFERENCE_JOINED, () => {
```
saghul, line 118: Maybe use once, so listeners are cleared after?

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,194 @@
...
+                displayName,
+                meetingUniqueId,
+                isBreakoutRoom
+            }
+    
+            this.sendIdentity(identityData);
+        });
+
+        // Note, this will only be called for normal rooms, not breakout rooms.
+        conference.on(CONFERENCE_UNIQUE_ID_SET, (meetingUniqueId) => {
```
saghul, line 150: Ditto.

modules/RTCStats/RTCStats.ts
```diff
@@ -0,0 +1,194 @@
...
+            this.sendIdentity({meetingUniqueId});
+        });
+
+        conference.on(CONFERENCE_LEFT, () => {
```
saghul, line 154: Ditto.

webpack.config.js
```diff
@@ -18,7 +18,10 @@ module.exports = (_env, argv) => {
...
+                hints: false
```
saghul, line 23: Why is this needed here?

saghul (commented):
A couple of minor things!

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: Done.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: Done.

webpack.config.js
andrei-gavrilescu: Don't think there's a reason, just remnants from the old PR, removed.

saghul (approved):

andrei-gavrilescu:

andrei-gavrilescu:

andrei-gavrilescu:

andrei-gavrilescu:

PR #2339, opened 2023-08-15
add feat Breakout Room Messaging/Broadcast message

Bloodiko:
This PR adds lib functionality to improve Text Messaging in the following ways: 

- onedirectional "Info"-messaging to all rooms (including main room for backlog reasons?)  AKA Broadcast
- onedirectional "Info"-messaging to a single (Breakout) Room 

- bidirectional "Private Messaging" across all rooms 

Keep in mind that broadcasting messages should only be allowed for moderators maybe and only work within the same conference, not across conferences. 

Current behaviour: 
Message to any room, except the joined room returns an error message from the websocket. 
Message to any participant outside the own room returns nothing. 

Proposed behaviour: 
Message to any room **within** the boundaries of the conference  by a moderator should be allowed, as well as broadcasting a message to all rooms in the conference. 
Private messages from any participant to another participant in a different room should be allowed. 

Please tell me what you think about this. 
I am not experienced enough to create a PR on the backend, so that side will be missing entirely.

### DIFF
--- a/modules/xmpp/ChatRoom.js
+++ b/modules/xmpp/ChatRoom.js
@@ -892,10 +892,9 @@ export default class ChatRoom extends Listenable {
      * Send text message to the other participants in the conference
      * @param message
      * @param elementName
-     * @param {string} [roomjid] the jid of the chat room where the message will be sent
      */
-    sendMessage(message, elementName, roomjid = this.roomjid) {
-        const msg = $msg({ to: roomjid,
+    sendMessage(message, elementName) {
+        const msg = $msg({ to: this.roomjid,
             type: 'groupchat' });
 
         // We are adding the message in a packet extension. If this element
@@ -917,10 +916,9 @@ export default class ChatRoom extends Listenable {
      * @param id id/muc resource of the receiver
      * @param message
      * @param elementName
-     * @param {string} [roomjid] the jid of the chat room where the recieving participant is
      */
-    sendPrivateMessage(id, message, elementName, roomjid = this.roomjid) {
-        const msg = $msg({ to: `${roomjid}/${id}`,
+    sendPrivateMessage(id, message, elementName) {
+        const msg = $msg({ to: `${this.roomjid}/${id}`,
             type: 'chat' });
 
         // We are adding the message in packet. If this element is different


modules/xmpp/ChatRoom.js
```diff
@@ -892,9 +892,10 @@ export default class ChatRoom extends Listenable {
      * Send text message to the other participants in the conference
      * @param message
      * @param elementName
+     * @param {string} [roomjid] the jid of the chat room where the message will be sent
      */
-    sendMessage(message, elementName) {
-        const msg = $msg({ to: this.roomjid,
+    sendMessage(message, elementName, roomjid = this.roomjid) {
```
saghul, line 897: I don't think this is the right thing to do.

modules/xmpp/ChatRoom.js
```diff
@@ -916,9 +917,10 @@ export default class ChatRoom extends Listenable {
      * @param id id/muc resource of the receiver
      * @param message
      * @param elementName
+     * @param {string} [roomjid] the jid of the chat room where the recieving participant is
      */
-    sendPrivateMessage(id, message, elementName) {
-        const msg = $msg({ to: `${this.roomjid}/${id}`,
+    sendPrivateMessage(id, message, elementName, roomjid = this.roomjid) {
```
saghul, line 922: Ditto.

modules/xmpp/BreakoutRooms.js
```diff
@@ -131,6 +132,47 @@ export default class BreakoutRooms {
         this._sendMessage(message);
     }
 
+    /**
+     * Sends a message to a breakout room.
+     *
+     * @param {string} roomJid - JID of the room to send the message to.
+     * @param {string} textContent - The message content.
+     */
+    sendMessageToRoom(roomJid, textContent) {
```
saghul, line 141: These new functions should check for the broadcast capability I think.

modules/xmpp/BreakoutRooms.js
```diff
@@ -7,6 +7,7 @@ const FEATURE_KEY = 'features/breakout-rooms';
 const BREAKOUT_ROOM_ACTIONS = {
     ADD: `${FEATURE_KEY}/add`,
     MOVE_TO_ROOM: `${FEATURE_KEY}/move-to-room`,
+    MESSAGE_TO_ROOM: `${FEATURE_KEY}/message-to-room`,
```
saghul, line 10: You need to have the prosody plugin handle these messages, but I haven't seen such a PR, are you still working on it?

saghul (changes requested):

modules/xmpp/BreakoutRooms.js
Bloodiko: realizing that i dont even know how prosody handles regular messages, I didnt knew where to start. 
Could you point me to the direction and i'll see if i'm able to create one ? 

i guess in this case until I added a PR for prosody ( in jitsi-meet ? or is it a different repo i didnt find yet ? ) i'll close this again, until its relevant.

saghul: Feel free to leave it open, perhaps better mark it as a daft. Here is the BR plugin: https://github.com/jitsi/jitsi-meet/blob/master/resources/prosody-plugins/mod_muc_breakout_rooms.lua you can take a look at how other messages are handled.

Bloodiko: Thanks, marked it as draft. 

I did find the breakout plugin, but its not really easy to understand how the Lua plugins interact with prosody main code. 

At least there's no plugin for the chat messages and other Basic xmpp functions. 
Or I just didn't find it.  

I hope it's not buried in the java code, I certainly didn't wanted to touch that. Even if it's just for understanding purposes.

saghul: Chat is plain XMPP, but you need extra functionality, hence the need for a plugin.

modules/xmpp/ChatRoom.js
Bloodiko: reverted as requested

modules/xmpp/ChatRoom.js
Bloodiko: Ditto.

modules/xmpp/BreakoutRooms.js
Bloodiko: added check for feature, using the already existing `isFeatureSupported` function

modules/xmpp/BreakoutRooms.js
Bloodiko: Gotta be honest here.  Lua is definitely not my main language, and its absolutely not clear how the objects are structured. Or at least I was unable to figure out where to look it up. 
I gathered as much references from other prosody plugins, as well as third party prosody plugins as I was able to find. 
PR is open. (https://github.com/jitsi/jitsi-meet/pull/13745)

I'll also open this one again.

PR #2343, opened 2023-08-17
Sends conference request over http before connecting to xmpp

damencho:


modules/xmpp/ChatRoom.js
```diff
@@ -266,8 +267,10 @@ export default class ChatRoom extends Listenable {
             if (this.password) {
                 pres.c('password').t(this.password).up();
             }
-            if (this.options.billingId) {
-                pres.c('billingid').t(this.options.billingId).up();
+
+            // send the machineId with the initial presence
+            if (this.moderator.targetUrl) {
```
bgrozev: Why not send it always, so we can eventually remove the backend code that reads it?

modules/xmpp/moderator.js
```diff
@@ -314,10 +314,19 @@ export default class Moderator {
                 // decide to send our IQ.
                 this.connection.flush();
             } else {
+                const conferenceRequest = this._createConferenceRequest(roomJid);
+
+                if (FeatureFlags.isJoinAsVisitorSupported()) {
+                    if (!conferenceRequest.properties) {
```
bgrozev, line 322: I don't think that's necessary. It's always initialized as an object in `_createConferenceRequest`

bgrozev: Also, why do we include `visitors-version` only over HTTP? I'd move this to `_createConferenceRequest`

JitsiConnection.js
```diff
@@ -22,6 +23,7 @@ export default function JitsiConnection(appID, token, options) {
     this.token = token;
     this.options = options;
     this.xmpp = new XMPP(options, token);
+    this.xmpp.connection.moderator = new Moderator(this.xmpp, this.options);
```
bgrozev, line 26: Why set it here instead of the `xmpp` or `xmpp.connection` constructor? They have all the necessary fields.

JitsiConnection.js
```diff
@@ -56,11 +56,26 @@ export default function JitsiConnection(appID, token, options) {
...
+ * to be used.
  */
 JitsiConnection.prototype.connect = function(options = {}) {
-    this.xmpp.connect(options.id, options.password);
+    // if we get redirected, we set disableFocus to skip sending the conference request twice
+    if (this.xmpp.connection.moderator.targetUrl && !this.options.disableFocus && options.name) {
+        this.xmpp.options.conferenRequestSent = false;
+        this.xmpp.connection.moderator.sendConferenceRequest(this.xmpp.getRoomJid(options.name))
+            .then(() => {
+                this.xmpp.options.conferenRequestSent = true;
```
bgrozev, line 73: typo: "conferenRequestSent"

modules/xmpp/ChatRoom.js
damencho: So sending a that information that way is out of spec so I wanted it to work only when this mode is enabled.

damencho: But if you think we will enable this http request by default in the future, let's remove it ... if some needs to disable it in the future can create a flag to disable it. WDYT?

bgrozev (approved):

modules/xmpp/moderator.js
```diff
@@ -39,549 +41,639 @@ function createExpBackoffTimer(step) {
...
-    function listener(event) {
-        if (event.data && event.data.sessionId) {
-            if (event.origin !== window.location.origin) {
-                logger.warn(`Ignoring sessionId from different origin: ${event.origin}`);
+        // FIXME: Message listener that talks to POPUP window
+        /**
+         *
+         * @param event
+         */
+        function listener(event) {
```
hristoterezov, line 95: I think this function can be outside of the constructor. WDYT?

damencho: Well I have kept the code as it is. We can improve it later.

modules/xmpp/ChatRoom.js
```diff
@@ -266,8 +267,10 @@ export default class ChatRoom extends Listenable {
...
+                pres.c('billingid').t(Settings.machineId).up();
```
hristoterezov, line 282: Why do we use stop using the `billingId` in favour of `machineId`? I don't understand this change. Could you please give me some context?

modules/xmpp/moderator.js
```diff
@@ -238,13 +242,6 @@ export default class Moderator {
                     .up();
             }
         }
-        if (FeatureFlags.isJoinAsVisitorSupported()) {
```
hristoterezov: Why we don't need this anymore?

modules/util/EventEmitterForwarder.js
```diff
@@ -30,9 +31,25 @@ EventEmitterForwarder.prototype.forward = function(...args) {
     args[0] = this.dest;
 
     // Using bind.apply to pass the arguments as Array-like object ("arguments")
-    this.src.addListener(
+    this.unsubscribers.set(
```
hristoterezov, line 34: This is awesome to remove the listeners! Thanks for doing the effort. I'm wondering why we this functionality didn't exist.

But I don't think this whole thing will work in a generic way! Listenables will return the function for removing the listener and this would work but a normal EventEmitter will just return itself...

So maybe it will be better to store the actual listener that we attach in the unsubscribers Map and then use it call removeListener. This way this should work everything.

JitsiConference.js
```diff
@@ -147,7 +147,7 @@ export default function JitsiConference(options) {
     this.connection = options.connection;
     this.xmpp = this.connection?.xmpp;
 
-    if (this.xmpp.isRoomCreated(options.name, options.customDomain)) {
```
hristoterezov, line 150: Why do we need this change?

modules/xmpp/moderator.js
damencho: Cause we move it in _createConferenceRequest so it will be available for both xmpp and http send method.

modules/util/EventEmitterForwarder.js
damencho: Ok, I will do the change force push it in the commit. Is this ok?

JitsiConference.js
damencho: This was causing problems ... let me try to remember it was few months ago ...

modules/xmpp/xmpp.js
```diff
@@ -164,6 +165,8 @@ export default class XMPP extends Listenable {
             shard: options.deploymentInfo.shard
         });
 
+        this.connection.moderator = new Moderator(this);
```
hristoterezov, line 168: I'm wondering what is the reason to put the moderator as part of the connection and not directly in the xmpp instance since we are creating it there?

damencho: connection as XMPP connection? Or JitsiConnection?

JitsiConference.js
damencho: No idea, ok I will drop the commit when force pushing the other changes.

damencho: We need to pass xmpp to Moderator, and in XmppConnection we pass just options no instance of other objects.
We need the xmpp.eventEmitter in Moderator.

modules/xmpp/xmpp.js
```diff
@@ -810,7 +810,7 @@ export default class XMPP extends Listenable {
     disconnect(ev) {
         if (this.disconnectInProgress) {
             return this.disconnectInProgress;
-        } else if (!this.connection) {
+        } else if (!this.connection || !this.connection.connected) {
```
hristoterezov, line 813: I'm wondering is it possible to call connect and without waiting for result call disconnect. 

It looks to me that after your change we won't call disconnect on Strophe at all and before we would have called it. Is this correct?

Do we care about this use case? The previous version sounds more correct to me.

damencho: > I'm wondering is it possible to call connect and without waiting for result call disconnect.

Nope, we have see a lot of problems because of that. Where we connect quickly and disconnect event is fired and everythng messed up we think we are connected, the library thinks it is disocnnected.



> It looks to me that after your change we won't call disconnect on Strophe at all and before we would have called it. Is this correct?

Well if it is not connected we will skip. Well maybe it sounds correct but there was a particular problem that was fixed with this.

damencho: And this is exactly with the visitors. As we always call disconnect there and this promise was never resolving and everything hangs.

modules/xmpp/xmpp.js
damencho: We need to pass xmpp to Moderator, and in XmppConnection we pass just options no instance of other objects.
We need the xmpp.eventEmitter in Moderator.

hristoterezov (approved):

PR #2350, opened 2023-09-13
Add SVC and Simulcast support for AV1 and VP9.

jallamsetty1:
Both Simulcast and SVC modes are supported for AV1 and VP9. The default mode for both of them is SVC and is enabled by default unless it is explicitly disabled from config.js.
```
videoQuality: {
      av1: {
            maxBitratesVideo: {
               low: 100000,
               standard: 300000,
               high: 1000000,
               ssHigh: 2500000
            },
            scalabilityModeEnabled: true,
            useSimulcast: false,
            useKSVC: true
        },
        h264: {
            maxBitratesVideo: {
                low: 200000,
                standard: 500000,
                high: 1500000,
                ssHigh: 2500000
            },
            scalabilityModeEnabled: true
        },
        vp8: {
            maxBitratesVideo: {
                low: 200000,
                standard: 500000,
                high: 1500000,
                ssHigh: 2500000
            },
            scalabilityModeEnabled: false
        },
        vp9: {
            maxBitratesVideo: {
                low: 100000,
                standard: 300000,
                high: 1200000,
                ssHigh: 2500000
            },
            scalabilityModeEnabled: true,
            useSimulcast: false,
            useKSVC: true
        }
}
```


jallamsetty1:
Depends on https://github.com/jitsi/jitsi-videobridge/pull/1998

modules/RTC/TPCUtils.js
```diff
@@ -1,25 +1,26 @@
...
 const SIM_LAYER_3_RID = '3';
+const VIDEO_CODECS = [ CodecMimeType.AV1, CodecMimeType.H264, CodecMimeType.VP8, CodecMimeType.VP9 ];
 
-export const HD_BITRATE = 2500000;
-export const HD_SCALE_FACTOR = 1;
-export const LD_SCALE_FACTOR = 4;
-export const SD_SCALE_FACTOR = 2;
+export const HD_SCALE_FACTOR = 1.0;
+export const LD_SCALE_FACTOR = 4.0;
+export const SD_SCALE_FACTOR = 2.0;
```
saghul, line 23: Doesn't it make sense to adjust these values when One enables 4K? Maybe it should be part of the config? I can see us wanting to use a scale factor of 6 in a Spot room configured for 4K for example.

modules/RTC/TPCUtils.js
```diff
@@ -345,15 +422,22 @@ export class TPCUtils {
...
      *
      * @param {JitsiLocalTrack} localVideoTrack The local video track.
+     * @param {CodecMimeType} codec - The codec currently in use.
      * @returns {Array<number>}
      */
-    calculateEncodingsBitrates(localVideoTrack) {
+    calculateEncodingsBitrates(localVideoTrack, codec) {
+        // SVC mode for VP9 and AV1 codecs.
+        if (this._isRunningInSvcMode(codec)) {
+            return [ this.codecSettings[codec].maxBitratesVideo.high, 0, 0 ];
```
saghul, line 431: Can this be called for a SS track? Shouldn't it be `ssHigh` in that case?

modules/RTC/TPCUtils.js
```diff
@@ -571,11 +670,90 @@ export class TPCUtils {
...
+     * @param {number} maxHeight - The height of the outgoing resolution that needs to be configured.
+     * @param {JitsiLocalTrack} localVideoTrack - The local video track.
+     * @returns {void>}
+     */
+    updateEncodingsScalabilityMode(encodings, codec, maxHeight, localVideoTrack) {
+        if (!this.codecSettings[codec].scalabilityModeEnabled) {
+            return;
+        }
+
+        // Configure L1T3 mode on each of the encodings when H.264 is the selected codec.
```
saghul, line 699: Why do we want that on every encoding? Does it make sense also for high layer?

modules/RTC/TPCUtils.js
```diff
@@ -571,11 +670,90 @@ export class TPCUtils {
...
+        const standardSimulcastEncodings = this._getVideoStreamEncodings(videoType, codec);
+        const codecBitrates = this.codecSettings[codec].maxBitratesVideo;
+
+        // Simulcast mode.
+        if (this.codecSettings[codec].useSimulcast) {
+            for (const encoding in encodings) {
+                if (encodings.hasOwnProperty(encoding)) {
+                    encodings[encoding].scaleResolutionDownBy
+                        = standardSimulcastEncodings[encoding].scaleResolutionDownBy;
+                    encodings[encoding].scalabilityMode = VideoEncoderScalabilityMode.L1T3;
```
saghul, line 718: Ditto.

modules/RTC/TPCUtils.js
```diff
@@ -571,11 +670,90 @@ export class TPCUtils {
...
+            }
+
+            return;
+        }
+
+        // SVC mode for desktop streams.
+        if (videoType === VideoType.DESKTOP) {
+            const lowFpsScreenshare = localVideoTrack.getVideoType() === VideoType.DESKTOP
+                && this.pc._capScreenshareBitrate
+                && !browser.isWebKitBased();
```
saghul, line 729: Can you pl leave a comment about what's special in WebKit?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -693,11 +692,19 @@ TraceablePeerConnection.prototype.getAudioLevels = function(speakerList = []) {
 
 /**
  * Checks if the browser is currently doing true simulcast where in three different media streams are being sent to the
- * bridge. Currently this happens only when VP8 is the selected codec.
+ * bridge. Currently this happens always for VP8 and only simulcast is enabled for VP9/AV1/H264.
```
saghul, line 695: Nit: only IF simulcast ...

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2481,17 +2488,20 @@ TraceablePeerConnection.prototype._setVp9MaxBitrates = function(description, isL
...
     const mLines = parsedSdp.media.filter(m => m.type === MediaType.VIDEO && m.direction !== direction);
+    const currentCodec = this.codecSettings.codecList[0];
+    const codecScalabilityModeSettings = this.tpcUtils.codecSettings[currentCodec];
 
     for (const mLine of mLines) {
-        if (this.codecSettings.codecList[0] === CodecMimeType.VP9) {
-            const bitrates = this.tpcUtils.videoBitrates.VP9 || this.tpcUtils.videoBitrates;
-            const hdBitrate = bitrates.high ? bitrates.high : HD_BITRATE;
-            const ssHdBitrate = bitrates.ssHigh ? bitrates.ssHigh : HD_BITRATE;
+        const isDoingKSvc = CodecMimeType.VP9 && !codecScalabilityModeSettings.scalabilityModeEnabled;
```
saghul, line 2495: What is ksvc?

modules/browser/BrowserCapabilities.js
```diff
@@ -226,6 +226,17 @@ export default class BrowserCapabilities extends BrowserDetection {
...
     }
 
+    /**
+     * Returns true if the browser supports the new Scalability Mode API for VP9/AV1 simulcast. H.264 simulcast will
+     * also be supported by the jvb for this version because the bridge is able to read the Dependency Descriptor RTP
+     * header extension to extract layers information for H.264 as well.
+     *
+     * @returns {boolean}
+     */
+    supportsScalabilityModeAPI() {
```
saghul, line 236: Can't this be detected at runtime somehow?

service/RTC/StandardVideoSettings.ts
```diff
@@ -0,0 +1,47 @@
+import browser from "../../modules/browser";
```
saghul, line 1: Style: use single quotes. Not sure why the linter didn't pick it up...

saghul (commented):
Left some comments!

service/RTC/VideoEncoderScalabilityMode.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    /**
+     * L2T2 mode: 2-layer spatial and 3-layer temporal scalability encoding.
+     */
+    L2T3: 'L2T3',
+
+    /**
+     * L2T3_Key mode: 2-layer spatial and 3-layer temporal scalability K-SVC encoding.
+     */
+    L2T3_Key: 'L2T3_Key',
```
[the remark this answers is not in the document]
jallamsetty1: Good catch! It is a typo, will fix it.

modules/RTC/TPCUtils.js
```diff
@@ -80,32 +105,76 @@ export class TPCUtils {
...
-                maxBitrate: browser.isFirefox() ? this.encodingBitrates.low : maxVideoBitrate,
+                maxBitrate: browser.isFirefox() ? codecBitrates.low : maxVideoBitrate,
                 rid: SIM_LAYER_3_RID,
                 scaleResolutionDownBy: browser.isFirefox() ? LD_SCALE_FACTOR : HD_SCALE_FACTOR
             }
         ];
+
+        if (this.codecSettings[codec].scalabilityModeEnabled) {
+            // Configure all 3 encodings for the simulcast case.
+            if (this.codecSettings[codec].useSimulcast || codec === CodecMimeType.H264) {
```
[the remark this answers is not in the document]
jallamsetty1: While VP9 and AV1 support both Simulcast and SVC mode, H.264 supports only the simulcast mode when scalabilityMode is set and DD header is negotiated. I will add these details to the comments.

modules/RTC/TPCUtils.js
```diff
@@ -80,32 +105,76 @@ export class TPCUtils {
...
+    }
+
+    /**
+     * Returns a boolean indicating whether the video encoder is running in SVC mode.
+     *
+     * @param {CodecMimeType} codec
+     * @returns boolean
+     */
+    _isRunningInSvcMode(codec) {
+        return (codec === CodecMimeType.VP9 || codec === CodecMimeType.AV1)
```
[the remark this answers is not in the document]
jallamsetty1: I am not sure I understand your question. 
VP8 only supports simulcast, i.e., 3 streams of different resolutions and each stream is temporally scalable. 
VP9 and AV1 support true SVC mode where one stream is both temporally and spatially scalable. When this returns true, the client makes sure that only the first encoding is correctly configured depending on what resolution is being requested keep the other 2 encodings disabled always.

modules/RTC/TPCUtils.js
jallamsetty1: The bitrate for the first encoding is updated here - https://github.com/jitsi/lib-jitsi-meet/pull/2350/files#diff-6cd0aeb8cb1acbd0118edb9d36b059e32f1984192a2c012bf5d6db68f02ab71aR726
This line is here to ensure that the other 2 encodings have a max bitrate of 0 so that they are disabled all the time.

modules/RTC/TPCUtils.js
jallamsetty1: Yes, since H.264 supports only simulcast, we want all stream to be temporally scalable, even the stream with the highest resolution.

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
saghul: I mean to convey that a function called `_isRunningInSvcMode` might not be very clear since VP8 also does SVC, except only temporal, right?

How about `_isRunningInFullSvcMode` or `_isRunningInRealSvcMode` or `_isRunningInNonGmoSvcMode` ...

modules/RTC/TPCUtils.js
jallamsetty1: Yes even for VP9 and AV1, when we are doing true simulcast, we want all of the streams to be temporally scalable and hence we set the scalability mode to L1T3.

modules/RTC/TPCUtils.js
saghul: I don't think that is obvious at all from an API prespective... is there a way to unify the logic for all encodings?

modules/RTC/TraceablePeerConnection.js
jallamsetty1: K-SVC mode is the current mode VP9 is running in Chrome where we configure 3 different encodings like we would for simulcast but the browser only sends out 1 single stream that is temporally and spatially scalable.

modules/RTC/TPCUtils.js
jallamsetty1: Yes, but I would think we should address it in a different PR since its not related to this feature.

saghul: Makes sense!

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
jallamsetty1: Ok, got it. I will change it to `_isRunningInFullSvcMode`.

modules/RTC/TPCUtils.js
jallamsetty1: Ok sure.

modules/RTC/TPCUtils.js
```diff
@@ -34,19 +35,40 @@ export class TPCUtils {
...
-            high: HD_BITRATE,
-            ssHigh: HD_BITRATE
-        };
+        this.codecSettings = STANDARD_CODEC_SETTINGS;
+        const videoQualitySettings = this.pc.options?.videoQuality;
+
+        if (videoQualitySettings) {
+            for (const codec of VIDEO_CODECS) {
+                const bitrateSettings = videoQualitySettings[codec]?.maxBitratesVideo
+                    ?? videoQualitySettings[codec.toUpperCase()]?.maxBitratesVideo
```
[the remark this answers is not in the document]
jallamsetty1: I am trying to make sure that the code is case insensitive for the codec value in config.js. If we have them in upper case in the constant, then we need to check if the codec value is in lower case in config.js.

modules/browser/BrowserCapabilities.js
jallamsetty1: I am not sure if we want to enable it based on feature detection. I think its better to go with a version that we have tested and we know works as expected since its available in 113 onwards but there were bugs . WDYT?

saghul: Gotcha, makes sense!

modules/RTC/TPCUtils.js
```diff
@@ -34,19 +35,93 @@ export class TPCUtils {
      */
     constructor(peerconnection) {
         this.pc = peerconnection;
-        const bitrateSettings = this.pc.options?.videoQuality?.maxBitratesVideo;
-        const standardBitrates = {
-            low: LD_BITRATE,
-            standard: SD_BITRATE,
-            high: HD_BITRATE,
-            ssHigh: HD_BITRATE
+        this.codecSettings = STANDARD_CODEC_SETTINGS;
```
saghul, line 38: We need to clone the object here. Just by using the reference we'll be mutating the STANDARD_CODEC_SETTINGS object.

modules/RTC/TPCUtils.js
```diff
@@ -1,25 +1,26 @@
...
 const SIM_LAYER_1_RID = '1';
 const SIM_LAYER_2_RID = '2';
 const SIM_LAYER_3_RID = '3';
+const VIDEO_CODECS = [ CodecMimeType.AV1, CodecMimeType.H264, CodecMimeType.VP8, CodecMimeType.VP9 ];
 
-export const HD_BITRATE = 2500000;
-export const HD_SCALE_FACTOR = 1;
-export const LD_SCALE_FACTOR = 4;
-export const SD_SCALE_FACTOR = 2;
+export const HD_SCALE_FACTOR = 1.0;
```
saghul, line 21: Can you pl add a TODO comment here to revisit for 4k?

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
saghul: But in the same line you are indexing it both ways. It looks super weird.

modules/RTC/TPCUtils.js
```diff
@@ -34,19 +35,93 @@ export class TPCUtils {
...
-            ssHigh: HD_BITRATE
+        this.codecSettings = STANDARD_CODEC_SETTINGS;
+        const videoQualitySettings = this.pc.options?.videoQuality;
+
+        if (videoQualitySettings) {
+            for (const codec of VIDEO_CODECS) {
+                const bitrateSettings = videoQualitySettings[codec]?.maxBitratesVideo
+                    ?? videoQualitySettings[codec.toUpperCase()]?.maxBitratesVideo
+                    ?? (videoQualitySettings.maxBitratesVideo
+                        && (videoQualitySettings.maxBitratesVideo[codec]
```
saghul, line 46: Ditto.

modules/RTC/TPCUtils.js
```diff
@@ -34,19 +35,93 @@ export class TPCUtils {
...
+    _calculateActiveEncodingParamsForSvc(localVideoTrack, newHeight, codec) {
+        const codecBitrates = this.codecSettings[codec].maxBitratesVideo;
+        const localTrack = localVideoTrack.getTrack();
+        const { height } = localTrack.getSettings();
+        const desktopShareBitrate = this.pc.options?.videoQuality?.desktopBitrate || DESKTOP_SHARE_RATE;
+        const isScreenshare = localVideoTrack.getVideoType() === VideoType.DESKTOP;
+        const scalabilityMode = isScreenshare && this._isScreenshareBitrateCapped(localVideoTrack)
+
+            // The endpoint needs to send only 1 spatial layer with temporal scalability for low fps SS.
+            ? VideoEncoderScalabilityMode.L1T3
```
saghul, line 93: Please rewirte this to not use nested inline conditionals, it's very hard to read and easy to make mistakes.

modules/RTC/TPCUtils.js
```diff
@@ -34,19 +35,93 @@ export class TPCUtils {
...
+
+            // For camera and high fps screenshare, we need both spatial and temporal scalability.
+            : this.codecSettings[codec].useL3T3
+                ? VideoEncoderScalabilityMode.L3T3 : VideoEncoderScalabilityMode.L3T3_KEY;
+        const maxBitrate = isScreenshare
+            ? this._isScreenshareBitrateCapped(localVideoTrack) ? desktopShareBitrate : codecBitrates.ssHigh
```
saghul, line 99: Ditto.

modules/RTC/TPCUtils.js
```diff
@@ -34,19 +35,93 @@ export class TPCUtils {
...
-        // Check if the max. bitrates for video are specified through config.js videoQuality settings.
-        // Right now only VP8 bitrates are configured on the simulcast encodings, VP9 bitrates have to be
-        // configured on the SDP using b:AS line.
-        this.videoBitrates = bitrateSettings ?? standardBitrates;
-        this.encodingBitrates = this.videoBitrates.VP8 ?? this.videoBitrates;
+        if (newHeight >= height || newHeight === 0 || isScreenshare) {
+            return config;
+        }
+
+        if (newHeight >= height / SD_SCALE_FACTOR) {
```
saghul, line 113: Is no rounding necessary?

modules/RTC/TPCUtils.js
```diff
@@ -34,19 +35,93 @@ export class TPCUtils {
...
+            config.maxBitrate = codecBitrates.standard;
+            config.scalabilityMode = this.codecSettings[codec].useL3T3
+                ? VideoEncoderScalabilityMode.L2T3 : VideoEncoderScalabilityMode.L2T3_KEY;
```
saghul, line 116: Sorry I'm a bit out of my depth here, but why are we setting the mode to L2T3 or L2T3_KEY if useL**3**T3 is set?

modules/RTC/TPCUtils.js
```diff
@@ -80,37 +158,103 @@ export class TPCUtils {
...
-                maxBitrate: reversedEncodings ? this.encodingBitrates.low : maxVideoBitrate,
+                maxBitrate: reversedEncodings ? codecBitrates.low : maxVideoBitrate,
                 rid: SIM_LAYER_3_RID,
                 scaleResolutionDownBy: reversedEncodings ? LD_SCALE_FACTOR : HD_SCALE_FACTOR
             }
         ];
+
+        if (this.codecSettings[codec].scalabilityModeEnabled) {
+            // Configure all 3 encodings when simulcast is requested through config.js for AV1 and VP9 and for H.264
+            // always since that is the only supported mode when DD header extension is negotiated for H.264.
```
saghul, line 197: I see no check for the DD header extension negotiation here. What if someone with a new LJM connects to a bridge without that? Is that possible?

modules/RTC/TPCUtils.js
```diff
@@ -80,37 +158,103 @@ export class TPCUtils {
...
+     * the highest resolution stream is available always. Safari is an exception here since it does not send the
+     * desktop stream at all if only the high resolution stream is enabled.
+     *
+     * @param {JitsiLocalTrack} localVideoTrack - The local video track.
+     * @returns {boolean}
+     */
+    _isScreenshareBitrateCapped(localVideoTrack) {
+        return localVideoTrack.getVideoType() === VideoType.DESKTOP
+            && this.pc._capScreenshareBitrate
+            && !browser.isWebKitBased();
```
saghul, line 257: Why not WebKit? Can you pl add a comment explaining?

modules/RTC/TPCUtils.js
```diff
@@ -346,21 +490,25 @@ export class TPCUtils {
...
+     * Returns the calculates max bitrates that need to be configured on the stream encodings based on the video
      * type and other considerations associated with screenshare.
      *
      * @param {JitsiLocalTrack} localVideoTrack The local video track.
+     * @param {CodecMimeType} codec - The codec currently in use.
+     * @param {number} newHeight The resolution requested for the video track.
      * @returns {Array<number>}
      */
-    calculateEncodingsBitrates(localVideoTrack) {
+    calculateEncodingsBitrates(localVideoTrack, codec, newHeight) {
```
saghul, line 501: The previous function is ` calculateEncodingsActiveState(localVideoTrack, newHeight, codec) {` can we please settle on an order for coded and height and use it consistently?

modules/RTC/TPCUtils.spec.js
```diff
@@ -208,4 +202,605 @@ describe('TPCUtils', () => {
             }
         });
     });
+
+
+    describe('Test encodings when default settings are used for', () => {
```
saghul, line 207: ❤️

modules/RTC/TraceablePeerConnection.js
```diff
@@ -693,11 +692,19 @@ TraceablePeerConnection.prototype.getAudioLevels = function(speakerList = []) {
...
  * Checks if the browser is currently doing true simulcast where in three different media streams are being sent to the
- * bridge. Currently this happens only when VP8 is the selected codec.
+ * bridge. Currently this happens always for VP8 and only if simulcast is enabled for VP9/AV1/H264.
  * @returns {boolean}
  */
 TraceablePeerConnection.prototype.doesTrueSimulcast = function() {
-    return this.isSimulcastOn() && this.getConfiguredVideoCodec() === CodecMimeType.VP8;
+    const currentCodec = this.getConfiguredVideoCodec();
+    const codecScalabilityModeSettings = this.tpcUtils.codecSettings[currentCodec];
+    const runsInSimulcastMode = currentCodec === CodecMimeType.AV1 || currentCodec === CodecMimeType.VP9
```
saghul, line 701: I think I've seen a similar check elsewhere. Would it make sense to have some `SIMULCAST_MODE_CODECS` array and check fi the codec in question is in there?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -693,11 +692,19 @@ TraceablePeerConnection.prototype.getAudioLevels = function(speakerList = []) {
...
+        ? codecScalabilityModeSettings.scalabilityModeEnabled && codecScalabilityModeSettings.useSimulcast
```
saghul, line 702: No nested ternary operator pl...

Note to self: I think there is an eslint rule for that...

service/RTC/StandardVideoSettings.ts
```diff
@@ -0,0 +1,47 @@
+import browser from '../../modules/browser';
+
+/**
+ * Standard scalability mode settings for different video codecs and the default bitrates.
+ */
+export const STANDARD_CODEC_SETTINGS = {
```
saghul, line 6: Maybe a simple way to avoid the codec casing stuff is to do:

```
const AV1_SETTINGS = { ... };
export const STANDARD_CODEC_SETTINGS = {
    av1: AV1_SETTINGS,
    AV1: AV1_SETTINGS,
    ...
}

```

saghul (commented):
Left some more comments, PTAL!

modules/RTC/TPCUtils.js
jallamsetty1: It has already been explained few lines above this :) https://github.com/jitsi/lib-jitsi-meet/pull/2350/files#diff-6cd0aeb8cb1acbd0118edb9d36b059e32f1984192a2c012bf5d6db68f02ab71aR248

modules/RTC/TPCUtils.js
jallamsetty1: Jonathan just discovered that DD header extension is negotiated whenever AV1 is offered by Jicofo. I am working on AV1-DD suppression logic (for VP8 and VP9) so I can add a check for that here. We decided not to merge this PR without the suppression logic since it affects scalability for VP8.

modules/RTC/TPCUtils.js
jallamsetty1: I am using useL3T3 in a broader sense, i.e., if L3T3 (and not L3T3_KEY) is used when full resolution is requested. If the highest layer needs to be suspended then the corresponding scalabilityMode for sending the SD stream would be L2T3 and not L2T3_KEY. I changed the config variable to `useL3T3Mode` if that helps.

saghul: Sorry, that didn't help :-) Since this translates into a setting that users can set in config.js we need a better name IMHO.

I cannot provide one because it's not 100% clear to me what it really does...

jallamsetty1: This setting is experimental until we figure out which mode is better for full SVC, `L3T3` or `L3T3_KEY`. For anyone actually interested in understanding the differences, I have the referenced the doc that explains the modes here https://github.com/jitsi/lib-jitsi-meet/pull/2350/files#diff-66113b9503c3b54c96306116bcd0c8e3bace7c11861d47f7f91887e0612ae491R3. I honestly can't think of a better name myself.

paweldomas: I may be missing something here, but maybe the config name should be about whether or not the K-SVC is used? for example `use_K_SVC` ?

Then:
```
config.scalabilityMode = this.codecSettings[codec].use_K_SVC
                ? VideoEncoderScalabilityMode.L2T3_KEY : VideoEncoderScalabilityMode.L2T3;
```

It seems in various places that the config is used to decide whether K-SVC is to be used or not, but currently it's meaning is the inversion of that.

paweldomas: In case having a default for the config is a problem, you can check for `undefined` to tell if the default of `true` in this case should be used.

jallamsetty1: We want to make K-SVC the default mode to be used unless its overridden through config.js. `use_K_SVC` makes sense and users will have to switch it to false in the config to try the other mode.

jallamsetty1: Done, renamed the setting and inverted the defaults.

modules/RTC/TPCUtils.js
```diff
@@ -345,44 +497,109 @@ export class TPCUtils {
...
+     * @returns {Array<VideoEncoderScalabilityMode> | undefined}
+     */
+    calculateEncodingsScalabilityMode(localVideoTrack, codec, maxHeight) {
+        if (!this.pc.isSpatialScalabilityOn() || !this.codecSettings[codec].scalabilityModeEnabled) {
+            return;
+        }
+
+        // Default modes for simulcast.
+        const scalabilityModes = [
+            VideoEncoderScalabilityMode.L1T3,
```
saghul, line 553: Is this value being the same intentional?

modules/RTC/TPCUtils.js
```diff
@@ -345,44 +497,109 @@ export class TPCUtils {
...
+     * and the codec in use.
+     *
+     * @param {JitsiLocalTrack} localVideoTrack The local video track.
+     * @param {CodecMimeType} codec - The codec currently in use.
+     * @param {number} maxHeight The resolution requested for the video track.
+     * @returns {number|undefined}
+     */
+    calculateEncodingsScaleFactor(localVideoTrack, codec, maxHeight) {
+        if (this.pc.isSpatialScalabilityOn() && this.isRunningInSimulcastMode(codec)) {
+            return;
```
saghul, line 584: Why not return 1?

modules/RTC/TPCUtils.js
```diff
@@ -396,6 +613,17 @@ export class TPCUtils {
...
 
+        // SVC mode for VP9 and AV1 codecs.
+        if (this._isRunningInFullSvcMode(codec)) {
+            const activeEncoding = parameters.encodings[0];
+
+            if (activeEncoding.active) {
+                return height / activeEncoding.scaleResolutionDownBy;
+            }
+
+            return null;
```
saghul, line 624: Why not make this 0?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -339,6 +334,13 @@ export default function TraceablePeerConnection(
      */
     this._senderMaxHeights = new Map();
 
+    /**
+     * Flag indicating bridge support for AV1 codec. On the bridge connection, it is supported only when support for
+     * AV1 Dependency Descriptor header extensions is offered by Jicofo. H.264 simulcast is also possible when these
+     * header extensions are negotiated.
+     */
+    this._supportsAv1HeaderExts = false;
```
saghul, line 342: Wouldn't renaming this to `supportsDDHeaderExt` make more sense?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1722,6 +1730,61 @@ TraceablePeerConnection.prototype._mungeCodecOrder = function(description) {
...
+ * @returns {RTCSessionDescription} the munged description.
+ */
+TraceablePeerConnection.prototype._updateAv1DdHeaders = function(description) {
+    if (!browser.supportsScalabilityModeAPI()) {
+        return description;
+    }
+
+    const parsedSdp = transform.parse(description.sdp);
+    const mLines = parsedSdp.media.filter(m => m.type === MediaType.VIDEO);
+    const extUri = 'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension';
```
saghul, line 1747: Maybe extract this to the top since it's a module level constant.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1722,6 +1730,61 @@ TraceablePeerConnection.prototype._mungeCodecOrder = function(description) {
...
+            const headerIndex = mLine.ext.findIndex(ext => ext.uri === extUri);
+            const shouldNegotiateHeaderExts = codec === CodecMimeType.AV1 || codec === CodecMimeType.H264;
+
+            if (!this._supportsAv1HeaderExts && headerIndex >= 0) {
+                this._supportsAv1HeaderExts = true;
+            }
+
+            if (this._supportsAv1HeaderExts && shouldNegotiateHeaderExts && headerIndex < 0) {
+                mLine.ext.push({
+                    value: 11,
```
saghul, line 1773: Why 11?

saghul (approved):
Amazing work! Left some nmon-blocking comments, which can be handled in subsequent PRs since I suppose we might need to tweak things a bit.

Also, is mobile ok with this change?

modules/RTC/TPCUtils.js
jallamsetty1: Yes for simulcast, each one of the simulcast streams will have temporal scalability only and hence L1T3.

jallamsetty1:
> Amazing work! Left some nmon-blocking comments, which can be handled in subsequent PRs since I suppose we might need to tweak things a bit.
> 
> Also, is mobile ok with this change?

Actually, thats a good idea. I will merge this and open a follow-up PR. I don't see many changes though. Yes, no issues with mobile.

saghul:
> > Amazing work! Left some nmon-blocking comments, which can be handled in subsequent PRs since I suppose we might need to tweak things a bit.
> 
> > 
> 
> > Also, is mobile ok with this change?
> 
> 
> 
> Actually, thats a good idea. I will merge this and open a follow-up PR. I don't see many changes though. Yes, no issues with mobile.

Let's go then! 🙌

modules/RTC/TPCUtils.js
jallamsetty1: Actually, this method returns null for cases when the send resolution cannot be determined. I think it makes sense to return null here.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: 11 is the header ext id used by Jicofo/bridge for the DD RTP header extension. I have declared it as a variable now.

modules/xmpp/JingleSessionPC.js
```diff
@@ -403,23 +401,25 @@ export default class JingleSessionPC extends JingleSession {
         pcOptions.capScreenshareBitrate = false;
         pcOptions.codecSettings = options.codecSettings;
         pcOptions.enableInsertableStreams = options.enableInsertableStreams;
-        pcOptions.videoQuality = options.videoQuality;
+
+        if (options.videoQuality) {
```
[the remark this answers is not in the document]
jallamsetty1: This converts all the videoQuality key values to lowercase so that the code is case insensitive.

PR #2353, opened 2023-09-19
fix(ProxyConnectionPC) fake IQ acknowledgements

saghul:


saghul:
I think I caught the problem. The connection is all fake: https://github.com/jitsi/lib-jitsi-meet/blob/dda9ec120aad5a496dce47cc8762948b3b1a7411/modules/proxyconnection/ProxyConnectionPC.js#L170

The function that does the IQ sending does not take any callback argument and call it: https://github.com/jitsi/lib-jitsi-meet/blob/dda9ec120aad5a496dce47cc8762948b3b1a7411/modules/proxyconnection/ProxyConnectionPC.js#L318

Faking a success should do I think.

This likely wasn't a problem before because the queue was running at all times, so it didn't matter that session-accept wasn't technically complete.

jallamsetty1 (approved):

PR #2356, opened 2023-09-21
fix: getting right resolution for mobile web

gabiborlea:
test at https://gabi.jitsi.net

### DIFF
--- a/modules/RTC/JitsiTrack.js
+++ b/modules/RTC/JitsiTrack.js
@@ -379,6 +379,20 @@ export default class JitsiTrack extends EventEmitter {
     }
 
     /**
+     * Returns the height of the track in 16:9 format
+     */
+    getHeight() {
+        return Math.min(this.track.getSettings().height, this.track.getSettings().width);
+    }
+
+    /**
+     * Returns the width of the track in 16:9 format
+     */
+    getWidth() {
+        return Math.max(this.track.getSettings().height, this.track.getSettings().width);
+    }
+
+    /**
      * Checks whether the MediaStream is active/not ended.
      * When there is no check for active we don't have information and so
      * will return that stream is active (in case of FF).
--- a/modules/RTC/JitsiTrack.js
+++ b/modules/RTC/JitsiTrack.js
@@ -379,7 +379,7 @@ export default class JitsiTrack extends EventEmitter {
     }
 
     /**
-     * Returns the height of the track in 16:9 format
+     * Returns the height of the track in normalized landscape format.
      */
     getHeight() {
         return Math.min(this.track.getSettings().height, this.track.getSettings().width);
--- a/modules/RTC/JitsiTrack.js
+++ b/modules/RTC/JitsiTrack.js
@@ -386,7 +386,7 @@ export default class JitsiTrack extends EventEmitter {
     }
 
     /**
-     * Returns the width of the track in 16:9 format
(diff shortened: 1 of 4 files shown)


saghul (dismissed):
LGTM, good find! Please hold on merging, I'd like @jallamsetty1 's expert eyes on it too :-)

saghul:
Jenkins add to whitelist.

modules/connectivity/ConnectionQuality.js
```diff
@@ -56,6 +57,25 @@ const MAX_TARGET_BITRATE = 2500;
...
+ * @param {string} target
+ * @returns {number}
+ */
+function getTargetForFormat(videoQualitySettings, simulcastFormat) {
+    if (browser.isMobileDevice() && !browser.isReactNative()) {
+        // The mobile video stream is 9:16 so the bitrate target should but the
+        // targets are for 16:9, so the height is correct but the width is different
+        const mobileStreamWidth = simulcastFormat.height * 9 / 16;
+
+        return videoQualitySettings[simulcastFormat.target] * mobileStreamWidth / simulcastFormat.width;
```
jallamsetty1, line 73: Can you explain this formula? It doesn't make sense to me. Maybe I am missing something. Going by this, for a capture of 360x640 (SD) stream, the target bitrate comes to 94 Kbps which is very low compared to 300 Kbps which is typically the target bitrate for the SD stream.

jallamsetty1: This also doesn't take into account the fact that the endpoint can be in landscape orientation when the video track is created or when layer suspension kicks in which the target resolution has the correct height. Maybe figure out the effective height here https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/connectivity/ConnectionQuality.js#L281. I don't think finding out a multiplying factor for the standard target bitrates will work.

gabiborlea: The video stream on mobile is usually in portrait and 9:16. So when the height is for example 360 the width for which the bitrate is calculated is 640. But the actual width will be 202, and the bitrate will be lower than the target one which is intended for 360x640. This is why on the mobile web the connection is always non-optimal or close to being non-optimal, the connectionQuality percent being around 30% always, when having the camera on and the call on jvb.

modules/RTC/JitsiTrack.js
```diff
@@ -378,6 +378,20 @@ export default class JitsiTrack extends EventEmitter {
         return this.videoType;
     }
 
+    /**
+     * Returns the height of the track in 16:9 format
```
saghul, line 382:
```suggestion
     * Returns the height of the track in normalized landscape format.
```

modules/RTC/JitsiTrack.js
```diff
@@ -378,6 +378,20 @@ export default class JitsiTrack extends EventEmitter {
...
 
+    /**
+     * Returns the height of the track in normalized landscape format.
+     */
+    getHeight() {
+        return Math.min(this.track.getSettings().height, this.track.getSettings().width);
+    }
+
+    /**
+     * Returns the width of the track in 16:9 format
```
saghul, line 389:
```suggestion
     * Returns the width of the track in normalized landscape format.
```

saghul (approved):

saghul:
Excellent!

PR #2366, opened 2023-10-04
fix(signaling): Merge ssrcOwners and _sourceName.

hristoterezov:
Additional benefits are:
 - ssrc -> source name will be updated on ssrc remap message from
 the bridge.
 - the map will be correctly cleaned when member leave (this logic was
 not working well for _sourceName map)
 - Looks cleaner and simpler.

### DIFF
--- a/modules/proxyconnection/CustomSignalingLayer.js
+++ b/modules/proxyconnection/CustomSignalingLayer.js
@@ -111,12 +111,6 @@ export default class CustomSignalingLayer extends SignalingLayer {
     /**
      * @inheritDoc
      */
-    setTrackSourceName(ssrc, sourceName) { // eslint-disable-line no-unused-vars
-    }
-
-    /**
-     * @inheritDoc
-     */
     updateSsrcOwnersOnLeave(id) {
         const ssrcs = Array.from(this.ssrcOwners)
             .filter(entry => entry[1] === id)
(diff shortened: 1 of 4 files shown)


modules/proxyconnection/CustomSignalingLayer.js
```diff
@@ -108,12 +108,6 @@ export default class CustomSignalingLayer extends SignalingLayer {
         return false;
     }
 
-    /**
-     * @inheritDoc
-     */
-    setTrackSourceName(ssrc, sourceName) { // eslint-disable-line no-unused-vars
-    }
-
```
jallamsetty1, line 116: The implementation of `setSSRCOwner` and `updateSsrcOwnersOnLeave` here and in `SignalingLayerImpl` have diverged. Since Spots's wireless connection uses a mock signaling connection, I think we can remove the impl of those methods here but it needs to be tested.

jallamsetty1: @saghul WDYT?

saghul: Agreed! I think it's safe, but we can test against alpha, once it lands.

hristoterezov: I don't see how anything would break but testing once it lands is always a good idea :)

jallamsetty1 (approved):

PR #2368, opened 2023-10-06
ref(CustomSignaling): Remove ssrcOwners&chatRoom.

hristoterezov:


saghul:
There are still a number of calls to these APIs in JingleSessionPC: https://github.com/search?q=repo%3Ajitsi%2Flib-jitsi-meet%20setSSRCOwner&type=code how is that supposed to work here?

PR #2371, opened 2023-10-10
feat(ScreenObtainer): add more control over screen obtainer

DanielMcAssey:
We have implemented these changes on our application, to allow for additional control of the screen obtainer on the browser.

It may not apply to Jitsi, however I thought I would propose these changes for more control over the library.

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -189,7 +189,9 @@ const ScreenObtainer = {
 
         const audio = this._getAudioConstraints();
         let video = {};
+        const constraintOpts = {};
         const {
+            desktopDisplaySurface,
             desktopSharingFrameRate,
             desktopSurfaceSwitching,
             desktopSelfBrowserSurface,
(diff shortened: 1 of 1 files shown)


modules/RTC/ScreenObtainer.js
```diff
@@ -212,13 +237,19 @@ const ScreenObtainer = {
             }
         }
 
+        // Allow a user to be shown a preference for what screen is to be captured.
+        if (browser.isSafari() && browser.isEngineVersionGreaterThan(11) && desktopDisplaySurface) {
```
saghul, line 241: We don't support Safari < 13 so you can skip the version check.

saghul (commented):
Changes generally LGTM, WDYT @jallamsetty1 ?

modules/RTC/ScreenObtainer.js
DanielMcAssey: Thank you! Updated to remove that

jallamsetty1:
Thanks for the PR @DanielMcAssey! The PR LGTM. It would be nice to bunch all these options together under `screenshareSettings` in config.js. We have been trying to do this for new settings so that config.js doesn't look cluttered.

DanielMcAssey:
Hey @jallamsetty1, thanks for the feedback. I have just updated it to put all the new settings under screenShareSettings

jallamsetty1:
Jenkins test this please.

jallamsetty1 (approved):

Issue #2370, opened 2023-10-10
Race condition may cause JitsiConference.prototype._doReplaceTrack to hang forever


athouary:
## Description
I consume `lib-jitsi-meet` in a custom user interface to connect and use a self-hosted `docker-jitsi-meet`. I've encountered some cases where the promise returned by`JitsiConference.addTrack` is pending forever, and after long researches, I've finally understood why.

The `JitsiConference.prototype._doReplaceTrack` promise resolves once the `replaceTrack` on all the existing Jingle Session (P2P and/or JVB) resolve. `JingleSessionPC` relies on `AsyncQueue` and pushes the `replaceTrack` call into it. In some rare case, when a 3rd user joins the conference and it goes from P2P to JVB, then the P2P JingleSession closes and the AsyncQueue is killed. This results in the `replaceTrack` to be pending forever.

## Current behavior
A call to `JitsiConference.addTrack`  or `JitsiConference.replaceTrack` may hang forever if the conference goes from P2P to JVB in the mean time.

## Expected Behavior
The promises returned by those functions should either be resolved or rejected. Actually, it should most probably resolve and ignore the P2P session since it's not relevant anymore.

## Possible Solution
We should probably call the all the pending task callbacks when killing an AsyncQueue, with a specific error that `replaceTrack` would catch. It would then resolve its promise (it seems ok not to manage to replace a track on a non-existing peer connection).

A more simple but dirtier solution would be to reject the promise after a timeout.

## Steps to reproduce

1. Client A joins a Jitsi conference
2. Client B joins a Jitsi conference
3. Client A adds local tracks to the conference
4. Client B adds local tracks to the conference
5. Client C joins the conference before the tracks are completely added for one or both other users

## Environment details
Reproduced on (non-exhaustive): 
* `lib-jitsi-meet` `v1699.0.0+683dc0fb`
* `docker-jitsi-meet` `stable-8960`
* Clients on Chrome 117

saghul:
> We should probably call the all the pending task callbacks when killing an AsyncQueue, with a specific error that `replaceTrack` would catch. It would then resolve its promise (it seems ok not to manage to replace a track on a non-existing peer connection).

This sounds about right!

PR #2373, opened 2023-10-11
fix(JingleSessionPC): call task callbacks upon clearing (#2370)

athouary:
This ensures that promises relying on the task callback will settle.

### DIFF
--- a/modules/util/AsyncQueue.js
+++ b/modules/util/AsyncQueue.js
@@ -4,6 +4,19 @@ import { queue } from 'async-es';
 const logger = getLogger(__filename);
 
 /**
+ * Error to be passed to a callback of a queued task when the queue is cleared.
+ */
+export class ClearedQueueError extends Error {
+    /**
+     * Creates new instance.
+     */
+    constructor(message) {
+        super(message);
+        this.name = 'ClearedQueueError';
+    }
+}
+
+/**
  * A queue for async task execution.
  */
 export default class AsyncQueue {
@@ -13,12 +26,16 @@ export default class AsyncQueue {
     constructor() {
         this._queue = queue(this._processQueueTasks.bind(this), 1);
         this._stopped = false;
+        this._taskCallbacks = new Map();
     }
 
     /**
      * Removes any pending tasks from the queue.
      */
     clear() {
+        for (const finishedCallback of this._taskCallbacks.values()) {
+            finishedCallback(new ClearedQueueError('The queue has been cleared'));
+        }
         this._queue.kill();
     }
 
@@ -31,6 +48,8 @@ export default class AsyncQueue {
         } catch (error) {
             logger.error(`Task failed: ${error?.stack}`);
             finishedCallback(error);
+        } finally {
+            this._taskCallbacks.delete(task);
         }
     }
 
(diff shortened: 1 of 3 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -1321,6 +1321,11 @@ export default class JingleSessionPC extends JingleSession {
             workFunction,
             error => {
                 if (error) {
+                    if (error instanceof ClearedQueueError) {
+                        // The session might have been terminated before the task was executed, making it obsolete.
+                        logger.debug(`${this} ICE restart task aborted: session terminated`);
+                        success();
```
saghul, line 1327: We should return here

modules/xmpp/JingleSessionPC.js
```diff
@@ -2194,6 +2199,11 @@ export default class JingleSessionPC extends JingleSession {
                 workFunction,
                 error => {
                     if (error) {
+                        if (error instanceof ClearedQueueError) {
+                            // The session might have been terminated before the task was executed, making it obsolete.
+                            logger.debug(`${this} renegotiation after addTrack aborted: session terminated`);
+                            resolve();
```
saghul, line 2205: Ditto, return here too

modules/xmpp/JingleSessionPC.js
```diff
@@ -2306,6 +2316,11 @@ export default class JingleSessionPC extends JingleSession {
                 workFunction,
                 error => {
                     if (error) {
+                        if (error instanceof ClearedQueueError) {
+                            // The session might have been terminated before the task was executed, making it obsolete.
+                            logger.debug('Replace track aborted: session terminated');
+                            resolve();
```
saghul, line 2322: Ditto

modules/xmpp/JingleSessionPC.js
```diff
@@ -2504,6 +2519,11 @@ export default class JingleSessionPC extends JingleSession {
                 workFunction,
                 error => {
                     if (error) {
+                        if (error instanceof ClearedQueueError) {
+                            // The session might have been terminated before the task was executed, making it obsolete.
+                            logger.debug(`${this} ${operationName} aborted: session terminated`);
+                            resolve();
```
saghul, line 2525: Ditto

saghul (changes requested):

modules/util/AsyncQueue.js
```diff
@@ -31,6 +48,8 @@ export default class AsyncQueue {
         } catch (error) {
             logger.error(`Task failed: ${error?.stack}`);
             finishedCallback(error);
```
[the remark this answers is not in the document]
saghul: You need to try-catch here too.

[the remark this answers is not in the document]
athouary: I don't understand why, an error during the callback here would not interrupt anything (there's nothing to process afterwards). Isn't it better to let the error bubble up instead of just logging it?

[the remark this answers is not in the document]
saghul: Ah right, sorry I hadn't looked at the whole context.

saghul (approved):
LGTM!

saghul:

athouary:
Hi @saghul, if everything is ok with this PR, when do think it will be merged?

Also, I've created a [PR on jitsi/rtcstats](https://github.com/jitsi/rtcstats/pull/33), I'm not sure I should ask to get it reviewed and merged.

PR #2374, opened 2023-10-12
feat(RTCStats) Suppress unnecessary error message when rtcstatsEnabled === false

AndyLeezard:
There are use cases where `RTCStats.init` is called without the argument property `initConfig.analytics.rtcstatsEnabled` being `true` - it is because by default it is false.

In those cases, calling `connection.initJitsiConference("conference", confOptions);` result in this error below:

Screenshot_235

RTCStats is sending a somewhat misleading console error message (as if the user deliberately called `RTCStats.start` without initializing `RTCStats.init`), whereas the browser technically did call `RTCStats.init` before `RTCStats.start` - only `initConfig.analytics.rtcstatsEnabled` was `false` by default.

This pull request handles cases where RTCStats is initialized without being enabled.

Thank you for reviewing my pull request.

### DIFF
--- a/modules/RTCStats/RTCStats.ts
+++ b/modules/RTCStats/RTCStats.ts
@@ -34,8 +34,8 @@ function connectionFilter(config) {
  * Config and conference changes are handled by the start method.
  */
 class RTCStats {
-    #initialized = false;
-    #trace: any = null;
+    _initialized = false;
+    _trace: any = null;
     public events: EventEmitter = new EventEmitter();
 
     /**
@@ -59,7 +59,7 @@ class RTCStats {
         // If rtcstats is not enabled or already initialized, do nothing.
         // Calling rtcsatsInit multiple times will cause the global objects to be rewritten multiple times,
         // with unforeseen consequences.
-        if (!rtcstatsEnabled || this.#initialized) return;
+        if (!rtcstatsEnabled || this._initialized) return;
 
         rtcstatsInit(
             { statsEntry: this.sendStatsEntry.bind(this) },
@@ -70,7 +70,7 @@ class RTCStats {
               eventCallback: (event) => this.events.emit(RTC_STATS_PC_EVENT, event)}
         );
 
-        this.#initialized = true;
+        this._initialized = true;
     }
 
     /**
@@ -107,7 +107,7 @@ class RTCStats {
         if (!rtcstatsEnabled) return;
 
         // If rtcstats proxy module is not initialized, do nothing.
-        if (!this.#initialized) {
+        if (!this._initialized) {
             logger.error('Calling start before RTCStats proxy module is initialized.');
 
             return;
@@ -127,12 +127,12 @@ class RTCStats {
             const endpointId = conference.myUserId();
             const meetingUniqueId = conference.getMeetingUniqueId();
 
-            this.#trace = traceInit(traceOptions);
+            this._trace = traceInit(traceOptions);
 
             // Connect to the rtcstats server instance. Stats (data obtained from getstats) won't be send until the
             // connect successfully initializes, however calls to GUM are recorded in an internal buffer even if not
             // connected and sent once it is established.
-            this.#trace.connect(isBreakoutRoom);
+            this._trace.connect(isBreakoutRoom);
 
             const identityData = {
                 ...confConfig,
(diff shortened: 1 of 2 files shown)


andrei-gavrilescu:
Hey Andy, nice catch and thanks for the contribution. You're right that the error message is misleading and shouldn't be logged if rtcstatsEnabled is false. Although I think there's a bit of confusion regarding the flow, rtcstatsEnabled is from the conference config and not a local state of RTCStats, so if that's set to false there should be no initialization going on and the `.start` method should simply return if that's the case.
With that being said, simply moving the check from https://github.com/jitsi/lib-jitsi-meet/blob/bae4a9c70ada6d4bba46bba2fd271ab5d0ec5bb9/modules/RTCStats/RTCStats.ts#L124 to the beginning of the start method should fix the problem, wdyt?
Something like this

```
start(conference: JitsiConference) {
        const { 
            options: { 
                config : confConfig = {},
                name: confName = ''
            } = {},
            _statsCurrentId : displayName = ''
        } = conference; 

const {
            analytics: {
                rtcstatsEnabled = false,
                rtcstatsEndpoint: endpoint = '',
                rtcstatsUseLegacy: useLegacy = false
            } = {}
        } = confConfig;

        // Reset the trace module in case it wasn't during the previous conference.
        // Closing the underlying websocket connection and deleting the trace obj.
        this.reset();

        // The new conference config might have rtcstats disabled, so we need to check again.
        if (!rtcstatsEnabled) return;
    
        
        // If rtcstats proxy module is not initialized, do nothing.
        if (!this._initialized) {
            logger.error('Calling start before RTCStats proxy module is initialized.');

            return;
        }

**Comment by AndyLeezard** [comment]
It looks like the argument `(conference: JitsiConference)` does not provide the property `rtcstatsEnabled` from its `options.config.analytics` - the conference does not save this info in itself either.
The property exists in type `IRTCStatsConfiguration`, which is defined at the moment of `JitsiMeetJS.init()` that calls `Statistics.init()`.

The way I see it is, it's either we elaborate the message while lowering its logger level.
For example: 

```javascript
start(conference: JitsiConference) {
if (!this._initialized) {
            logger.info("RTCStats - 'start' method requires calling 'init' method (JitsiMeetJS.init) with analytics.rtcstatsEnabled config option being true.");
            return;
}
```

Or let RTCStats remember if `rtcstatsEnabled` was enabled at the moment of initialization (my initial approach)

Or turning `rtcstatsEnabled` to `true` by default instead of `false`.

modules/RTCStats/RTCStats.ts
```diff
@@ -34,19 +34,19 @@ function connectionFilter(config) {
  * Config and conference changes are handled by the start method.
  */
 class RTCStats {
-    private _initialized: boolean = false;
-    private _trace: any = null;
+    #initialized = false;
+    #trace: any = null;
```
AndyLeezard, line 38: Use of JS-level runtime-safe private identifier prefix '#'.

AndyLeezard: Sure thing!
It was a merely optional suggestion.
It's rolled back.

modules/RTCStats/RTCStats.ts
```diff
@@ -60,13 +60,13 @@ class RTCStats {
...
-        rtcstatsInit( 
-            { statsEntry: this.sendStatsEntry.bind(this) }, 
+
+        rtcstatsInit(
+            { statsEntry: this.sendStatsEntry.bind(this) },
             { connectionFilter,
-              pollInterval,
-              useLegacy,
-              sendSdp,
+                pollInterval,
```
andrei-gavrilescu, line 67: Nit: looks like some extra indentation.

modules/RTCStats/RTCStats.ts
```diff
@@ -34,19 +34,19 @@ function connectionFilter(config) {
  * Config and conference changes are handled by the start method.
  */
 class RTCStats {
-    private _initialized: boolean = false;
-    private _trace: any = null;
+    _initialized = false;
```
andrei-gavrilescu, line 37: Nit: revert to using private, and it seems boolean type was removed.

andrei-gavrilescu (commented):
Some very minor nits, otherwise looks good! Thanks for looking into this :)

modules/RTCStats/RTCStats.ts
AndyLeezard: Right away. Reverted back to using the private modifier and type assertion on boolean.

andrei-gavrilescu:
Looks good!

andrei-gavrilescu (approved):

PR #2375, opened 2023-10-16
feat(xmpp) remove support for the deprecated options.bosh

saghul:
> 4 years seems like ample time :-)

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -119,7 +119,6 @@ export default class XMPP extends Listenable {
      * @param {Object} options
      * @param {String} options.serviceUrl - URL passed to the XMPP client which will be used to establish XMPP
      * connection with the server.
-     * @param {String} options.bosh - Deprecated, use {@code serviceUrl}.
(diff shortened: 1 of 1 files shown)


damencho:
Will it work with default config.js?
https://github.com/jitsi/jitsi-meet/blob/8a2e4bc6281ee2d7a0602d359a03c8dc9c2c34e0/config.js#L49

damencho:
I think we handle it in jitsi-meet, so maybe its good.

damencho:
I wonder whether the lib-jitsi-meet users are using the serviceUrl thingy ...

paweldomas:
> I wonder whether the lib-jitsi-meet users are using the serviceUrl thingy ...

What I'd suggest is to throw an error with a meaningful message which makes fixing it a breeze:
```
if (options.bosh && !options.serviceUrl) {
  throw new Error('The "bosh" option is no longer supported, please use "serviceUrl" instead');
}
```

saghul:
Updated PTAL!

paweldomas (approved):

Issue #2379, opened 2023-10-18
crashes on iOS 14 / 15 with d.getHeight is not a function.  'd.getHeight' is undefined


Corey-Maler:
*This Issue tracker is only for reporting bugs and tracking code related issues.*

Before posting, please make sure you check community.jitsi.org to see if the same or similar bugs have already been discussed. General questions, installation help, and feature requests can also be posted to community.jitsi.org.

## Description
Crashes on iOS 14 / 15 with error "l.getHeight is not a function (getHeight is undefined)"

After digging a bit, I found that it refers to this line https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/RTC/JitsiLocalTrack.js#L115

This change was introduced in https://github.com/jitsi/lib-jitsi-meet/pull/2356 this PR.

I assume it is a confusion here since `track` at that line refers to the native MediaStreamTrack which does not have `getHeight` method and since it is under condition, it does not work only for specific safari versions

## Current behavior
Cannot enable camera, "failed to access your camera" is displayed

(Technically)
`createLocalTrack` crashes with an error

## Expected Behavior
Camera works

(Technically)
`createLocalTrack` does not crash with an error

## Steps to reproduce
Open https://beta.meet.jit.si on any iOS 14 or 15, not even necessary to start a call

saghul:
@gabiborlea can you PTAL?

gabiborlea:
You are right, I'll create a PR to fix this!

gabiborlea:
Fixed with this PR: https://github.com/jitsi/lib-jitsi-meet/pull/2380.

PR #2380, opened 2023-10-18
fix(track): get track height

gabiborlea:


modules/RTC/JitsiLocalTrack.js
```diff
@@ -112,8 +112,8 @@ export default class JitsiLocalTrack extends JitsiTrack {
         // Safari returns an empty constraints object, construct the constraints using getSettings.
         if (!Object.keys(this._constraints).length && videoType === VideoType.CAMERA) {
             this._constraints = {
-                height: track.getHeight(),
-                width: track.getWidth()
+                height: track.getSettings().getHeight(),
```
saghul, line 115: Why not this.getHeight() ?

saghul (approved):
LGTM with a question.

modules/RTC/JitsiLocalTrack.js
Corey-Maler: I think I can help: track here is MediaStreamTrack, not jitsi track, which does not have `getHeight` and that caused issue https://github.com/jitsi/lib-jitsi-meet/issues/2379 

I would probably suggest checking if it is possible to rename variable as nativeTrack or something to prevent further confusion

Corey-Maler: oh, I see your point, sorry, missed `this` instead of `track`. Good point though!

gabiborlea: no particular reason for this.getHeight()

gabiborlea: I'll replace it

saghul (approved):

PR #2384, opened 2023-10-27
fix(SS): Reverse the encoding order for Firefox desktop streams.

jallamsetty1:
Fixes an issue where screenshare from Firefox appears blurry on other endpoints in the call.


damencho (approved):

jallamsetty1:
Know test failure.

PR #2391, opened 2023-11-13
Insert simulcast recv line for m-lines associated with all local sources on Firefox

jallamsetty1:
Currently, it is being done only for the first video source.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -284,130 +284,6 @@ export class TPCUtils {
     }
 
     /**
-     * Ensures that the ssrcs associated with a FID ssrc-group appear in the correct order, i.e.,
-     * the primary ssrc first and the secondary rtx ssrc later. This is important for unified
-     * plan since we have only one FID group per media description.
-     * @param {Object} description the webRTC session description instance for the remote
-     * description.
-     * @private
-     */
-    ensureCorrectOrderOfSsrcs(description) {
-        const parsedSdp = transform.parse(description.sdp);
-
-        parsedSdp.media.forEach(mLine => {
-            if (mLine.type === MediaType.AUDIO) {
-                return;
-            }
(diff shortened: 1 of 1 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -401,11 +377,29 @@ export class TPCUtils {
         const simulcastLine = browser.isFirefox() && browser.isVersionGreaterThan(71)
             ? `recv ${SIM_LAYER_RIDS.join(';')}`
             : `recv rid=${SIM_LAYER_RIDS.join(';')}`;
+        const sdp = transform.parse(desc.sdp);
+        const mLines = sdp.media.filter(m => m.type === MediaType.VIDEO);
+        const senderMids = Array.from(this.pc._localTrackTransceiverMids.values());
```
hristoterezov: Do we need `Array.from`? The return value from Map.values should be iterator which also has find method. Wouldn't that work?

jallamsetty1: Technically its not needed, added here for clarity. I don't see any harm in using it though :)

hristoterezov: Generally for performance? We are adding a redundant operation. 

Also I don't understand the clarity argument. Actually for me it was the opposite I started wondering what would the .values() call return so that we need to convert it to array. Would you explain what the array conversation clarifies in this case?

jallamsetty1: Actually that doesn't seem to be true. `values()` returns a Map iterator which doesn't support find.

hristoterezov: I was looking into https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Iterator/find . But just gave it a try and it seems to not work for some reason...

hristoterezov (approved):

PR #2399, opened 2023-11-20
Remove callstats support

bgrozev:
- ref: Remove PrecallTest.
- ref: Remove startCallStats.
- ref: Remove callstats support.


jallamsetty1 (dismissed):

JitsiConference.js
```diff
@@ -2706,14 +2671,10 @@ JitsiConference.prototype.sendFeedback = function(overallFeedback, detailedFeedb
...
 /**
- * Returns true if the callstats integration is enabled, otherwise returns
- * false.
- *
- * @returns true if the callstats integration is enabled, otherwise returns
- * false.
+ * @returns false, since callstats in not supported anymore.
  */
 JitsiConference.prototype.isCallstatsEnabled = function() {
-    return this.statistics.isCallstatsEnabled();
```
damencho, line 2716: Maybe we should add a deprecation notice in the js docs and add a print that this will be removed, wdyt?

damencho (approved):

PR #2400, opened 2023-11-20
fix: Double MUC_JOINED event.

hristoterezov:
When a self presence is received ( from mute or something else) during
leave operation, after presence unavailable is sent and before the final
presence unavaliable is received, we were interpreting it as muc joined
since the joined flag was immediately set to false.


damencho (approved):

damencho:
Hum ... seems that broke leaving lobby room, according to tests :)

damencho (approved):

PR #2405, opened 2023-12-05
fix(AsyncQueue): call the callback only if it's a function (#2370)

athouary:
Following my PR for #2370, I ran into `TypeError`s in my console where the task callback was called without being defined. The error is caught and logged so it's not critical, but it pollutes the console.


saghul:
Was the callback undefined?

athouary:
Yes it was. In several places some tasks are pushed into the queue with no callback.

saghul:
Then I think using optional chaining would look cleaner: func?.(arg);

saghul:

saghul (approved):

PR #2406, opened 2023-12-06
Output single TypeScript declaration file

DanielMcAssey:
Hey!
Opening this as a draft, but its essentially finished. Wanted to get thoughts on this,

1. Last step from generating a single declaration file was fixing the Listenable, to correctly implement EventEmitter instead of instantiating it. (As recommended by NodeJS) https://nodejs.org/api/events.html#events_events
2. Updated the gen-types command to output a single declaration file, which is now referenced in package.json
3. Updated editorconfig to match the 2 spaces in the json files

Now the main potentially breaking change is:

1. Listenable addEventListener does not return an unsubscribe function anymore, however I could not find a use of it in the code.
2. The types/auto directory is no longer generated, we could get around this by creating a new command, but I thought it was redundant.

What do you think @saghul ?

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -9,6 +9,9 @@ indent_style = space
 max_line_length = 80
 trim_trailing_whitespace = true
 
+[*.json]
+indent_size = 2
+
 [*.md]
 max_line_length = 120
 trim_trailing_whitespace = false
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ npm-*.log
 stats.json
 .vscode
 dist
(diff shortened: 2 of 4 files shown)


DanielMcAssey:
Looking at this, its probably a little more work, which I am happy to do, if what I described is okay?

saghul:
General, quick observation: I like the direction and would take it a step further and drop Listenable altogether since it no longer does anything, does it?

I also think a single types file is a good idea!

I'd like @hristoterezov and @jallamsetty1 to weigh in as well.

PR #2407, opened 2023-12-06
Unify events and output single TypeScript declaration

DanielMcAssey:
Found a less destructive way than my previous PR #2406 

Copy from that PR:

1. Last step from generating a single declaration file was fixing the Listenable, to correctly implement and extend EventEmitter instead of instantiating it. (As recommended by NodeJS) https://nodejs.org/api/events.html#events_events
2. Updated the gen-types command to output a single declaration file, which is now referenced in package.json

Now the main potentially breaking change is:

1. Updated to add addCancellableEventListener to return an unsubscribe function, only one instance found where it was being used.
2. The types/auto directory is no longer generated, we could get around this by creating a new command, but I thought it was redundant.

What do you think @saghul ?

### DIFF
--- a/modules/RTC/CodecSelection.spec.js
+++ b/modules/RTC/CodecSelection.spec.js
@@ -1,5 +1,3 @@
-import EventEmitter from 'events';
-
 import * as JitsiConferenceEvents from '../../JitsiConferenceEvents.ts';
 import Listenable from '../util/Listenable.js';
 import JingleSessionPC from '../xmpp/JingleSessionPC.js';
@@ -42,7 +40,6 @@ class MockConference extends Listenable {
         };
 
         this.activeMediaSession = undefined;
-        this.eventEmitter = new EventEmitter();
         this.mediaSessions = [];
         this.participants = [];
         this._signalingLayer = new MockSignalingLayerImpl();
--- a/.gitignore
+++ b/.gitignore
@@ -14,6 +14,7 @@ npm-*.log
 stats.json
 .vscode
 dist
+index.d.ts
 types/auto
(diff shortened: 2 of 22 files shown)


saghul:
I'd call that new EventManager EventEmitter and just alias the import. IMHO it would avoid confusion.

DanielMcAssey:
@saghul good idea! All sorted now I think.

saghul:

saghul:

saghul:
Generally LGTM but I'd like to hear what @jallamsetty1 and @hristoterezov think.

DanielMcAssey:
Just removed the on/off alias as it was redundant on the EventEmitter as the Node EventEmitter already does this

saghul:
Ping @jallamsetty1 @hristoterezov

saghul (dismissed):
LGTM. I'd do another pass (after this PR) and completely get rid of Listenable and EventManager, they really server very little purpose now.

modules/RTC/CodecSelection.spec.js
```diff
@@ -42,7 +40,6 @@ class MockConference extends Listenable {
         };
 
         this.activeMediaSession = undefined;
-        this.eventEmitter = new EventEmitter();
```
hristoterezov, line 45: Suggestion: lets use `this.emit` instead of `this.eventEmitter.emit` in this class? WDYT?

hristoterezov (commented):
LGTM! Needs rebasing tho!

DanielMcAssey:
Sorted! Just fixed the conflict

hristoterezov (approved):

PR #2409, opened 2023-12-08
feat(RTC): add support for creating non-standard tracks

DanielMcAssey:
We have this manually patched into our copy of lib-jitsi-meet, but thought it might be worth sharing.

This change allows applications that use lib-jitsi-meet to manually create JitiLocalTracks from some track metadata and mediastream, similar to what the library does internally, but exposing it.

We use it to create a track for our Canvas, that we then add to the conference, similar to below:

```
    const canvasStream = myCanvas.captureStream(desktopShareFPS);
    const newTracks = JitsiMeetJS.createLocalTracksFromMediaStreams([
      {
        stream: canvasStream,
        sourceId: 'my:canvas',
        sourceType: 'canvas',
        track: canvasStream.getVideoTracks()[0],
        videoType: 'desktop'
      }
    ]);
```

This would possibly fix https://github.com/jitsi/lib-jitsi-meet/pull/1593

### DIFF
--- a/JitsiMeetJS.spec.ts
+++ b/JitsiMeetJS.spec.ts
@@ -13,7 +13,6 @@ describe('JitsiMeetJS', () => {
                 stream: canvasStream,
                 sourceType: 'canvas',
                 mediaType: MediaType.VIDEO,
-                track: canvasStream.getVideoTracks()[0],
                 videoType: VideoType.DESKTOP
             };
             const newTracks = JitsiMeetJS.createLocalTracksFromMediaStreams([ trackInfo ]);
@@ -22,22 +21,19 @@ describe('JitsiMeetJS', () => {
             expect(newTracks.length).toBe(1);
         });
 
-        it('throws an error if track is not from the same stream', () => {
+        it('throws an error if track is not the correct media type', () => {
             const canvas = document.createElement('canvas');
-            const otherCanvas = document.createElement('canvas');
 
             const canvasStream = canvas.captureStream(5);
-            const otherCanvasStream = otherCanvas.captureStream(5);
             const trackInfo = {
                 stream: canvasStream,
                 sourceType: 'canvas',
-                mediaType: MediaType.VIDEO,
-                track: otherCanvasStream.getVideoTracks()[0],
+                mediaType: MediaType.AUDIO,
                 videoType: VideoType.DESKTOP
             };
 
             expect(() => JitsiMeetJS.createLocalTracksFromMediaStreams([ trackInfo ]))
-                .toThrowError(JitsiTrackErrors.TRACK_MISMATCHED_STREAM);
+                .toThrowError(JitsiTrackErrors.TRACK_NO_STREAM_TRACKS_FOUND);
         });
     });
 });
--- a/JitsiMeetJS.spec.ts
+++ b/JitsiMeetJS.spec.ts
@@ -1,0 +1,43 @@
+import JitsiMeetJS from './JitsiMeetJS';
+import { VideoType } from './service/RTC/VideoType';
+import { MediaType } from './service/RTC/MediaType';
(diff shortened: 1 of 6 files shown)


saghul:
Yes! I have needed this myself too and was on my list of things, Good to see!

JitsiMeetJS.ts
```diff
@@ -421,6 +421,16 @@ export default {
             });
     },
 
+    /**
+     * Manually create JitsiLocalTrack's from the provided track info, by exposing the RTC method
+     *
+     * @param {Array<Object>} tracksInfo - array of track information
```
saghul, line 427: Can you please document this?

saghul: One more thing: could we generate the sourceId internally?

DanielMcAssey: Sure! Ill do those changes, for sourceId, would a random number be sufficient? Or should it follow a certain pattern?

DanielMcAssey: I have just added everything, except the bit to generate source ID. is the general structure okay?

Added some tests too

JitsiMeetJS.ts
```diff
@@ -421,6 +429,35 @@ export default {
...
+            }
+
+            if (tracks.length > 1) {
+                throw new JitsiTrackError(JitsiTrackErrors.TRACK_TOO_MANY_TRACKS_IN_STREAM, null, null);
+            }
+
+            trackInfo.track = tracks[0];
+
+            if (!trackInfo.sourceId) {
+                trackInfo.sourceId = 'GENERATEDVALUEHERE';
```
jallamsetty1, line 454: `sourceId` is the ID of the desktop sharing source and is used only for remote control feature in the Electron app. What purpose does it serve there?

DanielMcAssey: No purpose, happy to remove if not needed

DanielMcAssey: Just updated the docs and removed the sourceId on the new function

JitsiMeetJS.spec.ts
```diff
@@ -0,0 +1,39 @@
+import JitsiMeetJS from './JitsiMeetJS';
+import { VideoType } from './service/RTC/VideoType';
+import { MediaType } from './service/RTC/MediaType';
+import { JitsiTrackErrors } from './JitsiTrackErrors';
+
+describe('JitsiMeetJS', () => {
```
saghul, line 6: Thanks for this! ❤️

JitsiMeetJS.ts
```diff
@@ -90,6 +91,13 @@ interface IJitsiMeetJSOptions {
     }
 }
 
+interface ICreateLocalTrackFromMediaStreamOptions {
+    stream: MediaStream,
+    sourceType: string,
```
saghul, line 96: What is this used for, other than documentation?

saghul (approved):
LGTM with a question. Excellent work!

JitsiMeetJS.ts
DanielMcAssey: Purely documentation and typings for the function

DanielMcAssey: Or if you mean the sourceType, then it seems to be a way for client apps to determine the type of stream, e.g. https://github.com/jitsi/lib-jitsi-meet/blob/dd81069cb0e18a387c80cae149bfe6f08b52a6c4/modules/proxyconnection/ProxyConnectionService.js#L286

saghul: Gotcha. We are planning to phase that one out, so I can take of it when the day comes. Thanks for checking!

jallamsetty1:
@DanielMcAssey Thanks for the PR. I will update the handbook.

PR #2417, opened 2023-12-18
fix(types): fix last few types

DanielMcAssey:
Need to modify these types to get lib-jitsi-meet working in a typescript project.

Some of the types were not compatible, removed some references to hand-crafted from one of the files, as the auto types cannot find them.


saghul:

PR #2419, opened 2023-12-18
fix(types): update to TS 5.3.3 and fix types

DanielMcAssey:
Took a bit longer than expected, as there were a lot of types to update.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,7 +59,7 @@
     "karma-webpack": "5.0.0",
     "process": "0.11.10",
     "string-replace-loader": "3.0.3",
-    "typescript": "4.3.5",
+    "typescript": "5.3.3",
     "webpack": "5.76.0",
     "webpack-bundle-analyzer": "4.4.2",
     "webpack-cli": "4.9.0"
@@ -69,7 +69,7 @@
     "build:webpack": "LIB_JITSI_MEET_COMMIT_HASH=$(git rev-parse --short HEAD 2>/dev/null) webpack",
     "build:webpack-dev": "webpack --mode development",
     "build:tsc": "tsc --build --clean && tsc",
-    "gen-types": "tsc --declaration --emitDeclarationOnly --out index.js",
+    "gen-types": "tsc --declaration --emitDeclarationOnly --outFile index.d.ts",
     "lint": "eslint .",
     "lint-fix": "eslint . --fix",
     "postinstall": "patch-package",
--- a/package.json
+++ b/package.json
@@ -87,9 +87,7 @@
     "index.d.ts"
   ],
   "overrides": {
-    "karma-rollup-preprocessor": {
-      "rollup": "npm:@rollup/wasm-node"
-    }
+    "rollup": "npm:@rollup/wasm-node"
   },
   "license": "Apache-2.0"
 }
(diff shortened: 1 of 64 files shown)


saghul:
Impressive work, thank you!

Quick question before I dive deeper: can't we do something to avoid those weird inline imports in the JSdocs?

DanielMcAssey:
> Impressive work, thank you!
> 
> Quick question before I dive deeper: can't we do something to avoid those weird inline imports in the JSdocs?

Not that I could find :(

It seems that something changed between 4 and 5, and referencing global imports. In an ideal world, when this is fully TypeScript, it wont matter, but its because of the way declaration is generated, it needs to reference the appropriate module.

saghul:
Sorry but that adds too much curn, it's not a change I'm comfortable adding.

DanielMcAssey:
Fair enough. I saw your PR, want me to update this one? ill update it to remove the import statements, but keep the fixed types, then move it to the types folder

DanielMcAssey:
@saghul I've just pushed an update, it removes the import statements, but still fixes the types and keeps it in the types folder.

Fingers crossed it will be easier for TypeScript modules to be used in JSDoc

saghul:
Thank you! I'll take another look!

package.json
```diff
@@ -84,8 +84,10 @@
   "types": "types/index.d.ts",
   "files": [
     "dist",
-    "types",
-    "index.d.ts"
+    "types"
   ],
+  "overrides": {
+    "rollup": "npm:@rollup/wasm-node"
```
saghul, line 90: Can you elaborate on why this is necessary?

DanielMcAssey: Its a weird bug in npm, see my first comment. But its to do with the test framework using rollup, if you look at the logs of one of the early failed builds, it complains it cant find the native rollup.

The fix is to add an override to package.json, as seen here: https://github.com/vitejs/vite/issues/15167

DanielMcAssey: Specifically this run: https://github.com/jitsi/lib-jitsi-meet/actions/runs/7251356239/job/19753537267

DanielMcAssey:
Unfortunately, I tried removing the roll-up hack (Blegh 🤢) hoping that the npm issue had been fixed, but it hasnt. I have re-added just so this builds successfully.

saghul (approved):

saghul:

saghul:
Let's see if this passes Jenkins. I wonder if there is a problem with Jitsi meet using an older TS... Hopefully not.

saghul:
After Jenkins passes, can you pl add a new commit basically undoing this: https://github.com/jitsi/lib-jitsi-meet/commit/ca40744fa28c5a7bfb93196f0d49760e3c019262 so we can test it that way too?

DanielMcAssey:
Added the type line back back

saghul:

saghul:
Damn, sorry this fell through the cracks @DanielMcAssey :-/ Would you like to revive it?

DanielMcAssey:
Its all good! I noticed someone has already kicked off some of the migration, I will wait till that flushes through, and then I will get this updated :) If that sounds good to you?

saghul:
Sounds good! I did coincidentally get the rollup stuff fixed too!

PR #2436, opened 2024-01-22
fix(video-quality): Adjust encodings bitrates/scalefactor for high capture resolutions

jallamsetty1:
When the client is configured to capture camera sources at 1080p or higher, adjust the bitrates and the encodings scale factor accordingly.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -6,24 +6,18 @@ import CodecMimeType from '../../service/RTC/CodecMimeType';
 import { MediaDirection } from '../../service/RTC/MediaDirection';
 import { MediaType } from '../../service/RTC/MediaType';
 import { getSourceIndexFromSourceName } from '../../service/RTC/SignalingLayer';
-import { STANDARD_CODEC_SETTINGS } from '../../service/RTC/StandardVideoSettings';
+import {
+    SIM_LAYERS,
+    STANDARD_CODEC_SETTINGS,
+    VIDEO_QUALITY_LEVELS
+} from '../../service/RTC/StandardVideoSettings';
 import VideoEncoderScalabilityMode from '../../service/RTC/VideoEncoderScalabilityMode';
 import { VideoType } from '../../service/RTC/VideoType';
 import browser from '../browser';
 
 const logger = getLogger(__filename);
 const DESKTOP_SHARE_RATE = 500000;
-const SIM_LAYER_1_RID = '1';
-const SIM_LAYER_2_RID = '2';
-const SIM_LAYER_3_RID = '3';
 const VIDEO_CODECS = [ CodecMimeType.AV1, CodecMimeType.H264, CodecMimeType.VP8, CodecMimeType.VP9 ];
-
-// TODO - need to revisit these settings when 4K is the captured resolution. We can change the LD scale factor to 6
-// instead of 4 so that the lowest resolution will be 360p instead of 540p.
-export const HD_SCALE_FACTOR = 1.0;
-export const LD_SCALE_FACTOR = 4.0;
-export const SD_SCALE_FACTOR = 2.0;
-export const SIM_LAYER_RIDS = [ SIM_LAYER_1_RID, SIM_LAYER_2_RID, SIM_LAYER_3_RID ];
 
 /**
  * Handles track related operations on TraceablePeerConnection when browser is
@@ -38,6 +32,9 @@ export class TPCUtils {
     constructor(peerconnection) {
         this.pc = peerconnection;
         this.codecSettings = clonedeep(STANDARD_CODEC_SETTINGS);
+        this.l0ScaleFactor = SIM_LAYERS[0].scaleFactor;
+        this.l1ScaleFactor = SIM_LAYERS[1].scaleFactor;
+        this.l2ScaleFactor = SIM_LAYERS[2].scaleFactor;
         const videoQualitySettings = this.pc.options?.videoQuality;
 
         if (videoQualitySettings) {
@@ -50,7 +47,7 @@ export class TPCUtils {
                         && videoQualitySettings.maxbitratesvideo[codec.toUpperCase()]);
 
                 if (bitrateSettings) {
-                    [ 'low', 'standard', 'high', 'ssHigh' ].forEach(value => {
+                    [ 'low', 'standard', 'high', 'fullHd', 'ultraHd', 'ssHigh' ].forEach(value => {
                         if (bitrateSettings[value]) {
                             this.codecSettings[codec].maxBitratesVideo[value] = bitrateSettings[value];
                         }
@@ -93,35 +90,49 @@ export class TPCUtils {
         const isScreenshare = localVideoTrack.getVideoType() === VideoType.DESKTOP;
         let scalabilityMode = this.codecSettings[codec].useKSVC
             ? VideoEncoderScalabilityMode.L3T3_KEY : VideoEncoderScalabilityMode.L3T3;
-        let maxBitrate = codecBitrates.high;
+        const qualityLevel = VIDEO_QUALITY_LEVELS.find(level => level.height === newHeight);
+        let maxBitrate;
 
         if (this._isScreenshareBitrateCapped(localVideoTrack)) {
             scalabilityMode = VideoEncoderScalabilityMode.L1T3;
             maxBitrate = desktopShareBitrate;
         } else if (localVideoTrack.getVideoType() === VideoType.DESKTOP) {
             maxBitrate = codecBitrates.ssHigh;
+        } else if (qualityLevel) {
+            maxBitrate = codecBitrates[qualityLevel.level];
         }
 
         const config = {
             active: newHeight > 0,
             maxBitrate,
             scalabilityMode,
-            scaleResolutionDownBy: HD_SCALE_FACTOR
+            scaleResolutionDownBy: this.l2ScaleFactor
         };
 
         if (newHeight >= height || newHeight === 0 || isScreenshare) {
             return config;
         }
 
-        if (newHeight >= height / SD_SCALE_FACTOR) {
-            config.maxBitrate = codecBitrates.standard;
+        if (!qualityLevel) {
+            return config;
+        }
+
+        config.scaleResolutionDownBy = height / newHeight;
+
+        // Configure the sender to send all 3 spatial layers for resolutions 720p and higher.
+        switch (qualityLevel.level) {
+        case 'ultraHd':
+        case 'fullHd':
+        case 'high':
+            config.scalabilityMode = this.codecSettings[codec].useKSVC
+                ? VideoEncoderScalabilityMode.L3T3_KEY : VideoEncoderScalabilityMode.L3T3;
+            break;
+        case 'standard':
             config.scalabilityMode = this.codecSettings[codec].useKSVC
                 ? VideoEncoderScalabilityMode.L2T3_KEY : VideoEncoderScalabilityMode.L2T3;
-            config.scaleResolutionDownBy = SD_SCALE_FACTOR;
-        } else {
-            config.maxBitrate = codecBitrates.low;
+            break;
+        default:
             config.scalabilityMode = VideoEncoderScalabilityMode.L1T3;
-            config.scaleResolutionDownBy = LD_SCALE_FACTOR;
         }
 
         return config;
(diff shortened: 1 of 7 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -93,35 +90,49 @@ export class TPCUtils {
...
-            config.maxBitrate = codecBitrates.standard;
+        if (!qualityLevel) {
+            return config;
+        }
+
+        config.scaleResolutionDownBy = height / newHeight;
+
+        // Configure the sender to send all 3 spatial layers for resolutions 720p and higher.
+        switch (qualityLevel.level) {
+        case 'ultraHd':
```
saghul, line 124: Maybe make these constants?

modules/RTC/TPCUtils.js
```diff
@@ -93,35 +90,49 @@ export class TPCUtils {
...
 
         const config = {
             active: newHeight > 0,
             maxBitrate,
             scalabilityMode,
-            scaleResolutionDownBy: HD_SCALE_FACTOR
+            scaleResolutionDownBy: this.l2ScaleFactor
         };
 
         if (newHeight >= height || newHeight === 0 || isScreenshare) {
```
saghul, line 112:
```suggestion
        if (newHeight >= height || !config.active || isScreenshare) {
```

modules/RTC/TPCUtils.js
```diff
@@ -93,35 +90,49 @@ export class TPCUtils {
...
             return config;
         }
 
-        if (newHeight >= height / SD_SCALE_FACTOR) {
-            config.maxBitrate = codecBitrates.standard;
+        if (!qualityLevel) {
```
saghul, line 116: Fold this into the previous condition?

modules/RTC/TPCUtils.js
```diff
@@ -93,35 +90,49 @@ export class TPCUtils {
...
+            return config;
+        }
+
+        config.scaleResolutionDownBy = height / newHeight;
```
saghul, line 120: Do we want to round this in a specific way? Maybe to make it an integer rather than allow fractional scaling?

modules/RTC/TPCUtils.js
```diff
@@ -579,25 +606,26 @@ export class TPCUtils {
...
+                id: SIM_LAYERS[2].rid,
                 direction: 'recv'
             }
         ];
 
         // Firefox 72 has stopped parsing the legacy rid= parameters in simulcast attributes.
         // eslint-disable-next-line max-len
         // https://www.fxsitecompat.dev/en-CA/docs/2019/pt-and-rid-in-webrtc-simulcast-attributes-are-no-longer-supported/
+        const ridLine = rids.map(val => val.id).join(';');
         const simulcastLine = browser.isFirefox() && browser.isVersionGreaterThan(71)
```
saghul, line 626: While we're here, do we support FF < 71?

modules/connectivity/ConnectionQuality.js
```diff
@@ -61,32 +34,32 @@ let startBitrate = 800;
...
+function getTarget(simulcast, resolution, millisSinceStart, bitrates) {
     let target = 0;
     let height = Math.min(resolution.height, resolution.width);
 
     // Find the first format with height no bigger than ours.
-    let simulcastFormat = kSimulcastFormats.find(f => f.height <= height);
+    let qualityLevel = VIDEO_QUALITY_LEVELS.find(f => f.height <= height);
 
-    if (simulcastFormat && simulcast && videoQualitySettings.codec === CodecMimeType.VP8) {
+    if (qualityLevel && simulcast) {
```
saghul, line 46: This reads weird now, because it only applies to VP8, but I don't understand how this condition implies that.

saghul (commented):
Left a couple of comments, nothing major. Excellent work, I'm exited to try 4k for realz!

modules/RTC/TPCUtils.js
jallamsetty1: Thanks for the suggestion, I will fold in the `qualityLevel` check as well, can't believe I missed that 🤦‍♀️

modules/RTC/TPCUtils.js
jallamsetty1: We do need fractional scaling. For example, if we the video is captured at 1080p and we need to scale the stream down to 720p, we need to set the `scaleResolutionDownBy` value to 1.5 and it works as expected. I feel implementing a rounding logic to round the float to one decimal value might introduce unexpected errors. Since the requested/captured resolutions are all constants, we should be safe here. WDYT?

modules/RTC/TPCUtils.js
jallamsetty1: Looks like we don't have a min. version check for FF in the library. I think we should do that in a separate PR.

modules/connectivity/ConnectionQuality.js
jallamsetty1: Well actually, this doesn't apply to VP8 only anymore. H.264, AV1 and VP9 can also run in simulcast mode (via config overrides) and it is now being correctly detected here https://github.com/jitsi/lib-jitsi-meet/pull/2436/files#diff-c56ee5241e6d324b646a7ec9e618327097f918fba28f70c5396c2ccef69919ecR317 so I fixed this check.

modules/RTC/TPCUtils.js
```diff
@@ -88,40 +87,51 @@ export class TPCUtils {
      */
     _calculateActiveEncodingParams(localVideoTrack, codec, newHeight) {
         const codecBitrates = this.codecSettings[codec].maxBitratesVideo;
-        const height = localVideoTrack.getHeight();
-        const desktopShareBitrate = this.pc.options?.videoQuality?.desktopbitrate || DESKTOP_SHARE_RATE;
+        const trackHeight = localVideoTrack.getHeight();
+        const effectiveNewHeight = newHeight > trackHeight ? trackHeight : newHeight;
```
saghul, line 91: Not sure I'm following this right. Here we are picking `newHeight` only if it's _smaller_.

So what happens if we are currently only sending the middle layer and someone starts requesting the high layer? Won't the effective new height be the previous value here?

Since we use the new height to compute the active value, it feels weird we might use the current value for computing the bitrate.

modules/RTC/TPCUtils.js
```diff
@@ -88,40 +87,51 @@ export class TPCUtils {
...
-        let maxBitrate = codecBitrates.high;
+        const qualityLevel = VIDEO_QUALITY_LEVELS.find(level => level.height === effectiveNewHeight);
+        let maxBitrate;
 
         if (this._isScreenshareBitrateCapped(localVideoTrack)) {
             scalabilityMode = VideoEncoderScalabilityMode.L1T3;
             maxBitrate = desktopShareBitrate;
         } else if (localVideoTrack.getVideoType() === VideoType.DESKTOP) {
             maxBitrate = codecBitrates.ssHigh;
+        } else if (qualityLevel) {
```
saghul, line 104: If there is no defined qualityLevel we'll be setting `maxBitrate` to `undefined`. Is that ok?

When can that happen? Should we at least add a log line there?

Rounding to the nearest quality level sounds like the reasonable thing to do here IMHO, if we have no exact match.

480 is notably missing from the list of levels, for example.

modules/RTC/TPCUtils.js
```diff
@@ -161,13 +171,30 @@ export class TPCUtils {
...
      */
-    _getVideoStreamEncodings(videoType, codec) {
+    _getVideoStreamEncodings(localTrack, codec) {
+        const captureResolution = localTrack.getHeight();
         const codecBitrates = this.codecSettings[codec].maxBitratesVideo;
+        const videoType = localTrack.getVideoType();
+        let maxBitrate = codecBitrates.high;
+        const qualityLevel = VIDEO_QUALITY_LEVELS.find(level => level.height === captureResolution);
+
+        if (qualityLevel) {
```
saghul, line 184: Ditto.

modules/RTC/TPCUtils.spec.js
```diff
@@ -291,7 +291,136 @@ describe('TPCUtils', () => {
...
+                scalabilityModes = tpcUtils.calculateEncodingsScalabilityMode(track, codec, height);
+                expect(scalabilityModes[0]).toBe(VideoEncoderScalabilityMode.L2T3_KEY);
+                expect(scalabilityModes[1]).toBe(undefined);
+                expect(scalabilityModes[2]).toBe(undefined);
+
+                scaleFactor = tpcUtils.calculateEncodingsScaleFactor(track, codec, height);
+                expect(scaleFactor[0]).toBe(6);
+            });
+
+            it('and requested resolution is 180', () => {
```
saghul, line 405: Following from the above, I think a couple of tests with values not in our range would be useful.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -284,7 +285,7 @@ export default function TraceablePeerConnection(
 
     this.interop = new Interop();
 
-    this.simulcast = new SdpSimulcast({ numOfLayers: SIM_LAYER_RIDS.length });
+    this.simulcast = new SdpSimulcast({ numOfLayers: SIM_LAYERS.length });
```
saghul, line 288: Should we drop the parameter and just get the constant internally in `SdpSimulcast`? It feels like they are very coupled.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2067,14 +2053,22 @@ TraceablePeerConnection.prototype._setMaxBitrates = function(description, isLoca
...
+            .find(track => this._localTrackTransceiverMids.get(track.rtcId) === mLine.mid.toString());
+
+        if ((isDoingVp9KSvc || this.tpcUtils._isRunningInFullSvcMode(currentCodec)) && localTrack) {
+            const qualityLevel = VIDEO_QUALITY_LEVELS.find(level => level.height === localTrack.getHeight());
+            let maxBitrate;
+
+            if (localTrack.getVideoType() === VideoType.DESKTOP) {
+                maxBitrate = codecScalabilityModeSettings.maxBitratesVideo.ssHigh;
+            } else {
+                maxBitrate = codecScalabilityModeSettings.maxBitratesVideo[qualityLevel.level];
```
saghul, line 2066: Can be undefined.

modules/connectivity/ConnectionQuality.js
```diff
@@ -61,32 +34,32 @@ let startBitrate = 800;
...
+ * @param bitrates {Object} the bitrates for the local video source.
  */
-function getTarget(simulcast, resolution, millisSinceStart, videoQualitySettings) {
+function getTarget(simulcast, resolution, millisSinceStart, bitrates) {
     let target = 0;
     let height = Math.min(resolution.height, resolution.width);
 
     // Find the first format with height no bigger than ours.
-    let simulcastFormat = kSimulcastFormats.find(f => f.height <= height);
+    let qualityLevel = VIDEO_QUALITY_LEVELS.find(f => f.height <= height);
```
saghul, line 44: The only case which would lead to qualityLevel being undefined here would be something < 180. Does that make any sense? Even then, wouldn't pretending it's 180 be ok?

service/RTC/StandardVideoSettings.ts
```diff
@@ -38,10 +60,72 @@ export const STANDARD_CODEC_SETTINGS = {
...
+    {
+        height: 720,
+        level: 'high'
+    },
+    {
+        height: 540,
+        level: 'standard'
+    },
+    {
+        height: 360,
```
saghul, line 100: PL add 480 here. Many phones will use that actually when 360 is requested.

saghul (changes requested):
Left some comments, PTAL. I think we need some better handling of finding the right quality level.

modules/RTC/TPCUtils.js
jallamsetty1: It wouldn't be catastrophic if maxBitrate is undefined, the browser would limit the bitrate to its defaults but you are right, rounding might be better especially for mobile. Will add 480 as well. Can you think of other resolutions that are applicable to mobile?

modules/RTC/TPCUtils.js
saghul: 🙏  I think it will help those not working with the code as closely as you do.

modules/RTC/TPCUtils.js
saghul: I don't think I have seen any other weird ones. I'd add some log if we round to the nearest, after trying an exact match. This way we'll eventually see it while examining logs.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Fixed

modules/connectivity/ConnectionQuality.js
jallamsetty1: Techincally, the resolution won't be less than 180 since its coming from here https://github.com/jitsi/lib-jitsi-meet/blob/master/service/RTC/Resolutions.js#L1. 
Even otherwise, I have added `VIDEO_QUALITY_LEVEL.NONE` with a bitrate of 0 to handle cases elsewhere.

saghul: 👍

saghul (approved):
Outstanding work, can't wait to roll this out! 🚀

PR #2442, opened 2024-01-31
Add synchronization around local track access to JitsiConference

paweldomas:
See commit msg for details

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -37,6 +37,7 @@ import AvgRTPStatsReporter from './modules/statistics/AvgRTPStatsReporter';
 import LocalStatsCollector from './modules/statistics/LocalStatsCollector';
 import SpeakerStatsCollector from './modules/statistics/SpeakerStatsCollector';
 import Statistics from './modules/statistics/statistics';
+import AsyncQueue2 from './modules/util/AsyncQueue2';
 import EventEmitter from './modules/util/EventEmitter';
 import { safeSubtract } from './modules/util/MathUtil';
 import RandomUtil from './modules/util/RandomUtil';
@@ -153,6 +154,14 @@ export default function JitsiConference(options) {
     this.eventEmitter = new EventEmitter();
     this.options = options;
     this.eventManager = new JitsiConferenceEventManager(this);
+
+    /**
+     * The queue is used to synchronize access to local tracks (this.rtc.localTracks). This prevents from unexpected
+     * behavior when a new track operation is scheduled, before the previous one has finished.
+     *
+     * @type {AsyncQueue2}
+     */
+    this.modificationQueue = new AsyncQueue2();
 
     /**
      * List of all the participants in the conference.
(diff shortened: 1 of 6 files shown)


JitsiConference.js
```diff
@@ -1071,60 +1083,52 @@ JitsiConference.prototype.getTranscriptionStatus = function() {
...
-        .then(() => {
-            // Presence needs to be sent here for desktop track since we need the presence to reach the remote peer
-            // before signaling so that a fake participant tile is created for screenshare. Otherwise, presence will
-            // only be sent after a session-accept or source-add is ack'ed.
-            if (track.getVideoType() === VideoType.DESKTOP) {
+        track.setSourceName(sourceName);
+        const addTrackPromises = [];
+
+        this.p2pJingleSession && addTrackPromises.push(this.p2pJingleSession.addTracks([ track ]));
+        this.jvbJingleSession && addTrackPromises.push(this.jvbJingleSession.addTracks([ track ]));
```
jallamsetty1, line 1119: As discussed, this is going to be problematic. In both jvb and p2p case, the first track needs to be added to be replaced on the `RTCRtpSender` associated with the first `recvonly` transceiver. When a secondary track is added, for jvb, a new transceiver is created and for p2p, a new transceiver is created only when there is no `recvonly` one. JingleSessionPC::addTracks needs to be modified to take this into account. Right now, it is designed to be used for secondary tracks only and therefore creates a new transceiver for jvb always.

paweldomas: By primary and secondary do you mean for each media type, for example:
- primary track will be the 1st video track added to the connection
- secondary track will be the 2nd video track added to the connection?

If there are two tracks, one audio and one video then there is one primary video track and one primary audio track?

hristoterezov:
@paweldomas Wouldn't all this be handled by the modification queue even before? What will actually cause the unexpected results?

paweldomas:
@hristoterezov the issue is that JitsiConference's track operations(add/replace/remove) as well as the JVB/P2P JingleSession initialization (accept,create offer) rely on `this.rtc.localTracks` state. You can see that variables get updated only once the operations scheduled on JingleSessionPC resolve. So if you perform either a track operation, or try to initialize new JingleSessionPC, without explicitly waiting for previous track operations to be done, then out of sync state will be taken into consideration.

Note that currently only application that uses lib-jitsi-meet is able to wait for other track operation to resolve, which is the main design flaw. JitsiConference by design can't do it's job correctly. To give an example:

1. JitsiConference stats and JVB JingleSessionPC is created. It accept the offer an so on. There are no local tracks.
2. User calls `addTrack(camera1);`, this queues addTracks in JingleSessionPC and this operation is async and takes time. Let's say 200ms.
3. 100ms later, another user joined and started a P2P connection. JitsiConference will not have camera1 in this.rtc.localTracks yet and the P2P session will start without including it in the local tracks of JingleSessionPC.
4. Another 100ms later, JVB JingleSessionPC is done with addTrack and now we end up in a state where JVB JingleSessionPC has camera1, but P2P JingleSessionPC does not.

The idea in this PR for a fix is to use the modification queue to schedule P2P initialization to be executed after the first addTrack. So that it will have `camera1` in `this.rtc.localTracks` when creating/answering the offer. It makes JitsiConference aware of all currently scheduled track operations.

PR #2446, opened 2024-02-05
feat: Drop conference duration component and use the data form.

damencho:


jallamsetty1 (approved):

damencho:
Yes. You normally update the client and the backend at the same time. This is how it is supposed to happen.

Romick2005:
Thank you for the explanation!

PR #2463, opened 2024-02-22
fix(JitsiTrack): Detect track ended event on Safari.

jallamsetty1:
Fixes https://github.com/jitsi/jitsi-meet/issues/14386


damencho (approved):

jallamsetty1:
The test failure is not related.

PR #2465, opened 2024-02-23
Avoid finally in promise chain. Log error if error happen in _l_lastVideoSenderUpdatePromise.

### LINKED ISSUE #2464: _lastVideoSenderUpdatePromise.finally is not a function
I got this error from react-native:
ERROR  [TypeError: this._lastVideoSenderUpdatePromise.finally is not a function (it is undefined)]

image

Romick2005:
Fixes #2464 _lastVideoSenderUpdatePromise.finally is not a function issue. Add more information about occurred errors.

### DIFF
--- a/modules/RTC/TraceablePeerConnection.js
+++ b/modules/RTC/TraceablePeerConnection.js
@@ -2261,7 +2261,7 @@ TraceablePeerConnection.prototype.setSenderVideoConstraints = function(frameHeig
 TraceablePeerConnection.prototype._updateVideoSenderParameters = function(nextFunction) {
     const nextPromise = this._lastVideoSenderUpdatePromise.catch(
       error => {
-          logger.error('[TraceablePeerConnection] Failed to proceed with last video sender update promise', error);
+          logger.error('Failed to proceed with last video sender update promise', error);
       }
     ).then(nextFunction);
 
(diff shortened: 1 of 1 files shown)


saghul:
What environment are you running in? Is Promise.prototype.finally not defined there?

Romick2005:
> What environment are you running in? Is Promise.prototype.finally not defined there?

react-native

saghul:
Then you need to apply the promise finally polyfill.

Romick2005:
> Then you need to apply the promise finally polyfill.

But this pr allow us to see errors also. And this is the only place where we can have more environments covered. Wouldn't it be better? Also this will cover some case with UnhandledPromiseRejection. So for example:
```
const p = Promise.reject(new Error('BOOM'));
p.catch(() => {});
p.finally(() => {});
```
will throw UnhandledPromiseRejectionWarning
`(node:22503) UnhandledPromiseRejectionWarning: Error: BOOM at Object.<anonymous> `

But this:
```
const p = Promise.reject(new Error('BOOM'));
p.catch(() => {}).finally(() => {});
```

So it is not also about polyfils, but for general code reliability.
@saghul What do you think?

saghul:
Error in that function can happen and are ok to ignore, hence the finally.

While what you say is generally good advice, it doesn't need to be followed in every case.

Romick2005:
Is this a good case to kill two rabbits with one shot? Allow easy react-native integration and have proper error handling and logs?

I do not agree with you. So I will patch it, sorry.

PR #2466, opened 2024-02-23
feat(JitsiConference): Separate conference name validation errors.

Romick2005:
Separate thrown errors:  "no conference name passed" and "it contains invalid characters like capital letters".

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -131,21 +131,26 @@ function _getCodecMimeType(codec) {
  *       and so on...
  */
 export default function JitsiConference(options) {
-    if (!options.name || options.name.toLowerCase() !== options.name.toString()) {
-        const errmsg
-            = 'Invalid conference name (no conference name passed or it '
-                + 'contains invalid characters like capital letters)!';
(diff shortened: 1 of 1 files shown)


JitsiConference.js
```diff
@@ -476,7 +481,7 @@ JitsiConference.prototype._init = function(options = {}) {
             userName: config.statisticsDisplayName ? config.statisticsDisplayName : this.myUserId(),
             confID: config.confID || `${this.connection.options.hosts.domain}/${this.options.name}`,
             siteID: config.siteID,
-            roomName: this.options.name,
```
saghul, line 479: Why are you changing this?

saghul:
Since the errors are indistinguishable (no code or other attribute) why is this change useful?

JitsiConference.js
Romick2005: My bad that should be roomName. Initially I used roomName = options.name, but then I notice that it is mention as conference name in some places and room name on others so renamed all occurrence and this code was corrupted (

Romick2005:
So you can close it if it is not important.

PR #2468, opened 2024-02-27
fix(tracks): Do not filter the tracks for p2p.

jallamsetty1:
The tracks should get removed if the startAudioMuted/startVideoMuted flags are set for the conference. Fixes the failing startMuted testcase.


paweldomas:
What are the symptoms of the issue?

jallamsetty1:
> What are the symptoms of the issue?

The symptoms are similar to the ones you were trying to fix. The audio/video tracks don't get added to the p2p connection sometimes.

jallamsetty1:
Jenkins test this please.

damencho (approved):

PR #2483, opened 2024-03-18
feat(build) remove export from build script

saghul:
this moves the determination of the git commit to webpack-shared-config LIB_JITSI_MEET_COMMIT_HASH is still kept for backward compatibility

fix(build) in case git fails use development


saghul:
Rebase of https://github.com/jitsi/lib-jitsi-meet/pull/1956

damencho (approved):

Issue #2489, opened 2024-03-26
TypeError: _this.isCollectingLocalStats is not a function (it is undefined)


Romick2005:
After `jitsiConnectionEstablished` I tried to call  ```JitsiMeetJS.createLocalTracks({ devices: ['audio'] });```
But got an error: `"error": [TypeError: _this.isCollectingLocalStats is not a function (it is undefined)]`

I am using latest react-native@0.73.6

damencho:
Can you reproduce the same error on alpha.jitsi.net?

Romick2005:
Using raw git code pointing main to the index.js. Directly.

saghul:
Did you initialize the library by calling `JitsiMeetJS.init` ?

Romick2005:
> Did you initialize the library by calling `JitsiMeetJS.init` ?

Yes, sir. Only one time on project initial setup.
lib-jitsi-meet version is `git reset --hard 006457f`.

```
JitsiMeetJS.init({
  disableAudioLevels: false,
  useIPv6: false,
  disableThirdPartyRequests: true,
});
```

Romick2005:
> Can you reproduce the same error on alpha.jitsi.net?

Sorry, but I am not sure how to check it there as I use react-native mobile app together with latest lib-jitsi-meet code.

damencho:
> > Can you reproduce the same error on alpha.jitsi.net?
> 
> Sorry, but I am not sure how to check it there as I using react-native mobile app together with latest lib-jitsi-meet code.

Yeah, I was wondering how is possible jitsi-meet web and mobile already using this and works without a problem, that's why I was asking about some test and I did not see the react-native line.

Romick2005:
More info.
It is createLocalTracks from JitsiMeetJS.ts with extra console.log:
![image](https://github.com/jitsi/lib-jitsi-meet/assets/17779660/5943fbe5-259b-4a93-9fe7-d14eb8fffebc)

This is my code and how it is used:
![image](https://github.com/jitsi/lib-jitsi-meet/assets/17779660/8ac2ddfb-975c-4a40-ab41-a347a099e7a3)

And this is output from console:
![image](https://github.com/jitsi/lib-jitsi-meet/assets/17779660/ed5bee54-b15a-4b1f-962e-fd4ef0dcb947)

Romick2005:
PS Could it be related with generator creating some different this for an object call? Because if I use it like regular promise it is working )
Working code sample:

[...]

PR #2490, opened 2024-03-26
Fix for TypeError: _this.isCollectingLocalStats is not a function (it is undefined)

Romick2005:
#2489 Move isCollectingLocalStats to the statistics. Mark JitsiMeetJS.isCollectingLocalStats as deprecated with warning to use Statistics.isCollectingLocalStats instead.

### DIFF
--- a/JitsiMeetJS.ts
+++ b/JitsiMeetJS.ts
@@ -24,7 +24,6 @@ import ProxyConnectionService
     from './modules/proxyconnection/ProxyConnectionService';
 import recordingConstants from './modules/recording/recordingConstants';
 import Settings from './modules/settings/Settings';
-import LocalStatsCollector from './modules/statistics/LocalStatsCollector';
 import Statistics from './modules/statistics/statistics';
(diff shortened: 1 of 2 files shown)


saghul:
Why is this change necessary?

Romick2005:
> Why is this change necessary?

In order to fix #2489. Also it helps to organize code in more obvious way as now method that depends on statistics is located directly in Statistics file. Don't you think it will look more as cleaner code and as a bonus without an error?

Romick2005:
> No, because on UMD builds Statistics is not exposed, JitsiMeetJS is the only exposed object.
> 
> In addition, current working code does not have this problem so I suspect is has to do with the environment perhaps.

Then we can just remove deprecated warning. And isCollectingLocalStats method would be exposed as before via JitsiMeetJS.isCollectingLocalStats.

Should I remove deprecation warning?

Romick2005:
Done sir. Please review and merge please.

PR #2492, opened 2024-03-26
fix(flags) don't enable SSRC rewriting by default

saghul:
We are going to work on making the tracks API backwards compatible and
re-enable it in the near future.


jallamsetty1:
Jenkins test this please.

jallamsetty1 (approved):

PR #2493, opened 2024-03-26
feat: Introduces connection event for custom properties.

damencho:
Used for shard and region values coming from the backend.

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -1110,5 +1110,16 @@ export default class XMPP extends Listenable {
         }
 
         this.sendDeploymentInfo = false;
+
+        const { region, shard } = aprops;
+
+        if (region || shard) {
+            // avoids sending empty values
+            this.eventEmitter.emit(JitsiConnectionEvents.PROPERTIES_UPDATED, JSON.parse(JSON.stringify({
+                region,
+                shard
+            })));
+        }
+
     }
 }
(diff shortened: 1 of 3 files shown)


jallamsetty1 (approved):

modules/xmpp/xmpp.js
```diff
@@ -1110,5 +1110,16 @@ export default class XMPP extends Listenable {
         }
 
         this.sendDeploymentInfo = false;
+
+        const { region, shard } = aprops;
+
+        if (region || shard) {
+            // avoids sending empty values
+            this.eventEmitter.emit(JitsiConnectionEvents.PROPERTIES_UPDATED, JSON.parse(JSON.stringify({
```
saghul, line 1118: Suggestion: use lodash clone or structuredClone.

saghul (approved):

modules/xmpp/xmpp.js
damencho: That will drop the keys with undefined values?

saghul: No. Do you want to?

damencho: Yes. That was a quick hack to avoid if this and if that to not add the keys.

saghul: Ah then go ahead!

PR #2499, opened 2024-04-05
fix(rtc) use exact deviceId constraints

saghul:
webrtc-adapter transforms "deviceId": "foo" to "deviceId": { ideal: "foo"}.

This makes the camera non-selectable if the resolution constraints are not met.

Setting the constraint to exact will sidestep that adapter behavior, and express stronger intent.


saghul:
Caveat emptor: I'm not 100% certain this won't introduce problems with all browsers, so I suggest a "merge, test in beta, revert if problematic" approach.

PR #2502, opened 2024-04-12
feat(jaas) add zero-config mode for JaaS customers

saghul:


saghul:
@jallamsetty1 @damencho I had the following idea today.

Ideally I'd like to fullfil the dream of literraly zero config to get a meeting going, which seems to be within reach for JaaS customers.

The only thing I haven't figured out is how to add the `room` query parameter to the connection URLs, since we don't know it at that moment. Ideas?

damencho:
Not really. For the http requests to jicofo I needed it so I added it to the api, to be able to set it when creating connection.

damencho:
Or maybe it was when calling connect

damencho:
https://github.com/jitsi/lib-jitsi-meet/blob/9b11d097e57d76eb8acc279a7c6320f9c9fdc543/JitsiConnection.js#L59

Maybe it is worth moving the logic for these params from jitsi-meet into ljm. 

https://github.com/jitsi/jitsi-meet/blob/d1590db82bf3fd1be8d4179db3df6c7ff7270048/react/features/base/connection/actions.any.ts#L151

saghul:
> Maybe it is worth moving the logic for these params from jitsi-meet into ljm.

Pretty much yeah.

Another option would be some `JitsiMeetJS.joinConference(options)` convenience function which creates a connection and then joins a conference. This way we could get everything we need there and piecemeal it to either the connection or the conference.

jallamsetty1:
> > Maybe it is worth moving the logic for these params from jitsi-meet into ljm.
> 
> Pretty much yeah.
> 
> Another option would be some `JitsiMeetJS.joinConference(options)` convenience function which creates a connection and then joins a conference. This way we could get everything we need there and piecemeal it to either the connection or the conference.

I like this approach better. A single call to JitsiMeetJS.joinConference(options) should create the connection and the conference and join the room automatically for those consumers of the library who are looking for a high-level integration. 

I have few other ideas around device selection and such so that the app doesn't have to create/add/replace tracks at the conference level. That should all be handled by the library itself.

PR #2509, opened 2024-04-18
ref(ssrc-rewriting): Fire track removed/added instead of owner changed.

jallamsetty1:
When an existing SSRC for a remote track gets re-mapped from one source to another, fire a TRACK_REMOVED event followed by TRACK_ADDED event instead of TRACK_OWNER_CHANGED event. This should simplify the application logic for track handling when SSRC rewriting is enabled.

### DIFF
--- a/JitsiTrackEvents.ts
+++ b/JitsiTrackEvents.ts
@@ -48,12 +48,7 @@ export enum JitsiTrackEvents {
     /**
      * Indicates that a new owner has been assigned to a remote track when SSRC rewriting is enabled.
      */
-    TRACK_OWNER_ADDED = 'track.owner_added',
-
-    /**
-     * Indicates that an owner has been removed from a remote track when SSRC rewriting is enabled.
-     */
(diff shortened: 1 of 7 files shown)


JitsiTrackEvents.ts
```diff
@@ -45,6 +45,16 @@ export enum JitsiTrackEvents {
      */
     NO_AUDIO_INPUT = 'track.no_audio_input',
 
+    /**
+     * Indicates that a new owner has been assigned to a remote track when SSRC rewriting is enabled.
+     */
+    TRACK_OWNER_ADDED = 'track.owner_added',
```
saghul, line 51: "added" and "removed" read weird here. How about a single `TRACK_OWNER_SET` ? We'd only need 1 then.

saghul (commented):
LGTM with a suggestion.

saghul (approved):

jallamsetty1:
Jenkins test this please.

PR #2518, opened 2024-04-26
Expose `track.conference` field in types

jesse-gibson-dd:
There's a field `jitsiTrack.conference` that keeps a reference to the track's owning conference. It does not have a corresponding TypeScript definition. This PR adds a type definition.

- `JitsiLocalTrack.conference`: `null` if it hasn't been added to a conference yet
- `JitsiRemoteTrack.conference`: non-nullable, always associated with a conference

(We're using this in our application.)


jesse-gibson-dd:
We should already have a CCLA signed for CoScreen/Datadog. Let me know if there's anything missing on our end.

saghul (approved):

PR #2520, opened 2024-05-01
feat(codec-selection): Use the new codec selection API

jallamsetty1:
https://github.com/w3ctag/design-reviews/issues/836. This allows the client to seamlessly switch between the codecs without having to trigger a renegotiation. This feature is behind the flag testing.enableCodecSelectionAPI in config.js

This is marked WIP since we need a JVB change for this feature to work as expected for all codecs. When the DD header extension headers are negotiated, Chrome includes the layer information for VP8 in the DD extension headers but the bridge currently reads this from the VP8 packet headers. Since renegotiation cycle is not triggered when this feature is enabled, client doesn't have a way of removing the DDs from being sent by Chrome when we switch to VP8. - this has been fixed in https://github.com/jitsi/jitsi-videobridge/pull/2136

### DIFF
--- a/JitsiConferenceEventManager.js
+++ b/JitsiConferenceEventManager.js
@@ -734,6 +734,11 @@ JitsiConferenceEventManager.prototype.setupStatisticsListeners = function() {
             JitsiConferenceEvents.BEFORE_STATISTICS_DISPOSED);
     });
 
+    conference.statistics.addEncodeTimeStatsListener((tpc, stats) => {
+        conference.eventEmitter.emit(
+            JitsiConferenceEvents.ENCODE_TIME_STATS_RECEIVED, tpc, stats);
+    });
+
     // if we are in startSilent mode we will not be sending/receiving so nothing to detect
     if (!conference.options.config.startSilent) {
         conference.statistics.addByteSentStatsListener((tpc, stats) => {
--- a/modules/RTC/CodecSelection.js
+++ b/modules/RTC/CodecSelection.js
@@ -30,15 +30,16 @@ export class CodecSelection {
      * @param {string} options.p2p settings (codec list, preferred and disabled) for the p2p connection.
      */
     constructor(conference, options) {
-        this.conference = conference;
-        this.options = options;
         this.codecPreferenceOrder = {};
-        this.visitorCodecs = [];
+        this.conference = conference;
         this.encodeTimeStats = new Map();
+        this.options = options;
+        this.screenshareCodec = {};
+        this.visitorCodecs = [];
 
         for (const connectionType of Object.keys(options)) {
             // eslint-disable-next-line prefer-const
-            let { disabledCodec, preferredCodec, preferenceOrder } = options[connectionType];
+            let { disabledCodec, preferredCodec, preferenceOrder, screenshareCodec } = options[connectionType];
             const supportedCodecs = new Set(this._getSupportedVideoCodecs(connectionType));
 
             // Default preference codec order when no codec preferences are set in config.js
@@ -90,6 +91,11 @@ export class CodecSelection {
 
             logger.info(`Codec preference order for ${connectionType} connection is ${selectedOrder}`);
             this.codecPreferenceOrder[connectionType] = selectedOrder;
+
+            // Set the preferred screenshare codec.
+            if (screenshareCodec && supportedCodecs.has(screenshareCodec)) {
+                this.screenshareCodec[connectionType] = screenshareCodec;
+            }
         }
 
         this.conference.on(
@@ -220,9 +226,7 @@ export class CodecSelection {
         if (!session) {
             return;
         }
-        const currentCodecOrder = session.peerconnection.getConfiguredVideoCodecs();
         const isJvbSession = session === this.conference.jvbJingleSession;
-
         let localPreferredCodecOrder = isJvbSession ? this.codecPreferenceOrder.jvb : this.codecPreferenceOrder.p2p;
 
         // E2EE is curently supported only for VP8 codec.
@@ -276,10 +280,7 @@ export class CodecSelection {
             return;
         }
 
-        // Reconfigure the codecs on the media session.
-        if (!selectedCodecOrder.every((val, index) => val === currentCodecOrder[index])) {
-            session.setVideoCodecs(selectedCodecOrder);
-        }
+        session.setVideoCodecs(selectedCodecOrder);
     }
 
     /**
@@ -305,5 +306,15 @@ export class CodecSelection {
      */
     getCodecPreferenceList(connectionType) {
         return this.codecPreferenceOrder[connectionType];
+    }
+
+    /**
+     * Returns the preferred screenshare codec for the given connection type.
+     *
+     * @param {String} connectionType The media connection type, 'p2p' or 'jvb'.
+     * @returns CodecMimeType
+     */
+    getScreenshareCodec(connectionType) {
+        return this.screenshareCodec[connectionType];
     }
 }
--- a/modules/RTC/CodecSelection.js
+++ b/modules/RTC/CodecSelection.js
@@ -138,8 +138,7 @@ export class CodecSelection {
     }
 
     /**
-     * Processes the encode time stats received for all the local video sources and switches the codec to a low
-     * complexity codec if needed.
+     * Processes the encode time stats received for all the local video sources.
      *
      * @param {TraceablePeerConnection} tpc - the peerconnection for which stats were gathered.
      * @param {Object} stats - the encode time stats for local video sources.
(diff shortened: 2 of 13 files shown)


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1601,6 +1601,18 @@ TraceablePeerConnection.prototype._assertTrackBelongs = function(
  * video in the local SDP.
  */
 TraceablePeerConnection.prototype.getConfiguredVideoCodec = function() {
+    const localVideoTrack = this.getLocalVideoTracks()[0];
+
+    if (this._usesCodecSelectionAPI && localVideoTrack) {
+        const rtpSender = this.findSenderForTrack(localVideoTrack.getTrack());
+
+        if (rtpSender) {
+            const { codecs } = rtpSender.getParameters();
```
saghul, line 1619: This API exists before the new codec selection one. Shouldn't we be using it without the guard?

modules/RTC/CodecSelection.js
```diff
@@ -127,6 +131,83 @@ export class CodecSelection {
         return supportedCodecs;
     }
 
+    /**
+     * Processes the encode time stats received for all the local video sources and switches the codec to a low
```
saghul, line 141: We don't do any codec change here yet AFAICT, pl update the comment.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -813,18 +813,27 @@ TraceablePeerConnection.prototype.getTargetVideoBitrates = function() {
 };
 
 /**
- * Tries to find {@link JitsiTrack} for given SSRC number. It will search both
- * local and remote tracks bound to this instance.
+ * Tries to find {@link JitsiTrack} for given SSRC number. It will search both local and remote tracks bound to this
+ * instance.
  * @param {number} ssrc
+ * @param {isPrimary} boolean - Whether the SSRC passed is a primary SSRC of a source.
```
saghul, line 819: It's not clear to me why we need the primary flag here. Wouldn't check the check always succeed if we always go throught he else branch?

modules/statistics/RTPStatsCollector.js
```diff
@@ -638,13 +643,36 @@ StatsCollector.prototype.processStatsReport = function() {
...
+                    && this.peerconnection.usesCodecSelectionAPI()
+                    && before?.totalEncodeTime
+                    && before?.framesEncoded) {
+                    const encodeTimeDelta = now.totalEncodeTime - before.totalEncodeTime;
+                    const framesEncodedDelta = now.framesEncoded - before.framesEncoded;
+                    const encodeTimePerFrameInMs = 1000 * encodeTimeDelta / framesEncodedDelta;
+                    const encodeTimeStats = {
+                        codec: codecShortType,
+                        encodeTime: encodeTimePerFrameInMs,
+                        isBandwidthLimited: now.qualityLimitationReason === 'bandwidth',
```
saghul, line 670: Why not surface the reason raw? Why do we not care anout the `cpu` reason, for example?

JitsiConferenceEventManager.js
```diff
@@ -739,6 +739,11 @@ JitsiConferenceEventManager.prototype.setupStatisticsListeners = function() {
             JitsiConferenceEvents.BEFORE_STATISTICS_DISPOSED);
     });
 
+    conference.statistics.addEncodeTimeStatsListener((tpc, stats) => {
```
saghul, line 742: Commenting here, semi-randomly: looks like this commit could be extracted from this PR to make it smaller, it can be applied to the current code, can't it?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1899,7 +1899,10 @@ TraceablePeerConnection.prototype.removeTrackFromPc = function(localTrack) {
  * @returns {boolean}
  */
 TraceablePeerConnection.prototype.usesCodecSelectionAPI = function() {
-    return this._usesCodecSelectionAPI;
+    // Browser throws an error when H.264 is set on the encodings. Therefore, munge the SDP when H.264 needs to be
```
saghul, line 1902: Is this reported somewhere? A link to a bug we can follow would be nice.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1184,7 +1185,10 @@ export default class JingleSessionPC extends JingleSession {
             logger.info(`${this} setVideoCodecs: ${codecList}`);
             this.peerconnection.setVideoCodecs(codecList);
 
-            if (this.usesCodecSelectionAPI) {
+            // Browser throws an error when H.264 is set on the encodings. Therefore, munge the SDP when H.264 needs to
```
saghul, line 1188: Pl add a comment linking this and the other (same) check.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1192,6 +1192,13 @@ export default class JingleSessionPC extends JingleSession {
                 return;
             }
 
+            // Skip renegotiation when the selected codec order matches with that of the remote SDP.
+            const currentCodecOrder = this.peerconnection.getConfiguredVideoCodecs();
+
+            if (codecList.every((val, index) => val === currentCodecOrder[index])) {
```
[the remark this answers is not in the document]
jallamsetty1: Nope, cannot use deepEqual here since `codecList` can be a subset of `currentCodecOrder`. All we need to check is if all the entries from `codecList` exist in `currentCodecOrder` in the same order. There can be extra entries in `currentCodecOrder` but that doesn't necessitate a renegotiation.

modules/RTC/CodecSelection.js
jallamsetty1: Ah right, I didn't add the switching logic in this PR since it was getting long, forgot to update the comment.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: `findSenderForTrack` still checks if `getSenders` on the peerconnection exists and only then returns the `RTCRtpSender` . Now that getSenders is implemented everywhere, I can go and update `findSenderForTrack` and remove the guard from everywhere else as well.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Before this change, this function was returning a track only if the passed SSRC was the primary SSRC and therefore stats were being processed only for the primary SSRC. If we modify this function to return the track for all the SSRCs in all the cases, this would result in extra processing. Let me check how bad it is going to be if we return the track for all SSRCs.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1174,14 +1176,18 @@ export default class JingleSessionPC extends JingleSession {
...
-     * @param {CodecMimeType} disabled the codec that needs to be disabled.
+     * @param {Array<CodecMimeType>} codecList the preferred codecs.
      */
     setVideoCodecs(codecList) {
+
         if (this._assertNotEnded()) {
             logger.info(`${this} setVideoCodecs: ${codecList}`);
             this.peerconnection.setVideoCodecs(codecList);
 
+            if (this.usesCodecSelectionAPI) {
```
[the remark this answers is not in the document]
jallamsetty1: It is set when the jingleSession is initialized - https://github.com/jitsi/lib-jitsi-meet/pull/2520/files#diff-2ab65ee8ec5b7568335948135d9d2a5b7010c054a1d821f9774a7b542689b282R403

JitsiConferenceEventManager.js
jallamsetty1: It can be applied to the current code but the client can act on them only if the codec selection API is used. It didn't make sense to capture/process the encodeTime stats if we are not acting on them.

modules/statistics/RTPStatsCollector.js
jallamsetty1: We could and I actually modified this to surface `qualityLimitationReason` in the patch that I am working on now. Will update it here :)

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1956,6 +1956,33 @@ TraceablePeerConnection.prototype._adjustRemoteMediaDirection = function(remoteD
...
 };
 
+/**
+ * Returns the codec to be used for screenshare based on the supported codecs and the preferred codec requested
+ * through config.js setting.
+ *
+ * @param {CodecMimeType} cameraCodec - the preferred codec to be used for the camera track.
+ * @returns {CodecMimeType}
+ */
+TraceablePeerConnection.prototype._getPreferredCodecForScreenshare = function(cameraCodec) {
```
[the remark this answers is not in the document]
jallamsetty1: Yes, it makes sense. I guess calling the param `cameraCodec` was a poor choice.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: I haven't reported it yet, I am not able to reproduce this using a fiddle. I will open a bug report and update it here then.

modules/RTC/CodecSelection.js
```diff
@@ -127,6 +137,83 @@ export class CodecSelection {
...
+            const track = tpc.getTrackBySSRC(ssrc);
+            let existingStats = statsPerTrack.get(track.rtcId);
+            const encodeResolution = Math.min(resolution.height, resolution.width);
+            const ssrcStats = {
+                encodeResolution,
+                encodeTime,
+                qualityLimitationReason
+            };
+
+            if (existingStats) {
```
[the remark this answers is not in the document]
jallamsetty1: Good catch, I missed it here but fixed it in the patch that I am working on. Will update it here.

modules/statistics/RTPStatsCollector.js
```diff
@@ -84,6 +84,10 @@ SsrcStats.prototype.setCodec = function(codec) {
     this.codec = codec || '';
 };
 
+SsrcStats.prototype.setEncodeStats = function(encodeStats) {
+    this.encodeStats = encodeStats || {};
```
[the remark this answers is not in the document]
jallamsetty1: Nope, they are not reset elsewhere. We'd be using this to show the codec for screenshare in the stats (currently not implemented but on my list) since we can have separate codecs for camera and screenshare.

saghul (approved):

PR #2522, opened 2024-05-07
feat(createLocalTracks) drop unused slow gUM event

saghul:
It's better handled by the application layer anyway.


jallamsetty1 (approved):
LGTM!

PR #2531, opened 2024-05-21
feat(ice-restart): Impl an exponential backoff timer for ICE restarts.

jallamsetty1:


JitsiConference.js
```diff
@@ -2854,8 +2856,16 @@ JitsiConference.prototype._onIceConnectionFailed = function(session) {
             reasonDescription: 'ICE FAILED'
         });
     } else if (session && this.jvbJingleSession === session) {
+        // Implement an exponential backoff timer for ICE restarts.
+        const delay = Math.min(1000 * Math.pow(2, this._iceRestarts), 60000);
+        const jitter = Math.random() * 100; // Adding jitter to avoid synchronized retries
+        const totalDelay = delay + jitter;
```
damencho, line 2862: In other places we use getJitterDelay, can we use it here?

jallamsetty1: Sure, didn't know we had such a thing. Thanks for pointing it out.

damencho (approved):

PR #2532, opened 2024-05-27
Fix URL.search is not implemented for react native.

Romick2005:
Investigating a code it seems like we do not heavily use any URL instance. In reality all that code need to do is just to work with url search string and modify it if necessary. 
Unfortunately react native does not support URL.search, but the lib-jitsi-meet library code use it only in one place by creating new instance of URL class, so I guess it is better to not convert string into url and then get url.search, but rather work at lower code level with strings. This would help library to have wider support (no need for URL.search polyfill), avoid URL dependency, have clear logic how restoring url is build on.

This is microimprovement, but I hope this will help the lib and have better RN support.

image


saghul:
There are many things RN doesn't provide out of the box, the most important one being WebRTC itself. In order to keep the library small, we chose to not bnake those in here but to rather rely on the host application providing an appropriate environment.

Here is a sample of everything we are currently polyfilling: https://github.com/jitsi/jitsi-meet/tree/master/react/features/mobile/polyfills

Since there is no change we are bringing all of that in, I don't think this PR improves the status quo in a meningful enough way.

Hope you understand.

PR #2537, opened 2024-06-12
Migrate some simple pre-ES6 classes to TS

saghul:
- **ref(EventEmitterForwarder) rewrite in TS as an ES6 class**
- **ref(ComponentsVersions) rewrite in TS as a ES6 class**


jallamsetty1 (approved):
🚀

PR #2539, opened 2024-06-21
fix(video-quality) Reset scale factor and bitrates for VP9 K-SVC.

jallamsetty1:
Fixes low resolution video from Spot running older electron versions when the codec changes from VP9->VP8->VP9.


saghul (approved):
👏

PR #2542, opened 2024-06-27
feat(quality) Add a QualityController class for runtime adjustments.

jallamsetty1:
Make run time adjustments to the client when adaptive mode is enabled.


JitsiConference.js
```diff
@@ -475,8 +448,35 @@ JitsiConference.prototype._init = function(options = {}) {
...
+            preferredCodec: _getCodecMimeType(config.p2p?.preferredCodec),
+            screenshareCodec: browser.isMobileDevice()
+                ? _getCodecMimeType(config.p2p?.mobileScreenshareCodec)
+                : _getCodecMimeType(config.p2p?.screenshareCodec)
+        }
+    };
+
+    this.qualityController = new QualityController(
+        this,
+        this.rtc,
```
saghul, line 477: Why not just pass the conference and extract `rtc` from it if necessary?

modules/qualitycontrol/CodecSelection.js
```diff
@@ -281,7 +204,43 @@ export class CodecSelection {
             return;
         }
 
-        session.setVideoCodecs(selectedCodecOrder);
+        session.setVideoCodecs(selectedCodecOrder, this.screenshareCodec?.jvb);
+    }
+
+    /**
+     * Changes the codec preference order when initiated by the {@link QualityController} to improve video quality.
```
saghul, line 211: Nitpick: this class doesn't know why this function does, that's the responsibility of the quality controller. So I wouldn't mention the quality aspect of it.

modules/qualitycontrol/CodecSelection.js
```diff
@@ -281,7 +204,43 @@ export class CodecSelection {
...
+     *
+     * @param {JitsiLocalTrack} localTrack - The local video track for which stats are being processed.
```
saghul, line 213: Bogus description.

modules/qualitycontrol/CodecSelection.js
```diff
@@ -281,7 +204,43 @@ export class CodecSelection {
...
+    changeCodecPreferenceOrder(localTrack, codec) {
+        const session = this.conference.getActiveMediaSession();
+        const connectionType = session.isP2P ? 'p2p' : 'jvb';
+        const codecOrder = this.codecPreferenceOrder[connectionType];
+        const videoType = localTrack.getVideoType();
+        const codecsByVideoType = VIDEO_CODECS_BY_COMPLEXITY[videoType]
+            .filter(val => Boolean(codecOrder.find(supportedCodec => supportedCodec === val)));
+        const codecIndex = codecsByVideoType.findIndex(val => val === codec.toLowerCase());
+
+        if (codecIndex < codecsByVideoType.length - 1) {
```
saghul, line 226: Note we'll also get here for `-1`. Given the `findIndex` above I guess the only options here are for the index to be within bounds or `-1`. Special care needs to be taken for the last element, of course.

modules/qualitycontrol/CodecSelection.js
```diff
@@ -281,7 +204,43 @@ export class CodecSelection {
...
+            const newCodec = codecsByVideoType[codecIndex + 1];
+
+            if (localTrack.getVideoType() === VideoType.CAMERA) {
```
saghul, line 229:
```suggestion
            if (videoType === VideoType.CAMERA) {
```

modules/qualitycontrol/CodecSelection.js
```diff
@@ -281,7 +204,43 @@ export class CodecSelection {
...
+            } else {
+                this.screenshareCodec[connectionType] = newCodec;
+            }
+
+            this.selectPreferredCodec(session);
+
+            return true;
+        }
+
+        return false;
```
saghul, line 243: I think the code will look nicer to read if you reverse the check and return early, this way you'll reduce one lavel of indentation.

modules/qualitycontrol/CodecSelection.spec.js
```diff
@@ -203,25 +236,94 @@ describe('Codec Selection', () => {
...
+
+                // Expect the local endpoint to continue sending VP9.
+                expect(jingleSession.setVideoCodecs).toHaveBeenCalledWith([ 'vp9', 'av1', 'vp8' ], undefined);
+            });
+        });
+
+        it('and does not change codec if the current codec is already the lowest complexity codec', () => {
+            const localTrack = new MockLocalTrack(720, 'camera');
+
+            qualityController._codecController.codecPreferenceOrder.jvb = [ 'vp8', 'vp9', 'av1' ];
```
saghul, line 305: You can use the public getter, can't you?

modules/qualitycontrol/QualityController.ts
```diff
@@ -0,0 +1,366 @@
...
+import RTC from "../RTC/RTC";
+import TraceablePeerConnection from "../RTC/TraceablePeerConnection";
+import JingleSessionPC from "../xmpp/JingleSessionPC";
+import { CodecSelection } from "./CodecSelection";
+import ReceiveVideoController from "./ReceiveVideoController";
+import SendVideoController from "./SendVideoController";
+import { VIDEO_CODECS_BY_COMPLEXITY, VIDEO_QUALITY_LEVELS } from '../../service/RTC/StandardVideoSettings';
+
+const logger = getLogger(__filename);
+const LIMITED_BY_CPU_TIMEOUT = 60000;
```
saghul, line 17: Can you please document these constants?

modules/qualitycontrol/QualityController.ts
```diff
@@ -0,0 +1,366 @@
...
+    encodeTime: number;
+    qualityLimitationReason: QualityLimitationReason;
+}
+
+interface IVideoConstraints {
+    maxHeight: number;
+    sourceName: string;
+}
+
+class FixedSizeArray {
```
saghul, line 59: Why is this necessary? You can achieve the same with `Object.seal`:

```
// create array with 42 empty slots
let a = new Array(42);

// fill array with some value because
// empty slots can not be changed after calling Object.seal
a.fill(undefined);

Object.seal(a);
// now a is a fixed-size array with mutable entries
```

modules/qualitycontrol/QualityController.ts
```diff
@@ -0,0 +1,366 @@
...
+        this._rtc = rtc;
+        this._codecController = new CodecSelection(conference, codecSettings);
+        this._enableAdaptiveMode = enableAdaptiveMode;
+
+        this._receiveVideoController = new ReceiveVideoController(conference, rtc);
+        this._sendVideoController = new SendVideoController(conference, conference);
+        this._encodeTimeStats = new Map();
+
+        this._conference.on(
+            JitsiConferenceEvents._MEDIA_SESSION_STARTED,
```
saghul, line 121: Side question: any reason why we don't make this public?

modules/qualitycontrol/QualityController.ts
```diff
@@ -0,0 +1,366 @@
...
+                this._receiveVideoController.onMediaSessionStarted(session);
+                this._sendVideoController.onMediaSessionStarted(session);
+            });
+        this._conference.on(
+                JitsiConferenceEvents._MEDIA_SESSION_ACTIVE_CHANGED,
+                () => this._sendVideoController.configureConstraintsForLocalSources());
+        this._conference.on(
+            JitsiConferenceEvents.CONFERENCE_VISITOR_CODECS_CHANGED,
+            (codecList: CodecMimeType[]) => this._codecController.updateVisitorCodecs(codecList));
+        this._conference.on(
```
saghul, line 133: Should these 2 be dbounced, to avoid a "storm" when joining last or when many people leave at once?

service/RTC/StandardVideoSettings.ts
```diff
@@ -74,6 +75,19 @@ export const STANDARD_CODEC_SETTINGS = {
     }
 };
 
+export const VIDEO_CODECS_BY_COMPLEXITY = {
```
saghul, line 78: What happens if someone configures H.264 as their preferred codec for either of the 2?

service/RTC/StandardVideoSettings.ts
```diff
@@ -74,6 +75,19 @@ export const STANDARD_CODEC_SETTINGS = {
...
+    'camera' : [
+        CodecMimeType.AV1,
+        CodecMimeType.VP9,
+        CodecMimeType.VP8
+    ],
+    'desktop' : [
```
saghul, line 84: This order reads weird. Doesn't it represent "the codec I want to be in, in ideal conditions" ? In that case, I thought we'd want to be in AV1 specially for SS, no?

saghul (commented):
Left some comments, PTAL! I'll do some testing with the Acer Craptop. Excellent work on abstracting the whole thing into a QualityController, that itself controls the 3 other entities! ❤️

modules/qualitycontrol/QualityController.ts
```diff
@@ -0,0 +1,366 @@
...
+                    this._limitedByCpuTimeout = undefined;
+                }
+
+                return;
+            }
+
+            // Do nothing if the limitation reason is bandwidth since the browser will dynamically adapt the outbound
+            // resolution based on available uplink bandwith. Otherwise,
+            // 1. Switch the codec to the lowest complexity one incrementally.
+            // 2. Switch to a lower lastN value, cutting the receive videos by half in every iteration until
```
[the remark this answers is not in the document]
jallamsetty1: IMHO we should stick to only one approach here that we feel works best for our application and not make it configurable. Making the adaptive mode highly configurable makes it overcomplicated, there will be many code paths that we won't be testing and we will have to keep maintaining them for the community.

modules/qualitycontrol/QualityController.ts
```diff
@@ -0,0 +1,366 @@
...
+            // LAST_N_THRESHOLD is reached.
+            // 3. Lower the receive resolution of individual streams up to 180p.
+            if (qualityLimitationReason === QualityLimitationReason.CPU
+                || (encodeResolution < Math.min(localTrack.maxEnabledResolution, localTrack.getCaptureResolution())
+                    && qualityLimitationReason !== QualityLimitationReason.BANDWIDTH)) {
```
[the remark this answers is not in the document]
jallamsetty1: This was to handle the case where the browser doesn't report quality limitation in the stats and yet you see a degraded outbound stream as observed on your Acer laptop. It all depends on how we tackle that case. Maybe we need more datapoints to determine that is the case. Compute pressure doesn't seem to be relevant at all. It would be great if you can test on your laptop and share your findings so that we can adjust the check here.

modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -201,29 +73,53 @@ export default class ReceiveVideoController {
...
     }
 
+    /**
+     * Updates the source based constraints based on the maxHeight set.
+     * @returns {void}
+     */
+    _updateIndividualConstraints() {
+        const individualConstraints = this._receiverVideoConstraints.constraints;
+
+        if (individualConstraints && Object.keys(individualConstraints).length) {
```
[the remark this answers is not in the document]
jallamsetty1: Copy pasted from elsewhere :) I modified it iterate over Object.values()

modules/qualitycontrol/SendVideoController.js
```diff
@@ -151,6 +83,60 @@ export default class SendVideoController {
         return this._preferredSendMaxFrameHeight;
     }
 
+    /**
+     * Configures the video encodings on the local sources when a media connection is established or becomes active.
+     *
+     * @returns {Promise<void[]>}
```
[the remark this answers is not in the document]
jallamsetty1: Fixed it

modules/xmpp/JingleSessionPC.js
```diff
@@ -1177,13 +1177,13 @@ export default class JingleSessionPC extends JingleSession {
...
      *
-     * @param {Array<CodecMimeType>} codecList the preferred codecs.
+     * @param {Array<CodecMimeType>} codecList - Preferred codecs for video.
+     * @param {CodecMimeType} screenshareCodec - The preferred screenshare codec.
      */
-    setVideoCodecs(codecList) {
-
+    setVideoCodecs(codecList, screenshareCodec) {
         if (this._assertNotEnded()) {
             logger.info(`${this} setVideoCodecs: ${codecList}`);
```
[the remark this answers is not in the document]
jallamsetty1: Done!

service/RTC/StandardVideoSettings.ts
jallamsetty1: The complexity order is from highest to lowest. If the current codec is AV1 for camera, we will switch to VP9 and then to VP8. Same with SS codecs. If we are already encoding in AV1 for SS, we move onto modifying lastN/receive resolution.

service/RTC/StandardVideoSettings.ts
jallamsetty1: The reason I haven't included H.264 is because I am not able to configure it using the codec selection API. So we never attempt to switch to it as part of runtime adaptation. However, if the client is currently encoding in H.264 and the QualityController attempts correction, it will pick the first one from the list here and walk down the list when needed.

service/RTC/StandardVideoSettings.ts
```diff
@@ -74,6 +75,19 @@ export const STANDARD_CODEC_SETTINGS = {
...
+        CodecMimeType.VP9,
```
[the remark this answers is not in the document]
jallamsetty1: Nope, its the other way around as explained above.

modules/qualitycontrol/QualityController.ts
```diff
@@ -0,0 +1,366 @@
...
+
+        if (!this._limitedByCpuTimeout) {
+            this._limitedByCpuTimeout = window.setTimeout(() => {
+                this._limitedByCpuTimeout = undefined;
+                const updatedStats = this._encodeTimeStats.get(trackId);
+                const latestSourceStats: ISourceStats = updatedStats.get(updatedStats.size() - 1);
+
+                // If the encoder is still limited by CPU, switch to a lower complexity codec.
+                if (latestSourceStats.qualityLimitationReason === QualityLimitationReason.CPU
+                    || encodeResolution <  Math.min(localTrack.maxEnabledResolution, localTrack.getCaptureResolution())) {
```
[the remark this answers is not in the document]
jallamsetty1: It could but only in the cases where the camera doesn't support the requested resolution for one of the lower spatial layers when simulcast is enabled. If the requested is 720p and the camera supports only up to 540p, then we take the `Math.min(requestedSresolution, captureResolution) ` so it should be fine here. However, if the camera supports 540p and the bridge requests 360p, 270p will be sent and that should trigger a codec switch. I will improve this check to handle this case. Thanks for pointing it out.

modules/qualitycontrol/QualityController.ts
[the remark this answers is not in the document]
saghul: Will do!

modules/qualitycontrol/QualityController.ts
jallamsetty1: I am still working on this since I need to test this with a larger call to figure out what the optimum wait time. We do not want it to be long since users might have to wait a long time for videos to start rendering if a codec switch on the sender is needed.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1596,17 +1599,17 @@ TraceablePeerConnection.prototype._assertTrackBelongs = function(
...
  */
-TraceablePeerConnection.prototype.getConfiguredVideoCodec = function() {
-    const localVideoTrack = this.getLocalVideoTracks()[0];
+TraceablePeerConnection.prototype.getConfiguredVideoCodec = function(localTrack) {
+    const localVideoTrack = localTrack ?? this.getLocalVideoTracks()[0];
+    const rtpSender = this.findSenderForTrack(localVideoTrack.getTrack());
 
-    if (this.usesCodecSelectionAPI() && localVideoTrack) {
-        const rtpSender = this.findSenderForTrack(localVideoTrack.getTrack());
+    if (this.usesCodecSelectionAPI() && rtpSender) {
```
jallamsetty1, line 1612: We need the guardrail around the sender even though `RTCPeerConnection::getSenders()` is supported on all our endpoints because we remove the track from the sender when we mute the video source. So `findSenderForTrack` will return `undefined` in this case.

saghul (approved):
Impressive work! ❤️  🚀

PR #2543, opened 2024-07-01
fix(JitsiConference,ChatRoom) make isFocus check more resilient

saghul:
Protect against an incoming session that is received before we have processed presence for the focus user.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2325,11 +2325,8 @@ JitsiConference.prototype._rejectIncomingCall = function(jingleSession, options)
     // Terminate the jingle session with a reason
     jingleSession.terminate(
         null /* success callback => we don't care */,
-        error => {
-            logger.warn(
-                'An error occurred while trying to terminate'
-                    + ' invalid Jingle session', error);
-        }, {
+        null, /* error callback => we don't care */
+        {
             reason: options && options.reason,
             reasonDescription: options && options.reasonDescription,
             sendSessionTerminate: true
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -2694,12 +2694,11 @@ JitsiConference.prototype.sendApplicationLog = function() { };
 /**
  * Checks if the user identified by given <tt>mucJid</tt> is the conference focus.
  * @param mucJid the full MUC address of the user to be checked.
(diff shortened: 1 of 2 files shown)


saghul:

saghul:
Alright, I see what the problem is...

The function in `JitsiConference` checks the MUC JIDs, like `extraprotocolsrelaxhighly@conference.alpha.jitsi.net/focus` but the JIDs stored in the `Moderator` class after the conference request are connection JIDs like `focus@auth.alpha.jitsi.net`.

Oh well, this whole approach might be flawed...

@damencho I took a quick look and there doesn't seem to be a way to get the MUC JID from the conference request, is that right?

damencho:
Yeah, it is the connection jid that jicofo returns there. But we should also have the connection jids in the room.

saghul:
> Yeah, it is the connection jid that jicofo returns there. But we should also have the connection jids in the room.

Do / can we have that **before** we have processed presence for the focus? Where can I check? I look at the conferenceRequest object but it's not there... 

An option I don't like is to use `${mucJid}/focus` but it's not pretty...

damencho:
> Do / can we have that **before** we have processed presence for the focus? Where can I check? I look at the conferenceRequest object but it's not there...

Nope, the connection jid we get with the presence.

damencho:
https://github.com/jitsi/lib-jitsi-meet/blob/ec98b02095f806ca284d656965e2aba05bc0b2f8/modules/xmpp/ChatRoom.js#L686

This without a resource should be matching the one coming from jicofo `focus@auth.alpha.jitsi.net`.

damencho:
But the code already is doing the correct thing on next line?

damencho:
Aaa its the other way around, to be able to say whether the room jid is a jid for jicofo before getting the presence ... hum well that sounds wrong, as we cannot tell that. 
We should not get the room jid if that participant is not in the room, we do not have its presence ...

damencho:
Hum, maybe we should just make jicofo return both jids ...

saghul:
> Hum, maybe we should just make jicofo return both jids ... 

This is what I was thinking, since jicofo knows the room for sure.

At the same time, this is a workaround so maybe we should it...

PR #2544, opened 2024-07-08
feat(rtcstats): move conference start time to ljm

andrei-gavrilescu:


modules/RTCStats/RTCStats.ts
```diff
@@ -84,6 +86,16 @@ class RTCStats {
...
 
+        // The statisticsId, statisticsDisplayName and _statsCurrentId (renamed to displayName) fields 
+        // that are sent through options might be a bit confusing. Depending on the context, they could 
+        // be intermixed inside ljm, for instance _statsCurrentId might refer to the email field which is stored
+        // in statisticsId or it could have the same value as callStatsUserName.
+        // The following is the mapping between the fields, and a short explanation of each:
+        // statisticsId -> email, this is only send by jitsi-meet if enableEmailInStats option is set.
+        // statisticsDisplayName -> nick, this is only send by jitsi-meet if enableDisplayNameInStats option is set.
+        // localId, this is the unique id that is used to track users throughout stats.
+        const localId = Settings?.callStatsUserName ?? '';
```
saghul, line 97: If no unique name is provided, shouldn't we generate sonething random, instead of using the empty string and potentially mixing things up?

saghul (approved):
LGTM with a question.

modules/RTCStats/RTCStats.ts
andrei-gavrilescu: I don't know if this would be the place to do that, the callStatsUserName is saved locally and the most common use case for it is to track session originating from the same medium (browser, app, etc).
If we were to generate something unique it won't do us much good as we already have a statsSessionId which is unique per session. Having it empty would at least signal that callStatsUserName isn't properly set in some envs.

saghul: Sounds reasonable!

andrei-gavrilescu:

andrei-gavrilescu:

PR #2546, opened 2024-07-19
ref(QualityController) Add recovery mechanism and adjust the resolution check.

jallamsetty1:
Impl a recovery mechanism for the lastN to be increased if the cpu limitation goes away and doesn't return after increasing lastN. Also, additionally improve the calculation of the expected resolution taking simulcast stream resolutions into account.

### DIFF
--- a/modules/qualitycontrol/QualityController.ts
+++ b/modules/qualitycontrol/QualityController.ts
@@ -11,6 +11,7 @@ import { CodecSelection } from "./CodecSelection";
 import ReceiveVideoController from "./ReceiveVideoController";
 import SendVideoController from "./SendVideoController";
 import { VIDEO_CODECS_BY_COMPLEXITY, VIDEO_QUALITY_LEVELS } from '../../service/RTC/StandardVideoSettings';
+import { VideoType } from '../../service/RTC/VideoType';
 
 const logger = getLogger(__filename);
 
@@ -20,6 +21,10 @@ const LIMITED_BY_CPU_TIMEOUT = 60000;
 
 // The min. value that lastN will be set to while trying to fix video qaulity issues.
 const MIN_LAST_N = 3;
+
+// Wait time to check if cpu limitation reappears after lastN rampup is attempted for recovery. If it does, all further
+// attempts to bring lastN to configured channelLastN value will be blocked.
+const LAST_N_RAMPUP_TIME = 60000;
 
 enum QualityLimitationReason {
     BANDWIDTH = 'bandwidth',
@@ -47,6 +52,7 @@ interface ISourceStats {
     localTrack: JitsiLocalTrack;
     qualityLimitationReason: QualityLimitationReason;
     timestamp: number;
+    tpc: TraceablePeerConnection;
 };
 
 interface ITrackStats {
@@ -60,7 +66,7 @@ interface IVideoConstraints {
     sourceName: string;
 }
 
-class FixedSizeArray {
+export class FixedSizeArray {
     private _data: ISourceStats[];
     private _maxSize: number;
   
@@ -93,11 +99,13 @@ class FixedSizeArray {
  * by controlling the codec, encode resolution and receive resolution of the remote video streams. It also makes
  * adjustments based on the outbound and inbound rtp stream stats reported by the underlying peer connection.
  */
-export default class QualityController {
+export class QualityController {
     private _codecController: CodecSelection;
     private _conference: JitsiConference;
     private _enableAdaptiveMode: boolean;
     private _encodeTimeStats: Map<string, FixedSizeArray>;
+    private _isLastNRampupBlocked: boolean;
+    private _lastNRampupTimeout: number | undefined;
     private _limitedByCpuTimeout: number | undefined;
     private _receiveVideoController: ReceiveVideoController;
     private _sendVideoController: SendVideoController;
@@ -113,6 +121,7 @@ export default class QualityController {
         this._conference = conference;
         this._codecController = new CodecSelection(conference, codecSettings);
         this._enableAdaptiveMode = enableAdaptiveMode;
+        this._isLastNRampupBlocked = false;
         this._receiveVideoController = new ReceiveVideoController(conference);
         this._sendVideoController = new SendVideoController(conference);
         this._encodeTimeStats = new Map();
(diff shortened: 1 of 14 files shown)


modules/qualitycontrol/QualityController.ts
```diff
@@ -231,6 +266,80 @@ export default class QualityController {
...
+     * @returns {void}
+     */
+    _performQualityOptimizations(sourceStats: ISourceStats): void {
+        // Do not attempt run time adjustments if the adaptive mode is disabled.
+        if (!this._enableAdaptiveMode) {
+            return;
+        }
+
+        const { encodeResolution, localTrack, qualityLimitationReason, tpc } = sourceStats;
+        const trackId = localTrack.rtcId.toString();
```
saghul, line 284: isn't this a string already?

modules/qualitycontrol/QualityController.ts
```diff
@@ -231,6 +266,80 @@ export default class QualityController {
...
+
+        if (encodeResolution === tpc.calculateExpectedSendResolution(localTrack)) {
+            if (this._limitedByCpuTimeout) {
+                window.clearTimeout(this._limitedByCpuTimeout);
+                this._limitedByCpuTimeout = undefined;
+            }
+
+            if (qualityLimitationReason === QualityLimitationReason.NONE
```
saghul, line 292: If we move from being CPU limited to being bandwidth limited (not sure that's possible!) we are not doing anything, which I guess it's ok?

modules/qualitycontrol/QualityController.ts
```diff
@@ -231,6 +266,80 @@ export default class QualityController {
...
+                && qualityLimitationReason !== QualityLimitationReason.BANDWIDTH)) {
+
+            if (this._lastNRampupTimeout) {
+                window.clearTimeout(this._lastNRampupTimeout);
+                this._lastNRampupTimeout = undefined;
+                this._isLastNRampupBlocked = true;
+            }
+            const codecSwitched = this._maybeSwitchVideoCodec(trackId);
+
+            if (!codecSwitched && !this._limitedByCpuTimeout) {
```
saghul, line 332: Won't the 2nd part of the confition always be true? We always reset it is set above.

saghul (commented):
Left some comments, PTAL! Good job on adding tests for this!

modules/qualitycontrol/QualityController.ts
jallamsetty1: Nope thats a different timeout. `_lastNRampupTimeout` is reset above and here we are checking for `_limitedByCpuTimeout`

modules/qualitycontrol/QualityController.ts
jallamsetty1: If the bandwidth limitation kicks in after the cpu one gets cleared, I do not think its a good idea to increase lastN. Though the limitation is in upstream and increasing lastN affect downstream b/w, I still think its better to wait for b/w limitation to go away before ramping up lastN.

modules/qualitycontrol/QualityController.ts
jallamsetty1: Nope, its rtcTrackIdCounter that is passed when creating track and that is a number.

modules/qualitycontrol/QualityController.ts
```diff
@@ -146,24 +155,50 @@ export default class QualityController {
...
+            }
+            newLastN = videoStreamsReceived / 2;
+
+            if (newLastN < MIN_LAST_N) {
+                newLastN = MIN_LAST_N;
+            }
+
+        // Increment lastN by 1 every LAST_N_RAMPUP_TIME (60) secs.
+        } else if (lastN < channelLastN) {
+            newLastN++;
```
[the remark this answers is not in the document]
jallamsetty1: I don't think we should let number of remote participants influence lastN rampup. In the above example, if we stopped lastN rampup at 5 and after a while a bunch of users join the call at once, we would start receiving the sources slowly even though there is no cpu restriction as opposed to receiving all of them as soon as they join the call.

modules/qualitycontrol/QualityController.ts
```diff
@@ -146,24 +155,50 @@ export default class QualityController {
...
+        // Do nothing if lastN is already set to zero, i.e., the endpoint is in audio-only mode.
+        if (lastN === 0) {
             return false;
         }
 
-        let newLastN = videoStreamsReceived / 2;
+        let newLastN = lastN;
+
+        // If channelLastN is not set or set to -1 in config.js, the client will ramp up lastN to only up to 25.
+        let { channelLastN = 25 } = this._conference.options.config;
```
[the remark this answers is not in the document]
jallamsetty1: Yes! I was being lazy, we use this constant in multiple places. I will make this a constant and default video quality settings.

modules/qualitycontrol/QualityController.ts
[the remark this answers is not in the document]
saghul: Sounds good.

modules/qualitycontrol/QualityController.ts
saghul: Sounds good, just wanted to double check.

modules/qualitycontrol/QualityController.ts
```diff
@@ -146,24 +155,50 @@ export default class QualityController {
...
 
-        if (lastN === -1 || newLastN < MIN_LAST_N) {
-            newLastN = MIN_LAST_N;
+        channelLastN = channelLastN === -1 ? 25 : channelLastN;
+        if (cpuLimited) {
+            const videoStreamsReceived = this._conference.getForwardedSources().length;
+            if (lastN !== -1 && lastN <= MIN_LAST_N) {
```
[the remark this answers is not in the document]
jallamsetty1: Done

modules/qualitycontrol/QualityController.ts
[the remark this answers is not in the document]
jallamsetty1: Done

modules/qualitycontrol/QualityController.ts
```diff
@@ -21,6 +22,10 @@ const LIMITED_BY_CPU_TIMEOUT = 60000;
 // The min. value that lastN will be set to while trying to fix video qaulity issues.
 const MIN_LAST_N = 3;
 
+// Wait time to check if cpu limitation reappears after lastN rampup is attempted for recovery. If it does, all further
+// attempts to bring lastN to configured channelLastN value will be blocked.
+const LAST_N_RAMPUP_TIME = 60000;
```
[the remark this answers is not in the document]
jallamsetty1: Done

modules/qualitycontrol/QualityController.ts
```diff
@@ -146,24 +155,50 @@ export default class QualityController {
...
+     * @param {boolean} cpuLimited - whether the endpoint is cpu limited or not.
+     * @returns boolean - Returns true if an action was taken, false otherwise.
      */
-    _maybeLowerLastN(): boolean {
+    _lowerOrRaiseLastN(cpuLimited: boolean): boolean {
         const lastN = this.receiveVideoController.getLastN();
-        const videoStreamsReceived = this._conference.getForwardedSources().length;
 
-        if (lastN !== -1 && lastN <= MIN_LAST_N) {
+        // Do nothing if lastN is already set to zero, i.e., the endpoint is in audio-only mode.
```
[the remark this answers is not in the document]
jallamsetty1: Hmm .. it doesn't seem to be that straight forward since audio-only feature is fully implemented in the application. However, I realized that when audio-only mode is enabled, client's local video sources are muted so the stats won't be processed and when either camera or screenshare is started, the client is automatically kicked out of audio-only mode. Therefore, we don't have to make assumptions about whether client is in that mode while calculating lastN. I have removed that check all together now.

modules/qualitycontrol/QualityController.ts
```diff
@@ -113,18 +114,18 @@ export class QualityController {
     /**
      * 
      * @param {JitsiConference} conference - The JitsiConference instance.
-     * @param {Object} codecSettings - Codec settings for jvb and p2p connections passed to config.js.
-     * @param {boolean} enableAdaptiveMode - Whether the client needs to make runtime adjustments for better video
-     * quality.
+     * @param {Object} options - video quality settings passed through config.js.
      */
-    constructor(conference: JitsiConference, codecSettings: any, enableAdaptiveMode: boolean) {
+    constructor(conference: JitsiConference, options) {
```
saghul, line 119: type for options?

modules/qualitycontrol/QualityController.ts
```diff
@@ -163,25 +164,20 @@ export class QualityController {
...
-        channelLastN = channelLastN === -1 ? 25 : channelLastN;
+        channelLastN = channelLastN === LAST_N_UNLIMITED ? DEFAULT_LAST_N : channelLastN;
         if (cpuLimited) {
             const videoStreamsReceived = this._conference.getForwardedSources().length;
-            if (lastN !== -1 && lastN <= MIN_LAST_N) {
-                return false;
-            }
-            newLastN = videoStreamsReceived / 2;
 
+            newLastN = videoStreamsReceived / 2;
```
saghul, line 180: Round to the nearest integer?

modules/qualitycontrol/QualityController.ts
```diff
@@ -380,7 +377,7 @@ export class QualityController {
                     trackStats: [ ssrcStats ]
                 };
 
-                statsPerTrack.set(track.rtcId, existingStats);
+                statsPerTrack.set(trackId, existingStats);
```
saghul, line 380: Why are we setting the string as a key, which then forces us to use `.toString()` everywhere?

saghul (approved):
LGTM with a couple of minor nits, awesome work!

modules/qualitycontrol/QualityController.ts
jallamsetty1: Done

modules/qualitycontrol/QualityController.ts
jallamsetty1: Done

modules/qualitycontrol/QualityController.ts
jallamsetty1: I initially declared the key for the encode time stats to be a string and later realized that `rtcId` is a number. I changed it back. At that time I thought a string made more sense if its a number/string but I see that its a number always. I changed it to a number number.

saghul (approved):
🙌

PR #2552, opened 2024-08-08
fix(ICE): treat completed as connected

paweldomas:
When WebRTC ICE gathering policy is set to gather once instead of continually, the controlling agent goes to completed instead of connected (no more candidates to check). This doesn't happen in Chrome or other browsers, but is reproducible with node.js wrapper around WebRTC which runs with the default settings.

This is causing a bug where the initiator side of a P2P connection does not fully switch to P2P mode and keeps on sending data on both P2P and JVB connections.


jallamsetty1 (approved):

corby-github:
👍 Thank you for this fix!

PR #2557, opened 2024-08-20
fix(chat) change jid to participantId for reactions

he-patrick:
- jitsi-meet uses participantId to handle received reactions


JitsiConferenceEventManager.js
```diff
@@ -380,10 +380,11 @@ JitsiConferenceEventManager.prototype.setupChatRoomListeners = function() {
         XMPPEvents.REACTION_RECEIVED,
 
         (jid, reactionList, messageId) => {
+            participantId = Strophe.getResourceFromJid(jid);
```
mihhu: Please make sure you add const in front of this so it's defined, as it's a new variable without context

PR #2560, opened 2024-08-23
fix(TPC) fix getting video SSRCs if local one is not present yet

saghul:


saghul:
I caught this one in the wild, not sure how the code can end up there.

damencho (approved):

PR #2561, opened 2024-08-23
fix(ResumeTask) reset resume timeout after running timer

saghul:
It will avoid having an extra "Canceling connection resume task" log line every time we schedule a timer.

### DIFF
--- a/modules/xmpp/ResumeTask.js
+++ b/modules/xmpp/ResumeTask.js
@@ -127,6 +127,8 @@ export default class ResumeTask {
      * @returns {void}
      */
     _resumeConnection() {
+        this._resumeTimeout = undefined;
+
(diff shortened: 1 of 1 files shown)


damencho (approved):

modules/xmpp/ResumeTask.js
```diff
@@ -127,6 +127,8 @@ export default class ResumeTask {
      * @returns {void}
      */
     _resumeConnection() {
+        this._resumeTimeout = undefined;
```
Romick2005, line 130: Should you also clear timeout? Like: clearTimeout(this._resumeTimeout); or it is not necessary?

saghul: Not necessary since this function is called in the timeout handler.

PR #2562, opened 2024-08-23
fix(XmppConnection) limit retries to 3, bail afterwards

saghul:


saghul:
Draft, gotta test it works, will do later.

damencho (approved):

PR #2565, opened 2024-09-04
fix(xmpp): Fixes skipping messages when disco-info is processed late.

damencho:
When using bosh and xmpp conference-request (no jiconop on server-side) we can drop av moderation and speaker-stats messages.

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -149,6 +149,10 @@ export default class XMPP extends Listenable {
         // Cache of components used for certain features.
         this._components = [];
 
+        // We want to track messages that are received before we process the disco-info components
+        // re-order of receiving we may drop some messages
+        this._preComponentsMsgs = [];
+
         initStropheNativePlugins();
 
         const xmppPing = options.xmppPing || {};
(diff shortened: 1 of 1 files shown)


modules/xmpp/xmpp.js
```diff
@@ -508,8 +514,9 @@ export default class XMPP extends Listenable {
         this._maybeSendDeploymentInfoStat(true);
 
         if (this._components.length > 0) {
```
saghul, line 516: Drop the if since it's always initialized to an array?

saghul (dismissed):
LGTM, good catch!

saghul (approved):

jallamsetty1 (approved):

modules/xmpp/xmpp.js
damencho: By default components is empty and we will iterate over the messages and don't do anything. So its not a big deal with the check ...

hristoterezov: I understand that functionally it wont change anything. I'm asking technically what is the advantage to iterate over the messages and call the function so that it will quickly exit versus just having the check?

hristoterezov (approved):

PR #2572, opened 2024-09-16
ref(SDP) Convert to ES6 class.

jallamsetty1:


saghul (approved):
🚀  Bonus points if you extract `urn:xmpp:jingle:apps:dtls:0` and similar constants so we have smaller identifiers to visually parse.

saghul (approved):

damencho (approved):

PR #2575, opened 2024-09-19
feat(visitors): Adds an option to skip redirections.

damencho:
This is needed for breakout rooms where the main room has visitors.


saghul (approved):

damencho:
Close in favor of https://github.com/jitsi/lib-jitsi-meet/pull/2576

PR #2576, opened 2024-09-20
feat(moderator): Skips redirect events when moving to main from breakout.

damencho:


hristoterezov (approved):

authenticateAndUpgradeRole.js
```diff
@@ -111,7 +115,9 @@ export default function authenticateAndUpgradeRole({
                         // we execute this logic in JitsiConference where we bind the current conference as `this`
                         // At this point we should have the new session ID
                         // stored in the settings. Send a new conference IQ.
-                        this.room.xmpp.moderator.sendConferenceRequest(this.room.roomjid).finally(resolve);
+                        this.room.xmpp.moderator.sendConferenceRequest(this.room.roomjid)
+                            .catch(e => logger.trace('sendConferenceRequest rejected', e))
+                            .finally(resolve);
```
Romick2005, line 120: Why do you need `finally` here? Isn't it better to have just `then`?

modules/xmpp/ChatRoom.js
```diff
@@ -239,6 +239,9 @@ export default class ChatRoom extends Listenable {
             const preJoin
                 = this.options.disableFocus
                     ? Promise.resolve()
+                        .finally(() => {
```
Romick2005, line 242: Same here: Why do you need finally here? Isn't it better to have just then?

modules/xmpp/moderator.js
```diff
@@ -335,28 +335,31 @@ export default class Moderator extends Listenable {
...
                     .catch(error => {
                         logger.warn(`Error: ${error}`);
-                        this._handleError(roomJid);
+                        this._handleError(roomJid, undefined, undefined, resolve, reject);
                     });
             }
         }).then(() => {
             this.conferenceRequestSent = true;
+        })
+        .finally(() => {
```
Romick2005, line 361: Would be great if you could add catch -> log some error and then add then to proceed with this. In general I would suggest to avoid using finally as mush as possible as it does not work right away in RN

PR #2577, opened 2024-09-25
ref(SDPDiffer) Convert to ES6 class.

jallamsetty1:
Make it work directly with unified-plan SDP and add more unit tests.

Please look at the individual commits.


modules/RTC/TraceablePeerConnection.js
```diff
@@ -1123,117 +1130,80 @@ TraceablePeerConnection.prototype._removeRemoteTrack = function(toBeRemoved) {
...
-     * Groups mapped by primary SSRC number
-     * @type {Map<number,Array<SSRCGroupInfo>>}
-     */
-    const groupsMap = new Map();
-
-    if (typeof desc !== 'object' || desc === null
-        || typeof desc.sdp !== 'string') {
-        logger.warn('An empty description was passed as an argument');
-
+    if (!localSDP || typeof localSDP !== 'string') {
```
saghul, line 1141: Is this an error? Should a warning be logged?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1123,117 +1130,80 @@ TraceablePeerConnection.prototype._removeRemoteTrack = function(toBeRemoved) {
...
         || mline.direction === MediaDirection.SENDRECV);
 
-    let index = 0;
-
-    for (const mLine of media) {
-        if (!Array.isArray(mLine.ssrcs)) {
-            continue; // eslint-disable-line no-continue
-        }
+    if (!Array.isArray(media)) {
+        return;
```
saghul, line 1149: Shouldn't we set the new map and return nothing?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1123,117 +1130,80 @@ TraceablePeerConnection.prototype._removeRemoteTrack = function(toBeRemoved) {
...
+    }
 
-        if (Array.isArray(mLine.ssrcGroups)) {
-            for (const group of mLine.ssrcGroups) {
-                if (typeof group.semantics !== 'undefined' && typeof group.ssrcs !== 'undefined') {
-                    // Parse SSRCs and store as numbers
-                    const groupSSRCs = group.ssrcs.split(' ').map(ssrcStr => parseInt(ssrcStr, 10));
-                    const primarySSRC = groupSSRCs[0];
+    for (const localTrack of this.localTracks.values()) {
```
saghul, line 1152: `localTracks` is already an array, you can iterate it without the `.values()`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1123,117 +1130,80 @@ TraceablePeerConnection.prototype._removeRemoteTrack = function(toBeRemoved) {
...
-                    // Note that group.semantics is already present
-                    group.ssrcs = groupSSRCs;
+            ssrcs.forEach(ssrc => ssrcInfo.ssrcs.push(ssrc.id));
 
-                    // eslint-disable-next-line max-depth
-                    if (!groupsMap.has(primarySSRC)) {
-                        groupsMap.set(primarySSRC, []);
-                    }
-                    groupsMap.get(primarySSRC).push(group);
+            if (Array.isArray(ssrcGroups)) {
```
saghul, line 1173: We check `ssrcs` with `if (ssrcs?.length) {` but here we do it differently. I feel we should check consistently since they are both arrays.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1123,117 +1130,80 @@ TraceablePeerConnection.prototype._removeRemoteTrack = function(toBeRemoved) {
...
+                for (const group of ssrcGroups) {
+                    group.ssrcs = group.ssrcs.split(' ').map(ssrcStr => parseInt(ssrcStr, 10));
+                    ssrcInfo.groups.push(group);
                 }
-            }
 
-            const simGroup = mLine.ssrcGroups.find(group => group.semantics === 'SIM');
+                const simGroup = ssrcGroups.find(group => group.semantics === 'SIM');
```
saghul, line 1179: Extract `SIM` to a constant?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1123,117 +1130,80 @@ TraceablePeerConnection.prototype._removeRemoteTrack = function(toBeRemoved) {
...
-                const ssrcGroups = groupsMap.get(ssrcNumber);
+            const oldSsrcInfo = this.localSSRCs.get(localTrack.rtcId);
+            const oldSsrc = this._extractPrimarySSRC(oldSsrcInfo);
+            const newSsrc = this._extractPrimarySSRC(ssrcInfo);
 
-                for (const group of ssrcGroups) {
-                    ssrcInfo.groups.push(group);
-                }
+            if (oldSsrc !== newSsrc) {
+                oldSsrc && logger.error(`${this} Overwriting SSRC for track=${localTrack}] with ssrc=${newSsrc}`);
```
saghul, line 1199: What's the effect of this? If it's not kosher to do why are we doing it?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1123,117 +1130,80 @@ TraceablePeerConnection.prototype._removeRemoteTrack = function(toBeRemoved) {
...
-     * @type {Map<number,Array<SSRCGroupInfo>>}
-     */
-    const groupsMap = new Map();
-
-    if (typeof desc !== 'object' || desc === null
-        || typeof desc.sdp !== 'string') {
-        logger.warn('An empty description was passed as an argument');
-
+    if (!localSDP || typeof localSDP !== 'string') {
         return ssrcMap;
```
saghul, line 1142: Set the map and return nothing?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1516,7 +1484,12 @@ TraceablePeerConnection.prototype.addTrack = function(track, isInitiator = false
     const webrtcStream = track.getOriginalStream();
 
     try {
-        this.tpcUtils.addTrack(track, isInitiator);
+        transceiver = this.tpcUtils.addTrack(track, isInitiator);
```
saghul, line 1487: This try block feels too wide no? What is the operation that can throw an error? Ideally we keep it as narrow as possible.

modules/sdp/LocalSdpMunger.js
```diff
@@ -28,77 +25,75 @@ export default class LocalSdpMunger {
...
-        const pcId = this.tpc.id;
-
-        return `${streamId}-${pcId} ${trackId}-${pcId}`;
     }
 
     /**
-     * Updates or adds a 'msid' attribute in the format '<endpoint_id>-<mediaType>-<trackIndex>-<tpcId>'
-     * example - d8ff91-video-0-1
-     * All other attributes like 'cname', 'label' and 'mslabel' are removed since these are not processed by Jicofo.
+     * Updates or adds a 'msid' attribute in the format. All other attributes like 'cname', 'label' and 'mslabel' are
```
saghul, line 31: "Updates or adds a 'msid' attribute in the format." not sure what this means. I feel like a part was chopped and it no longer makes sense. Perhaps it should just say "Updates or adds a 'msid' attribute."

modules/sdp/LocalSdpMunger.js
```diff
@@ -28,77 +25,75 @@ export default class LocalSdpMunger {
...
-        // Always overwrite msid since we want the msid to be in this format even if the browser generates one.
-        for (const source of sources) {
-            const msid = mediaSection.ssrcs.find(ssrc => ssrc.id === source && ssrc.attribute === 'msid');
-
-            if (msid) {
-                trackId = msid.value.split(' ')[1];
+    _transformMediaIdentifiers(mediaSection, ssrcMap) {
+        const mediaType = mediaSection.mLine.type;
+        const mediaDirection = mediaSection.mLine.direction;
+        const sources = [ ...new Set(mediaSection.mLine.ssrcs?.map(s => s.id)) ];
```
saghul, line 42: Is the intention to get an array of unique IDs?

modules/sdp/LocalSdpMunger.js
```diff
@@ -28,77 +25,75 @@ export default class LocalSdpMunger {
...
+        let sourceName;
+
+        if (ssrcMap.size) {
+            for (const [ id, trackSsrcs ] of ssrcMap.entries()) {
+                if (isEqual([ ...sources ].sort(), [ ...trackSsrcs.ssrcs ].sort())) {
```
saghul, line 47: You are sorting a copy of the `sources` array on each loop iteration... that's 1 too-many copies. You can just sort the array on creation and compare against it.

saghul, line 47: Why do you copy `trackSsrcs` would the inplace sort eb detrimental?

modules/sdp/SDP.js
```diff
@@ -20,8 +21,9 @@ export default class SDP {
      *
      * @param {string} sdp - The SDP generated by the browser when SDP->Jingle conversion is needed, an empty string
      * when Jingle->SDP conversion is needed.
+     * @param {boolean} isP2P - Whether this SDP belongs to a p2p peerconnection.
      */
-    constructor(sdp) {
+    constructor(sdp, isP2P = false) {
```
saghul, line 26: The `isP2P` option feels weird to me. From the prespective of a class that works with SDP, it's probably best to name it by the behavior it encapsulates, rather than when you want to actually use that behavior.

If my reading is correct something along the lines of `useMid` or bundleGroupIsMid` would do?

modules/sdp/SDP.js
```diff
@@ -86,9 +89,9 @@ export default class SDP {
      * @returns {boolean}
      */
     containsSSRC(ssrc) {
-        const souceMap = this.getMediaSsrcMap();
+        const sourceMap = this.getMediaSsrcMap();
 
-        return Object.values(souceMap).some(media => media.ssrcs[ssrc]);
+        return [ ...sourceMap.values() ].some(media => media.ssrcs[ssrc]);
```
saghul, line 94: Fun observation: Iterators now have `some` but only on very recent browsers :-/

modules/sdp/SDP.js
```diff
@@ -462,32 +472,91 @@ export default class SDP {
...
+                            source.append(param);
+                        }
+                        description.append(source);
+                    }
+
+                    const ssrcGroupLines = SDPUtil.findLines(mediaItem, 'a=ssrc-group:');
+
+                    ssrcGroupLines.forEach(line => {
+                        const idx = line.indexOf(' ');
+                        const semantics = line.substr(0, idx).substr(13);
```
saghul, line 518: Where do these 13 and 14 numbers come from?

modules/sdp/SDPDiffer.js
```diff
@@ -1,220 +1,122 @@
...
+            const mySsrcs = Object.keys(mySource.ssrcs);
+            const othersSsrcs = Object.keys(othersSource.ssrcs);
 
-    for (let i = 0, l = array1.length; i < l; i++) {
-        // Check if we have nested arrays
-        if (array1[i] instanceof Array && array2[i] instanceof Array) {
-            // recurse into the nested arrays
-            if (!array1[i].equals(array2[i])) {
-                return false;
+            if (othersSsrcs.length && !isEqual([ ...mySsrcs ].sort(), [ ...othersSsrcs ].sort())) {
```
saghul, line 46: Do `Object.keys(mySource.ssrcs).sort()` so you don't need to create a copy.

modules/xmpp/JingleSessionPC.js
```diff
@@ -1009,7 +1009,7 @@ export default class JingleSessionPC extends JingleSession {
             sid: this.sid
         });
 
-        new SDP(offerSdp).toJingle(
+        new SDP(offerSdp, this.isP2P).toJingle(
```
saghul, line 1012: Do we foresee any extra options to this constructor?

saghul (commented):
Left some comments! 😮‍💨

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Good point! Yes, but it has no affect on the functionality and that is why I didn't notice it. Thanks for pointing it out.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Isn't it a Map?

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Sure, I will include this change in the follow-up PR.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: It has no affect on functionality since we fixed it recently but it's good to know when this happens and was therefore  logging it as an error. Technically it shouldn't be happening but recently we saw this happen in Chrome 123 or so. I can change it to debug.

modules/sdp/LocalSdpMunger.js
jallamsetty1: Fixed the comments.

modules/sdp/LocalSdpMunger.js
jallamsetty1: Yes

modules/sdp/LocalSdpMunger.js
jallamsetty1: I was trying to preserve the order of SSRCs from the SDP to Jingle. If I sort `sources` in place, the newly added lines for `name` and `videoType` will be out of order.

modules/sdp/SDP.js
jallamsetty1: Now that we are handling the many m-lines to one media description conversion in SDP class itself, we need to know if the SDP is for p2p or not. It is not just for bundle group line, in the follow up PR, that option is used to do Jingle->SDP differently for P2P vs JVB SDPs.

modules/sdp/SDP.js
jallamsetty1: Good to know, I will look that up.

modules/sdp/SDP.js
jallamsetty1: length of 'a=ssrc-group:'

modules/xmpp/JingleSessionPC.js
jallamsetty1: Nope, this is all we would need and that is why I didn't create an options parameter for this.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: This technically wouldn't happen since the browser will always produce an  SDP for createOffer/createAnswer. The call itself will fail and user kicked out when that happens so I thought it's ok not to display the error here.

saghul: In that case, let's throw an exception, as if it was an assertion. If the invariant is ever broken, better do it as loud as possible.

modules/RTC/TraceablePeerConnection.js
saghul: Sorry, you're right. I was reading RTC.js where it's an Array 😅

modules/sdp/LocalSdpMunger.js
saghul: Is it possible to get duplicated?

modules/sdp/LocalSdpMunger.js
saghul: Aha. In that case it might be better to have a `const sortedSources = sources.slice().sort();` to avoid a copy on each loop iteration, which would be a bunch of them in case of many participants, no?

modules/sdp/SDP.js
saghul: Ack.

modules/sdp/SDP.js
saghul: Actually, it may work today, since Babel uses core-js to polyfill all missing ES features. Can you please give it a try? It should basically be: ` return sourceMap.values().some(media => media.ssrcs[ssrc]);`

modules/sdp/SDP.js
saghul: I wish there was a better way...

modules/xmpp/JingleSessionPC.js
saghul: Ack.

saghul (commented):
A couple of minor comments, almost there! Great work Jaya!

modules/sdp/SDP.js
jallamsetty1: I added a helper function in SDPUtil for this.

modules/sdp/SDP.js
jallamsetty1: I wouldn't worry too much about it here in this specific instance since this method is going away in my follow up PR but I will keep this in mind.

modules/sdp/LocalSdpMunger.js
```diff
@@ -28,77 +22,78 @@ export default class LocalSdpMunger {
...
-                trackId = msid.value.split(' ')[1];
+    _transformMediaIdentifiers(mediaSection, ssrcMap) {
+        const mediaType = mediaSection.mLine.type;
+        const mediaDirection = mediaSection.mLine.direction;
+        const sources = [ ...new Set(mediaSection.mLine.ssrcs?.map(s => s.id)) ];
+        let sourceName;
+
+        if (ssrcMap.size) {
+            for (const [ id, trackSsrcs ] of ssrcMap.entries()) {
+                const sortedSources = sources.slice().sort();
```
saghul, line 45: Put this outside of the loop since it doesn't change within.

jallamsetty1: Done

saghul (approved):
🔥🔥🔥

PR #2578, opened 2024-09-25
feat(statistics): add pre call test interface

andrei-gavrilescu:


modules/statistics/PreCallTest.ts
```diff
@@ -0,0 +1,48 @@
+import PreCallTest from '@jitsi/precall-test';
+
+
+export interface PreCallResult {
+    throughput: number; // Maximum bandwidth reached in kbps  (kilo bits per second).
+    fractionalLoss: number; // Packet loss percentage over all the test traffic.
+    rtt: number;  // Round trip time in milliseconds.
+    jitter: number;
```
saghul, line 8: Document this too pl.

modules/statistics/PreCallTest.ts
```diff
@@ -0,0 +1,48 @@
...
+ * Run a pre-call test to check the network conditions. It uses a TURN server to establish
+ * a connection between two PeerConnections using the server as a relay. Afterwards it sends 
+ * some test traffic through a data channel to measure the network conditions, these are 
+ * recorded and returned through a Promise.
+ * 
+ * @param {Array<IceServer>} - The ICE servers to use for the test, these are passes to the PeerConnection constructor. 
+ * @returns {Promise<PreCallResult | string>} - A Promise that resolves with the test results or rejects with an error message.
+ */
+export default async function runPreCallTest(iceServers: Array<IceServer>): Promise<PreCallResult | string> {
+    // On initialization, the PreCallTest object simply does some checks and some browsers verifications,
```
saghul, line 31: Can we have some static `init()` method perhaps? It could keep track of whether it has already run and avoid doing the work over and oer again.

modules/statistics/PreCallTest.ts
```diff
@@ -0,0 +1,48 @@
...
+    // these seem to be reusable, so we'll keep the object around.
+    preCallTest || (preCallTest = new PreCallTest())
+
+    return new Promise((resolve, reject) => {
+        // It's not explicitly stated in the code, but if message is not null, something went wrong,
```
saghul, line 36: Let's change the library please, this is not a good API.

webpack-shared-config.js
```diff
@@ -80,8 +80,8 @@ module.exports = (minimize, analyzeBundle) => {
         },
         performance: {
             hints: minimize ? 'error' : false,
-            maxAssetSize: 1.08 * 1024 * 1024,
-            maxEntrypointSize: 1.08 * 1024 * 1024
+            maxAssetSize: 1.09 * 1024 * 1024,
```
saghul, line 83: Let's bump it to 1.25 to avoid having to do it one by one.

saghul (changes requested):

modules/statistics/PreCallTest.ts
andrei-gavrilescu: I agree, just didn't want to mess with the library to much, cause I don't know the extent of the changes required. Should  I start a refactoring effort on precall_test?

saghul: By the looks we don't need a full refactor, but a better way to report errors, to start with.

andrei-gavrilescu: alright

modules/statistics/PreCallTest.ts
saghul: I meant the initial check we can reuse, not the whole test. This "init" thing would do just that.

andrei-gavrilescu: not sure I follow completely, you mean something like?
`const init = (function() = {
    preCallTest = new PreCallTest())
})()` And basically initialize on import? so I can remove the check from runPreCallTest?

saghul: No, I mean something like:

```
await PrecallTest.globalInit();
const p = new PreCallTest();
...
```

modules/statistics/PreCallTest.ts
andrei-gavrilescu: Added a new version with a promise based interface

saghul (approved):

PR #2586, opened 2024-10-09
fix(JingleSession) Remove ice restart option and functionality.

jallamsetty1:
Reinvite method in Jicofo is set to restart-jingle (terminate and recreate the whole jingle session) instead of sending 'transport-replace'. The replaceTransport option has never worked because of a bug in Chrome related to TCC sequence numbering after the DTLS transport layer is reset to start a new DTLS handshake with fresh transport on the bride.

### DIFF
--- a/modules/xmpp/strophe.jingle.js
+++ b/modules/xmpp/strophe.jingle.js
@@ -4,13 +4,7 @@ import { cloneDeep } from 'lodash-es';
 import { $iq, Strophe } from 'strophe.js';
 
 import { MediaType } from '../../service/RTC/MediaType';
-import {
-    ACTION_JINGLE_TR_RECEIVED,
-    ACTION_JINGLE_TR_SUCCESS,
-    createJingleEvent
-} from '../../service/statistics/AnalyticsEvents';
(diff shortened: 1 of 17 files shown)


modules/xmpp/strophe.jingle.js
```diff
@@ -255,40 +249,6 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
             this.eventEmitter.emit(XMPPEvents.CALL_ENDED, sess, reasonCondition, reasonText);
             break;
         }
-        case 'transport-replace': {
```
bgrozev, line 258: Suggest adding a log that transport-replace in not supported, just to help people debug if it happens to be used somehow

damencho (approved):

fippo:
Can you remind me of the upstream issue number? Agree that the general behavior after a DTLS restart is pretty underspecified but maybe we can at least get the TWCC sequence number clarified :-)

jallamsetty1:
> Can you remind me of the upstream issue number? Agree that the general behavior after a DTLS restart is pretty underspecified but maybe we can at least get the TWCC sequence number clarified :-)

Sure, here it is - https://issues.webrtc.org/issues/42222276

PR #2590, opened 2024-10-24
feat(JingleSession) Convert Jingle->SDP directly w/o interop layer.

jallamsetty1:


modules/RTC/TraceablePeerConnection.js
```diff
@@ -769,45 +783,37 @@ TraceablePeerConnection.prototype.getRemoteTracks = function(endpointId, mediaTy
...
-                if (fidLines.length) {
-                    const secondarySsrc = fidLines[0].split(' ')[2];
-
-                    lines += `${fidLines[0]}\r\n`;
-                    ssrcLines = ssrcLines.concat(SDPUtil.findLines(media, `a=ssrc:${secondarySsrc}`));
-                }
-                removeSsrcInfo[idx] += `${ssrcLines.join('\r\n')}\r\n`;
-                removeSsrcInfo[idx] += lines;
-            }
+    for (const [ sourceName, sourceInfo ] of this._remoteSsrcMap.entries()) {
```
saghul, line 812: You don't need the `.entries()` to iterate a Map.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -769,45 +783,37 @@ TraceablePeerConnection.prototype.getRemoteTracks = function(endpointId, mediaTy
...
- * Parses the remote description and returns the sdp lines of the sources associated with a remote participant.
+ * Returns the remote sourceInfo for a given source name.
+ *
+ * @param {string} sourceName - The source name.
+ * @returns {TPCSourceInfo}
+ */
+TraceablePeerConnection.prototype.getRemoteSourceInfoBySourceName = function(sourceName) {
+    const info = this._remoteSsrcMap?.get(sourceName);
+
+    return info ? cloneDeep(info) : null;
```
saghul, line 794:
```suggestion
    return cloneDeep(this._remoteSsrcMap?.get(sourceName) ?? null);
```

saghul: I assume `cloneDeep(null) === null`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -328,12 +326,28 @@ export default function TraceablePeerConnection(
...
+     * Holds the SSRC map for the remote tracks mapped by their source names.
+     *
+     * @type {Map<string, TPCSourceInfo>}
+     * @property {string} mediaType - The media type of the track.
+     * @property {string} msid - The track's MSID.
+     * @property {Array<string>} ssrcs - The SSRCs associated with the track.
+     * @property {Array<TPCGroupInfo>} groups - The SSRC groups associated with the track.
+     * @property {string} msid - The track's MSID.
+     */
+    this._remoteSsrcMap = null;
```
saghul, line 349: Looking at how we need to handle the code later on, doesn't it make sense to always have this initialized to a an empty Map?

Having its size be 0 would be equivalente to the current null value, but I bet it sill simplify some code and make things easier to work with.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1145,7 +1154,7 @@ TraceablePeerConnection.prototype._processAndExtractSourceInfo = function(localS
     const media = session.media.filter(mline => mline.direction === MediaDirection.SENDONLY
         || mline.direction === MediaDirection.SENDRECV);
 
-    if (!Array.isArray(media)) {
+    if (!media?.length) {
```
saghul, line 1157: Is the optional chaining necessary? `media` will always be an array since filter always returns one.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1837,6 +1842,27 @@ TraceablePeerConnection.prototype.removeTrackFromPc = function(localTrack) {
...
+/**
+ * Updates the remote source map with the given source map for adding or removing sources.
+ *
+ * @param {Map<string, TPCSourceInfo>} sourceMap - The map of source names to their corresponding SSRCs.
+ * @param {boolean} isAdd - Whether the sources are being added or removed.
+ * @returns {void}
+ */
+TraceablePeerConnection.prototype.updateRemoteSources = function(sourceMap, isAdd) {
+    if (this._remoteSsrcMap) {
+        for (const [ sourceName, ssrcInfo ] of sourceMap.entries()) {
```
saghul, line 1854: No need for `.entries()`

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1837,6 +1842,27 @@ TraceablePeerConnection.prototype.removeTrackFromPc = function(localTrack) {
...
+            if (isAdd) {
+                this._remoteSsrcMap.set(sourceName, ssrcInfo);
+            } else {
+                this._remoteSsrcMap.delete(sourceName);
+            }
+        }
+    } else if (isAdd) {
```
saghul, line 1861: This branch wouldn't be necessary if the Map is always initialized.

modules/sdp/SDP.js
```diff
@@ -24,7 +24,30 @@ export default class SDP {
      * @param {boolean} isP2P - Whether this SDP belongs to a p2p peerconnection.
      */
     constructor(sdp, isP2P = false) {
-        const media = sdp.split('\r\nm=');
+        this._updateSessionAndMediaSections(sdp);
+        this.isP2P = isP2P;
+        this.raw = this.session + this.media.join('');
```
saghul, line 29: Question: would it make sense to make `raw ` a getter which creates its value when accessed? Looks like we need to keep it in sync with `session` and `media`, but a getter would help simplify that. The downside would be if we access it many many times, since its value would be computed every time.

modules/sdp/SDP.js
```diff
@@ -34,64 +57,104 @@ export default class SDP {
...
+            const { mediaType, msid, ssrcList, groups } = source;
+            let idx;
+
+            if (isAdd) {
+                // For P2P, check if there is an m-line with the matching mediaType that doesn't have any ssrc lines.
+                // Update the existing m-line if it exists, otherwise create a new m-line and add the sources.
+                idx = this.media.findIndex(mLine => mLine.includes(`m=${mediaType}`) && !mLine.includes('a=ssrc'));
+                if (!this.isP2P || idx === -1) {
+                    this.addMlineForNewSource(mediaType, true);
+                    idx = this.media.length - 1;
```
saghul, line 84: Why does the index need adjusting?

modules/xmpp/JingleSessionPC.js
```diff
@@ -1995,6 +1852,14 @@ export default class JingleSessionPC extends JingleSession {
                 logger.debug(`Existing SSRC re-mapped ${ssrc}: new owner=${owner}, source-name=${source}`);
 
                 this._signalingLayer.setSSRCOwner(ssrc, owner, source);
+                const oldSourceName = track.getSourceName();
+                const sourceInfo = this.peerconnection.getRemoteSourceInfoBySourceName(oldSourceName);
+
+                // Update the SSRC map on the peerconnection.
+                if (sourceInfo) {
+                    this.peerconnection._remoteSsrcMap.delete(oldSourceName);
```
saghul, line 1860: Having to dig in the private methods here looks nasty shall this have some API?

saghul (commented):
My review is 50% hope, 50% substance 😅

modules/RTC/TraceablePeerConnection.js
jallamsetty1: I have a follow up PR that will get rid of localSSRCs/remoteSSRCs and start using _localSSCMap and _remoteSSRCMap for all other helper functions, I have them initialized to an empty map there. I should have done that here itself :) Modified it.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: Yes its not, force of habit for some reason :)

modules/sdp/SDP.js
jallamsetty1: I started off by introducing a getter for raw but it is indeed being accessed many times in JIngleSessionPC resuting in unnecessary recomputations. I did run into another issue as well where the SDP was getting messed up.

modules/sdp/SDP.js
jallamsetty1: 1. For JVB, we are creating a new m-line always.
2. For P2P, we try to use an existing recvonly m-line (because of a secondary local video source) for adding the new remote source, create new one otherwise.
We have to update the relevant m-line so we adjust the m-line index accordingly.

modules/xmpp/JingleSessionPC.js
jallamsetty1: Oops, I added TPC.updateRemoteSources for this exact purpose later but forgot to update it here. Thanks for pointing it out.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -328,12 +326,26 @@ export default function TraceablePeerConnection(
...
-     * Holds the SSRC map for the local tracks.
+     * Holds the SSRC map for the local tracks mapped by their source names.
      *
-     * @type {Map<string, TPCSSRCInfo>}
+     * @type {Map<string, TPCSourceInfo>}
+     * @property {string} msid - The track's MSID.
+     * @property {Array<string>} ssrcs - The SSRCs associated with the track.
+     * @property {Array<TPCGroupInfo>} groups - The SSRC groups associated with the track.
      */
     this._localSsrcMap = null;
```
saghul, line 336: Any reason not to initialize this to a new map too?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -680,7 +692,7 @@ TraceablePeerConnection.prototype.getLocalVideoSSRCs = function(localTrack) {
         return ssrcs;
     }
 
-    const ssrcGroup = this.isSpatialScalabilityOn() ? 'SIM' : 'FID';
+    const ssrcGroup = this.isSpatialScalabilityOn() ? SSRC_GROUP_SEMANTICS.SIM : SSRC_GROUP_SEMANTICS.FID;
```
saghul, line 695: ❤️

modules/RTC/TraceablePeerConnection.js
```diff
@@ -769,45 +781,35 @@ TraceablePeerConnection.prototype.getRemoteTracks = function(endpointId, mediaTy
 };
 
 /**
- * Parses the remote description and returns the sdp lines of the sources associated with a remote participant.
+ * Returns the remote sourceInfo for a given source name.
+ *
+ * @param {string} sourceName - The source name.
+ * @returns {TPCSourceInfo}
```
saghul, line 787: It can also return null, but I think undefined is also good, see below.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -769,45 +781,35 @@ TraceablePeerConnection.prototype.getRemoteTracks = function(endpointId, mediaTy
...
+ */
+TraceablePeerConnection.prototype.getRemoteSourceInfoBySourceName = function(sourceName) {
+    return cloneDeep(this._remoteSsrcMap?.get(sourceName) ?? null);
```
saghul, line 790:
```suggestion
    return cloneDeep(this._remoteSsrcMap.get(sourceName));
```

saghul: Now that the map always exists there is no need to optional-chain the call. Also, any reason to prefer a null result vs undefined? The map will return undefined if the key is not in there, which should suffice and simplifies the code a bit?

modules/sdp/SDP.js
saghul: Isn't it mostly being accessed _after_ being computed? In that case we'd have the same amount of computations, but without the risk of missing updating raw when we updated something else.

Perhaps we can look into it after this PR.

modules/sdp/SDP.js
```diff
@@ -136,6 +200,96 @@ export default class SDP {
...
+        const newSession = transform.parse(this.raw);
+        const newMedia = [];
+
+        newSession.media.forEach(mLine => {
+            const type = mLine.type;
+
+            if (type === MediaType.APPLICATION) {
+                const newMline = cloneDeep(mLine);
+
+                newMline.mid = newMedia.length.toString();
```
saghul, line 219: Is this what we really want? The lengfth as a string? Not its index?

modules/sdp/SDP.js
```diff
@@ -136,6 +200,96 @@ export default class SDP {
...
+                const newMline = cloneDeep(mLine);
+
+                newMline.ssrcs = [];
+                newMline.ssrcGroups = [];
+                newMline.mid = newMedia.length.toString();
+                newMline.bundleOnly = undefined;
+                newMline.direction = idx ? 'sendonly' : 'sendrecv';
+
+                // Add the sources and the related FID source group to the new m-line.
+                const group = mLine.ssrcGroups?.find(g => g.ssrcs.includes(ssrc.id.toString()));
```
saghul, line 248: You are converting the ssrc id into a string many times here, calculate it once: `const idStr = ssrc.id.toString();` and then compare with it.

saghul (commented):
A couple of comments and LGTM!

modules/sdp/SDP.js
jallamsetty1: No, there are instances where it is simply passed to sdp-transofrm for parsing and also for renegotiations. I agree that optimizing this could be a separate PR.

modules/sdp/SDP.js
jallamsetty1: Yes, that is the requirement. The `mid` for the new m-line needs to be a string.

modules/RTC/TraceablePeerConnection.js
jallamsetty1: I have some more work planned around this, i.e., getting rid of localSSRCs and remoteSSRCs in favor of these. Lets restrict this PR to only the SDP conversion, WDYT?

saghul: It just looks odd that one is a map and the other one is null. If it's not a huge change it would be nice to at least have them aligned. No worries if it would break things, we can do it later.

modules/sdp/SDP.js
saghul: I mean for the MID to the the length. I guess it's the index for the last added element, right?

jallamsetty1: Yes.

PR #2594, opened 2024-11-05
fix(breakout): Clear up request sent on authentication.

damencho:
If guest authenticates in meeting to become a moderator and creates a breakout room, was not inviting jicofo and fails to join breakout room. Fixes jitsi/jitsi-meet#15024.


saghul (approved):
This looks a bit backwards. Why can't LJM reset it on its own?

damencho:
This is in ljm.

PR #2599, opened 2024-11-18
ref(TPC) Move all the utility functions to TPCUtils.

jallamsetty1:
Move all utility functions to TPCUtils that include
1. munging the SDP.
2. getting current codecs
3. getting expected transceiver direction.
4. getting expected settings for video encodings.

Move all functions that call RTCPeerConnection APIs directly back to TPC.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -470,10 +470,10 @@ export class TPCUtils {
             mLine.ssrcs = reorderedSsrcs;
         });
 
-        return new RTCSessionDescription({
+        return {
             type: description.type,
             sdp: transform.write(parsedSdp)
-        });
+        };
     }
 
     /**
@@ -626,10 +626,10 @@ export class TPCUtils {
             }
         });
 
-        return new RTCSessionDescription({
+        return {
             type: desc.type,
             sdp: transform.write(sdp)
-        });
+        };
     }
 
     /**
@@ -703,10 +703,10 @@ export class TPCUtils {
             }
         }
 
-        return new RTCSessionDescription({
+        return {
             type: description.type,
             sdp: transform.write(parsedSdp)
-        });
+        };
     }
 
     /**
@@ -776,10 +776,10 @@ export class TPCUtils {
             fmtpOpus.config = mungedConfig.trim();
         }
 
-        return new RTCSessionDescription({
+        return {
             type: description.type,
             sdp: transform.write(parsedSdp)
-        });
+        };
     }
 
     /**
@@ -842,10 +842,10 @@ export class TPCUtils {
             }
         }
 
-        return new RTCSessionDescription({
+        return {
             type: description.type,
             sdp: transform.write(parsedSdp)
-        });
+        };
     }
 
     /**
@@ -892,9 +892,9 @@ export class TPCUtils {
             }
         });
 
-        return new RTCSessionDescription({
+        return {
             type: description.type,
             sdp: transform.write(parsedSdp)
-        });
+        };
     }
 }
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -5,7 +5,6 @@ import transform from 'sdp-transform';
 import { CodecMimeType } from '../../service/RTC/CodecMimeType';
 import { MediaDirection } from '../../service/RTC/MediaDirection';
 import { MediaType } from '../../service/RTC/MediaType';
-import { getSourceIndexFromSourceName } from '../../service/RTC/SignalingLayer';
 import {
     SIM_LAYERS,
     STANDARD_CODEC_SETTINGS,
@@ -15,13 +14,17 @@ import {
 import { VideoEncoderScalabilityMode } from '../../service/RTC/VideoEncoderScalabilityMode';
 import { VideoType } from '../../service/RTC/VideoType';
 import browser from '../browser';
+import SDPUtil from '../sdp/SDPUtil';
 
 const logger = getLogger(__filename);
+const DD_HEADER_EXT_URI
+    = 'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension';
+const DD_HEADER_EXT_ID = 11;
 const VIDEO_CODECS = [ CodecMimeType.AV1, CodecMimeType.H264, CodecMimeType.VP8, CodecMimeType.VP9 ];
 
 /**
- * Handles track related operations on TraceablePeerConnection when browser is
- * running in unified plan mode.
+ * Handles all the utility functions for the TraceablePeerConnection class, like calculating the encoding parameters,
+ * determining the media direction, calculating bitrates based on the current codec settings, etc.
  */
 export class TPCUtils {
     /**
@@ -29,10 +32,22 @@ export class TPCUtils {
      *
      * @param peerconnection - the tpc instance for which we have utility functions.
      */
-    constructor(peerconnection) {
+    constructor(peerconnection, options) {
         this.pc = peerconnection;
+        this.options = options;
         this.codecSettings = cloneDeep(STANDARD_CODEC_SETTINGS);
-        const videoQualitySettings = this.pc.options?.videoQuality;
+
+        /**
+         * Flag indicating bridge support for AV1 codec. On the bridge connection, it is supported only when support for
+         * Dependency Descriptor header extensions is offered by Jicofo. H.264 simulcast is also possible when these
+         * header extensions are negotiated.
+         */
+        this.supportsDDHeaderExt = false;
+
+        /**
+         * Reads videoQuality settings from config.js and overrides the code defaults for video codecs.
+         */
+        const videoQualitySettings = this.options.videoQuality;
 
         if (videoQualitySettings) {
             for (const codec of VIDEO_CODECS) {
@@ -86,7 +101,7 @@ export class TPCUtils {
         const codecBitrates = this.codecSettings[codec].maxBitratesVideo;
         const trackCaptureHeight = localVideoTrack.getCaptureResolution();
         const effectiveNewHeight = newHeight > trackCaptureHeight ? trackCaptureHeight : newHeight;
-        const desktopShareBitrate = this.pc.options?.videoQuality?.desktopbitrate || codecBitrates.ssHigh;
+        const desktopShareBitrate = this.options.videoQuality?.desktopbitrate || codecBitrates.ssHigh;
         const isScreenshare = localVideoTrack.getVideoType() === VideoType.DESKTOP;
         let scalabilityMode = this.codecSettings[codec].useKSVC
             ? VideoEncoderScalabilityMode.L3T3_KEY : VideoEncoderScalabilityMode.L3T3;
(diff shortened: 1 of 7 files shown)


modules/RTC/TPCUtils.js
```diff
@@ -15,24 +14,40 @@ import {
...
+ * determining the media direction, calculating bitrates based on the current codec settings, etc.
  */
 export class TPCUtils {
     /**
      * Creates a new instance for a given TraceablePeerConnection
      *
      * @param peerconnection - the tpc instance for which we have utility functions.
      */
-    constructor(peerconnection) {
+    constructor(peerconnection, options) {
```
saghul, line 35: Make options default to `{}`, it will help you later.

saghul: Also document it above.

modules/RTC/TPCUtils.js
```diff
@@ -588,6 +476,100 @@ export class TPCUtils {
...
+        const rtpSender = this.pc.findSenderForTrack(localVideoTrack.getTrack());
+
+        if (this.pc.usesCodecSelectionAPI() && rtpSender) {
+            const { codecs } = rtpSender.getParameters();
+
+            return codecs[0].mimeType.split('/')[1].toLowerCase();
+        }
+
+        const sdp = this.pc.remoteDescription?.sdp;
+        const defaultCodec = CodecMimeType.VP8;
```
saghul, line 496: Constant outside of here?

modules/RTC/TPCUtils.js
```diff
@@ -588,6 +476,100 @@ export class TPCUtils {
...
+     * @returns {CodecMimeType} The codec that is set as the preferred codec for the given local video track.
+     */
+    getConfiguredVideoCodec(localTrack) {
+        const localVideoTrack = localTrack ?? this.pc.getLocalVideoTracks()[0];
+        const rtpSender = this.pc.findSenderForTrack(localVideoTrack.getTrack());
+
+        if (this.pc.usesCodecSelectionAPI() && rtpSender) {
+            const { codecs } = rtpSender.getParameters();
+
+            return codecs[0].mimeType.split('/')[1].toLowerCase();
```
saghul, line 492: Is there a chance the codec list is empty?

modules/RTC/TPCUtils.js
```diff
@@ -588,6 +476,100 @@ export class TPCUtils {
...
+            return defaultCodec;
+        }
+        const parsedSdp = transform.parse(sdp);
+        const mLine = parsedSdp.media
+            .find(m => m.mid.toString() === this.pc.localTrackTransceiverMids.get(localVideoTrack.rtcId));
+        const payload = mLine.payloads.split(' ')[0];
+        const { codec } = mLine.rtp.find(rtp => rtp.payload === Number(payload));
+
+        if (codec) {
+            return Object.values(CodecMimeType).find(value => value === codec.toLowerCase());
```
saghul, line 508: Is there a chance we don't find it?

modules/RTC/TPCUtils.js
```diff
@@ -671,122 +653,248 @@ export class TPCUtils {
...
         }
-        if (!transceiver) {
-            return Promise.reject(
-                new Error(`Replace track failed - no transceiver for old: ${oldTrack}, new: ${newTrack}`));
-        }
-        logger.debug(`${this.pc} Replacing ${oldTrack} with ${newTrack}`);
 
-        return transceiver.sender.replaceTrack(track)
-            .then(() => Promise.resolve(transceiver));
+        return new RTCSessionDescription({
```
saghul, line 706: The RTCSessionDescription constructor is deprecated and it's encouraged to use bare objects: https://developer.mozilla.org/en-US/docs/Web/API/RTCSessionDescription/RTCSessionDescription

modules/RTC/TPCUtils.js
```diff
@@ -671,122 +653,248 @@ export class TPCUtils {
...
+            let mungedConfig = '';
+
+            for (const key of Object.keys(fmtpConfig)) {
+                mungedConfig += `${key}=${fmtpConfig[key]}; `;
+            }
+
+            fmtpOpus.config = mungedConfig.trim();
+        }
+
+        return new RTCSessionDescription({
```
saghul, line 779: Ditto.

modules/RTC/TPCUtils.js
```diff
@@ -671,122 +653,248 @@ export class TPCUtils {
...
+                mLine.bandwidth = undefined;
             }
         }
 
-        return Promise.allSettled(promises)
-            .then(settledResult => {
-                const errors = settledResult
-                    .filter(result => result.status === 'rejected')
-                    .map(result => result.reason);
+        return new RTCSessionDescription({
```
saghul, line 845: Ditto.

modules/RTC/TPCUtils.js
```diff
@@ -671,122 +653,248 @@ export class TPCUtils {
...
+                        value: DD_HEADER_EXT_ID,
+                        uri: DD_HEADER_EXT_URI
+                    });
+                } else if (!shouldNegotiateHeaderExts && headerIndex >= 0) {
+                    mLine.ext.splice(headerIndex, 1);
+                }
+            }
+        });
+
+        return new RTCSessionDescription({
```
saghul, line 895: Ditto.

modules/RTC/TPCUtils.spec.js
```diff
@@ -35,7 +35,7 @@ describe('TPCUtils', () => {
 
         it('sort ssrcs associated with all FID ssrc-groups', () => {
             const pc = new MockPeerConnection();
-            const tpcUtils = new TPCUtils(pc);
+            const tpcUtils = new TPCUtils(pc, { });
```
saghul, line 38: If you use a default, we don't need to pass the weird empty object.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -1466,30 +1333,54 @@ TraceablePeerConnection.prototype._updateAv1DdHeaders = function(description) {
...
+            sendEncodings: []
+        };
+
+        if (!browser.isFirefox()) {
+            transceiverInit.sendEncodings = this.tpcUtils.getStreamEncodings(track);
+        }
+
+        try {
+            transceiver = this.peerconnection.addTransceiver(mediaStreamTrack, transceiverInit);
+        } catch (error) {
```
saghul, line 1365: Suggestion: make this function async so you don't need to do this, execptions will reject the promsie already.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2144,17 +1906,61 @@ TraceablePeerConnection.prototype.calculateExpectedSendResolution = function(loc
...
+    const transceiver = localTrack?.track && localTrack.getOriginalStream()
+        ? this.peerconnection.getTransceivers().find(t => t.sender?.track?.id === localTrack.getTrackId())
+        : this.peerconnection.getTransceivers().find(t => t.receiver?.track?.kind === mediaType);
+    const parameters = transceiver?.sender?.getParameters();
+
+    // Resolve if the encodings are not available yet. This happens immediately after the track is added to the
+    // peerconnection on chrome in unified-plan. It is ok to ignore and not report the error here since the
+    // action that triggers 'addTrack' (like unmute) will also configure the encodings and set bitrates after that.
+    if (!parameters?.encodings?.length) {
+        return Promise.resolve();
```
saghul, line 1937: Ditto, why not make it async?

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2182,20 +1988,50 @@ TraceablePeerConnection.prototype.configureVideoSenderEncodings = function(local
...
+ * and update the DD Header extensions for AV1.
+ *
+ * @param {RTCSessionDescription} description - The description to be munged.
+ * @returns {RTCSessionDescription} - The munged description.
+ */
+TraceablePeerConnection.prototype._mungeDescription = function(description) {
+    let mungedDescription = description;
+
+    this.trace('RTCSessionDescription::preTransform', dumpSDP(description));
+    mungedDescription = this.tpcUtils.mungeOpus(description);
```
saghul, line 2016: Each of these operations re-parses the SDP. Since we are refactoring, doesn't it make sense to parse it in this functiom, and pass the transform object along, so each of the tpcUtils functions can apply transformations, and then after all of them, re-generate the SDP?

saghul (commented):
Left some comments!

modules/RTC/TPCUtils.js
```diff
@@ -15,24 +14,40 @@ import {
...
 const logger = getLogger(__filename);
+const DD_HEADER_EXT_URI
+    = 'https://aomediacodec.github.io/av1-rtp-spec/#dependency-descriptor-rtp-header-extension';
+const DD_HEADER_EXT_ID = 11;
 const VIDEO_CODECS = [ CodecMimeType.AV1, CodecMimeType.H264, CodecMimeType.VP8, CodecMimeType.VP9 ];
 
 /**
- * Handles track related operations on TraceablePeerConnection when browser is
- * running in unified plan mode.
+ * Handles all the utility functions for the TraceablePeerConnection class, like calculating the encoding parameters,
```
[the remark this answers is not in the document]
jallamsetty1: Yes that is the intention. I see that the methods in TPC fall into one of the 3 categories below.
1. Methods that call RTCPeerConnection APIs directly or handle events fired on RTCPeerConnection.
2. Methods that operate on the SDPs (munging), get current codecs, calc encoding params, etc.
3. Methods that add/remove sources to the RTCPeerConnection, keeps their state updated and configures the audio and video sources.
These three can be encapsulated inside a JitsiPeerConnection class where classes for #1 and #2 are made private and only #3 can be exposed to JingleSession and statistics.

modules/RTC/TPCUtils.js
jallamsetty1: Do you know what version supports the bare objects? Since we need to check it across all browsers and add version checks, I suggest leaving it out of this one and doing it as a follow up PR. WDYT?

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
saghul: Sounds like a plan!

modules/RTC/TPCUtils.js
saghul: 👍

modules/RTC/TPCUtils.js
saghul: All of them support bare object and have so for years AFAIK. Only the constructor was deprecated.

modules/RTC/TPCUtils.js
```diff
@@ -574,6 +575,57 @@ export class TPCUtils {
...
     }
 
+    /**
+     * Injects a 'SIM' ssrc-group line for simulcast into the given session description object to make Jicofo happy.
+     * This is needed only for Firefox since it does not generate it when simulcast is enabled.
+     *
+     * @param desc A session description object (with 'type' and 'sdp' fields)
+     * @return A session description object with its sdp field modified to contain an inject ssrc-group for simulcast.
+     */
+    injectSsrcGroupForUnifiedSimulcast(desc) {
```
saghul, line 585: Why does this have "unified" in the name now?

modules/RTC/TPCUtils.js
```diff
@@ -574,6 +575,57 @@ export class TPCUtils {
         } ];
     }
 
+    /**
+     * Injects a 'SIM' ssrc-group line for simulcast into the given session description object to make Jicofo happy.
+     * This is needed only for Firefox since it does not generate it when simulcast is enabled.
```
saghul, line 580: We call this regardless of the browser, is it really only required for FF?

modules/RTC/TPCUtils.js
```diff
@@ -512,17 +513,17 @@ export class TPCUtils {
...
     /**
      * Returns the codecs in the current order of preference as configured on the peerconnection.
      *
-     * @param {RTCSessionDescription} - The local description to be used.
+     * @param {string} - The local SDP to be used.
      * @returns {Array}
      */
-    getConfiguredVideoCodecs(description) {
-        const currentSdp = description?.sdp ?? this.pc.localDescription?.sdp;
+    getConfiguredVideoCodecs(sdp) {
```
saghul, line 525: I suggest you break this function into 2:

```
getConfiguredVideoCodecs(sdp) {
}

_getConfiguredVideoCodecsImpl(parsedSdp) {
}
```

With the former calling the latter, after parsing the SDP.

This will allow you to call the 2nd one in the munge code and doing it all in 1 pass!

modules/RTC/TPCUtils.spec.js
saghul: You can remove this one now.

modules/RTC/TraceablePeerConnection.js
```diff
@@ -2182,20 +1926,55 @@ TraceablePeerConnection.prototype.configureVideoSenderEncodings = function(local
...
+/**
+ * Munges the provided description to update the codec order, set the max bitrates (for VP9 K-SVC), set stereo flag
+ * and update the DD Header extensions for AV1.
+ *
+ * @param {RTCSessionDescription} description - The description to be munged.
+ * @returns {RTCSessionDescription} - The munged description.
+ */
+TraceablePeerConnection.prototype._mungeDescription = function(description) {
+    this.trace('RTCSessionDescription::preTransform', dumpSDP(description));
+    let mungedSdp = transform.parse(description.sdp);
```
saghul, line 1952: Why are we not using SdpTransformWrap here?

saghul (approved):
🚀

saghul:
Pl next time use squash-merge, you can keep the individual commit messages if you want, but this was 1 logical change.

PR #2602, opened 2024-11-28
Improve Retry Mechanism and Listener Handling in ResumeTask

arslanarso:
# Improve Retry Mechanism and Listener Handling in `ResumeTask`

This PR introduces two key improvements to the `ResumeTask` implementation:

### 1. **Move `this._resumeRetryN += 1` to `_scheduleResume`:**
- The retry counter is now incremented only when a resume task is actually scheduled.
- This ensures that the retry count accurately reflects the number of retry attempts, avoiding unnecessary increments when `_scheduleResume` is not invoked.

### 2. **Remove unnecessary call to `_scheduleResume` in `schedule`:**
- Previously, `_scheduleResume` was invoked in `schedule` regardless of the network status, potentially causing attempts to reconnect even when the network was offline.
- This line has been removed to ensure that reconnect attempts are driven solely by the network listener, making the logic more consistent and reliable.

---

These changes improve the accuracy and robustness of the retry mechanism and eliminate potential conflicts in listener handling.

### DIFF
--- a/modules/xmpp/ResumeTask.js
+++ b/modules/xmpp/ResumeTask.js
@@ -56,7 +56,6 @@ export default class ResumeTask {
         this._cancelResume();
         this._removeNetworkOnlineListener();
 
-        this._resumeRetryN += 1;
 
         this._networkOnlineListener
             = NetworkInfo.addCancellableListener(
@@ -82,6 +81,8 @@ export default class ResumeTask {
             // NO-OP
             return;
         }
+
+        this._resumeRetryN += 1;
 
         // The retry delay will be:
         //   1st retry: 1.5s - 3s


saghul:
Both of the changes break the logic in subtle ways.

> ### 1. **Move `this._resumeRetryN += 1` to `_scheduleResume`:**
> 
>     * The retry counter is now incremented only when a resume task is actually scheduled.
> 
>     * This ensures that the retry count accurately reflects the number of retry attempts, avoiding unnecessary increments when `_scheduleResume` is not invoked.

In order for the retry count check to work, the counter needs to be accurate when we call `schedule()`, see how it's used in `XmppConnection.js`. The point of the counter is to avoid scheduling more than X times.

> ### 2. **Remove unnecessary call to `_scheduleResume` in `schedule`:**
> 
>     * Previously, `_scheduleResume` was invoked in `schedule` regardless of the network status, potentially causing attempts to reconnect even when the network was offline.
> 
>     * This line has been removed to ensure that reconnect attempts are driven solely by the network listener, making the logic more consistent and reliable.

It's not unnecessary. There are network glitches that will cause the connection to break while the network state doesn't transition to offline, so we cannot assume `schedule` was called in an offline state.

arslanarso:
> > ### 2. **Remove unnecessary call to `_scheduleResume` in `schedule`:**
> > ```
> > * Previously, `_scheduleResume` was invoked in `schedule` regardless of the network status, potentially causing attempts to reconnect even when the network was offline.
> > 
> > * This line has been removed to ensure that reconnect attempts are driven solely by the network listener, making the logic more consistent and reliable.
> > ```
> 
> It's not unnecessary. There are network glitches that will cause the connection to break while the network state doesn't transition to offline, so we cannot assume `schedule` was called in an offline state.




I observed that with this line NetworkInfo.isOnline() && this._scheduleResume();, the _scheduleResume function is triggered even when the internet is disconnected. Normally, it should not trigger the _scheduleResume function immediately when the internet is lost.

What solution should be applied here, in your opinion?

saghul:
I think you are looking at this the wrong way. The real question is why does `NetworkInfo.isOnline() ` return `true` if you are not online?

At the end of the day, it doesn't matter much. If your internet is down, but we retry, the connection will fail, and when the retry-count is exhausted, the reload mechanism will kick in. Otherwise you are going to be left in the meeting with no audio / video and wondering what's going on.

What exactly are you trying to fix here?

PR #2605, opened 2024-12-06
Allow Electron to use getDisplayMedia

### LINKED ISSUE #2597: Electron ScreenSharing Audio Echo
## Description
---

Related to: https://github.com/jitsi/lib-jitsi-meet/blob/bc446e999e4128031f265276e60f9a3808150760/modules/RTC/ScreenObtainer.js#L95

This is more of a discussion, but Electron now has support for **getDisplayMedia** (Bye bye programmatic screen selection 🥹 ), and it seems **getUserMedia** is no longer recommended for Screen Capture.

One downside of **getUserMedia** is that echo cancellation doesn't get applied (See: https://github.com/electron/electron/issues/27337), however its quite nice for programmatic selection of screen sharing, **getDisplayMedia** on Electron requires setting **setDisplayMediaRequestHandler** (See: https://www.electronjs.org/docs/latest/api/desktop-capturer), but it means we can make use of the new constraints syntax (min, max, ideal, exact).

However because of potential impact on 3rd party clients that use Screen Sharing, the way you interact and share your screen will differ (No programmatic request, as it requires user interaction).

So is it worth supporting getDisplayMedia in Electron, using some kind of flag?

## Current behavior
---
Audio echo when presenting a screen with audio in Electron

## Expected Behavior
---
No audio echo when presenting a screen with audio in Electron

## Possible Solution
---
Use **getDisplayMedia** instead of **getUserMedia**

## Steps to reproduce
---
Start a meeting in Electron, share desktop with audio, participants can hear themselves.

# Environment details
---
Electron v31, latest lib-jitsi-meet, custom client.

DanielMcAssey:
This PR allows Electron to use getDisplayMedia, its hidden behind a flag `electronUseGetDisplayMedia`.

If there's a better place for the flag, or naming convention, do let me know.

Later on I assume you will be deprecating the old method, and it'll be a flag to use the old method, before removing it entirely?

Closes #2597

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -46,11 +46,13 @@ const ScreenObtainer = {
      * @private
      */
     _createObtainStreamMethod() {
-        if (browser.isElectron()) {
+        const supportsGetDisplayMedia = browser.supportsGetDisplayMedia();
+
+        if (browser.isElectron() && !this.options.electronUseGetDisplayMedia) {
             return this.obtainScreenOnElectron;
-        } else if (browser.isReactNative() && browser.supportsGetDisplayMedia()) {
+        } else if (browser.isReactNative() && supportsGetDisplayMedia) {
             return this.obtainScreenFromGetDisplayMediaRN;
-        } else if (browser.supportsGetDisplayMedia()) {
+        } else if (supportsGetDisplayMedia) {
             return this.obtainScreenFromGetDisplayMedia;
         }
         logger.log('Screen sharing not supported on ', browser.getName());


saghul (approved):
LGTM. The bulk of the work needs to happen in the Electron SDK since we still beed to integrate with the picker.

jallamsetty1 (approved):

DanielMcAssey:
Just bumping this, if possible to get it merged :)

saghul:
Sorry about the delay, done!

DanielMcAssey:
Thank you!

PR #2609, opened 2024-12-12
feat(RTC) drop PERMISSION_PROMPT_IS_SHOWN event

saghul:
It was never accurate since browsers have changed their behavior over time and can be implemented in the application more reliably.


damencho (approved):

saghul:
This needs to land first: https://github.com/jitsi/jitsi-meet/pull/15370

jallamsetty1:
I guess we can get rid of this one as well - https://github.com/jitsi/lib-jitsi-meet/blob/master/JitsiMeetJS.ts#L74

jallamsetty1 (approved):

damencho (approved):

PR #2615, opened 2025-01-13
feat(subject): Avoids setting subject if already set.

damencho:


jallamsetty1 (approved):

modules/xmpp/ChatRoom.js
```diff
@@ -1014,6 +1014,12 @@ export default class ChatRoom extends Listenable {
      * @param subject
      */
     setSubject(subject) {
+        if ((subject || subject === '')
```
hristoterezov, line 1017: I'm wondering shouldn't we handle the cases when subject is `undefined`?

hristoterezov (changes requested):
Left a comment!

hristoterezov (approved):

PR #2616, opened 2025-01-15
Address some inconsistencies related to configs.js

hristoterezov:


hristoterezov:
Related to https://github.com/jitsi/jitsi-meet/pull/15490

PR #2619, opened 2025-01-21
fix(JPC) skip processing remote audio sources when startSilent

saghul:
Avoids having any actual remote audio tracks, which can mess with the mobile audio session.

### DIFF
--- a/modules/xmpp/JingleSessionPC.js
+++ b/modules/xmpp/JingleSessionPC.js
@@ -553,6 +553,13 @@ export default class JingleSessionPC extends JingleSession {
 
             for (const description of descriptionsWithSources) {
                 const mediaType = $(description).attr('media');
+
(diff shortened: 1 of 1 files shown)


modules/xmpp/JingleSessionPC.js
```diff
@@ -1835,6 +1842,12 @@ export default class JingleSessionPC extends JingleSession {
         if (!FeatureFlags.isSsrcRewritingSupported()) {
             return;
         }
+
+        if (mediaType === MediaType.AUDIO && this.options.startSilent) {
+
+            return;
```
paweldomas, line 1848: is the session going to eventually catch up with the skipped sources? or does startSilent mean no audio for the entire session?

saghul: The latter. A full session restart is necessary for getting the missing sources.

PR #2622, opened 2025-01-24
fix(codec) Debounce the call that calc codec intersection set.

jallamsetty1:
Calculate codec intersection set only once per second even when there is a burst of joins/leaves. Also, check for current codec before chaining a codec change operation when codec selection API is used.

### DIFF
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -512,10 +512,10 @@ export class TPCUtils {
         const rtpSender = this.pc.findSenderForTrack(localVideoTrack.getTrack());
 
         if (this.pc.usesCodecSelectionAPI() && rtpSender) {
-            const { codecs } = rtpSender.getParameters();
+            const { encodings } = rtpSender.getParameters();
 
-            if (codecs?.length) {
-                return codecs[0].mimeType.split('/')[1].toLowerCase();
+            if (encodings[0].codec) {
+                return encodings[0].codec.mimeType.split('/')[1].toLowerCase();
             }
         }
 
--- a/modules/qualitycontrol/CodecSelection.spec.js
+++ b/modules/qualitycontrol/CodecSelection.spec.js
@@ -50,43 +50,57 @@ describe('Codec Selection', () => {
             };
 
             qualityController = new QualityController(conference, options);
+            jasmine.clock().install();
             spyOn(jingleSession, 'setVideoCodecs');
         });
 
-        it('and remote endpoints use the new codec selection logic', () => {
+        afterEach(() => {
+            jasmine.clock().uninstall();
+        });
+
+        it('and remote endpoints use the new codec selection logic', async () => {
             // Add a second user joining the call.
             participant1 = new MockParticipant('remote-1');
             conference.addParticipant(participant1, [ 'vp9', 'vp8' ]);
 
+            await nextTick(1000);
+
             expect(jingleSession.setVideoCodecs).toHaveBeenCalledTimes(1);
+            expect(jingleSession.setVideoCodecs).toHaveBeenCalledWith([ 'vp9', 'vp8' ], 'vp9');
 
             // Add a third user joining the call with a subset of codecs.
             participant2 = new MockParticipant('remote-2');
             conference.addParticipant(participant2, [ 'vp8' ]);
 
-            expect(jingleSession.setVideoCodecs).toHaveBeenCalledWith([ 'vp8' ], 'vp9');
-
-            // Make p2 leave the call
+            // Make p2 leave the call.
             conference.removeParticipant(participant2);
-            expect(jingleSession.setVideoCodecs).toHaveBeenCalledTimes(3);
+
+            await nextTick(1000);
+
+            expect(jingleSession.setVideoCodecs).toHaveBeenCalledTimes(2);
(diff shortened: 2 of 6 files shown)


modules/qualitycontrol/QualityController.ts
```diff
@@ -159,6 +161,26 @@ export class QualityController {
             (tpc: TraceablePeerConnection, stats: Map<number, IOutboundRtpStats>) => this._processOutboundRtpStats(tpc, stats));
     }
 
+    /**
+     * Creates a debounced function that delays the execution of the provided function until after the specified delay
```
saghul, line 165: Isn't this basically lodash's throtlle? https://lodash.com/docs/4.17.15#throttle

modules/qualitycontrol/QualityController.ts
```diff
@@ -159,6 +161,26 @@ export class QualityController {
...
+     * @param {Function} func - The function to be debounced.
+     * @param {number} delay - The delay in milliseconds.
+     * @returns {Function} - The debounced function.
+     */
+    _debounce(func: Function, delay: number) {
+        return function (...args) {
+            if (!this._timer) {
+                this._timer = setTimeout(() => {
+                    func.apply(this, args);
+                    this._timer = null;
```
saghul, line 178: If we end up keeing this, I'd reset the timer first, or it will get stuck if the function throws an exception.

modules/qualitycontrol/QualityController.ts
saghul: Wait, I see you imported it but then didn't use it?

modules/qualitycontrol/CodecSelection.spec.js
```diff
@@ -239,11 +270,14 @@ describe('Codec Selection', () => {
...
             conference.addParticipant(participant1, [ 'av1', 'vp9', 'vp8' ]);
+
+            await nextTick(1000);
+            expect(jingleSession.setVideoCodecs).toHaveBeenCalledTimes(1);
             expect(jingleSession.setVideoCodecs).toHaveBeenCalledWith([ 'av1', 'vp9', 'vp8' ], undefined);
 
             participant2 = new MockParticipant('remote-2');
             conference.addParticipant(participant2, [ 'av1', 'vp9', 'vp8' ]);
-            expect(jingleSession.setVideoCodecs).toHaveBeenCalledWith([ 'av1', 'vp9', 'vp8' ], undefined);
+            expect(jingleSession.setVideoCodecs).toHaveBeenCalledTimes(1);
```
saghul, line 280: Don't we also want to know that the codec selection ended up as expected?

saghul (commented):
Left some comments, PTAL!

modules/RTC/TPCUtils.js
```diff
@@ -512,10 +512,10 @@ export class TPCUtils {
         const rtpSender = this.pc.findSenderForTrack(localVideoTrack.getTrack());
 
         if (this.pc.usesCodecSelectionAPI() && rtpSender) {
-            const { codecs } = rtpSender.getParameters();
+            const { encodings } = rtpSender.getParameters();
```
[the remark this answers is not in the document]
jallamsetty1: With [mixed-codec simulcast](https://issues.webrtc.org/issues/362277533) support, the new way of determining current codec will be accurate. Even otherwise, it is a better way since we started setting codec on the encodings now.

modules/qualitycontrol/QualityController.ts
jallamsetty1: Oh yes, I was testing with lodash throttle with both the leading and trailing options set to true and ran into this limitation - If leading and trailing options are true, func is invoked on the trailing edge of the timeout only if the throttled function is invoked more than once during the wait timeout. 
I can go back and test lodash throttle with only the trailing option set to true and I believe the end result will be the same. Will update this.

modules/qualitycontrol/CodecSelection.spec.js
jallamsetty1: I think I removed this unintentionally, will add it back.

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
saghul: Is that also supported in old(er) chromes? 😅

modules/qualitycontrol/CodecSelection.spec.js
saghul: 👍

modules/qualitycontrol/QualityController.ts
jallamsetty1: Actually when I pass `{ leading: false, trailing: true }` to debounce's throttle, my testcases fail, its missing some of the calls when we expect them.

saghul: We are waiting for 1000 and advancing by 1000 I bet there are shenanigans there? Perhaps we should test by waiting a bit longer?

modules/RTC/TPCUtils.js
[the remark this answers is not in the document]
jallamsetty1: Lol ... we have the feature check at line 514 and run this code only on Chrome 126 and newer.

modules/qualitycontrol/QualityController.ts
saghul: Ok let's have it this way 👍

saghul (approved):

PR #2626, opened 2025-01-27
feat: Updates userRegion from backend.

damencho:


JitsiConference.js
```diff
@@ -560,6 +560,8 @@ JitsiConference.prototype._init = function(options = {}) {
     // creates dominant speaker detection that works only in p2p mode
     this.p2pDominantSpeakerDetection = new P2PDominantSpeakerDetection(this);
 
+    // We keep this for some time, till all backends are updated to use
```
jallamsetty1, line 563: Should we rephrase this to so that it grabs attention?
TODO: Drop this after the change to use the region from the http requests to prosody is propagated to all the deployments in production.

jallamsetty1 (approved):

saghul (approved):

PR #2632, opened 2025-02-06
feat(ScreenObtainer) add fallback option for Electron

saghul:
If the enclosing Jitsi Meet app declares gDM in Electron is supported, ignore the `electronUseGetDisplayMedia` config flag and use the gDM flow. This facilitates having backwards compatibility since the Electron app won't know if the deployment it's about to join has the gDM flow support or not.

### DIFF
--- a/modules/RTC/ScreenObtainer.js
+++ b/modules/RTC/ScreenObtainer.js
@@ -298,24 +298,20 @@ const ScreenObtainer = {
                     errorStack: error?.stack
                 };
 
-                logger.error('getDisplayMedia error', JSON.stringify(constraints), JSON.stringify(errorDetails));
+                logger.warn('getDisplayMedia error', JSON.stringify(constraints), JSON.stringify(errorDetails));
 
                 if (errorDetails.errorMsg?.indexOf('denied by system') !== -1) {
                     // On Chrome this is the only thing different between error returned when user cancels
                     // and when no permission was given on the OS level.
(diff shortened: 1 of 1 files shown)


modules/RTC/ScreenObtainer.js
```diff
@@ -94,7 +94,12 @@ const ScreenObtainer = {
      * @param {Object} options - Optional parameters.
      */
     obtainScreenOnElectron(onSuccess, onFailure, options = {}) {
-        if (window.JitsiMeetScreenObtainer && window.JitsiMeetScreenObtainer.openDesktopPicker) {
+        if (typeof window.JitsiMeetScreenObtainer?.openDesktopPicker === 'function') {
+            // Detect if we have the fallback option.
+            if (window.JitsiMeetScreenObtainer?.gDMSupported) {
+                return this.obtainScreenFromGetDisplayMedia(onSuccess, onFailure);
```
jallamsetty1, line 100: I just remembered that we apply the min fps constraint on the stream returned by `obtainScreenFromGetDisplayMedia` so that the 0 Hz mode for SS doesn't kick in automatically for browser. Do you think we need to run that on Electron clients as well? I am not sure if that will cause any issues in Electron though.

jallamsetty1 (approved):

modules/RTC/ScreenObtainer.js
saghul: I'd expect not to cause issues since it will be the same code path than for chrome.

PR #2638, opened 2025-02-13
fix(codec) Fix AV1/VP9 support on Firefox.

jallamsetty1:
Firefox added support for AV1 with DD and VP9 SVC is also working as expected in Firefox 136.

### DIFF
--- a/modules/qualitycontrol/CodecSelection.js
+++ b/modules/qualitycontrol/CodecSelection.js
@@ -72,11 +72,10 @@ export class CodecSelection {
             }
 
             // Push VP9 to the end of the list so that the client continues to decode VP9 even if its not
-            // preferable to encode VP9 (because of browser bugs on the encoding side or added complexity on mobile
-            // devices). Currently, VP9 encode is supported on Chrome and on Safari (only for p2p).
+            // preferable to encode VP9 (because of browser bugs on the encoding side or other reasons).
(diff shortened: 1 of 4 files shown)


modules/qualitycontrol/CodecSelection.js
```diff
@@ -72,11 +72,10 @@ export class CodecSelection {
             }
 
             // Push VP9 to the end of the list so that the client continues to decode VP9 even if its not
-            // preferable to encode VP9 (because of browser bugs on the encoding side or added complexity on mobile
-            // devices). Currently, VP9 encode is supported on Chrome and on Safari (only for p2p).
+            // preferable to encode VP9 (because of browser bugs on the encoding side or other reasons).
             const isVp9EncodeSupported = browser.supportsVP9() || (browser.isWebKitBased() && connectionType === 'p2p');
 
-            if (!isVp9EncodeSupported || this.conference.isE2EEEnabled()) {
```
saghul, line 79: Doesn't the E2EE part still stand?

jallamsetty1: It gets enforced [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/qualitycontrol/CodecSelection.js#L161) when the codec is selected so the check here is redundant.

saghul (approved):

PR #2640, opened 2025-02-14
fixed selectEndpoints resets constraints properly

### LINKED ISSUE #2134: selectEndpoints with no selected endpoints should remove previously selected endpoint
## Description

When creating an application directly with the lib and on top jitsi-meet,
I should be able to remove all participants from the `selectedEndpoints`

## Current behavior
when calling `selectEndpoints` with an empty array, a previously selected participants is still prioritised

## Expected Behavior
when calling `selectEndpoints` with an empty array, no participants should be priorisited

## Possible Solution
the `selectEndpoints` method should remove check about !this._selectedEndpoints.length and the `constraints` getter  should return the default constraints instead of the previous one.

## Steps to reproduce
Given a running conference running with the lib,
 - call `selectParticipants` on the conference object with an array containing a single participant : participants are selected
 - then call `selectParticipants` with an empty array: Participant are still selected.

note:
I also do not understand why when multiple participants are selected, we don't seem to send the list of selected participants. see [here](https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/qualitycontrol/ReceiveVideoController.js#L68)

laky241:
Fixes #2134 

When calling selectEndpoints, previously selected participants were still present instead of being removed.

Modified setReceiverConstraints() in ReceiveVideoController.js
If selectedEndpoints is an empty array, it now resets the constraints properly instead of keeping old values.

**this._receiverVideoConstraints.selectedEndpoints = undefined;**

 This ensures that when selectEndpoints is called

It completely resets selectedEndpoints instead of keeping old data.

### DIFF
--- a/modules/qualitycontrol/ReceiveVideoController.js
+++ b/modules/qualitycontrol/ReceiveVideoController.js
@@ -219,6 +219,11 @@ export default class ReceiveVideoController {
             this._receiverVideoConstraints.lastN = this._lastN;
             this._receiveResolutionLimitedByCpu && this._updateIndividualConstraints();
 
+            if (constraints.selectedEndpoints && constraints.selectedEndpoints.length === 0) {
+                this._receiverVideoConstraints.selectedEndpoints = undefined;
+            }
+            this._rtc.setReceiverVideoConstraints(this._receiverVideoConstraints);
+
             // Send the contraints on the bridge channel.
             this._rtc.setReceiverVideoConstraints(this._receiverVideoConstraints);
 
--- a/modules/qualitycontrol/ReceiveVideoController.js
+++ b/modules/qualitycontrol/ReceiveVideoController.js
@@ -222,7 +222,6 @@ export default class ReceiveVideoController {
             if (constraints.selectedEndpoints && constraints.selectedEndpoints.length === 0) {
                 this._receiverVideoConstraints.selectedEndpoints = undefined;
             }
-            this._rtc.setReceiverVideoConstraints(this._receiverVideoConstraints);
 
             // Send the contraints on the bridge channel.
             this._rtc.setReceiverVideoConstraints(this._receiverVideoConstraints);


modules/qualitycontrol/ReceiveVideoController.js
```diff
@@ -219,6 +219,11 @@ export default class ReceiveVideoController {
             this._receiverVideoConstraints.lastN = this._lastN;
             this._receiveResolutionLimitedByCpu && this._updateIndividualConstraints();
 
+            if (constraints.selectedEndpoints && constraints.selectedEndpoints.length === 0) {
+                this._receiverVideoConstraints.selectedEndpoints = undefined;
+            }
+            this._rtc.setReceiverVideoConstraints(this._receiverVideoConstraints);
```
saghul, line 225: This line is duplicated below. You just need to call it once.

saghul (approved):
LGTM but I defer to @jallamsetty1 for the final approval.

jallamsetty1:
Hi @laky241 I am afraid you are using the older format that was deprecated long time back. Please check setReceiverConstraints under https://jitsi.github.io/handbook/docs/dev-guide/dev-guide-ljm-api. The property should now be `selectedSources`.

saghul:
But if the code is still there, shouldn't it behave correctly?

jallamsetty1:
> But if the code is still there, shouldn't it behave correctly?

You mean if they are testing it against a bridge that supports it? In that case, I don't know how they are running into this issue. The code does a deep compare and passes the sends the new constraints if they don't match the last set of constraints.

jallamsetty1:
> @jallamsetty1, I understand now that selectedEndpoints is deprecated and selectedSources should be used instead.
> 
> However, I’m unsure about the next step. Since the old code is still there, should I update it to selectedSources, or is this issue not relevant anymore?
> 
> Please let me know how you’d like me to proceed!

What do you mean by "Since the old code is still there" ? the code explicitly checks if you are using the older format that is not supported and throws an error - https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/qualitycontrol/ReceiveVideoController.js#L206

My recommendation is to update your client to use `selectedSources` and retest.

PR #2643, opened 2025-02-16
Not Having Formatted Code

naman9271:
## the codebase need to be formatted as it


saghul:
Similar answer to https://github.com/jitsi/lib-jitsi-meet/pull/2649#issuecomment-2662363187

We already have pretty comprehensive linting rules that help with consistent styling.

The current rules you are applying go against out chosen defaults, even in the bare minimum like using single quote strings vs double quoted strings.

Changinag that now adds nothing but churn.

naman9271:
ok @saghul i am sorry for my silly PRs .from next time i will raise some valuable PRs .... Thanks for feedback this matters a lot

saghul:
No worries, thank you for understanding!

PR #2652, opened 2025-02-19
fix(SDP) Inject trackID if missing from signaled source info.

jallamsetty1:
Reconstruct the msid to include both streamId and trackId if its missing from the source signaling. Fixes issue on Chromium < 117 where the remote offer gets rejected and the client gets kicked out of the call.

### DIFF
--- a/modules/sdp/SDP.js
+++ b/modules/sdp/SDP.js
@@ -95,8 +95,18 @@ export default class SDP {
             updatedMidIndices.push(idx);
 
             if (isAdd) {
+                let updatedMsid = msid;
+
+                // If the msid is not in the <stream ID> <track ID> format, create msid in that format.
+                // Chrome's older versions (upto 117) expect the msid to be in this format.
+                if (msid.split(' ').length !== 2
+                    && browser.isChromiumBased()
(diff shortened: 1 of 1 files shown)


hristoterezov (approved):

modules/sdp/SDP.js
```diff
@@ -95,8 +95,18 @@ export default class SDP {
             updatedMidIndices.push(idx);
 
             if (isAdd) {
+                let updatedMsid = msid;
+
+                // If the msid is not in the <stream ID> <track ID> format, create msid in that format.
+                // Chrome's older versions (upto 117) expect the msid to be in this format.
+                if (msid.split(' ').length !== 2
+                    && browser.isChromiumBased()
```
saghul, line 103: Why the browser check? Doesn't it make sense to do it always?

saghul (approved):
LGTM with a question.

modules/sdp/SDP.js
jallamsetty1: As a precaution. The newer versions go into a renegotiation loop when I use the same string for both streamId and trackId and that was the reason I didn't implement this fix the first time around.

Issue #2637, opened 2025-02-13
Migrating Files To Typescript


naman9271:
# Migrating  Files to Typescript

---

- [x] Yes, I am willing to submit a PR

naman9271:
Hey @jallamsetty1  @saghul  @bgrozev , Please review this issue and assign me so that i open a PR for this

damencho:
We do not assign people. Just open a PR and we can review it.

naman9271:
@Saghul, I have submitted the PRs for the migration of a single file as per your earlier suggestion. Additionally, I have migrated 5+ more files locally apart from these, but to maintain clarity and avoid clutter, I believe it would be best to have the initial PRs reviewed first. Please take a look when you have the time.

naman9271:
Hey @jallamsetty1, @saghul , I believe this issue may have been closed inadvertently, as there are still 10 open PRs related to it. Additionally, I have locally migrated more files. Could you kindly review the pending PRs? Thanks!

saghul:
We'll get through them, thank you for sending them!

naman9271:
"Thank you! @saghul I appreciate your time and effort in reviewing them. Let me know if any changes are needed."

PR #2654, opened 2025-02-20
Migrated `modules/RTC/BridgeChannel.ts`

naman9271:
### Fixes #2637 

**Migrated the `modules/RTC/BridgeChannel` from js to ts **

### DIFF
--- a/modules/RTC/BridgeChannel.ts
+++ b/modules/RTC/BridgeChannel.ts
@@ -27,7 +27,7 @@ export default class BridgeChannel {
     private _areRetriesEnabled: boolean = false;
     private _closedFromClient: boolean = false;
     private _wsUrl?: string;
-    private _retryTimeout?: ReturnType<typeof setTimeout>;
+    private _retryTimeout?: number;
 
     /**
      * Binds "ondatachannel" event listener on the given RTCPeerConnection
--- a/modules/RTC/BridgeChannel.ts
+++ b/modules/RTC/BridgeChannel.ts
@@ -27,7 +27,7 @@ export default class BridgeChannel {
     private _areRetriesEnabled: boolean = false;
     private _closedFromClient: boolean = false;
     private _wsUrl?: string;
-    private _retryTimeout?: number;
+    private _retryTimeout?: ReturnType<typeof setTimeout>;
 
     /**
      * Binds "ondatachannel" event listener on the given RTCPeerConnection
--- a/modules/RTC/BridgeChannel.ts
+++ b/modules/RTC/BridgeChannel.ts
@@ -1,9 +1,15 @@
 import { safeJsonParse } from '@jitsi/js-utils/json';
 import { getLogger } from '@jitsi/logger';
 
-import RTCEvents from '../../service/RTC/RTCEvents';
+import { EventEmitter } from 'events';
 import { createBridgeChannelClosedEvent } from '../../service/statistics/AnalyticsEvents';
+import { SourceName } from '../../service/RTC/SignalingLayer';
+import { BridgeVideoType } from '../../service/RTC/BridgeVideoType';
+import RTCEvents from '../../service/RTC/RTCEvents';
 import Statistics from '../statistics/statistics';
+import JitsiConference from '../../JitsiConference'
+import ReceiverVideoConstraints from "../qualitycontrol/ReceiveVideoController";
+
 
 const logger = getLogger(__filename);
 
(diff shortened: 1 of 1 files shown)


naman9271:
@saghul please review this

modules/RTC/BridgeChannel.ts
```diff
@@ -268,26 +285,26 @@ export default class BridgeChannel {
...
+        channel.onerror = (event:ErrorEvent):void => {
             // WS error events contain no information about the failure (this is available in the onclose event) and
             // the event references the WS object itself, which causes hangs on mobile.
             if (this._mode !== 'websocket') {
                 logger.error(`Channel error: ${event.message}`);
             }
         };
 
-        channel.onmessage = ({ data }) => {
+        channel.onmessage = ({ data }):void=> {
```
manzil-infinity180, line 307: missing type for data ?

naman9271: done missed by mistake

modules/RTC/BridgeChannel.ts
```diff
@@ -4,6 +4,12 @@ import { getLogger } from '@jitsi/logger';
 import RTCEvents from '../../service/RTC/RTCEvents';
 import { createBridgeChannelClosedEvent } from '../../service/statistics/AnalyticsEvents';
 import Statistics from '../statistics/statistics';
+import { EventEmitter } from "events";
```
saghul, line 7: Use single quotes and move this up so the imports are properly sorted.

modules/RTC/BridgeChannel.ts
```diff
@@ -12,6 +18,17 @@ const logger = getLogger(__filename);
...
+
+    private _channel: RTCDataChannel | WebSocket | null = null;
+    private _conference: JitsiConference;
+    private _connected: boolean | undefined = undefined;
+    private _eventEmitter: EventEmitter;
+    private _mode: "datachannel" | "websocket" | null = null;
+    private _areRetriesEnabled: boolean = false;
+    private _closedFromClient: boolean = false;
+    private _wsUrl?: string;
+    private _retryTimeout?: ReturnType<typeof setTimeout>;
```
saghul, line 30: Is this not TimerID?

modules/RTC/BridgeChannel.ts
```diff
@@ -86,7 +103,7 @@ export default class BridgeChannel {
      *
      * @returns {void}
      */
-    _initWebSocket() {
+    _initWebSocket():void {
```
saghul, line 106: Here and elsewhere, leave a space after the colon.

modules/RTC/BridgeChannel.ts
naman9271: Yes @saghul  , but However `TimerID` is not a standard type in typescript.
In browser env. it returns a `number` whereas in node env. it returns  a `NodeJS.Timeout`
so I write this type as `ReturnType<typeof setTimeout>` which means `number | NodeJS.Timeout`  which comes in standard types of typescript. 
- let me know if there is any need of any change in this.

naman9271:
@saghul I have fixed github actions and the changes you said please review

modules/RTC/BridgeChannel.ts
saghul: This doesn't work on Node, so we are fine not being compatible with it.

saghul (commented):

modules/RTC/BridgeChannel.ts
naman9271: Yes and after I put its type to number `_retryTimeout` giving error as it is not recognising the type thier as number as you can see in [CI Log](https://github.com/jitsi/lib-jitsi-meet/actions/runs/13920830651)  and TS couldn't find any Type as `Timeout` and for this i go through the its docs again and it returnn a Timeout object in NodeJs env and numeric ID (which is number) in browser env. and the possible bug i am thinking so far is TypeScript infers the return type of `SetTimeout` differently here (maybe as NodeJs env.) 

**So the possible fixes i am thinking so far are**
- put it type as `ReturnType<typeof setTimeout>`  **or**
- put it type as `number` and explicity cast the TimeoutID in its instances like `this._retryTimeout = setTimeout(reload, timeoutS * 1000) as unknown as number`

- Please tell is my thinking is right or not or what would be the more better approach for this

Thanks

saghul: Let's keep it as is then, thanks for giving it a try!

saghul (approved):

saghul:

PR #2660, opened 2025-02-25
Migrated  `e2eping/e2eping`

naman9271:
Fixes #2637

### DIFF
--- a/modules/e2eping/e2eping.ts
+++ b/modules/e2eping/e2eping.ts
@@ -279,13 +279,9 @@ export default class E2ePing {
         // Listen to E2E PING requests and responses from other participants
         // in the conference.
         if (payload.type === E2E_PING_REQUEST) {
-            if (payload.id !== undefined) {
-                this.handleRequest(participant.getId(), { id: payload.id });
-            }
+            this.handleRequest(participant.getId(), payload);
         } else if (payload.type === E2E_PING_RESPONSE) {
-            if (payload.id !== undefined) {
-                this.handleResponse(participant.getId(), { id: payload.id });
-            }
+            this.handleResponse(participant.getId(), payload);
         }
     }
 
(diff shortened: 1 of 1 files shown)


naman9271:
@saghul please review this PR

modules/e2eping/e2eping.ts
```diff
@@ -252,13 +269,17 @@ export default class E2ePing {
      * @param participant - The message sender.
      * @param payload - The payload of the message.
      */
-    messageReceived(participant, payload) {
+    messageReceived(participant: JitsiParticipant, payload: { type: string; id?: number }): void {
         // Listen to E2E PING requests and responses from other participants
         // in the conference.
         if (payload.type === E2E_PING_REQUEST) {
-            this.handleRequest(participant.getId(), payload);
+            if (payload.id !== undefined) {
```
saghul, line 276: Why this check?

modules/e2eping/e2eping.ts
```diff
@@ -252,13 +269,17 @@ export default class E2ePing {
...
+                this.handleRequest(participant.getId(), { id: payload.id });
+            }
         } else if (payload.type === E2E_PING_RESPONSE) {
-            this.handleResponse(participant.getId(), payload);
+            if (payload.id !== undefined) {
```
saghul, line 280: Ditto.

saghul (commented):

modules/e2eping/e2eping.ts
naman9271: @saghul ,I added this check to ensure that payload.id is defined before calling handleRequest, as there could be cases where payload.id is undefined,while handleRequest not expecting it to undefined so to properly handle this, I have now defined appropriate TypeScript interfaces to handle this more efficiently . Kindly Review

saghul (approved):

PR #2663, opened 2025-02-25
Migrate `connectivity/IceFailedHandling`

naman9271:
Fixes #2637 

## Major Change

   - The `_canceled` property has been explicitly initialized to `false`. This change ensures that the task is not considered canceled when first created, providing a clear starting state for the task's execution.

   ```typescript
   private _canceled: boolean = false; // Default value set to false
   ```


naman9271:
@saghul Please Review this PR

jallamsetty1 (approved):

PR #2667, opened 2025-02-25
fix(SDP) Add trackID if its missing in msid.

jallamsetty1:
This is needed for older versions of Chrome. Also skip performance optimizations if encode resolution is missing in the stats.


damencho (approved):

modules/qualitycontrol/QualityController.ts
```diff
@@ -301,8 +301,12 @@ export class QualityController {
         if (!this._enableAdaptiveMode) {
             return;
         }
-
         const { encodeResolution, localTrack, qualityLimitationReason, tpc } = sourceStats;
+
+        // Older browser versions might not report the resolution in the stats.
+        if (isNaN(encodeResolution)) {
```
saghul, line 307: Use Number.isNaN. Unlike isNaN, it will never throw an exception.

saghul (approved):
LGTM with a comment.

PR #2674, opened 2025-03-07
Migrates `Jitsi participant`

naman9271:
Fixes #2637 

## Major Changes
- `npm install @types/strophe --save-dev`


JitsiParticipant.ts
```diff
@@ -1,15 +1,40 @@
...
 export default class JitsiParticipant {
 
+    private _jid: string;
+    private _id: string;
+    private _conference: JitsiConference;
+    private _displayName: string;
+    private _supportsDTMF: boolean;
+    private _tracks: JitsiTrack[];
+    private _role: string;
+    private _status: string | undefined;
```
saghul, line 25: Why not do _status?: string;

JitsiParticipant.ts
```diff
@@ -251,92 +286,92 @@ export default class JitsiParticipant {
      * special system participants may want to join hidden (like for example the
      * recorder).
      */
-    isHiddenFromRecorder() {
-        return this._identity?.user?.['hidden-from-recorder'] === 'true';
+    isHiddenFromRecorder(): boolean {
+        return this._identity?.['user']?.['hidden-from-recorder'] === 'true';
```
saghul, line 290: Why change the user check?

package.json
```diff
@@ -44,6 +44,7 @@
     "@types/jasmine": "4.0.3",
     "@types/lodash-es": "4.17.12",
     "@types/sdp-transform": "2.4.5",
+    "@types/strophe": "^1.2.37",
```
saghul, line 47: Nice!

saghul (commented):

JitsiParticipant.ts
naman9271: @saghul I have encountered this error 
image

 Since `_identity` is inferred as a generic object, TypeScript does not recognize the user property.
 I have updtaed that to avoid this issue.
 or i have another approach as :
 
```
return 'user' in this._identity
    && (this._identity as { user?: { ['hidden-from-recorder']?: string } })
    .user?.['hidden-from-recorder'] === 'true';
    }
```
- Or **please let me know** if there's a preferred way to address this

naman9271: @saghul I have encountered this error which is `_identity` is inferred as a generic object, TypeScript does not recognize the user property.
image

- Or i have another approach to fix this error which is 
```
    return 'user' in this._identity
    && this._identity?.user?.['hidden-from-recorder'] === 'true';
```

- Please let me know if this approach works for you or if you have any suggestions for a better solution.

saghul (approved):
A minor comment about dependency pinning, LGTM otherwise.

saghul:
Looks like this needs a rebase too.

JitsiParticipant.ts
```diff
@@ -251,92 +287,93 @@ export default class JitsiParticipant {
...
+    isReplaced(): boolean | undefined {
         return this._isReplaced;
     }
 
     /**
      * @returns {Boolean} Whether this participant replaces another participant
      * from the meeting.
      */
-    isReplacing() {
+    isReplacing(): boolean | undefined {
```
saghul, line 314: Ditto.

JitsiParticipant.ts
```diff
@@ -251,92 +287,93 @@ export default class JitsiParticipant {
...
         return this._isReplacing;
     }
 
     /**
      * @returns {Boolean} Whether this participant has joined without audio.
      */
-    isSilent() {
+    isSilent(): boolean | undefined {
```
saghul, line 321: Ditto.

package.json
```diff
@@ -40,6 +40,7 @@
     "@types/jasmine": "4.0.3",
     "@types/lodash-es": "4.17.12",
     "@types/sdp-transform": "2.4.5",
+    "@types/strophe": "1.2.37",
```
saghul, line 43: This is part of another PR isn't it? It will create a conflict.

saghul (commented):
Please check the comments.

package.json
naman9271: No it is introduced in this PR itself

JitsiParticipant.ts
```diff
@@ -251,92 +287,93 @@ export default class JitsiParticipant {
...
+    isModerator(): boolean {
         return this._role === 'moderator';
     }
 
     /**
      * @returns {Boolean} Wheter this participants will be replaced by another
      * participant in the meeting.
      */
-    isReplaced() {
+    isReplaced(): boolean | undefined {
```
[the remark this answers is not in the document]
naman9271: Done

JitsiParticipant.ts
```diff
@@ -251,92 +287,93 @@ export default class JitsiParticipant {
      * special system participants may want to join hidden (like for example the
      * recorder).
      */
-    isHiddenFromRecorder() {
-        return this._identity?.user?.['hidden-from-recorder'] === 'true';
+    isHiddenFromRecorder(): boolean {
+        return 'user' in this._identity
```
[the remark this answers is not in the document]
naman9271: @saghul Sorry for the earlier confusion, and thank you for your time and patience. I’ve now defined an interface `IIdentity` for _identity which is
```
export interface IIdentity {
    user?: {
        ['hidden-from-recorder']?: string;
    };
}
```
- Let me know if there's anything else that needs improvement

JitsiParticipant.ts
saghul: I think you can use any or unknown.

naman9271: @saghul ,I have implemented the changes using the any type. Initially, I considered this approach but hesitated since it's generally best to minimize the use of any. However, given the context, it seems like a reasonable solution.

saghul (approved):

saghul:

PR #2678, opened 2025-03-10
feat(rtcstats): Adds a default log collector.

damencho:


JitsiMeetJS.ts
```diff
@@ -244,6 +251,13 @@ export default {
      * @see Logger.addGlobalTransport
      */
     addGlobalLogTransport(globalTransport) {
```
andrei-gavrilescu, line 253: not sure if this is used by ljm customers, but if they did that would mean that they'd disable sending stats to rtcstats right?

modules/RTCStats/RTCStats.ts
```diff
@@ -110,6 +113,11 @@ class RTCStats {
             return;
         }
 
+        // Make an attempt to flush in case a lot of logs have been cached
+        if (this._defaultLogCollector) {
+            this._defaultLogCollector.flush();
```
andrei-gavrilescu, line 118: could you chaining

JitsiMeetJS.ts
```diff
@@ -86,6 +86,9 @@ interface IJitsiMeetJSOptions {
     flags?: {
         runInLiteMode?: boolean;
         ssrcRewritingEnabled?: boolean;
+    },
+    logging?: {
+        disableLogCollector?: boolean;
```
andrei-gavrilescu, line 91: I think there's an analytics option called rtcstatsStoreLogs, maybe we can reuse that?

modules/RTCStats/RTCStats.ts
```diff
@@ -191,6 +205,27 @@ class RTCStats {
...
+        }
+
+        return this._defaultLogCollector;
+    }
+
+    /**
+     * Clears the collector and stops it.
+     */
+    clearDefaultLogCollector() {
+        this._defaultLogCollector?.stop();
```
andrei-gavrilescu, line 226: does stop flush the logs?

JitsiMeetJS.ts
damencho: Yep, if they don't send them, yeah. 
Like jitsi-meet is sendgin them:
https://github.com/jitsi/jitsi-meet/blob/59420ea1a8204201db0b3203818594ef4ac66462/react/features/base/logging/JitsiMeetLogStorage.ts#L75

JitsiMeetJS.ts
damencho: Hum ... https://github.com/jitsi/jitsi-meet/blob/59420ea1a8204201db0b3203818594ef4ac66462/config.js#L1807
I'm confused ... this will disable it in jitsi-meet, that's why I added it:
https://github.com/jitsi/jitsi-meet/blob/e5fa25892ec33bec4a87d0153c0d2273ef4e4bdf/react/features/base/logging/middleware.ts#L155

So we can disable it via two configs. So probably I need to check both, otherwise we can disable it via one place in jitsi-meet and still be sending logs ... 

WDYT?

modules/RTCStats/RTCStats.ts
damencho: Yes.

JitsiMeetJS.ts
damencho: I will switch it to the other one.

damencho: Done!

JitsiMeetJS.ts
andrei-gavrilescu: I wonder if it's common for them to use it, because if it is then it probably defeats the purpose. cause we still have to tell them to add sending logic in their app.

damencho: Well then you have a custom one and handling the stuff yourself and do not need the default one ... that is the idea ... like jitsi-meet pushes one and we use that one ... 
I don't have a better idea how to handle that in another way.

andrei-gavrilescu: Can't you have 2 loggers? so basically make ljm always send to rtcstats unless disabled by the config?

damencho: But then, how to handle that with jitsi-meet without having duplicate logs?

damencho: Yeah, I'm confused ... I will take another look at it today.

andrei-gavrilescu: Maybe we can disable it from jitsi meet,  and only send logs to rtcstats through ljm rtcstats logger?

damencho: I have an  idea.

damencho: I will do the change in a bit.

damencho: > Maybe we can disable it from jitsi meet, and only send logs to rtcstats through ljm rtcstats logger?

And you already came to the same conclusion before me :) I missed your comment while writing mine, sorry :)

damencho: @andrei-gavrilescu Done. 
https://github.com/jitsi/jitsi-meet/pull/15763

Once we merge this one, I will update the jitsi-meet PR with the correct ljm version.

andrei-gavrilescu: Looks good!

andrei-gavrilescu (approved):

PR #2685, opened 2025-03-17
feat(deps) use our forked version of strophe.js

saghul:
- Based off 1.5.0, same as we were using before
- Remove optional dependencies
- Remove unnecessary dependencies

In addition, in this repo:
- Bundle the external plugins we are using as they are single JS files

## Types of changes

## Checklist:


damencho:
You have applied the patch in our fork? Why don't we drop the patch(strophe.js+1.5.0.patch) now from patches folder?

saghul:
Yes, I copies the patched version. Ops, I'll remove the patch indeed!

saghul:
@damencho Updated! See the Strophe changes: https://github.com/strophe/strophejs/compare/master...jitsi:strophejs:v1.5-jitsi

damencho (approved):

Issue #2693, opened 2025-03-19
Migrate Files in TypeScript


yashop7:

saghul:
No need to keep this issue. We'll handle PRs individually.

PR #2694, opened 2025-03-19
Migrating  /modules/recording/JibriSession

yashop7:
Fixes #2693

### DIFF
--- a/modules/recording/JibriSession.ts
+++ b/modules/recording/JibriSession.ts
@@ -1,11 +1,11 @@
 import { $iq } from 'strophe.js';
 
 import recordingXMLUtils from './recordingXMLUtils';
-
-type JitsiParticipant = any;
+import JitsiParticipant from '../../JitsiParticipant';
 
 export interface IJibriSessionOptions {
     connection?: any;
+    focusMucJid?: string;
     mode?: string;
     sessionID?: string;
     status?: string;
@@ -44,6 +44,7 @@ export default class JibriSession {
     private _liveStreamViewURL?: string;
     private _initiator?: JitsiParticipant | string;
     private _terminator?: JitsiParticipant | string;
+    private _focusMucJid?: string;
 
     /**
      * Initializes a new JibriSession instance.
@@ -55,9 +56,9 @@ export default class JibriSession {
         this._mode = options.mode;
         this._jibriJid = null;
         this._statusFromJicofo = '';
-
         this._setSessionID(options.sessionID);
         this.setStatus(options.status);
+        this._focusMucJid = options.focusMucJid;
     }
 
     /**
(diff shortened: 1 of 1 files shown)


yashop7:
@saghul, could you please review the changes and let me know if they meet your expectations?

modules/recording/JibriSession.ts
```diff
@@ -2,16 +2,55 @@ import { $iq } from 'strophe.js';
 
 import recordingXMLUtils from './recordingXMLUtils';
 
+type JitsiParticipant = any;
```
saghul, line 5: There is an open PR which introduces this type, let's wait until it lands so you can rebase and leverage it.

modules/recording/JibriSession.ts
```diff
@@ -2,16 +2,55 @@ import { $iq } from 'strophe.js';
...
+
+interface IJibriSessionOptions {
```
saghul, line 7: Export the interfaces so calling code can see them.

modules/recording/JibriSession.ts
```diff
@@ -2,16 +2,55 @@ import { $iq } from 'strophe.js';
...
 export default class JibriSession {
+    private _connection?: any;
+    private _mode?: string;
+    private _jibriJid: string | null = null;
+    private _statusFromJicofo: string = '';
+    private _sessionID?: string;
+    private _status?: string;
+    private _error?: string;
+    private _liveStreamViewURL?: string;
+    private _initiator?: JitsiParticipant | string;
```
saghul, line 45: Is that right? Can it be a participant OR a string?

modules/recording/JibriSession.ts
```diff
@@ -2,16 +2,55 @@ import { $iq } from 'strophe.js';
...
+    private _terminator?: JitsiParticipant | string;
```
saghul, line 46: ditto

saghul (commented):
Some quick comments.

modules/recording/JibriSession.ts
yashop7: I think yes, as it can be Participant or resource id

saghul: Sounds odd, can you please check where the difference comes from?

yashop7: I believe the probable reason behind this is JitsiParticipant, which is triggered by the Participants, while the string is triggered by the server externally

saghul: I looked into it and the case is when the local participant does it, since we have no JitsiParticipant object for ourselves.

modules/recording/JibriSession.ts
yashop7: Added the Type as well + added focusMucJid in the Constructor as Recording Manager class was using it in creating Session


Screenshot 2025-03-31 at 5 59 05 PM

saghul (approved):

saghul:

PR #2712, opened 2025-03-23
Migrated `./modules/utils/Listenable`

naman9271:
FIxes #2637


modules/util/Listenable.ts
```diff
@@ -1,11 +1,19 @@
 import EventEmitter from './EventEmitter';
 
+export type EventListener = (...args: any[]) => void;
```
saghul, line 3: This shouldn't be necessary now that you migrated EventEmitter, right?

naman9271: yes @saghul but i think it is neccessary as functions like `addCancellableListener` ,`removeListener` , `addListener` has parameter `listener` which has its type meanwhile i ahve updtaed the `addEventListener` and `removeEvent Listener` type ....PTAL

saghul: Right, but shoul;dn't this type be declared in EventEmitter and exposed there? After all it's the EventEmitter API the one that uses it.

naman9271: Oops! i just noticed that , I have make declared type in EventEmitter Please have a look

saghul (approved):

saghul:

PR #2719, opened 2025-03-23
Migrated `./modules/e2ee/crypro-utilis`

naman9271:
Fixes #2637

### DIFF
--- a/modules/e2ee/crypto-utils.ts
+++ b/modules/e2ee/crypto-utils.ts
@@ -4,8 +4,8 @@
  *
  * See https://tools.ietf.org/html/draft-omara-sframe-00#section-4.3.1
  */
-export async function deriveKeys(material) {
-    const info = new ArrayBuffer();
+export async function deriveKeys(material: CryptoKey): Promise<{ material: CryptoKey; encryptionKey: CryptoKey }> {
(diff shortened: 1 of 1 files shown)


naman9271:
For this to merge would have to migrate Context.js, as it is getting error in build ,so putting this PR on draft

naman9271:
.

webpack.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = (_env, argv) => {
                 library: 'JitsiMeetJS',
                 libraryTarget: 'umd',
                 path: path.join(process.cwd(), 'dist', 'umd') } },
-        {
```
saghul: Drop this please.

naman9271: @saghul actually it is failing on building , The second webpack configuration for the worker doesn't include the module rules for TypeScript processing. so I added  this 
I see the same issue in migration of context.js and worker.js #2800 #2799 

or would you like to fix it by adding below rule instead of ...config ??
```
   module: config.module,
   resolve: config.resolve,

saghul: Yeah that makes sense, thanks!

naman9271: changes done please take a look

saghul:

PR #2725, opened 2025-03-23
Migrating `modules/xmpp/StropheLastSuccess`

yashop7:
Fixes #2693

### DIFF
--- a/modules/xmpp/StropheLastSuccess.ts
+++ b/modules/xmpp/StropheLastSuccess.ts
@@ -1,13 +1,9 @@
 /**
  * Attaches to the {@link Strophe.Connection.rawInput} which is called whenever any data is received from the server.
  */
-export interface IXmppConnection {
-    connected: boolean;
-}
 
-export interface IStropheConnection {
-    rawInput: (...args: any[]) => void;
-}
+import XmppConnection from './XmppConnection';
+
 
 export default class LastRequestTracker {
     private _lastSuccess: number | null;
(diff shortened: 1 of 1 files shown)


modules/xmpp/StropheLastSuccess.ts
```diff
@@ -1,7 +1,18 @@
 /**
  * Attaches to the {@link Strophe.Connection.rawInput} which is called whenever any data is received from the server.
  */
+export interface IXmppConnection {
```
saghul, line 4: This feels like the wrong place to expose this since we do have an XmppConnection class.

modules/xmpp/StropheLastSuccess.ts
```diff
@@ -1,7 +1,18 @@
...
+    connected: boolean;
+}
+
+export interface IStropheConnection {
```
saghul, line 8: Ditto.

modules/xmpp/StropheLastSuccess.ts
yashop7: imported the Correct type, Can you please check if it is the right

modules/xmpp/StropheLastSuccess.ts
```diff
@@ -16,10 +23,10 @@ export default class LastRequestTracker {
      * @param {XmppConnection} xmppConnection - The XMPP connection which manages the given {@code stropheConnection}.
      * @param {Object} stropheConnection - Strophe connection instance.
      */
-    startTracking(xmppConnection, stropheConnection) {
+    startTracking(xmppConnection: XmppConnection, stropheConnection: any): void {
```
saghul, line 26: You can get the strophe connection type with Strophe.Connection.

yashop7: Yeah, @saghul, It goes unnoticed from my side 👍

saghul (approved):

saghul:

PR #2726, opened 2025-03-23
Migrating `modules/xmpp/strophe.util`

yashop7:
Fixes #2693

### DIFF
--- a/modules/xmpp/strophe.util.ts
+++ b/modules/xmpp/strophe.util.ts
@@ -47,6 +47,11 @@ const lastErrorStatusRegExpr: RegExp
 export default function(): void {
 
     Strophe.log = function(level: number, msg: any): void {
+        // Our global handler reports uncaught errors to the stats which may
+        // interpret those as partial call failure.
+        // Strophe log entry about secondary request timeout does not mean that
+        // it's a final failure(the request will be restarted), so we lower it's
+        // level here to a warning.
         logger.trace('Strophe', level, msg);
         if (typeof msg === 'string'
                 && msg.indexOf('Request ') !== -1
@@ -55,8 +60,11 @@ export default function(): void {
             level = Strophe.LogLevel.WARN;
         }
 
+        /* eslint-disable no-case-declarations */
         switch (level) {
         case Strophe.LogLevel.DEBUG:
+            // The log message which reports successful status is logged on
+            // Strophe's DEBUG level.
             if (lastErrorStatus !== -1
                     && resetLastErrorStatusRegExpr.test(msg)) {
                 logger.debug('Reset lastErrorStatus');
@@ -77,8 +85,16 @@ export default function(): void {
             logger.error(`Strophe: ${msg}`, msg);
             break;
         }
+
+        /* eslint-enable no-case-declarations */
     };
 
+    /**
+     * Returns error status (HTTP error code) of the last BOSH request.
+     *
+     * @return {number} HTTP error code, '0' for unknown or "god knows what"
+     * (this is a hack).
+     */
     Strophe.getLastErrorStatus = function(): number {
         return lastErrorStatus;
     };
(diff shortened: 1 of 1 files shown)


modules/xmpp/strophe.util.ts
```diff
@@ -4,6 +4,26 @@
 import { getLogger } from '@jitsi/logger';
 import { Strophe } from 'strophe.js';
 
+export interface IStrophe {
```
saghul, line 7: Do we need this? This package now depends on `@types/strophe` which should define these, doesn't it?

saghul: Not sure what you mean. My question is why do you need to define your own type if the `Strophe` type is already defined?

yashop7: imported type from `strophe`

saghul: Did you push? I see no change.

yashop7: Yeah, I pushed the changes before, 
I don't know why they do not appear here, 
imported the Strophe type in place of the Interface

Screenshot 2025-03-30 at 6 11 09 PM

modules/xmpp/strophe.util.ts
```diff
@@ -22,55 +23,48 @@ const logger = getLogger('modules/xmpp/strophe.util');
...
-        // it's a final failure(the request will be restarted), so we lower it's
-        // level here to a warning.
+    (Strophe as IStrophe).log = function(level: number, msg: any): void {
         logger.trace('Strophe', level, msg);
         if (typeof msg === 'string'
                 && msg.indexOf('Request ') !== -1
                 && msg.indexOf('timed out (secondary), restarting') !== -1) {
             // eslint-disable-next-line no-param-reassign
-            level = Strophe.LogLevel.WARN;
+            level = (Strophe as IStrophe).LogLevel.WARN;
```
saghul, line 56: The previous accessor should work just fine (here and elsewhere). What problem do you see?

modules/xmpp/strophe.util.ts
```diff
@@ -3,6 +3,7 @@
  */
 import { getLogger } from '@jitsi/logger';
 import { Strophe } from 'strophe.js';
+import type { Strophe as IStrophe } from 'strophe.js';
```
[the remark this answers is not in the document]
yashop7: No, it is not necessary, I will remove it, it was a mistake from the previous commits. Earlier, I made a custom interface IStrophe, which I removed, But rather than removing the IStrophe from other places, I directly imported it from Strophe

modules/xmpp/strophe.util.ts
```diff
@@ -59,11 +55,8 @@ export default function() {
             level = Strophe.LogLevel.WARN;
         }
 
-        /* eslint-disable no-case-declarations */
         switch (level) {
         case Strophe.LogLevel.DEBUG:
-            // The log message which reports successful status is logged on
```
saghul, line 65: Leave the comment in please.

modules/xmpp/strophe.util.ts
```diff
@@ -84,21 +77,13 @@ export default function() {
             logger.error(`Strophe: ${msg}`, msg);
             break;
         }
-
-        /* eslint-enable no-case-declarations */
     };
 
-    /**
```
saghul, line 91: Leave the coment.

modules/xmpp/strophe.util.ts
```diff
@@ -22,35 +22,31 @@ const logger = getLogger('modules/xmpp/strophe.util');
...
     = /request errored, status: (\d+), number of errors: \d+/;
 
 /**
  *
  */
-export default function() {
+export default function(): void {
 
-    Strophe.log = function(level, msg) {
-        // Our global handler reports uncaught errors to the stats which may
```
saghul, line 49: Why are you removing comments?!

saghul (changes requested):
Do NOT remove code comments please.

saghul (approved):

PR #2730, opened 2025-03-23
Mgrating `./modules/xmpp/ResumeTask`

naman9271:
Fixes #2637


modules/xmpp/ResumeTask.ts
```diff
@@ -8,42 +8,55 @@ import { getJitterDelay } from '../util/Retry';
 
 const logger = getLogger('modules/xmpp/ResumeTask');
 
+export interface IStropheConnection {
```
saghul, line 11: Please import the correct type.

naman9271: Oops, I misunderstood that! Sorry about it, and I'm working on fixing it now.

naman9271: Make changes please have a look @saghul

modules/xmpp/ResumeTask.ts
```diff
@@ -5,45 +5,63 @@ import {
     default as NetworkInfo
 } from '../connectivity/NetworkInfo';
 import { getJitterDelay } from '../util/Retry';
+import Strophe from 'strophe';
+
+declare module 'strophe' {
```
saghul, line 10: I'm confused, why do you need to define this? It's declared in types/strophe.

naman9271: My Bad, Actually I got confused i had imported `strophe` instead of `strophe.js` by mistake because of which that chunk have been created

saghul:

PR #2732, opened 2025-03-23
migrated `modules/xmpp/strophe.disco`

naman9271:
Fixes #2637


saghul:
This is a vendored dependency, so let's leave it alone.

naman9271:
> This is a vendored dependency, so let's leave it alone.

@saghul there is also a file `strophe.jingle` so do we need to migrate it or just leave it as it is ??

saghul:
Yep, let's leave it as is for now.

PR #2733, opened 2025-03-23
Migrate `modules/xmpp/MockClasses`

yashop7:
Fixes #2693

### DIFF
--- a/modules/xmpp/MockClasses.ts
+++ b/modules/xmpp/MockClasses.ts
@@ -1,16 +1,7 @@
 import { Strophe } from 'strophe.js';
-
 import Listenable from '../util/Listenable';
 
 /* eslint-disable no-empty-function */
-
-export interface ISocket {
-    readyState: number;
-}
-
-export interface IStropheProto {
-    socket: ISocket | undefined;
-}
 
 /**
  * Mock {@link ChatRoom}.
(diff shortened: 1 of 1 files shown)


modules/xmpp/MockClasses.ts
```diff
@@ -4,21 +4,34 @@ import Listenable from '../util/Listenable';
 
 /* eslint-disable no-empty-function */
 
+export interface ISocket {
```
saghul, line 7: This is the WebSocket interface, can't we use that?

modules/xmpp/MockClasses.ts
```diff
@@ -60,21 +73,23 @@ export class MockStropheConnection extends Listenable {
...
+    simulateConnectionState(newState: number): void {
         if (newState === Strophe.Status.CONNECTED) {
             this._proto.socket = {
                 readyState: WebSocket.OPEN
             };
         } else {
             this._proto.socket = undefined;
         }
-        this._connectCb(newState);
+        if (this._connectCb) {
```
saghul, line 84: Use optional chaining instead.

yashop7: Added it

modules/xmpp/MockClasses.ts
yashop7: yes @saghul , made the changes

saghul (approved):

PR #2724, opened 2025-03-23
Migrating `modules/proxyconnection/ProxyConnectionPC`

yashop7:
Fixes #2693

PR #2737, opened 2025-03-23
Migrating `modules/xmpp/BreakoutRooms`

yashop7:
Fixes #2693

### DIFF
--- a/modules/xmpp/BreakoutRooms.ts
+++ b/modules/xmpp/BreakoutRooms.ts
@@ -17,39 +17,14 @@ const BREAKOUT_ROOM_EVENTS = {
 
 const logger = getLogger('modules/xmpp/BreakoutRooms');
 
-export interface IChatRoom {
-    eventEmitter: {
-        emit: (event: string, ...args: any[]) => void;
-    };
-    isModerator: () => boolean;
-    myroomjid: string;
-    options: {
-        hiddenDomain?: string;
-    };
-    xmpp: {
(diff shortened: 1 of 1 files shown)


modules/xmpp/BreakoutRooms.ts
```diff
@@ -17,17 +17,67 @@ const BREAKOUT_ROOM_EVENTS = {
 
 const logger = getLogger('modules/xmpp/BreakoutRooms');
 
+export interface IChatRoom {
```
saghul, line 20: This is not the place to export this interface. That should happen in ChatRoom.

modules/xmpp/BreakoutRooms.ts
```diff
@@ -17,17 +17,67 @@ const BREAKOUT_ROOM_EVENTS = {
...
+        removeListener: (event: string, listener: (payload: IBreakoutRoomsMoveToRoomPayload | IBreakoutRoomsUpdatePayload) => void) => void;
+    };
+}
+
+export interface IBreakoutRoomMessage {
+    [key: string]: any;
+    type: string;
+}
+
+export interface IParticipant {
```
saghul, line 45: Not here.

modules/xmpp/BreakoutRooms.ts
yashop7: You are right, @saghul. I have removed the interfaces, but I'm including them in the PR draft until I migrate ChatRoom to TS

naman9271:
@yashop7 [Comment](https://github.com/jitsi/lib-jitsi-meet/pull/2724#issuecomment-2879692428)

PR #2741, opened 2025-03-25
fix: Correct return type of `getIdentity` function to `JwtIdentity`

yashop7:
- Ensures consistency with `.d.ts` file & proper type handling of jwt identity
- Updated the return type of `getIdentity()` from `object` to `JwtIdentity`

### DIFF
--- a/JitsiParticipant.ts
+++ b/JitsiParticipant.ts
@@ -12,16 +12,16 @@ export interface ISourceInfo {
 
 export interface IJwtIdentity {
     callee?: {
-        avatar: string;
-        id: string;
-        name: string;
+        avatar?: string;
+        id?: string;
+        name?: string;
     };
     group?: string;
     user?: {
-        avatar: string;
-        email: string;
-        id: string;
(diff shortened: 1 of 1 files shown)


JitsiParticipant.ts
```diff
@@ -10,6 +10,22 @@ export interface ISourceInfo {
     videoType: string;
 }
 
+export interface IJwtIdentity {
```
saghul, line 13: I bel;ieve every property here is optional.

yashop7: Yes, @saghul, you are right, updated `IJwtIdentity`

saghul:

JitsiParticipant.ts
saghul: All sub-fields are optional too.

naman9271:
@yashop7 [Comment](https://github.com/jitsi/lib-jitsi-meet/pull/2724#issuecomment-2879692428)

PR #2747, opened 2025-03-31
feat(ts) add typedoc

saghul:
The end goal is to fix our types to make typedoc generated docs as useful as possible.

The generated docs will be published to GH pages on every push.

Sample:

Screenshot 2025-03-31 at 12 52 37

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,4 +25,5 @@ jobs:
     - name: Check if the git repository is clean
       run: exit $( git status --porcelain --untracked-files=no | head -255 | wc -l )
     - run: npm run lint
+    - run: npm run typedoc
     - run: npm run test
--- a/.github/workflows/deploy-docs.yml
+++ b/.github/workflows/deploy-docs.yml
@@ -1,0 +1,41 @@
+name: Deploy GH pages
+
+on:
+  workflow_dispatch:
+  push:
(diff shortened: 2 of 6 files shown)


damencho (approved):

naman9271:
That's interesting, @saghul! , I think we should also add this in the entry point of the tsconfig so that these types get documented too.
```
      "service/**/*.ts",
      "modules/**/*.ts"
```
Would it make sense to include this now, or should we wait until the entire migration is complete ?

saghul:
I don't think so. The entry points are our API. Any type exposed there should be exposed indeed, but not the internals.

Note this is a first approximation. The plan is to keep improving them!

naman9271:
So should i include this in the GSoC idea of migration of lib-jitsi-meet or just have done by automated typings package generation ??

saghul:
I'd say improving the output by tweaking the existing types and prioritising the classes in the API should be part of the proposal, yes.

naman9271:
Got it! @saghul Sir So, instead of adding new APIs, I'll focus on improving the existing TypeScript types and JSDoc, I'll also ensure that the Typedoc-generated documentation becomes more meaningful by improving type definitions. Let me know if there's anything specific you'd like me to focus on in this!

saghul:
Sounds like a plan!

PR #2754, opened 2025-04-06
feat(xmpp): Switch unload to pagehide event listener

catFurr:
## Description

Switched the [beforeunload](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event) event listener to a [pagehide](https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event) event listener which is now baseline and well supported.

The `config.disableBeforeUnloadHandlers` is no longer needed. This simplifies configuration for maintainers, while following specification best practices.

This change allows the use of beforeunload handler in a more appropriate manner.

## Related Issue

https://github.com/jitsi/jitsi-meet/issues/15890
https://community.jitsi.org/t/prompt-before-closing-during-a-call/20081
https://community.jitsi.org/t/prevent-refresh-or-back-during-call-give-some-confirmation-prompt/84949/2

## Motivation and Context

The `unload` and `beforeunload` event handlers have been [marked deprecated](https://developer.chrome.com/docs/web-platform/page-lifecycle-api#legacy_lifecycle_apis_to_avoid) and websites are recommended to switch to better supported events.

The MDN docs mention
> If you're specifically trying to detect page unload events, the pagehide event is the best option.

This change allows the implementation of a confirmation dialog before meetings are exited.
> It is therefore recommended that developers listen for beforeunload only when users have unsaved changes so that the dialog mentioned above can be used to warn them about impending data loss, and remove the listener again when it is not needed. Listening for beforeunload sparingly can minimize the effect on performance.

## How Has This Been Tested?

- All `npm run test` pass.
- Deployed to jitsi-meet instance and connected to conference with no issues.
- meeting ends gracefully when page closed/navigated away.

no effect on UI.

## Types of changes

## Checklist:

### DIFF
--- a/modules/xmpp/xmpp.js
+++ b/modules/xmpp/xmpp.js
@@ -216,7 +216,7 @@ export default class XMPP extends Listenable {
         // they wanted to utilize the connected connection in an unload handler
         // of their own. However, it should be fairly easy for them to do that
         // by registering their unload handler before us.
-        $(window).on(`${this.options.disableBeforeUnloadHandlers ? '' : 'beforeunload '}unload`, ev => {
+        $(window).on('pagehide', ev => {
             this.disconnect(ev).catch(() => {
                 // ignore errors in order to not brake the unload.
             });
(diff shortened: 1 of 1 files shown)


damencho:
I don't think removing the option to disable unload it is correct there is some logic around it in jitsi-meet.
Also, this needs to be tested with react-native. 
The reason for the unload event is to flush the websocket or sendBeacon when using bosh, this avoids having ghosts in meetings when participants leave.

catFurr:
Tested using BOSH with many users joining and leaving, I cannot see any issues related to ghosts.
Do you have any suggestions for me related to testing this? Please let me know how I can test to ensure this change is completely safe, Thanks.

I'll test with a native build and let you know how it goes 👍🏽

PR #2762, opened 2025-04-10
feat(jaas) add simple zero-config mechanism for JaaS users

saghul:
Technically anyone can use it as long as they provide connection options, but we don't need any of that for JaaS.

### DIFF
--- a/JitsiMeetJS.ts
+++ b/JitsiMeetJS.ts
@@ -1,16 +1,19 @@
 import Logger from '@jitsi/logger';
+import { merge } from 'lodash-es';
 
+import JitsiConference from './JitsiConference';
 import * as JitsiConferenceErrors from './JitsiConferenceErrors';
 import * as JitsiConferenceEvents from './JitsiConferenceEvents';
 import JitsiConnection from './JitsiConnection';
 import * as JitsiConnectionErrors from './JitsiConnectionErrors';
-import * as JitsiConnectionEvents from './JitsiConnectionEvents';
+import { JitsiConnectionEvents } from './JitsiConnectionEvents';
 import JitsiMediaDevices from './JitsiMediaDevices';
 import * as JitsiMediaDevicesEvents from './JitsiMediaDevicesEvents';
 import JitsiTrackError from './JitsiTrackError';
 import * as JitsiTrackErrors from './JitsiTrackErrors';
 import * as JitsiTrackEvents from './JitsiTrackEvents';
 import * as JitsiTranscriptionStatus from './JitsiTranscriptionStatus';
+import JitsiLocalTrack from './modules/RTC/JitsiLocalTrack';
 import RTC from './modules/RTC/RTC';
 import RTCStats from './modules/RTCStats/RTCStats';
 import * as RTCStatsEvents from './modules/RTCStats/RTCStatsEvents';
@@ -27,6 +30,7 @@ import Settings from './modules/settings/Settings';
 import LocalStatsCollector from './modules/statistics/LocalStatsCollector';
 import runPreCallTest, { IIceServer, IPreCallResult } from './modules/statistics/PreCallTest';
 import Statistics from './modules/statistics/statistics';
+import Deferred from './modules/util/Deferred';
 import ScriptUtil from './modules/util/ScriptUtil';
 import * as VideoSIPGWConstants from './modules/videosipgw/VideoSIPGWConstants';
 import AudioMixer from './modules/webaudio/AudioMixer';
@@ -104,6 +108,16 @@ interface ICreateLocalTrackFromMediaStreamOptions {
     stream: MediaStream;
     track: any;
     videoType?: VideoType;
+}
+
+export interface IJoinConferenceOptions {
+    conferenceOptions?: any;
+    connectionOptions?: any;
+    jaas?: {
+        release?: boolean;
+        useStaging?: boolean;
+    };
+    tracks?: JitsiLocalTrack[];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


JitsiMeetJS.ts
```diff
@@ -495,6 +509,104 @@ export default {
...
+
+        // Provide a solid default config in case of JaaS.
+        if (appId.startsWith('vpaas-magic-cookie')) {
+            // Auto-initialize the library.
+            const autoInit = Symbol.for('kLJMAutoInit');
+
+            if (typeof window[autoInit] === 'undefined') {
+                window[autoInit] = true;
+
+                // TODO(saghul): review rtcstats initialization.
```
saghul, line 537: @andrei-gavrilescu I really don't like how this turned out, unless I add this here rtcstats won't be initialized, but I think it should, if the connection initialization is what happens first.

JitsiMeetJS.ts
```diff
@@ -495,6 +509,104 @@ export default {
...
+            // Make sure this runs after the promise was resolved so that local track events can be
+            // listened to.
+            queueMicrotask(() => {
+                for (const track of options.tracks || []) {
+                    conf.addTrack(track);
+                }
+
+                conf.join();
+
+                // TODO(saghul): set some sender / receiver constraints?
```
saghul, line 592: @jallamsetty1 WDYT? IMHO our current default of 180p is bad, it makes for a bad first impression. How about defaulting to 720?

Since that might be disruptive we could set the constraints here, when using this "simplified" API.

JitsiMeetJS.ts
```diff
@@ -495,6 +509,104 @@ export default {
...
+            d.reject(error);
+        }
+
+        conn.addEventListener(
+            JitsiConnectionEvents.CONNECTION_ESTABLISHED, onConnectionEstablished);
+        conn.addEventListener(
+            JitsiConnectionEvents.CONNECTION_FAILED, onConnectionFailed);
+        conn.connect({ name: roomName });
+
+        return d as unknown as Promise<JitsiConference>;
```
saghul, line 607: What should we return here? Some options:

- Just return the conference and rely on garbage collection to kill the connection. Not a fan...
- Add a handler to close the connection when the conference ends. It would break breakout rooms, so not great either.
- Return an object with 2 properties: connection and conference, and perhaps a `destroy()` or `leave()` helper method that does The Right Thing. This might as well be the inception of `JitsiSession`!

saghul (commented):
@damencho @hristoterezov @jallamsetty1 PTAL.

saghul:
Example of this PR in action: https://github.com/jitsi/ljm-getting-started/pull/1

JitsiMeetJS.ts
jallamsetty1: This is a bit tricky. We do not have the source info at this point so the client will have to set the default constraints to 720p. That means when a user joins the call, the bridge will ask all the other endpoints in the call to start sending 720p even if there are being viewed or not. This could create a lot of unnecessary traffic and overload the bridges.
Even if the application updates the receiver constraints, there may be some delay and depending on how they implement it, if they don't reset the default constraints back to 0, the bridge may end up sending a lot of unnecessary traffic down to the endpoints.
That is why the default is set to 180p in the bridge. Maybe better documentation would help here?

JitsiMeetJS.ts
```diff
@@ -106,6 +110,16 @@ interface ICreateLocalTrackFromMediaStreamOptions {
...
 }
 
+export interface IJoinConferenceOptions {
+    conferenceOptions?: any;
+    connectionOptions?: any;
+    jaas?: {
+        release?: boolean;
+        useStaging?: boolean;
+    };
+    tracks?: JitsiLocalTrack[];
```
jallamsetty1, line 120: Is the tracks part wired up yet? I don't see JitsiConference adding the tracks from` this.options.config.tracks `automatically to the media sessions being created. Is that for later?

JitsiMeetJS.ts
saghul: The status quo is bad and documentation won't change that. Imagine this, you are evaulating video products and see some upscaled 180p stream looking terrible. Would you dig deeper?

Perhaps we can find some middle gound with 360p, or 720p but limiting lastN?

JitsiMeetJS.ts
saghul: @jallamsetty1 thoughts on this?

JitsiMeetJS.ts
saghul: Some data from other comparable libraries:

- mediaSoup: sends the highest layer if not configured to a specific one
- janus: same as mediasoup
- livekit: it has an adaptive mode which will select the receive resolution based on the attached element size and visibility, this is nice!
- twilio: same as livekit, when "adaptive simulcast" mode is used

IMHO any option is better than our current one.

JitsiMeetJS.ts
andrei-gavrilescu: If I rmmbr correctly the reason why we need to initialize rtcstats before LJM gets loaded, is that rtcstats proxies calls to the WebRTC/GUM api, something like  `window[${prefix}RTCPeerConnection] = peerconnectionproxy`, and LJM does something similar where it uses a reference to `window[${prefix}RTCPeerConnection]`, 
If rtcstats doesn't get initialized before that, then LJM gets the reference to the non proxied RTCPeerConnection ctor and uses it in subsequent calls.
That's the reason we need to pass it on init, and the reason we need to pass the config to every connection as well is that it can change in between connections where LJM doesn't get reloaded, such as electron.

andrei-gavrilescu: We could have rtcstats always proxy the WertcAPI, and simply have it not connect to the server.
So we would only need to pass the config once, the implication would be that in cases where rtcstats is disabled, the proxies would still be active but not sending anything.

andrei-gavrilescu: looking through LJM a bit, that might no longer be the case, seems like we use RTCPeerConneciton directly now, so removing it might work, need to do some checks.

JitsiMeetJS.ts
jallamsetty1: I agree that we need a better one. I voted for something similar to what livekit and Twilio do when we first implemented receiver constraints but product wanted the application to control what is being requested based on the layout instead of the library doing it automatically. We need to involve product and have a much broader discussion.

JitsiMeetJS.ts
saghul: Thank you!

saghul (commented):
@jallamsetty1 I've added some changes, PTAL!

PR #2763, opened 2025-04-11
fix(e2ee) avoid restarting media sessions more than once

saghul:
After enabling E2EE for the first time, keep the media transform inplace, but make it a no-op.

This simplifies the flow a little since it doesn't require a media session restart.

Way back when, there were performance concerns when having a transform, which is what prompted us to implement the restar mechanism, but a lot has changed since and both audio and video frames don't need to jump threads these days. We are still not enabling an empty transform at all times out of precaution.

### DIFF
--- a/modules/e2ee/Context.js
+++ b/modules/e2ee/Context.js
@@ -1,5 +1,4 @@
 /* eslint-disable no-bitwise */
-/* global BigInt */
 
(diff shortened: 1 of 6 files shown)


jallamsetty1 (approved):
LGTM

PR #2765, opened 2025-04-11
updated workflows

naman9271:
### Changes Made
- Replaced `npm install` with `npm ci` in all relevant GitHub workflow files.

### Benefits
- **Speed:** `npm ci` is faster than `npm install` as it skips dependency resolution and directly installs from the lock file.
- **Best Practice:** Aligns with recommended practices for CI environments.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,7 +16,7 @@ jobs:
       run: |
         node -v
         npm -v
-    - run: npm install
+    - run: npm ci
     - run: npm pack
     - name: Check git status
       run: git status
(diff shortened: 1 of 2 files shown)


naman9271:
@saghul PTAL !!

saghul:
I'm not a fan of this change because running install helps us catch out of date package-lock.json.

We want resolution to happen in the CI. There is an extra step that catches a dirty git tree, in case the lock file got updated.

naman9271:
@saghul, I was recently going through the documentation and best practices for setting up CI/CD pipelines, and I came across the recommendation to use `npm ci` instead of` npm install` in CI environments. Based on that, I raised this PR. Now i understand it more deeply.

Apologies for that, I am closing this PR.

saghul:
No worries, thanks for looking into it!

PR #2766, opened 2025-04-14
feat(rtcstats): move rtcstats init

andrei-gavrilescu:
LJM no longer proxies RTCPeerConnection so we can move the rtcstats initialization process in a more convenient place.
Also it seems that the recent migration to .ts of the JitsiConnection module removed the       `RTCStats.startWithConnection(this)` call, disabling rtcstats, re added that.


saghul:
> Also it seems that the recent migration to .ts of the JitsiConnection module removed the `RTCStats.startWithConnection(this)` call, disabling rtcstats, re added that.

Missed that during the review, sorry about that!

saghul (approved):
Awesome, thanks for taking care of this!

PR #2769, opened 2025-04-18
fix(FF): scalability.

hristoterezov:
Disables AV1 and VP9 for FF due to scalability not working. Also in order to make VP8 scalability to work, DD extension headers had to be disabled.

### DIFF
--- a/JitsiConference.js
+++ b/JitsiConference.js
@@ -457,7 +457,8 @@ JitsiConference.prototype._init = function(options = {}) {
             preferredCodec: _getCodecMimeType(config.videoQuality?.preferredCodec),
             screenshareCodec: browser.isMobileDevice()
                 ? _getCodecMimeType(config.videoQuality?.mobileScreenshareCodec)
-                : _getCodecMimeType(config.videoQuality?.screenshareCodec)
+                : _getCodecMimeType(config.videoQuality?.screenshareCodec),
+            enableAV1ForFF: config.testing?.enableAV1ForFF
         },
         p2p: {
             preferenceOrder: browser.isMobileDevice()
@@ -467,7 +468,8 @@ JitsiConference.prototype._init = function(options = {}) {
             preferredCodec: _getCodecMimeType(config.p2p?.preferredCodec),
             screenshareCodec: browser.isMobileDevice()
                 ? _getCodecMimeType(config.p2p?.mobileScreenshareCodec)
-                : _getCodecMimeType(config.p2p?.screenshareCodec)
+                : _getCodecMimeType(config.p2p?.screenshareCodec),
+            enableAV1ForFF: true // For P2P no simulcast is needed, therefore AV1 can be used.
         }
     };
 
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -723,8 +723,9 @@ export class TPCUtils {
 
         return videoCodec === CodecMimeType.VP8 // VP8 always
 
-            // K-SVC mode for VP9 when no scalability mode is set. Though only one outbound-rtp stream is present,
-            // three separate encodings have to be configured.
+            // For FF: scalabilityModeEnabled is not supported and we have to use simulcast.
+            // For other browsers we use K-SVC mode for VP9 when no scalability mode is set. Although
+            // only one outbound-rtp stream is present, three separate encodings have to be configured.
             || (!this.codecSettings[videoCodec].scalabilityModeEnabled && videoCodec === CodecMimeType.VP9)
 
             // When scalability is enabled, always for H.264, and only when simulcast is explicitly enabled via
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -942,7 +942,10 @@ export class TPCUtils {
 
             if (isSender && mLine.ext?.length) {
                 const headerIndex = mLine.ext.findIndex(ext => ext.uri === DD_HEADER_EXT_URI);
-                const shouldNegotiateHeaderExts = codec === CodecMimeType.AV1 || codec === CodecMimeType.H264;
+                const shouldNegotiateHeaderExts = codec === CodecMimeType.AV1 || codec === CodecMimeType.H264
+
+                    // Removing DD ext header lines from the sdp breaks the scalability for FF with version 136+.
+                    || (codec === CodecMimeType.VP8 && browser.isFirefox());
 
                 if (!this.supportsDDHeaderExt && headerIndex >= 0) {
                     this.supportsDDHeaderExt = true;
--- a/modules/RTC/TPCUtils.js
+++ b/modules/RTC/TPCUtils.js
@@ -728,6 +728,11 @@ export class TPCUtils {
             // only one outbound-rtp stream is present, three separate encodings have to be configured.
             || (!this.codecSettings[videoCodec].scalabilityModeEnabled && videoCodec === CodecMimeType.VP9)
 
+            // FF uses simulcast with AV1.
+            || (!this.codecSettings[videoCodec].scalabilityModeEnabled
+                && this.codecSettings[videoCodec].useSimulcast
+                && videoCodec === CodecMimeType.AV1)
+
             // When scalability is enabled, always for H.264, and only when simulcast is explicitly enabled via
             // config.js for VP9 and AV1 since full SVC is the default mode for these 2 codecs.
             || (this.codecSettings[videoCodec].scalabilityModeEnabled
--- a/modules/browser/BrowserCapabilities.js
+++ b/modules/browser/BrowserCapabilities.js
@@ -246,15 +246,6 @@ export default class BrowserCapabilities extends BrowserDetection {
     }
 
     /**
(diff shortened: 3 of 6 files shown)


saghul (approved):

modules/RTC/TPCUtils.js
```diff
@@ -723,10 +723,14 @@ export class TPCUtils {
...
         return videoCodec === CodecMimeType.VP8 // VP8 always
 
-            // K-SVC mode for VP9 when no scalability mode is set. Though only one outbound-rtp stream is present,
-            // three separate encodings have to be configured.
+            // For FF: scalabilityModeEnabled is not supported and we have to use simulcast.
+            // For other browsers we use K-SVC mode for VP9 when no scalability mode is set. Although
+            // only one outbound-rtp stream is present, three separate encodings have to be configured.
             || (!this.codecSettings[videoCodec].scalabilityModeEnabled && videoCodec === CodecMimeType.VP9)
 
+            // FF uses simulcast with AV1.
```
saghul, line 731: Do we know how that performs?

modules/RTC/TPCUtils.js
```diff
@@ -942,7 +946,10 @@ export class TPCUtils {
 
             if (isSender && mLine.ext?.length) {
                 const headerIndex = mLine.ext.findIndex(ext => ext.uri === DD_HEADER_EXT_URI);
-                const shouldNegotiateHeaderExts = codec === CodecMimeType.AV1 || codec === CodecMimeType.H264;
+                const shouldNegotiateHeaderExts = codec === CodecMimeType.AV1 || codec === CodecMimeType.H264
+
+                    // Removing DD ext header lines from the sdp breaks the scalability for FF with version 136+.
+                    || (codec === CodecMimeType.VP8 && browser.isFirefox());
```
saghul, line 952: Should we check for the FF version?

modules/RTC/TPCUtils.js
hristoterezov: I haven't done comparison on CPU usage, etc with VP8 if this is what you mean. Otherwise during testing I haven't noticed any issues. 

If I have to speculate it should be fine since we had it enabled and on FF we don't have other codec that will use SVC, all of them will be using simulcast....

Do you want me to compare it more formally with VP8 and post the results? How do we normally do this?

modules/RTC/TPCUtils.js
hristoterezov: No need. This function will be called Only for FF 136+ anyway because the previous versions do not support DD.

jallamsetty1: We should get rid of this method all together since the bridge now gets scalability info from DD for all the video codecs. Earlier it used to parse them only for AV1 and H.264. Once DD is offered by Jicofo and negotiated by the  client, I see no reason to remove it from the remote desc. @JonathanLennox pls correct me if I am wrong.

modules/browser/BrowserCapabilities.js
```diff
@@ -233,8 +233,7 @@ export default class BrowserCapabilities extends BrowserDetection {
      * @returns {boolean}
      */
     supportsScalabilityModeAPI() {
-        return (this.isChromiumBased() && this.isEngineVersionGreaterThan(112))
-            || (this.isFirefox() && this.isVersionGreaterThan(135));
+        return this.isChromiumBased() && this.isEngineVersionGreaterThan(112);
```
jallamsetty1, line 236: We will have to bring the check back in for the reason mentioned above where we default to Simulcast for Firefox until the Firefox bug is fixed.

modules/RTC/TPCUtils.js
jallamsetty1: I'd rather use the setting from `useSimulcast` from codec settings to determine if the encoder is configured to use simulcast. This way it is easy to just flip the flag in config.js to test if the Firefox bug is fixed.
`this.codecSettings[videoCodec].scalabilityModeEnabled && this.codecSettings[videoCodec].useSimulcast`

jallamsetty1 (changes requested):

modules/RTC/TPCUtils.js
hristoterezov: Awesome! 

I suggest to remove it as a separate PR.

modules/browser/BrowserCapabilities.js
hristoterezov: Do you mean we should bring it back once the bug is fixed? Meaning no actionable item is related to this comment? 


Also just FYI: There actually isn't a FF bug. I think they just don't support scalabilityModeEnabled prop at all. They never implemented it. At least this is how I understand the current situation with FF.

modules/RTC/TPCUtils.js
```diff
@@ -723,10 +723,16 @@ export class TPCUtils {
...
 
-            // K-SVC mode for VP9 when no scalability mode is set. Though only one outbound-rtp stream is present,
-            // three separate encodings have to be configured.
+            // For FF: scalabilityModeEnabled is not supported and we have to use simulcast.
+            // For other browsers we use K-SVC mode for VP9 when no scalability mode is set. Although
+            // only one outbound-rtp stream is present, three separate encodings have to be configured.
             || (!this.codecSettings[videoCodec].scalabilityModeEnabled && videoCodec === CodecMimeType.VP9)
 
+            // FF uses simulcast with AV1.
+            || (!this.codecSettings[videoCodec].scalabilityModeEnabled
```
jallamsetty1, line 732: Why do we need this extra check here? Can't we just remove the check for H.264 on line 739?

hristoterezov: Oh. Are you saying remove the av1, vp9 and h264 checks and leave only:
Vp8 || (!this.codecSettings[videoCodec].scalabilityModeEnabled && this.codecSettings[videoCodec].useSimulcast)

hristoterezov: ok. I'm confused. what is your proposal for changing the code.

modules/browser/BrowserCapabilities.js
```diff
@@ -251,7 +251,19 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+        // TODO: enable for FF with version 136+ once media.webrtc.simulcast.vp9.enabled is set to true by default.
+        return !(this.isWebKitBased() || this.isFirefox());
+    }
+
+    /**
+     * Returns true if K-SVC is supported for VP9.
+     *
+     * @returns {boolean}
+     */
+    supportsKSVCForVP9() {
```
jallamsetty1, line 274: Also, the terms `supportsKSVCForVP9` and `supportsKSVCForAV1` are misleading here since K-SVC for VP9 means the old Chrome mode where the browser will send a single scalable stream but the client had to configure 3 different streams as if it were doing simulcast. That mode was never supported by Firefox or Safari. You should rename it to just supportsSVC.

jallamsetty1 (approved):

modules/qualitycontrol/CodecSelection.js
```diff
@@ -39,10 +39,16 @@ export class CodecSelection {
...
         for (const connectionType of Object.keys(options)) {
-            // eslint-disable-next-line prefer-const
-            let { disabledCodec, preferredCodec, preferenceOrder, screenshareCodec } = options[connectionType];
+            let { disabledCodec, preferredCodec, preferenceOrder } = options[connectionType];
+            const { enableAV1ForFF = false, screenshareCodec } = options[connectionType];
             const supportedCodecs = new Set(this._getSupportedVideoCodecs(connectionType));
 
+            // Adding this flag for testing purposes since AV1 is not properly supported by FF (it doesn't support SVC
+            // and it doesn't send temporal layers with simulcast).
+            if (browser.isFirefox() && !enableAV1ForFF) {
```
jallamsetty1, line 48: Can we disable AV1 the way we did it for VP9? Like AV1 decode will be allowed but not encode by pushing it to the end of the list? That way none of the endpoints doing AV1 will have to switch to VP9 when Firefox joins the call.

JitsiConference.js
```diff
@@ -467,7 +468,8 @@ JitsiConference.prototype._init = function(options = {}) {
             preferredCodec: _getCodecMimeType(config.p2p?.preferredCodec),
             screenshareCodec: browser.isMobileDevice()
                 ? _getCodecMimeType(config.p2p?.mobileScreenshareCodec)
-                : _getCodecMimeType(config.p2p?.screenshareCodec)
+                : _getCodecMimeType(config.p2p?.screenshareCodec),
+            enableAV1ForFF: options.testing?.enableAV1ForFF
```
jallamsetty1, line 472: Do we have to disable AV1 for p2p? Spatial scalability shouldn't matter for p2p case right?

modules/qualitycontrol/CodecSelection.js
hristoterezov: Done

JitsiConference.js
hristoterezov: Isn't it the temporal scalability we are worried about?

hristoterezov: Done

PR #2771, opened 2025-04-24
fix(codec) Always use DD when its supported by Jicofo and browser.

jallamsetty1:
The bridge is now able to use DD headers for VP8 and VP9 as well so there is no need to remove it from remote desc when the call switches from AV1 to VP8/VP9.


saghul (commented):
How does LJM feature-detect this?

Will the JVB choke if a new LJM connects to an older JVB?

jallamsetty1:
> How does LJM feature-detect this?
> 
> Will the JVB choke if a new LJM connects to an older JVB?

The change in JVB for handling DD headers for VP8 and VP9 was merged last May so we are not worried about running into this case.

saghul (approved):

PR #2772, opened 2025-04-29
fix(remoteControl): Failure to start

hristoterezov:
It appears that when the gDM flow is used we do not set properly the sourceId of the shared stream.  When the user has multiple screens we use the sourceId to identify the display that is currently shared. Therefore in electron we can't identify the display that is used and we fail.

## Types of changes

## Checklist:


saghul (approved):
Nice one!

PR #2776, opened 2025-05-12
feat: Drops a fix for ignoring startmuted locally.

damencho:


saghul:
Won't this be a problem for an old LJM connecting to a new(er) jicofo?

damencho:
Nope, we will keep for some time the start muted that jicofo sends via jingle. This PR is not needed in my opinion, but we will discuss it later today wiot @jallamsetty1, she is working on more jitsi-meet and ljm changes about start muted.

saghul:
Is this still relevant?

jallamsetty1:
> Is this still relevant?

Nope, the fix was implemented in the backend. we don't need this anymore.

PR #2782, opened 2025-05-23
feat(conference-request): Attach token to the conference request sent to jicofo.

damencho:


modules/xmpp/moderator.js
```diff
@@ -323,7 +324,10 @@ export default class Moderator extends Listenable {
                 fetch(this.targetUrl, {
                     method: 'POST',
                     body: JSON.stringify(this._createConferenceRequest(roomJid)),
-                    headers: { 'Content-Type': 'application/json' }
+                    headers: {
+                        'Content-Type': 'application/json',
+                        'Authorization': `Bearer ${this.xmpp.token}`
```
saghul, line 329: Shouldn't we skip the header if there is no token?

saghul (approved):

saghul (approved):

bgrozev (approved):

PR #2787, opened 2025-05-28
[feat] Migrate `detection/NoAudioSignalDetection`

naman9271:
Fixes #2637

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -1,7 +1,7 @@
 export {};
 
 declare global {    
-    type Timeout = NodeJS.Timeout | ReturnType<typeof setTimeout>;
+    type Timeout = ReturnType<typeof setTimeout>;
     interface Window {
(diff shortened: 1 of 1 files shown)


modules/detection/NoAudioSignalDetection.ts
```diff
@@ -17,38 +20,47 @@ const SILENCE_PERIOD_MS = 4000;
  * @fires DetectionEvents.NO_AUDIO_INPUT
  */
 export default class NoAudioSignalDetection extends EventEmitter {
+    private _conference: JitsiConference;
+    private _timeoutTrigger: NodeJS.Timeout | null;
```
saghul, line 24: We might need to define the timeout type somewhere global, but this is not it.

saghul (approved):

saghul:

PR #2792, opened 2025-05-30
Migrate `modules/browser`

naman9271:
Fixes #2637 

- I started migration of `jitsi conference` but before that it requires the migration of this file as there are some methods like `isFirefox` which is unable to detect from browser.
- and while its migration i faced a same issue like ts compliler unable to detect methods that are defined in `'@jitsi/js-utils'`

image


jallamsetty1:
> @saghul , @jallamsetty1 could you please look into this, like all the OS methods are undetectable by ts-compiler and i have tried directly import the class where th emethods are defined it still not able to detect them......is there anything i am missing image

It is because the type definitions for these methods are missing in js-utils. I started migrating BrowserCapabilities to TS but I completely dropped the ball  - https://github.com/jitsi/js-utils/pull/69. I will try to get that merged soon so that you can proceed with these changes.

PR #2799, opened 2025-05-31
Migrate `e2ee/context`

naman9271:
Fixes #2637


naman9271:
fo this `worker.js` has also need to be migrated so for then i put it on draft!

webpack.config.js
```diff
@@ -18,7 +18,7 @@ module.exports = (_env, argv) => {
                 library: 'JitsiMeetJS',
                 libraryTarget: 'umd',
                 path: path.join(process.cwd(), 'dist', 'umd') } },
-        {
```
saghul: Fix this like we discussed in the other PR.

saghul:

saghul (approved):

damencho:
Not sure why but this failed when building ljm with
```
modules/e2ee/Worker.ts(98,43): error TS2322: Type 'ArrayBuffer' is not assignable to type 'boolean'.
```

naman9271:
> Not sure why but this failed when building ljm with

Yeah i see, I have raised a PR ( #2896 ) for this ... Please take a look.

saghul:
I think it was due to the order in which the PRs was merged.

The merge queue would help but it has other drawbacks...

PR #2800, opened 2025-05-31
Migrated `e2ee/worker`

naman9271:
Fixes #2637

### DIFF
--- a/modules/e2ee/Worker.ts
+++ b/modules/e2ee/Worker.ts
@@ -4,9 +4,9 @@
 
 import { Context } from './Context';
 
-const contexts = new Map(); // Map participant id => context
+const contexts: Map<string, Context> = new Map(); // Map participant id => context
 
-let sharedContext;
+let sharedContext: Context | undefined;
 
 let enabled = false;
 
@@ -16,7 +16,7 @@ let enabled = false;
  * @param {string} participantId - The participant whose context we need.
  * @returns {Object} The context.
  */
-function getParticipantContext(participantId) {
+function getParticipantContext(participantId: string): Context {
     if (sharedContext) {
         return sharedContext;
     }
@@ -28,7 +28,7 @@ function getParticipantContext(participantId) {
         contexts.set(participantId, context);
     }
 
-    return contexts.get(participantId);
+    return contexts.get(participantId)!;
 }
 
 /**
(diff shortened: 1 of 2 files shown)


modules/e2ee/Worker.ts
```diff
@@ -65,25 +99,33 @@ onmessage = event => {
         }
     } else if (operation === 'encode' || operation === 'decode') {
         const { readableStream, writableStream, participantId } = event.data;
-        const context = getParticipantContext(participantId);
 
-        handleTransform(context, operation, readableStream, writableStream);
+        if (readableStream && writableStream && participantId) {
```
saghul: This is too defensive, throw an exception if they are not there, as that would be a prigramming error.

modules/e2ee/Worker.ts
```diff
@@ -65,25 +99,33 @@ onmessage = event => {
...
         contexts.forEach(context => context.setEnabled(enabled));
     } else if (operation === 'setKey') {
         const { participantId, key, keyIndex } = event.data;
-        const context = getParticipantContext(participantId);
 
-        if (key) {
-            context.setKey(key, keyIndex);
-        } else {
-            context.setKey(false, keyIndex);
+        if (participantId && keyIndex !== undefined) {
```
saghul: Ditto

modules/e2ee/Worker.ts
```diff
@@ -65,25 +99,33 @@ onmessage = event => {
...
+                context.setKey(new Uint8Array(key), keyIndex);
+            } else {
+                context.setKey(false, keyIndex);
+            }
         }
     } else if (operation === 'cleanup') {
         const { participantId } = event.data;
 
-        contexts.delete(participantId);
+        if (participantId) {
```
saghul: Ditto

webpack.config.js
```diff
@@ -18,9 +18,9 @@ module.exports = (_env, argv) => {
                 library: 'JitsiMeetJS',
                 libraryTarget: 'umd',
                 path: path.join(process.cwd(), 'dist', 'umd') } },
-        {
+        { ...config,
```
saghul: Fix like we discussed.

saghul (approved):

saghul:

PR #2801, opened 2025-06-01
Migrate `modules\videosipgw\JitsiVideoSIPGWSession.ts`

naman9271:
Fixes #2637 

- Changed type of listener from function to EventListener 
`Argument of type 'Function' is not assignable to parameter of type 'EventListener'.
  Type 'Function' provides no match for the signature '(...args: any[]): void'.ts(`

image

### DIFF
--- a/modules/videosipgw/JitsiVideoSIPGWSession.ts
+++ b/modules/videosipgw/JitsiVideoSIPGWSession.ts
@@ -1,13 +1,11 @@
 import { getLogger } from '@jitsi/logger';
 import { $iq } from 'strophe.js';
-
+import ChatRoom from '../xmpp/ChatRoom';
 import Listenable from '../util/Listenable';
 
 import * as VideoSIPGWConstants from './VideoSIPGWConstants';
 
 const logger = getLogger('modules/videosipgw/JitsiVideoSIPGWSession');
-
-import ChatRoom from '../xmpp/ChatRoom';
 
 /**
  * The event name for current sip video session state changed.
--- a/modules/videosipgw/JitsiVideoSIPGWSession.ts
+++ b/modules/videosipgw/JitsiVideoSIPGWSession.ts
@@ -6,18 +6,23 @@ import Listenable from '../util/Listenable';
 import * as VideoSIPGWConstants from './VideoSIPGWConstants';
 
 const logger = getLogger('modules/videosipgw/JitsiVideoSIPGWSession');
+import ChatRoom  from '../xmpp/ChatRoom';
 
 /**
  * The event name for current sip video session state changed.
  * @type {string} event name for sip video session state changed.
  */
-const STATE_CHANGED = 'STATE_CHANGED';
+const STATE_CHANGED: string = 'STATE_CHANGED';
 
 /**
  * Jitsi video SIP GW session. Holding its state and able to start/stop it.
  * When session is in OFF or FAILED stated it cannot be used anymore.
  */
 export default class JitsiVideoSIPGWSession extends Listenable {
+    sipAddress: string;
+    displayName: string;
+    chatRoom: ChatRoom;
+    state: string | undefined;
 
     /**
      * Creates new session with the desired sip address and display name.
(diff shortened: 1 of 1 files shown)


modules/videosipgw/JitsiVideoSIPGWSession.ts
```diff
@@ -6,18 +6,23 @@ import Listenable from '../util/Listenable';
 import * as VideoSIPGWConstants from './VideoSIPGWConstants';
 
 const logger = getLogger('modules/videosipgw/JitsiVideoSIPGWSession');
+import ChatRoom  from '../xmpp/ChatRoom';
```
saghul: Imports should go at the top. Please rebase to make sure the linter picks it up.

modules/videosipgw/JitsiVideoSIPGWSession.ts
```diff
@@ -6,18 +6,23 @@ import Listenable from '../util/Listenable';
...
 
 /**
  * Jitsi video SIP GW session. Holding its state and able to start/stop it.
  * When session is in OFF or FAILED stated it cannot be used anymore.
  */
 export default class JitsiVideoSIPGWSession extends Listenable {
+    sipAddress: string;
+    displayName: string;
+    chatRoom: ChatRoom;
+    state: string | undefined;
```
saghul: Why not `state?: string;`. ?

modules/videosipgw/JitsiVideoSIPGWSession.ts
```diff
@@ -28,7 +33,7 @@ export default class JitsiVideoSIPGWSession extends Listenable {
      * that participant.
      * @param {ChatRoom} chatRoom - The chat room this session is bound to.
      */
-    constructor(sipAddress, displayName, chatRoom) {
+    constructor(sipAddress: string, displayName: string, chatRoom: any) {
```
saghul: you have the type for `chatRoom`

modules/videosipgw/JitsiVideoSIPGWSession.ts
```diff
@@ -129,8 +134,13 @@ export default class JitsiVideoSIPGWSession extends Listenable {
...
      * @param {string} action - The action to send ('start' or 'stop').
      */
-    _sendJibriIQ(action) {
-        const attributes = {
+    private _sendJibriIQ(action: string): void {
+        const attributes: {
+            xmlns: string;
+            action: string;
+            sipaddress: string;
+            displayname?: string;
```
saghul: why not add the displayname directly here? that way you won't need to define the shape of the object.

modules/videosipgw/JitsiVideoSIPGWSession.ts
```diff
@@ -148,10 +158,11 @@ export default class JitsiVideoSIPGWSession extends Listenable {
...
             iq,
             () => {}, // eslint-disable-line no-empty-function
-            error => {
+            (error: any) => {
                 logger.error(
                     `Failed to ${action} video SIP GW session, error: `, error);
                 this.setState(VideoSIPGWConstants.STATE_FAILED);
-            });
+            },
+            undefined);
```
saghul, line 159: What is this and why does it need to be explicitly passed?

saghul (changes requested):
.

modules/videosipgw/JitsiVideoSIPGWSession.ts
naman9271: this function is expecting 4 arguments and timeout:number is missing so i put undefined there to fix that error
image

naman9271:
@saghul Done all the changes suggested

modules/videosipgw/JitsiVideoSIPGWSession.ts
```diff
@@ -7,17 +7,23 @@ import * as VideoSIPGWConstants from './VideoSIPGWConstants';
 
 const logger = getLogger('modules/videosipgw/JitsiVideoSIPGWSession');
 
+import ChatRoom from '../xmpp/ChatRoom';
```
saghul: This is still not were it should be.

Let's see if the linter catches it.

naman9271: Fixed.

saghul (approved):

saghul:

PR #2803, opened 2025-06-01
Migrate `modules\util\Deferred.js`

naman9271:
Fixes #2637

### DIFF
--- a/globals.d.ts
+++ b/globals.d.ts
@@ -1,6 +1,7 @@
 export {};
 
 declare global {
+    type Timeout = ReturnType<typeof setTimeout>;
     interface Window {
         connectionTimes: any;
     }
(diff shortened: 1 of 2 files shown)


globals.d.ts
```diff
@@ -1,6 +1,7 @@
 export {};
 
 declare global {
+    type Timeout = ReturnType<typeof setTimeout>;
```
saghul, line 4: Does using `window.setTimeout` avoid having to define this?

saghul:

saghul:
Can you fix the conflict pl?

globals.d.ts
naman9271: @saghul actually this is already merged under another PR so i think its better we use this

naman9271:
> Can you fix the conflict pl?

conflict resolved.

saghul:

PR #2811, opened 2025-06-05
feat(lint, ts) Add member-ordering rule for TS

jallamsetty1:
## Types of changes

## Checklist:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,5 +1,29 @@
 module.exports = {
+    parser: '@typescript-eslint/parser',
     extends: [
         '@jitsi/eslint-config'
-    ]
+    ],
+    plugins: [
+        '@typescript-eslint'
+    ],
+    rules: {
+        '@typescript-eslint/member-ordering': [
+            'error',
+            {
+                default: [
+                    'signature',
+                    'private-static-field',
+                    'protected-static-field',
+                    'public-static-field',
+                    'private-instance-field',
+                    'protected-instance-field',
+                    'public-instance-field',
+                    'constructor',
+                    'private-instance-method',
+                    'protected-instance-method',
+                    'public-instance-method'
+                ]
+            }
+        ]
+    }
 };
--- a/package.json
+++ b/package.json
@@ -16,7 +16,7 @@
   "author": "",
   "readmeFilename": "README.md",
   "dependencies": {
-    "@jitsi/js-utils": "2.2.1",
+    "@jitsi/js-utils": "2.4.6",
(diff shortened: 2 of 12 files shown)


.eslintrc.js
```diff
@@ -1,5 +1,29 @@
 module.exports = {
+    parser: '@typescript-eslint/parser',
     extends: [
         '@jitsi/eslint-config'
-    ]
+    ],
+    plugins: [
+        '@typescript-eslint'
+    ],
+    rules: {
```
saghul, line 9: I think it's simpler and avoids having to depend directly on the parser if you use an override like so: https://github.com/jitsi/jitsi-meet/blob/f060ab9d26478dc3fe49f17e38a7994225a295fc/react/.eslintrc.js#L11

.eslintrc.js
```diff
@@ -1,5 +1,29 @@
...
+        '@typescript-eslint/member-ordering': [
+            'error',
+            {
+                default: [
```
saghul, line 13: Why do we want to define this order? Don't the defaults work? Too much churn?

package.json
```diff
@@ -42,6 +42,8 @@
     "@types/lodash-es": "4.17.12",
     "@types/sdp-transform": "2.4.5",
     "@types/strophe": "1.2.37",
+    "@typescript-eslint/eslint-plugin": "8.33.1",
```
saghul: We should already be transitively depending on these, see above a potential way to avoid depending on them directly.

.eslintrc.js
jallamsetty1: The default order is the reverse of what we wanted, public first followed by protected and private and will result in a lot of churn.

.eslintrc.js
jallamsetty1: done

.eslintrc.js
saghul: 👍

tsconfig.json
```diff
@@ -14,6 +14,11 @@
       "ES2024.Promise"
     ],
     "outDir": "dist/esm/",
+    "paths": {
+      "@jitsi/js-utils": [
```
saghul: Do we need this be a use it cannot find the types? In that case we probably want to add a "types" entry to the package json of js-utils and avoid this.

jallamsetty1: Oh looks like its not needed after all. I was trying few different things to make it work. I will remove this.

saghul (approved):

PR #2819, opened 2025-06-07
migrate Sdp transform

naman9271:
Fixes #2637

PR #2820, opened 2025-06-07
Migrate Rtx modifier

naman9271:
Fixes #2637

### DIFF
--- a/modules/sdp/RtxModifier.ts
+++ b/modules/sdp/RtxModifier.ts
@@ -10,6 +10,30 @@ import { SdpTransformWrap, parseSecondarySSRC } from './SdpTransformUtil';
 const logger = getLogger('modules/sdp/RtxModifier');
 
 /**
+ * Helper interfaces for type safety.
+ */
+export interface IPrimarySsrcInfo {
+    cname?: string;
+    id: number;
+    msid?: string;
+}
+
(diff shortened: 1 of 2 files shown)


modules/sdp/RtxModifier.ts
```diff
@@ -9,6 +9,30 @@ import { SdpTransformWrap, parseSecondarySSRC } from './SdpTransformUtil';
...
+/**
+ * Helper interfaces for type safety.
+ */
+export interface IPrimarySsrcInfo {
+    cname?: string;
+    id: number;
+    msid?: string;
+}
+
+export interface IMLineWrap {
```
jallamsetty1: Can we migrate SDPTransformUtil.js to TS first and use its type here instead of creating a separate interface for it. This way we don't run into the risk of the interface and implementation diverging and we don't have to keep updating interface here when the class is modified.

naman9271: @jallamsetty1 yes sure i have already raised a PR for SDPTransformUtils.js (#2819 ) Please review

naman9271:
@jallamsetty1 , @saghul conflicts resolved PTAL.

naman9271:
@jallamsetty1 @saghul PTAl.

PR #2826, opened 2025-06-07
Migrate `modules\detection\TrackVADEmitter.ts`

naman9271:
Fixes #2637

### DIFF
--- a/modules/detection/TrackVADEmitter.ts
+++ b/modules/detection/TrackVADEmitter.ts
@@ -1,8 +1,19 @@
+import JitsiLocalTrack from '../RTC/JitsiLocalTrack';
 import RTC from '../RTC/RTC';
 import EventEmitter from '../util/EventEmitter';
 import { createAudioContext } from '../webaudio/WebAudioUtils';
 
 import { VAD_SCORE_PUBLISHED } from './DetectionEvents';
+
+export interface IVadProcessor {
+    calculateAudioFrameVAD: (pcmSample: Float32Array | number[]) => number;
+    getRequiredPCMFrequency: () => number;
+    getSampleLength: () => number;
+}
+
+export type AudioContextType = AudioContext;
+export type ScriptProcessorNodeType = ScriptProcessorNode;
+export type MediaStreamAudioSourceNodeType = MediaStreamAudioSourceNode;
 
 /**
  * Connects an audio JitsiLocalTrack to a vadProcessor using WebAudio ScriptProcessorNode.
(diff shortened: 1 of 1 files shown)


naman9271:
`_onAudioProcess` property renamed to `_onAudioProcessHandler` as it is duplicate name both used in a method and property as well so to resolve error i have renamed it

modules/detection/TrackVADEmitter.ts
```diff
@@ -187,26 +212,25 @@ export default class TrackVADEmitter extends EventEmitter {
...
     /**
      * Get the associated track label.
      *
      * @returns {string}
      */
-    getTrackLabel() {
-        return this._localTrack.getDeviceLabel();
+    getTrackLabel(): string {
+        return (this._localTrack as any).getDeviceLabel();
     }
```
naman9271: in this i used `any` as there is not any function `getDeviceLabel` defined in the `JitsiLocalTrack`, if any better solution please tell.

modules/detection/TrackVADEmitter.ts
```diff
@@ -144,15 +169,15 @@ export default class TrackVADEmitter extends EventEmitter {
             });
         }
 
-        this._bufferResidue = completeInData.slice(i, completeInData.length);
+        this._bufferResidue = new Float32Array(completeInData.slice(i, completeInData.length));
```
saghul, line 172: `slice` already creates a new array, why the extra array now?

modules/detection/TrackVADEmitter.ts
```diff
@@ -163,11 +188,11 @@ export default class TrackVADEmitter extends EventEmitter {
      *
      * @returns {void}
      */
-    _disconnectAudioGraph() {
+    _disconnectAudioGraph(): void {
         // Even thought we disconnect the processing node it seems that some callbacks remain queued,
         // resulting in calls with and uninitialized context.
         // eslint-disable-next-line no-empty-function
-        this._audioProcessingNode.onaudioprocess = () => {};
+        this._audioProcessingNode.onaudioprocess = () => {}; // eslint-disable-line  @typescript-eslint/no-empty-function
```
saghul: Update the comment above instead.

modules/detection/TrackVADEmitter.ts
saghul: I think we can drop the function altogether.

modules/detection/TrackVADEmitter.ts
```diff
@@ -187,26 +212,25 @@ export default class TrackVADEmitter extends EventEmitter {
      *
      * @returns {string}
      */
-    getDeviceId() {
+    getDeviceId(): string {
```
saghul: Same with this one, I think we can drop it.

modules/detection/TrackVADEmitter.ts
naman9271: yes it is creating a new array but i explicity cretaed as it is giving error that is missing number[] so i did this..... is tehre any better solution fo rthis please let me know. and I ahev solevd all the changes suggested in my latest commit
image

saghul:

naman9271:
@saghul , @jallamsetty1 fixed conflixts ...PTAL.

naman9271:
@jallamsetty1 @saghul PTAL.

jallamsetty1:
Jenkins test this please.

PR #2793, opened 2025-05-30
Update jitsi conference to ES6 Class

naman9271:
Fixes #2637

PR #2829, opened 2025-06-13
feat(file-sharing): Adds file sharing options.

damencho:


naman9271:
hey @damencho could you please hold this PR for merging as its changes lots of file and i have already opened a PR from them so i got conflict in them.
Thanks : )

damencho:
@naman9271 for which PRs I need to wait before rebasing this one?

naman9271:
Hi @damencho, can merge it now....... #2793 have merge succesfully.
Thanks for understanding : )

modules/xmpp/FileSharing.ts
```diff
@@ -0,0 +1,108 @@
...
+    /**
+     * Returns the file sharing identity type (service name).
+     *
+     * @returns {string} the file sharing service name.
+     */
+    getIdentityType() {
+        return IDENTITY_TYPE;
+    }
+
+    addFile(metadata: any) {
```
saghul: Pl document the type since this becomes part of the API.

saghul: Do you reckon it would be useful to have an API to add multiple files in a single shot?

saghul (approved):
LGTM with a couple of suggestions.

modules/xmpp/FileSharing.ts
damencho: Nope, we upload all files one by one: 
https://github.com/jitsi/jitsi-meet/blob/b0472e46774b6f70efa9bd98259766ac91ab6073/react/features/file-sharing/middleware.web.ts#L56

And when the upload returns 200 we add it to the list of files and broadcast the change to the other participants.

saghul: Can you document the metadata with a TS interface? This way it will bubble up to JM.

PR #2831, opened 2025-06-17
feat: Enable updating local audio track with new constraints

hlvhe:


modules/RTC/JitsiLocalTrack.js
```diff
@@ -977,4 +977,55 @@ export default class JitsiLocalTrack extends JitsiTrack {
     unmute() {
         return this._queueSetMuted(false);
     }
+
+    /**
+     * Applies media constraints to the current MediaStreamTrack.
+     *
+     * @param {Object} constraints - Media constraints to apply.
```
jallamsetty1: Can you document the acceptable properties of this parameter so that they can be extracted to a TS interface when we migrate this file from JS to TS?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -977,4 +977,55 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            ...initialSettings,
+            ...constraints
+        };
+        let mediaStreamData;
+
+        try {
+            const mediaStreamsData
+                = await RTCUtils.obtainAudioAndVideoPermissions({
+                    devices: [ mediaType ],
+                    micDeviceId: constraintsToApply.deviceId,
```
jallamsetty1: Goal is to make `applyConstraints` a generic method that can be called on video tracks as well. So lets check the mediaType and pass `micDeviceId` or `cameraDeviceId` accordingly.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -977,4 +977,55 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+                    constraints: { [mediaType]: constraintsToApply }
+                });
+
+            mediaStreamData = mediaStreamsData[0];
+
+            if (!mediaStreamData) {
+                throw new Error('No media stream data');
```
jallamsetty1: Can you add some more information in the error message here? pls log the track and the constraints passed.

modules/RTC/RTCUtils.js
```diff
@@ -140,22 +140,19 @@ function getConstraints(um = [], options = {}) {
...
         if (!constraints.audio || typeof constraints.audio === 'boolean') {
-            constraints.audio = {};
+            constraints.audio = {
+                autoGainControl: !disableAGC && !disableAP,
+                echoCancellation: !disableAEC && !disableAP,
+                noiseSuppression: !disableNS && !disableAP
+            };
+            if (stereo) {
+                Object.assign(constraints.audio, { channelCount: 2 });
+            }
```
jallamsetty1, line 150: Don't we need to sanitize the constraints in the else case and make sure we don't have unsupported constraints that might result in a gUM error?

JitsiMeetJS.ts
```diff
@@ -538,7 +538,6 @@ export default {
                     domain: jaasDomain,
                     muc: `conference.${appId}.${jaasDomain}`
                 },
-                conferenceRequestUrl: `https://${jaasDomain}/${appId}/conference-request/v1?room=${roomName}`,
```
jallamsetty1, line 541: Did you delete this by mistake?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -977,4 +977,55 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+            }
+        } catch (error) {
+            this.stream.getTracks().forEach(track => {
```
saghul: Tracks don't have a start method.

modules/RTC/JitsiLocalTrack.js
```diff
@@ -977,4 +977,55 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+                }
+            });
+
+            return;
+        }
+
+        await this.conference._removeLocalTrackFromPc(this);
+
+        this._setStream(mediaStreamData.stream);
+        this.track = mediaStreamData.track;
```
saghul: How will user code know the stream and track have been replaced?

JitsiMeetJS.ts
jallamsetty1: Oh I see that the LJM getting started sample was modified to use JitsiMeetJS.joinConference() instead of JitsiConference.join(). Unfortunately, we cannot delete this line here. Maybe we need to keep both modes in the sample ? Or make changes to the sample locally for testing purposes and not check those in? @saghul WDYT? We were hoping to update the sample to use the new method in JitsiLocalTrack.

saghul: How does this line cause any trouble?

modules/RTC/JitsiLocalTrack.js
jallamsetty1: Good catch! I missed it. Maybe move this.stopStream() to after the track is removed from PC?

modules/RTC/JitsiLocalTrack.js
jallamsetty1: We only need to reattach the new MediaStream to the corresponding containers for video tracks(which we haven't tested yet) but other than that I don't see why this would be needed for audio tracks. This is how `setEffect` is also implemented.

saghul: Right, not for audio tracks, but something is needed for video tracks? Also note that RN can't re-attach itself to the containers...

jallamsetty1: Oh ok, lets restrict this to audio tracks for now so that Hugo can proceed with the rest of the work for the GSOC project?

saghul: Sure thing!

JitsiMeetJS.ts
hlvhe: With ljm-getting-started, there is a CORS error: Access to fetch at 'https://8x8.vc/vpaas-magic-cookie-AppID/conference-request/v1?room=room' from origin 'http://127.0.0.1:8000' has been blocked by CORS policy: Request header field authorization is not allowed by Access-Control-Allow-Headers in preflight response.

modules/RTC/JitsiLocalTrack.js
hlvhe: The constraints are only applied when the MediaStreamTrack is stopped. I will try to reuse the previous track in case of an error.

JitsiMeetJS.ts
saghul: @aaronkvanmeerten Can we get that limitation lifted please?

modules/RTC/JitsiLocalTrack.js
saghul: 👍

modules/RTC/JitsiLocalTrack.js
```diff
@@ -977,4 +977,61 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+
+            await this.conference._addLocalTrackToPc(this);
+        } catch (error) {
+            logger.error(
+                `applyConstraints failed for track ${this} with constraints: ${JSON.stringify(
+                    constraints
+                )}: `,
+                error
+            );
+            throw error;
```
jallamsetty1: Aren't we supposed to create a new MediaStream with the old constraints and add it back to the pc if gUM with the new constraint fails?

hlvhe: In Chrome, cloning the previous track seems to prevent new constraints from being applied by gUM. If applyConstraints fails, would it be better to throw an error so that a new MediaStream can be requested from outside the method, similar to how it’s handled in setEffect?

jallamsetty1: It sounds counterintuitive to expect the calling function to create a new media stream and add it back to the conference if this method throws an error. The whole point of adding this method is to hide this complexity from the higher level APIs.

saghul: Also, FWIW, clone() is not implemented on RN so let's avoid it.

jallamsetty1:
Jenkins test this please.

jallamsetty1:
Jenkins test this please.

PR #2834, opened 2025-06-23
Migrate `JitsiConference` to TS

naman9271:
Fixes: #2693

### DIFF
--- a/JitsiConference.ts
+++ b/JitsiConference.ts
@@ -5,20 +5,25 @@ import { Strophe } from 'strophe.js';
 import * as JitsiConferenceErrors from './JitsiConferenceErrors';
 import JitsiConferenceEventManager from './JitsiConferenceEventManager';
 import * as JitsiConferenceEvents from './JitsiConferenceEvents';
+import type JitsiConnection from './JitsiConnection';
 import JitsiParticipant from './JitsiParticipant';
 import JitsiTrackError from './JitsiTrackError';
 import * as JitsiTrackErrors from './JitsiTrackErrors';
 import * as JitsiTrackEvents from './JitsiTrackEvents';
 import authenticateAndUpgradeRole from './authenticateAndUpgradeRole';
+import type JitsiLocalTrack from './modules/RTC/JitsiLocalTrack';
+import type JitsiRemoteTrack from './modules/RTC/JitsiRemoteTrack';
+import type JitsiTrack from './modules/RTC/JitsiTrack';
 import RTC from './modules/RTC/RTC';
 import { SS_DEFAULT_FRAME_RATE } from './modules/RTC/ScreenObtainer';
+import type TraceablePeerConnection from './modules/RTC/TraceablePeerConnection';
 import browser from './modules/browser';
 import ConnectionQuality from './modules/connectivity/ConnectionQuality';
 import IceFailedHandling from './modules/connectivity/IceFailedHandling';
 import * as DetectionEvents from './modules/detection/DetectionEvents';
 import NoAudioSignalDetection from './modules/detection/NoAudioSignalDetection';
 import P2PDominantSpeakerDetection from './modules/detection/P2PDominantSpeakerDetection';
-import VADAudioAnalyser from './modules/detection/VADAudioAnalyser';
+import VADAudioAnalyser, { IVADProcessor } from './modules/detection/VADAudioAnalyser';
 import VADNoiseDetection from './modules/detection/VADNoiseDetection';
 import VADTalkMutedDetection from './modules/detection/VADTalkMutedDetection';
 import { E2EEncryption } from './modules/e2ee/E2EEncryption';
@@ -27,23 +32,33 @@ import Jvb121EventGenerator from './modules/event/Jvb121EventGenerator';
 import FeatureFlags from './modules/flags/FeatureFlags';
 import { LiteModeContext } from './modules/litemode/LiteModeContext';
 import { QualityController } from './modules/qualitycontrol/QualityController';
-import RecordingManager from './modules/recording/RecordingManager';
+import { IReceiverVideoConstraints } from './modules/qualitycontrol/ReceiveVideoController';
+import JibriSession from './modules/recording/JibriSession';
+import RecordingManager, { IRecordingOptions } from './modules/recording/RecordingManager';
 import Settings from './modules/settings/Settings';
 import AvgRTPStatsReporter from './modules/statistics/AvgRTPStatsReporter';
 import LocalStatsCollector from './modules/statistics/LocalStatsCollector';
+import SpeakerStats from './modules/statistics/SpeakerStats';
 import SpeakerStatsCollector from './modules/statistics/SpeakerStatsCollector';
 import Statistics from './modules/statistics/statistics';
-import EventEmitter from './modules/util/EventEmitter';
+import EventEmitter, { EventListener } from './modules/util/EventEmitter';
+import Listenable from './modules/util/Listenable';
 import { isValidNumber, safeSubtract } from './modules/util/MathUtil';
 import RandomUtil from './modules/util/RandomUtil';
 import { getJitterDelay } from './modules/util/Retry';
 import $ from './modules/util/XMLParser';
 import ComponentsVersions from './modules/version/ComponentsVersions';
+import JitsiVideoSIPGWSession from './modules/videosipgw/JitsiVideoSIPGWSession';
 import VideoSIPGW from './modules/videosipgw/VideoSIPGW';
 import * as VideoSIPGWConstants from './modules/videosipgw/VideoSIPGWConstants';
+import BreakoutRooms from './modules/xmpp/BreakoutRooms';
+import type ChatRoom from './modules/xmpp/ChatRoom';
+import FileSharing from './modules/xmpp/FileSharing';
+import type JingleSessionPC from './modules/xmpp/JingleSessionPC';
 import MediaSessionEvents from './modules/xmpp/MediaSessionEvents';
+import RoomMetadata from './modules/xmpp/RoomMetadata';
 import SignalingLayerImpl from './modules/xmpp/SignalingLayerImpl';
-import {
+import XMPP, {
     FEATURE_E2EE,
     FEATURE_JIGASI,
     JITSI_MEET_MUC_TYPE
@@ -74,6 +89,95 @@ import {
 } from './service/statistics/AnalyticsEvents';
 import { XMPPEvents } from './service/xmpp/XMPPEvents';
 
+export interface IConferenceOptions {
+    config: {
+        analytics?: {
+            rtcstatsEnabled?: boolean;
+            rtcstatsEndpoint?: string;
+        };
+        applicationName?: string;
+        avgRtpStatsN?: number;
+        channelLastN?: number;
+        confID?: string;
+        createVADProcessor?: () => IVADProcessor;
+        deploymentInfo?: {
+            userRegion?: string;
+        };
+        disableAudioLevels?: boolean;
+        e2eping?: {
+            enabled?: boolean;
+        };
+        enableNoAudioDetection?: boolean;
+        enableNoisyMicDetection?: boolean;
+        enableTalkWhileMuted?: boolean;
+        hiddenDomain?: string;
+        p2p?: {
+            backToP2PDelay?: number;
+            codecPreferenceOrder?: string[];
+            disabledCodec?: string;
+            enabled?: boolean;
+            mobileCodecPreferenceOrder?: string[];
+            mobileScreenshareCodec?: string;
+            preferredCodec?: string;
+            screenshareCodec?: string;
+        };
+        startAudioMuted?: number;
+        startLastN?: number;
+        startVideoMuted?: number;
+        statisticsDisplayName?: string;
+        statisticsId?: string;
+        testing?: {
+            allowMultipleTracks?: boolean;
+            enableAV1ForFF?: boolean;
+            enableFirefoxP2p?: boolean;
+            forceInitiator?: boolean;
+            forceResponder?: boolean;
+            lastNRampupTime?: number;
+            p2pTestMode?: boolean;
+        };
+        transcriptionLanguage?: string;
+        videoQuality?: {
+            codecPreferenceOrder?: string[];
+            disabledCodec?: string;
+            enableAdaptiveMode?: boolean;
+            mobileCodecPreferenceOrder?: string[];
+            mobileScreenshareCodec?: string;
+            preferredCodec?: string;
+            screenshareCodec?: string;
+        };
+    };
+    connection: JitsiConnection;
+    customDomain?: string;
+    name: string;
+}
+
+export interface IStartMutedPolicy {
+    audio: boolean;
+    video: boolean;
+}
+
+export interface IConferenceProperties {
+    'audio-limit-reached'?: string;
+    'bridge-count'?: string;
+    'video-limit-reached'?: string;
+    'visitor-codecs'?: string;
+    'visitor-count'?: number;
+}
+
+export interface IStatisticsOptions {
+    aliasName?: string;
+    applicationName?: string;
+    confID?: string;
+    roomName?: string;
+    userName?: string;
+}
+
+export interface IStopSessionOptions {
+    reason?: string;
+    reasonDescription?: string;
+    requestRestart?: boolean;
+    sendSessionTerminate?: boolean;
+}
 
 const logger = getLogger('JitsiConference');
 
@@ -82,12 +186,12 @@ const logger = getLogger('JitsiConference');
  * {@link ACTION_JINGLE_SI_TIMEOUT} analytics event is sent (in ms).
  * @type {number}
  */
-const JINGLE_SI_TIMEOUT = 5000;
+const JINGLE_SI_TIMEOUT: number = 5000;
 
 /**
  * Default source language for transcribing the local participant.
  */
-const DEFAULT_TRANSCRIPTION_LANGUAGE = 'en-US';
+const DEFAULT_TRANSCRIPTION_LANGUAGE: string = 'en-US';
 
 /**
  * Checks if a given string is a valid video codec mime type.
@@ -96,9 +200,9 @@ const DEFAULT_TRANSCRIPTION_LANGUAGE = 'en-US';
  * @returns {CodecMimeType|null} mime type if valid, null otherwise.
  * @private
  */
-function _getCodecMimeType(codec) {
+function _getCodecMimeType(codec: string): Nullable<CodecMimeType> {
     if (typeof codec === 'string') {
-        return Object.values(CodecMimeType).find(value => value === codec.toLowerCase());
+        return Object.values(CodecMimeType).find(value => value === codec.toLowerCase()) || null;
     }
 
     return null;
@@ -136,11 +240,75 @@ function _getCodecMimeType(codec) {
  *       {@link JitsiConference.onMemberLeft}
  *       and so on...
  */
-export default class JitsiConference {
+export default class JitsiConference extends Listenable {
+    private _transcribingEnabled?: boolean;
+    private _visitorCodecs?: string[];
+    private _hasVisitors?: boolean;
+    private _sessionInitiateTimeout?: number;
+    private _desktopSharingFrameRate?: number;
+    private _numberOfParticipantsOnJoin?: number;
+    private _delayedIceFailed?: IceFailedHandling;
+    private _audioAnalyser?: VADAudioAnalyser;
+    private _noAudioSignalDetection?: NoAudioSignalDetection;
+    private _signalingLayer: SignalingLayerImpl;
+    private _conferenceJoinAnalyticsEventSent?: number;
+    private _e2eEncryption?: E2EEncryption;
+    private _liteModeContext?: LiteModeContext;
+    private _audioSenderLimitReached?: boolean;
+    private _videoSenderLimitReached?: boolean;
+    private _firefoxP2pEnabled: boolean;
+    private _iceRestarts: number;
+    private _unsubscribers: Array<() => void>;
+    public _statsCurrentId: string;
+    public connection: JitsiConnection;
+    public xmpp: XMPP;
+    public eventEmitter: EventEmitter;
+    public options: IConferenceOptions;
+    public eventManager: JitsiConferenceEventManager;
+    public participants: Map<string, JitsiParticipant>;
+    public componentsVersions: ComponentsVersions;
+    public jvbJingleSession?: JingleSessionPC;
+    public lastDominantSpeaker?: string;
+    public dtmfManager?: object;
+    public somebodySupportsDTMF: boolean;
+    public authEnabled: boolean;
+    public startMutedPolicy: IStartMutedPolicy;
+    public isMutedByFocus: boolean;
+    public mutedByFocusActor?: string;
+    public isVideoMutedByFocus: boolean;
+    public mutedVideoByFocusActor?: string;
+    public wasStopped: boolean;
+    public properties: IConferenceProperties;
+    public connectionQuality: ConnectionQuality;
+    public avgRtpStatsReporter?: AvgRTPStatsReporter;
+    public isJvbConnectionInterrupted: boolean;
+    public speakerStatsCollector: SpeakerStatsCollector;
+    public deferredStartP2PTask?: number;
+    public backToP2PDelay: number;
+    public isP2PConnectionInterrupted: boolean;
+    public p2p: boolean;
+    public p2pJingleSession?: JingleSessionPC;
+    public videoSIPGWHandler: VideoSIPGW;
+    public recordingManager: RecordingManager;
+    public room?: ChatRoom;
+    public e2eping?: E2ePing;
+    public rtc?: RTC;
+    public qualityController?: QualityController;
+    public statistics?: Statistics;
+    public jvb121Status?: Jvb121EventGenerator;
+    public p2pDominantSpeakerDetection?: P2PDominantSpeakerDetection;
+    public authIdentity?: string;
+    public p2pEstablishmentDuration?: number;
+    public jvbEstablishmentDuration?: number;
+    public isDesktopMutedByFocus: boolean;
+    public mutedDesktopByFocusActor?: string;
+
+
     /**
-     * @param {Object} options
+     * @param {IConferenceOptions} options
      */
-    constructor(options) {
+    constructor(options: IConferenceOptions) {
+        super();
         if (!options.name || options.name.toLowerCase() !== options.name.toString()) {
             const errmsg
                 = 'Invalid conference name (no conference name passed or it '
@@ -165,7 +333,6 @@ export default class JitsiConference {
             throw new Error(errmsg);
         }
 
-        this.eventEmitter = new EventEmitter();
         this.options = options;
         this.eventManager = new JitsiConferenceEventManager(this);
 
@@ -249,7 +416,7 @@ export default class JitsiConference {
          */
         this.deferredStartP2PTask = null;
 
-        const delay = Number.parseInt(options.config.p2p?.backToP2PDelay, 10);
+        const delay = Number.parseInt(String(options.config.p2p?.backToP2PDelay || 5), 10);
 
         /**
          * A delay given in seconds, before the conference switches back to P2P
@@ -328,6 +495,8 @@ export default class JitsiConference {
          * Number of times ICE restarts that have been attempted after ICE connectivity with the JVB was lost.
          */
         this._iceRestarts = 0;
+        
+        this._unsubscribers = [];
     }
 
     /**
@@ -342,8 +511,8 @@ export default class JitsiConference {
      * @returns {string}
      * @static
      */
-    static resourceCreator(jid, isAuthenticatedUser) {
-        let mucNickname;
+    static resourceCreator(jid: string, isAuthenticatedUser: boolean): string {
+        let mucNickname: string;
 
         if (isAuthenticatedUser) {
             // For authenticated users generate a random ID.
@@ -370,7 +539,7 @@ export default class JitsiConference {
      * @param options {object}
      * @param options.connection {JitsiConnection} overrides this.connection
      */
-    _init(options) {
+    _init(options: IConferenceOptions): void {
         this.eventManager.setupXMPPListeners();
 
         const { config } = this.options;
@@ -474,6 +643,7 @@ export default class JitsiConference {
 
         if (!this.statistics) {
             this.statistics = new Statistics(this, {
+                // @ts-ignore
                 aliasName: this._statsCurrentId,
                 applicationName: config.applicationName,
                 confID: config.confID ?? `${this.connection.options.hosts.domain}/${this.options.name}`,
@@ -581,7 +751,7 @@ export default class JitsiConference {
      * @param replaceParticipant {boolean} whether the current join replaces
      * an existing participant with same jwt from the meeting.
      */
-    join(password = '', replaceParticipant = false) {
+    join(password: string = '', replaceParticipant: boolean = false) {
         if (this.room) {
             this.room.join(password, replaceParticipant).then(() => this._maybeSetSITimeout());
         }
@@ -595,7 +765,7 @@ export default class JitsiConference {
    * authenticating and upgrading the role of the local participant/user finishes
    * and (2) has a cancel method that allows the caller to interrupt the process.
    */
-    authenticateAndUpgradeRole(options) {
+    authenticateAndUpgradeRole(options) { // TODO: fix types after migration of authenticateAndUpgradeRole.js
         return authenticateAndUpgradeRole.call(this, {
             ...options,
             onCreateResource: JitsiConference.resourceCreator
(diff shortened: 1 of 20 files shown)


globals.d.ts
```diff
@@ -3,7 +3,12 @@ export {};
 declare global {
     type Timeout = ReturnType<typeof setTimeout>;
     interface Window {
-        connectionTimes: any;
+                JitsiMeetJS?: {
```
saghul, line 6: Where are you getting this from?

modules/RTC/JitsiLocalTrack.js
```diff
@@ -356,7 +356,7 @@ export default class JitsiLocalTrack extends JitsiTrack {
      * Sends mute status for a track to conference if any.
      *
      * @param {boolean} mute - If track is muted.
-     * @private
+     * @public
```
saghul: This change is unrelated, please remove it from this PR.

globals.d.ts
naman9271: image

I encounter this here so i have added this in global.d.ts 😅

naman9271:
ping @saghul, @jallamsetty1 .

JitsiConference.ts
```diff
@@ -2892,7 +3072,7 @@ export default class JitsiConference {
      * @param {string} [to=''] - The ID of the recipient endpoint, or empty string to broadcast.
      * @param {boolean} [sendThroughVideobridge=false] - Whether to send through jitsi-videobridge.
      */
-    sendMessage(message, to = '', sendThroughVideobridge = false) {
+    sendMessage(message: any, to = '', sendThroughVideobridge = false): void {
         const messageType = typeof message;
```
naman9271, line 3040: here on using `object` type to message giving error and unable to find its proper type do you know what its type here ?? or should i use `Record<string, unknown>` for this.

saghul:
Can you please drop all the changes unrelated to JitsiConference from the PR? It's already quite large...

JitsiConference.ts
```diff
@@ -136,10 +240,69 @@ function _getCodecMimeType(codec) {
  *       and so on...
  */
 export default class JitsiConference {
-    /**
-     * @param {Object} options
-     */
-    constructor(options) {
+    connection: JitsiConnection;
```
saghul: Add visibility annotations. Default to private and then let's see if we need to "expose" some.

JitsiConference.ts
```diff
@@ -136,10 +240,69 @@ function _getCodecMimeType(codec) {
...
+    xmpp: XMPP;
```
saghul: Make sure you are working with latest master since Jaya landed the sorting in the linting rules.

JitsiConference.ts
```diff
@@ -321,7 +484,7 @@ export default class JitsiConference {
          */
         this._videoSenderLimitReached = undefined;
 
-        this._firefoxP2pEnabled = browser.isVersionGreaterThan(109)
+        this._firefoxP2pEnabled = browser.isVersionGreaterThan(109) // TODO: to be fixed
```
saghul: What needs to be fixed?

JitsiConference.ts
```diff
@@ -370,19 +533,24 @@ export default class JitsiConference {
...
      */
-    _init(options) {
+    _init(options: IConferenceOptions): void {
         this.eventManager.setupXMPPListeners();
 
         const { config } = this.options;
 
         this._statsCurrentId = config.statisticsId ?? Settings.callStatsUserName;
-        this.room = this.xmpp.createRoom(
+        this.xmpp.createRoom(
```
saghul: This is wrong, createRoom doesn't return a Promise.

JitsiConference.ts
```diff
@@ -370,19 +533,24 @@ export default class JitsiConference {
...
             this.options.name, {
                 ...config,
                 statsId: this._statsCurrentId
             },
             JitsiConference.resourceCreator
-        );
+        ).then(room => {
```
saghul: Ditto.

JitsiConference.ts
```diff
@@ -562,7 +731,7 @@ export default class JitsiConference {
 
         // Publish the codec preference to presence.
         this.setLocalParticipantProperty('codecList',
-             this.qualityController.codecController.getCodecPreferenceList('jvb'));
+             this.qualityController.codecController.getCodecPreferenceList('jvb').join(','));
```
saghul: Why are you encoding it as a string now?

JitsiConference.ts
```diff
@@ -730,7 +900,7 @@ export default class JitsiConference {
             // won't be usable anyway. This is done on success automatically
             // by the ChatRoom instance.
             this.getParticipants().forEach(
-        participant => this.onMemberLeft(participant.getJid())
+        participant => this.onMemberLeft(participant.getJid(), 'conference-left')
```
saghul: Why did you add this?

JitsiConference.ts
```diff
@@ -876,7 +1046,7 @@ export default class JitsiConference {
    * @param {MediaType} [mediaType] Optional media type (audio or video).
    * @returns {Array<JitsiLocalTrack>}
    */
-    getLocalTracks(mediaType) {
+    getLocalTracks(mediaType?: MediaType): JitsiLocalTrack[] {
```
saghul, line 1048: Do we rely on passing undefined here, anywhere in the code?

globals.d.ts
saghul: Sorry I cannot seem to open the image, can you copy the code here?

saghul:
I see we do quite a bit of `T | null` and `T | undefined`. Let's take chance to define some helper types that we can use across this project:

```
type Nullable<T> = T | null;
type Optional<T> = T | undefined;
```

And use them in this PR and also moving forward.

JitsiConference.ts
```diff
@@ -916,7 +1086,7 @@ export default class JitsiConference {
    * @param {string} eventId - The event ID.
    * @param {Function} handler - Handler for the event.
    */
-    on(eventId, handler) {
+    on(eventId: string, handler: (...args: any[]) => void): void {
```
saghul: Rather than implement all these methods here, now that this is an ES6 class, please make it inherit from `Listenable` and that should be it.

JitsiConference.ts
```diff
@@ -1025,11 +1195,11 @@ export default class JitsiConference {
...
    * Send presence command.
    * @param {string} name - The name of the command.
-   * @param {Object} values - With keys and values that will be sent.
+   * @param {Record<string, unknown>} values - With keys and values that will be sent.
    */
-    sendCommand(name, values) {
+    sendCommand(name: string, values: Record<string, unknown>): void {
         if (this.room) {
-            this.room.addOrReplaceInPresence(name, values) && this.room.sendPresence();
+            this.room.addOrReplaceInPresence(name, values) && this.room.sendPresence(false);
```
saghul: Let's add a default value in sendPresence instead.

JitsiConference.ts
```diff
@@ -1105,18 +1275,18 @@ export default class JitsiConference {
...
-    getTranscriptionStatus() {
+    getTranscriptionStatus(): string {
         return this.room.transcriptionStatus;
     }
 
     /**
    * Adds JitsiLocalTrack object to the conference.
    * @param {JitsiLocalTrack} track - The JitsiLocalTrack object.
-   * @returns {Promise<JitsiLocalTrack>}
+   * @returns {Promise<void | JitsiLocalTrack>}
```
saghul: Why void? I suppose we want to say undefined? See my note about Optional.

JitsiConference.ts
```diff
@@ -1105,18 +1275,18 @@ export default class JitsiConference {
...
    * @throws {Error} If the specified track is a video track and there is already
    * another video track in the conference.
    */
-    addTrack(track) {
+    addTrack(track: JitsiLocalTrack): Promise<void | JitsiLocalTrack> {
```
saghul: ditto

JitsiConference.ts
```diff
@@ -1153,7 +1323,7 @@ export default class JitsiConference {
                     .then(() => {
                         this._setupNewTrack(track);
                         mediaType === MediaType.VIDEO && this._sendBridgeVideoTypeMessage(track);
-                        this._updateRoomPresence(this.getActiveMediaSession());
+                        this._updateRoomPresence(this.getActiveMediaSession(), undefined);
```
saghul: Add a default value to the ctx parameter in _updateRoomPresence instead.

JitsiConference.ts
```diff
@@ -1170,7 +1340,7 @@ export default class JitsiConference {
                 // before signaling so that a fake participant tile is created for screenshare. Otherwise, presence will
                 // only be sent after a session-accept or source-add is ack'ed.
                 if (track.getVideoType() === VideoType.DESKTOP) {
-                    this._updateRoomPresence(this.getActiveMediaSession());
+                    this._updateRoomPresence(this.getActiveMediaSession(), undefined);
```
saghul: Ditto.

JitsiConference.ts
```diff
@@ -1197,9 +1367,9 @@ export default class JitsiConference {
 
     /**
    * Fires TRACK_MUTE_CHANGED change conference event.
-   * @param {JitsiTrack} track - The JitsiTrack object related to the event.
+   * @param {JitsiLocalTrack} track - The JitsiTrack object related to the event.
```
saghul, line 1328: I think this can fire for both local and remote tracks.

JitsiConference.ts
```diff
@@ -1243,12 +1413,12 @@ export default class JitsiConference {
    * Clear JitsiLocalTrack properties and listeners.
    * @param {JitsiLocalTrack} track - The JitsiLocalTrack object.
    */
-    onLocalTrackRemoved(track) {
+    onLocalTrackRemoved(track: JitsiLocalTrack): void {
         track.setConference(null);
         this.rtc.removeLocalTrack(track);
-        track.removeEventListener(JitsiTrackEvents.TRACK_MUTE_CHANGED, track.muteHandler);
+        track.removeAllListeners(JitsiTrackEvents.TRACK_MUTE_CHANGED);
```
saghul: @jallamsetty1 we need to check if this is kosher, as it will remove listeners added by the application, for example.

JitsiConference.ts
```diff
@@ -1243,12 +1413,12 @@ export default class JitsiConference {
...
         if (track.isAudioTrack()) {
-            track.removeEventListener(JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED, track.audioLevelHandler);
+            track.removeAllListeners(JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED);
```
saghul: Ditto.

JitsiConference.ts
```diff
@@ -1321,7 +1491,7 @@ export default class JitsiConference {
                 if ((oldTrackBelongsToConference && oldTrack?.isVideoTrack()) || newTrack?.isVideoTrack()) {
                     this._sendBridgeVideoTypeMessage(newTrack);
                 }
-                this._updateRoomPresence(this.getActiveMediaSession());
+                this._updateRoomPresence(this.getActiveMediaSession(), undefined);
```
saghul: Use default value.

JitsiConference.ts
```diff
@@ -1362,7 +1532,7 @@ export default class JitsiConference {
             logger.info('_doReplaceTrack - no P2P JingleSession');
         }
 
-        return Promise.all(replaceTrackPromises);
+        return Promise.all(replaceTrackPromises).then(() => { /* No-op */ });
```
saghul: Don't do this, we want the caller to be able to handle the result.

saghul: If not, you can write it like so:

```
await Promise.all(replaceTrackPromises);

return;
```

JitsiConference.ts
```diff
@@ -1475,7 +1645,7 @@ export default class JitsiConference {
             addPromises.push(this.addTrack(track));
         }
 
-        return Promise.allSettled(addPromises);
+        return Promise.allSettled(addPromises).then(() => { /* No-op */ });
```
saghul: Same as above.

JitsiConference.ts
```diff
@@ -1507,15 +1677,15 @@ export default class JitsiConference {
...
+    getRole(): string {
         return this.room.role;
     }
 
     /**
      * Returns whether or not the current conference has been joined as a hidden user.
      * @returns {boolean|null} True if hidden, false otherwise. Will return null if no connection is active.
      */
-    isHidden() {
+    isHidden(): boolean | null {
```
saghul: Let's make this return a boolean only. If there is no connection return false.

JitsiConference.ts
```diff
@@ -1529,7 +1699,7 @@ export default class JitsiConference {
      * @returns {boolean|null} true if local user is moderator, false otherwise. If
      * we're no longer in the conference room then <tt>null</tt> is returned.
      */
-    isModerator() {
+    isModerator(): boolean | null {
```
saghul: Same, return only a boolean.

JitsiConference.ts
```diff
@@ -1547,32 +1717,33 @@ export default class JitsiConference {
...
+    getLastN(): number {
         return this.qualityController.receiveVideoController.getLastN();
     }
 
     /**
      * Obtains the forwarded sources list in this conference.
      * @return {Array<string>|null}
      */
-    getForwardedSources() {
+    getForwardedSources(): string[] | null {
```
saghul: let's force this to always return an array, an empty one if there are no sources.

JitsiConference.ts
```diff
@@ -1584,8 +1755,8 @@ export default class JitsiConference {
      * @throws Error or RangeError if the given value is not a number or is smaller
      * than -1.
      */
-    setLastN(lastN) {
-        if (!Number.isInteger(lastN) && !Number.parseInt(lastN, 10)) {
+    setLastN(lastN: number): void {
```
saghul, line 1727: If we are annotaing the parameter as a number I don't think we need to parse it

JitsiConference.ts
```diff
@@ -1796,7 +1967,7 @@ export default class JitsiConference {
 
         // maybeStart only if we had finished joining as then we will have information for the number of participants
         if (this.isJoined()) {
-            this._maybeStartOrStopP2P();
+            this._maybeStartOrStopP2P(false);
```
saghul: Add a default value.

JitsiConference.ts
```diff
@@ -1973,7 +2144,7 @@ export default class JitsiConference {
         }
 
         if (isSelfPresence) {
-            this.leave().finally(() => this.xmpp.disconnect());
+            this.leave().finally(() => this.xmpp.disconnect(undefined));
```
saghul: Add a default value.

saghul (changes requested):
Made it half-way or so. Good stuff Naman! Please apply all the comments I mentioned all throughout the file.

JitsiConference.ts
naman9271: yes we have to pass undefined here, because there are instances of it using undefined

JitsiConference.ts
naman9271: Done

JitsiConference.ts
naman9271: I typed it void because the promise is returning that type as you can see in the below ss.

image

saghul: Tweaking the source of the issue is likely better than just adjusting the type here.

globals.d.ts
```diff
@@ -2,6 +2,8 @@ export {};
 
 declare global {
     type Timeout = ReturnType<typeof setTimeout>;
+    type Nullable<T> = T | null;
```
saghul: don't put them here, it would pollute the global namespace. put them in a new types.d.ts file please.

naman9271: Okayy fixing this

JitsiConference.ts
naman9271: No actually it showing a type error for that as we can't assign `JitsiLocalTrack | JitsiRemoteTrack`  to `JitsiLocalTrack`

JitsiConference.ts
naman9271: Okay looking into this

globals.d.ts
naman9271: `const globalNSConnectionTimes = window.JitsiMeetJS?.app?.connectionTimes ?? {};`

here it is using as a window property

JitsiConference.ts
naman9271: `setLocalParticipantProperty(name: string, value: string | string[])` 
now i updated the type of value to `string | string[]`  from `string` to fix this

JitsiConference.ts
naman9271: @saghul actually i am seeing but the implementation is returning Promise<void> in some cases. so i think we have to 
- return type as `Promise<void | JitsiLocalTrack>` or
- have to explicity return track after if statements

what do you say ??

naman9271:
@saghul I have make changes and ask some suggestions for the above all comments PTAL.

naman9271:
> Can you please drop all the changes unrelated to JitsiConference from the PR? It's already quite large...

#### `JitsiConnection`
- added missing properties in interface
- fix xmpp as per your suggestion of `@internal`

#### `JitsiParticipant`
- changed some properties from private to public........ (want a suggestion - should i make getter function for them as well and use @internal ??)

#### `global.d.ts`
- added type for JitsiMeetJS 

#### `types.d.ts`
- added nullable and optional type

#### `JitsiLocalTrack`
- just fixed lint error of having private methods above of public methods

#### `e2eeping`
- added missing properties in interface

#### `CodecSelection`
- fix proper type from `string` to `CodecMimeType`

#### `QuallityController`
- dropped duplicate interface `IVideoConstraints` as it is also present in `SendVideoController`

#### `Chatroom`
- added default value for a parameter in a method

#### `xmpp`
- added default value for a parameter in a method

#### `JingleSessionPC`
- changed property from private to @internal so that we can access it in JitsiConference

---


@saghul I have typed the other changes.

globals.d.ts
saghul: Ah ok.

JitsiConnection.ts
```diff
@@ -62,7 +69,7 @@ export default class JitsiConnection {
         // Initialize the feature flags so that they are advertised through the disco-info.
         FeatureFlags.init(options.flags || {});
 
-        this.xmpp = new XMPP(options, token);
+        this._xmpp = new XMPP(options, token);
```
saghul, line 72: You don't need to access it like this, leave it as `this.xmpp` since it's a getter.

JitsiConnection.ts
```diff
@@ -240,4 +247,12 @@ export default class JitsiConnection {
 
         return data;
     }
+
+    /**
+     * Get the XMPP instance.
+     * @internal
```
saghul, line 253: Did you check that the generated .d.ts files don't document it?

JitsiParticipant.ts
```diff
@@ -34,6 +32,9 @@ export default class JitsiParticipant {
     private _sources: Map<MediaType, Map<string, ISourceInfo>>;
     private _botType?: string;
     private _connectionJid?: string;
+    public _displayName: string;
```
saghul: Why do these need to be public?

JitsiParticipant.ts
```diff
@@ -240,19 +241,19 @@ export default class JitsiParticipant {
     }
 
     /**
-     * @returns {Array.<JitsiTrack>} The list of media tracks for this
+     * @returns {Array.<JitsiTrack|JitsiRemoteTrack>} The list of media tracks for this
```
saghul: JitsiParticipant encapsulates every other participant except us, so it's only concerned with remote tracks.

JitsiParticipant.ts
```diff
@@ -240,19 +241,19 @@ export default class JitsiParticipant {
...
      */
-    getTracks(): JitsiTrack[] {
+    getTracks(): (JitsiTrack | JitsiRemoteTrack)[] {
         return this._tracks.slice();
     }
 
     /**
      * @param {MediaType} mediaType
-     * @returns {Array.<JitsiTrack>} an array of media tracks for this
+     * @returns {Array.<JitsiTrack|JitsiRemoteTrack>} an array of media tracks for this
```
saghul: ditto

modules/RTC/JitsiLocalTrack.ts
```diff
@@ -664,6 +651,19 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
 
 
+    /**
+     * Sends mute status for a track to conference if any.
+     *
+     * @param {boolean} mute - If track is muted.
+     * @private
+     * @returns {void}
+     */
+    _sendMuteStatus(mute: boolean): void {
```
saghul: Does it need to be public? If so, annotate it with `@internal`

JitsiConference.ts
```diff
@@ -370,13 +534,13 @@ export default class JitsiConference {
      * @param options {object}
      * @param options.connection {JitsiConnection} overrides this.connection
      */
-    _init(options) {
+    async _init(options: IConferenceOptions): Promise<void> {
```
saghul: This function is not async.

JitsiConference.ts
```diff
@@ -370,13 +534,13 @@ export default class JitsiConference {
...
         this.eventManager.setupXMPPListeners();
 
         const { config } = this.options;
 
         this._statsCurrentId = config.statisticsId ?? Settings.callStatsUserName;
-        this.room = this.xmpp.createRoom(
+        this.room = await this.xmpp.createRoom(
```
saghul: This function is not async.

JitsiConference.ts
```diff
@@ -136,10 +240,70 @@ function _getCodecMimeType(codec) {
  *       and so on...
  */
 export default class JitsiConference {
```
saghul: Extend listenable to avoid basically implementing its content over again here.

saghul (commented):
Left some more comments.

JitsiConnection.ts
naman9271: image

image

No,  they are documenting this it is showing on typedocs as well

JitsiParticipant.ts
naman9271: becuase it's instance used in JItsiConference

```
 if (participant._displayName === displayName) {
            return;
        }

        participant._displayName = displayName;

File: JitsiConference.ts (hunk 39 above)
Comment (naman9271) [reply JitsiConference.ts]: @saghul ok I extend this but the `on` , `off` , `removelistener` , `addlistener` are methods in jitsiconference and properties in Listenable so should also drop these methods from JitsiConference ??

File: JitsiConference.ts (hunk 7 above)
Comment (naman9271) [reply JitsiConference.ts]: @saghul create room is returning promise as mentioned in JsDoc comment .... am i wrong ?
image

File: JitsiConference.ts (hunk 38 above)
Comment (naman9271) [reply JitsiConference.ts]: I have revert it to `.then()` to assign the resolved ChatRoom to `this.room` when the Promise resolves, rather than trying to assign the Promise directly.

```
  this.xmpp.createRoom(
            this.options.name, {
                ...config,
                statsId: this._statsCurrentId
            },
            JitsiConference.resourceCreator
        ).then(room => {
            this.room = room;
        });

JitsiConnection.ts
saghul: Did you set the tsconfig option to strip internal types?

JitsiConference.ts
saghul: Not really. createRoom doesn't return a promise!

JitsiConference.ts
saghul: Yes drop them.

JitsiConference.ts
naman9271: @saghul create room is returning promise as mentioned in JsDoc comment .... I am little confused by this please guide.

image

JitsiConnection.ts
naman9271: Oops I forget that, Modified tsconfig now

saghul: Does it work with the change?

naman9271: yeah it is working fine, @internal types are not exposed in d.ts

saghul: Great! Let's start using this then!

JitsiParticipant.ts
saghul: Then mark it internal

JitsiConference.ts
saghul: The jsdoc is wrong. It ends up calling this: https://github.com/jitsi/lib-jitsi-meet/blob/49eb29a846b44e2ee43f729aa8a5bf2e016dce5e/modules/xmpp/strophe.emuc.ts#L69

Which is not async.

naman9271: Done.. Fixed actualy got confused by JsDoc

jallamsetty1:
Jenkins test this please.

jallamsetty1:
Jenkins test this please.

jallamsetty1:
Jenkins test this please.

saghul:
@naman9271 Tests are failing, can you please take a look? Also please rebase to make sure nothing was missed.

saghul:
There are linting issues still. Please focus on this one as top priority @naman9271

naman9271:
cc\ @saghul @jallamsetty1 PTAL

saghul:

jallamsetty1:
Hi @naman9271 I see another conflict, can you please resolve this?

naman9271:
@jallamsetty1 conflicts resolved

jallamsetty1:
Jenkins test this please.

JitsiConference.ts
```diff
@@ -3622,11 +3762,11 @@ export default class JitsiConference {
      * names as defined by https://xmpp.org/extensions/xep-0166.html#def-reason
      * @param {string} options.reasonDescription - Text description that will be
      * included in the session terminate message.
-     * @param {boolean} requestRestart - Whether this is due to a session restart, in which case
+     * @param {boolean} options.requestRestart - Whether this is due to a session restart, in which case
      * media will not be resumed on the JVB.
      * @private
      */
-    _stopP2PSession(options = {}) {
+    _stopP2PSession(options?: { reason?: string; reasonDescription?: string; requestRestart?: boolean; }): void {
```
jallamsetty1: @naman9271  Can you default the `options` param to `{}` like before? The p2p -> jvb switchover test cases are failing because the code is throwing an exception as it tries to destructure the `options` in the next line.

2025-07-29T16:52:07.003Z [modules/xmpp/strophe.util] <vr.Strophe.log>:  Strophe: TypeError: Cannot read properties of undefined (reading 'reason')

naman9271: @jallamsetty1 changes done ...please take a look

jallamsetty1:
Jenkins test this please.

jallamsetty1 (approved):

JitsiConference.ts
```diff
@@ -1257,12 +1384,12 @@ export default class JitsiConference {
    * Clear JitsiLocalTrack properties and listeners.
    * @param {JitsiLocalTrack} track - The JitsiLocalTrack object.
    */
-    onLocalTrackRemoved(track) {
+    onLocalTrackRemoved(track: JitsiLocalTrack): void {
         track.setConference(null);
         this.rtc.removeLocalTrack(track);
-        track.removeEventListener(JitsiTrackEvents.TRACK_MUTE_CHANGED, track.muteHandler);
+        track.removeEventListener(JitsiTrackEvents.TRACK_MUTE_CHANGED, this._fireMuteChangeEvent.bind(this, track));
```
saghul: This doesn't work because `this._fireMuteChangeEvent.bind(this, track)` will return a new reference, which doesn't match the one you added. You need to save it, that's why it was assigned to `track.muteHandler`.

JitsiConference.ts
```diff
@@ -1257,12 +1384,12 @@ export default class JitsiConference {
...
         if (track.isAudioTrack()) {
-            track.removeEventListener(JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED, track.audioLevelHandler);
+            track.removeEventListener(JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED, this._fireAudioLevelChangeEvent.bind(this));
```
saghul: Same here.

JitsiConference.ts
saghul: When addint, you can use `addCancellableListener` which returns the remove function, so alll you need to do it call it.

JitsiConference.ts
```diff
@@ -1418,12 +1547,10 @@ export default class JitsiConference {
 
 
         // Add event handlers.
-        newTrack.muteHandler = this._fireMuteChangeEvent.bind(this, newTrack);
-        newTrack.addEventListener(JitsiTrackEvents.TRACK_MUTE_CHANGED, newTrack.muteHandler);
+        newTrack.addEventListener(JitsiTrackEvents.TRACK_MUTE_CHANGED, this._fireMuteChangeEvent.bind(this, newTrack));
```
saghul: Please see my comments on the remove function.

JitsiConference.ts
saghul: Please address this.

JitsiConference.ts
```diff
@@ -1737,9 +1865,11 @@ export default class JitsiConference {
      * Mutes a participant.
      * @param {string} id The id of the participant to mute.
      */
-    muteParticipant(id, mediaType) {
-        if (!mediaType) {
-            logger.error('muteParticipant: mediaType is required');
+    muteParticipant(id: string, mediaType: MediaType): void {
+        const muteMediaType = mediaType ? mediaType : MediaType.AUDIO;
```
saghul: If the parameeter is optional, annotate it as such and give it a default value, rather than. doing it like this.

JitsiConference.ts
```diff
@@ -2160,12 +2290,17 @@ export default class JitsiConference {
...
      * @param {JitsiRemoteTrack} removedTrack - The track that was removed.
      */
-    onRemoteTrackRemoved(removedTrack) {
+    onRemoteTrackRemoved(removedTrack: JitsiRemoteTrack): void {
         this.getParticipants().forEach(participant => {
             const tracks = participant.getTracks();
 
             for (let i = 0; i < tracks.length; i++) {
-                if (tracks[i] === removedTrack) {
+                // Compare by a unique property to avoid type incompatibility
```
saghul: This doesn't make sense. Both are track instances so we know in advance they do have those functions defined.

JitsiConference.ts
saghul: Any is goood!

JitsiConference.ts
```diff
@@ -3411,6 +3550,7 @@ export default class JitsiConference {
 
         this.isP2PConnectionInterrupted = false;
         this.p2pJingleSession
+        // @ts-ignore
```
saghul: Why is this necessary?

saghul (changes requested):
Good job, almost there! I left some comments, the most important ones being the event listener leaks because we don't remove thee 2 listeners added to local tracks.

JitsiConference.ts
naman9271: yeah actually for this i think it needed migration of  xmppConnection as `it is saying jingle is not defined in xmppConnection` so i put ts-ignore

JitsiConference.ts
naman9271: yeah i fixed it actually it first don't recognising the getId function

JitsiConference.ts
saghul: Sounds good!

JitsiConference.ts
saghul: 👍  Please drop the comment too.

JitsiConference.ts
```diff
@@ -1387,9 +1390,9 @@ export default class JitsiConference extends Listenable {
     onLocalTrackRemoved(track: JitsiLocalTrack): void {
         track.setConference(null);
         this.rtc.removeLocalTrack(track);
-        track.removeEventListener(JitsiTrackEvents.TRACK_MUTE_CHANGED, this._fireMuteChangeEvent.bind(this, track));
+        this.unsubscribers.forEach(remove => remove());
         if (track.isAudioTrack()) {
-            track.removeEventListener(JitsiTrackEvents.TRACK_AUDIO_LEVEL_CHANGED, this._fireAudioLevelChangeEvent.bind(this));
+            this.unsubscribers.forEach(remove => remove());
```
jallamsetty1: You don't need to do this again since you have already removed the handlers in the line above this check.

modules/RTC/JitsiTrack.js
```diff
@@ -3,9 +3,9 @@ import { getLogger } from '@jitsi/logger';
 import * as JitsiTrackEvents from '../../JitsiTrackEvents';
 import { MediaType } from '../../service/RTC/MediaType';
 import browser from '../browser';
+import Listenable from '../util/Listenable';
```
jallamsetty1, line 6: Did you forget to modify the JitsiTrack class definition ? It is supposed to extend Listenable instead of EventEmitter for you to be able to call addCancellableListener.

jallamsetty1:
Jenkins test this please.

jallamsetty1:
Jenkins test this please.

jallamsetty1:
Jenkins test this please.

naman9271:
@jallamsetty1 ptal.

saghul:
jenkins please test this please.

saghul:
@naman9271 All looks good here! Can you pl rebase one more time?

naman9271:
> @naman9271 All looks good here! Can you pl rebase one more time?

@saghul done PTAL.

jallamsetty1:
Jenkins test this please.

PR #2842, opened 2025-06-30
feat(av-moderation) add support for moderating desktop tracks

damencho:


service/RTC/MediaType.ts
```diff
@@ -10,6 +10,11 @@ export enum MediaType {
      */
     AUDIO = 'audio',
 
+    /**
+     * The desktop video type.
+     */
+    DESKTOP = 'desktop',
```
saghul, line 16: @damencho In the end I added this because of ChatRoom's `muteParticipant`.

WDYT is best here? Remote mute is not quite part of AVM, but it kinda is, so I could revert the change and change the type we are passing to that function.

JitsiConferenceEventManager.js
```diff
@@ -105,6 +105,30 @@ export default class JitsiConferenceEventManager {
             }
         );
 
+        chatRoom.addListener(XMPPEvents.DESKTOP_MUTED_BY_FOCUS,
```
saghul, line 108: @jallamsetty1 Can you please have a glance over the desktop mute code?

service/RTC/MediaType.ts
damencho: I can drop my change about the new type and we can go with this ...

modules/RTC/RTC.js
```diff
@@ -559,6 +559,25 @@ export default class RTC extends Listenable {
...
+    * @param value The mute value.
+    * @returns {Promise}
+    */
+    setDesktopMute(value) {
+        const mutePromises = [];
+
+        this.getLocalTracks(MediaType.DESKTOP)
+            .forEach(videoTrack => {
+                // this is a Promise
+                mutePromises.push(value ? videoTrack.mute() : videoTrack.unmute());
```
jallamsetty1: I see that you haven't used `setDesktopMute` yet to unmute the track. This won't work for desktop tracks. On mute, desktop tracks are removed from RTCPeerConnection but not from TPC. So they need to be replaced with the new desktop stream on unmute.

jallamsetty1: For this to work, we need to change the _setMuted impl for JitsiLocalTrack. We can never force unmute desktop share so we shouldn't allow it here. WDYT?

saghul: Good eye! Upon closer inspection we _only_ use this family of functions to muting, never for unmuting. So I removed the parameter to make it clear and avoid footguns.

modules/RTC/RTC.js
```diff
@@ -542,16 +541,33 @@ export default class RTC extends Listenable {
...
+    }
+
+    /**
+    * Set mute for all local desktop video streams attached to the conference.
+    * @returns {Promise}
+    */
+    setDesktopMute() {
+        const mutePromises = [];
+
+        this.getLocalTracks(MediaType.DESKTOP)
```
jallamsetty1: I just remembered that the screen share tracks have their type set to `video`, `videoType` is the one that  distinguishes them from camera tracks. So this would always return an empty array.

Also, do we want `setVideoMute` to mute only camera tracks and not the desktop tracks? If so, `setVideoMute` also needs to be modified.

saghul: I'll give it a test, good point!

saghul: Updated, PTAL!

jallamsetty1: LGTM

PR #2845, opened 2025-07-02
fix(TPC) Enable L1T3 for p2p streams

jallamsetty1:
This should make p2p video more resilient to packet loss.


saghul:
How does this help? It it because it adds the tempral layers so fps can be throttled in case of bad bw?

jallamsetty1:
> How does this help? It it because it adds the tempral layers so fps can be throttled in case of bad bw?

Yes exactly!

fippo:
https://www.youtube.com/watch?v=Zwjjb9Cx3ZA has aged nicely. One of the advantages is that you don't have to recover lost frames on the higher temporal layers, you just drop them.

PR #2851, opened 2025-07-20
Migrate `modules\connectivity\ConnectionQuality.js`

naman9271:


jallamsetty1:
Jenkins test this please.

modules/connectivity/ConnectionQuality.ts
```diff
@@ -84,20 +87,65 @@ function rampUp(millisSinceStart) {
...
+    };
+    serverRegion?: string;
+}
+
+export interface IConnectionQualityOptions {
+    config: {
+        disableLocalStats: boolean;
+        disableLocalStatsBroadcast: boolean;
+        pcStatsInterval: number;
+        startBitrate: number;
```
jallamsetty1: This is no longer being set in the config.js, this can be dropped from here.

jallamsetty1: Also, pls drop these lines as well - https://github.com/jitsi/lib-jitsi-meet/blob/master/modules/connectivity/ConnectionQuality.js#L145

modules/connectivity/ConnectionQuality.ts
```diff
@@ -84,20 +87,65 @@ function rampUp(millisSinceStart) {
...
+    };
+}
+
+export interface IRemoteStatsData {
```
jallamsetty1: Since `IRemoteStatsData` is a subset of `ILocalStats`, you can avoid duplication by defining it this way:
```
export type IRemoteStatsData = Pick<
    ILocalStats,
    'bitrate' | 'connectionQuality' | 'jvbRTT' | 'maxEnabledResolution' | 'packetLoss' | 'serverRegion'
>;
```
Also, I would rename it to `IRemoteStats` to be consistent.

modules/connectivity/ConnectionQuality.ts
naman9271: Done

modules/connectivity/ConnectionQuality.ts
naman9271: Done

jallamsetty1:
@naman9271 pls fix the linter error.

jallamsetty1:
Jenkins test this please.

jallamsetty1 (approved):

jallamsetty1:
Jenkins test this please.

PR #2853, opened 2025-07-20
Migrate `authenticateAndUpgradeRole.ts`

naman9271:


saghul:
Can you move this into JitsiConference instead? I think it was outside for some legacy reasons at one point, but I see no reason why it cannot be where. it belongs now.

naman9271:
@saghul i have moved it to JitsiCOnference please take a look

JitsiConference.ts
```diff
@@ -210,6 +215,149 @@ function _getCodecMimeType(codec: string): Nullable<CodecMimeType> {
...
+    id: string;
+    onCreateResource?: (...args: any[]) => any;
+    onLoginSuccessful?: () => void;
+    password: string;
+}
+interface IProcessWithCancel extends Promise<void> {
+    cancel: () => void;
+}
+
+/* eslint-disable no-invalid-this */
```
saghul: Not like this, make the function a method of conference, of course!

naman9271: ahh got it .. please take a look i have fixed this in lastest commit

JitsiConference.ts
```diff
@@ -6,11 +6,16 @@ import * as JitsiConferenceErrors from './JitsiConferenceErrors';
 import JitsiConferenceEventManager from './JitsiConferenceEventManager';
 import * as JitsiConferenceEvents from './JitsiConferenceEvents';
 import type JitsiConnection from './JitsiConnection';
+import {
+    CONNECTION_DISCONNECTED,
```
saghul: You only need JitsiConnectionEvents, the rest are within it.

JitsiConference.ts
```diff
@@ -759,20 +791,123 @@ export default class JitsiConference extends Listenable {
...
+ * Jicofo in order to obtain a session ID (which is then stored in the local
+ * storage). The user's role of the parent conference will be upgraded to
+ * moderator (by Jicofo). It's also used to join the conference when starting
+ * from anonymous domain and only authenticated users are allowed to create new
+ * rooms.
+ *
+ * @param options - Options for authentication and upgrade.
+ * @returns A thenable which settles when the process finishes and has a cancel method.
+ */
+ authenticateAndUpgradeRole(this: JitsiConference, {
```
saghul: Why do you put it here rather than with the other instance methods?!

saghul: You don't need to type `this`. Also, you can run the linter locally rather than pushing untested code.

naman9271: Ah, my bad. I’ve fixed the changes in the latest commit, and I’ll make sure to keep this in mind for the future.

JitsiConference.ts
```diff
@@ -2307,22 +2334,125 @@ export default class JitsiConference extends Listenable {
...
+        (process as IProcessWithCancel).cancel = () => {
+            canceled = true;
+            rejectPromise({});
+            xmpp?.disconnect();
+        };
+
+        return process;
     }
 
+    /* eslint-enable no-invalid-this */
```
saghul: Drop this comment.

JitsiConference.ts
```diff
@@ -2307,22 +2334,125 @@ export default class JitsiConference extends Listenable {
...
+        // 2. Let the API client/consumer know as soon as the XMPP user has been
+        //    successfully logged in.
+        onLoginSuccessful
+    }: IAuthenticateAndUpgradeRoleOptions): Promise<void> {
+        let canceled = false;
+        let rejectPromise: (reason?: IUpgradeRoleError | {}) => void;
+        let xmpp: XMPP = new XMPP(this.connection.options, undefined);
+
+        const process = new Promise<void>((resolve, reject) => {
+        // The process is represented by a Thenable with a cancel method. The
```
saghul: Is the indentation alright here?

JitsiConference.ts
```diff
@@ -2307,22 +2334,125 @@ export default class JitsiConference extends Listenable {
         }
     }
 
+
+    /* eslint-disable no-invalid-this */
```
saghul: Drop this

naman9271: Done please take a look

saghul:

naman9271:
cc: @saghul

saghul (approved):

PR #2866, opened 2025-07-22
feat(visitors): Support private messages

hristoterezov:
## Summary
- Adds support for visitors to send private messages to main participants
- Extends private message API to support full JID addressing for visitor messages
- Enhances message events with visitor metadata (display name, visitor status, original JID)

## Changes
- Modified `sendPrivateTextMessage` to accept `useFullJid` parameter for direct JID addressing
- Updated private message event handling to include visitor display name and status
- Enhanced message parsing to extract visitor information from nick and addresses elements

## Test plan
- [ ] Verify visitors can send private messages to participants
- [ ] Confirm message events include proper visitor metadata
- [ ] Test both regular participant and visitor private message flows
- [ ] Validate message routing works correctly with full JID addressing


modules/xmpp/ChatRoom.js
```diff
@@ -1304,8 +1305,28 @@ export default class ChatRoom extends Listenable {
             const messageId = $(msg).attr('id') || uuidv4();
 
             if (type === 'chat') {
+                // Check if this is a visitor message (has nick element like group messages)
+                const nickEl = $(msg).find('>nick');
```
damencho, line 1309: We need to discuss this. As I'm working on another feature where all messages can have nick, not only those from visitors.

damencho: Nevermind ... we use it in other places too, I will change it in the new implementation.

PR #2869, opened 2025-07-30
feat(AudioController) Adds support for subscribing to audio sources.

jallamsetty1:
JVB support for this was added here - https://github.com/jitsi/jitsi-videobridge/commit/6fc76e46bb109f14ceea785a3fcd306323bc2c80

### DIFF
--- a/JitsiConference.ts
+++ b/JitsiConference.ts
@@ -1717,6 +1717,26 @@ export default class JitsiConference extends Listenable {
     }
 
     /**
+     * Sets the audio subscription mode for the local user.
+     *
+     * @param {ReceiverAudioSubscription} mode - The audio subscription mode to set.
+     * @returns {void}
+     */
+    setAudioSubscriptionMode(mode) {
+        this.qualityController.audioController.setAudioSubscription(mode);
+    }
+
+    /**
+     * Sets deafened status for the local user.
+     *
+     * @param {boolean} isDeafened - Whether the user should stop receiving remote audio.
+     * @returns {void}
+     */
+    setIsDeafened(isDeafened) {
+        this.qualityController.audioController.setIsDeafened(isDeafened);
+    }
+
+    /**
      * Selects a new value for "lastN". The requested amount of videos are going
      * to be delivered after the value is in effect. Set to -1 for unlimited or
      * all available videos.
--- a/JitsiConference.ts
+++ b/JitsiConference.ts
@@ -67,6 +67,7 @@ import { BridgeVideoType } from './service/RTC/BridgeVideoType';
 import { CodecMimeType } from './service/RTC/CodecMimeType';
 import { MediaType } from './service/RTC/MediaType';
 import RTCEvents from './service/RTC/RTCEvents';
+import { IReceiverAudioSubscriptionMessage } from './service/RTC/ReceiverAudioSubscription';
 import { SignalingEvents } from './service/RTC/SignalingEvents';
 import { getMediaTypeFromSourceName, getSourceNameForJitsiTrack } from './service/RTC/SignalingLayer';
 import { VideoType } from './service/RTC/VideoType';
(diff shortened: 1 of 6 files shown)


jallamsetty1:
Jenkins test this please.

modules/qualitycontrol/ReceiveAudioController.ts
```diff
@@ -0,0 +1,97 @@
...
+        if ((message.mode == ReceiverAudioSubscription.NONE || message.mode == ReceiverAudioSubscription.ALL)
+            && this._subscriptionMode == message.mode) {
+            logger.debug(`Ignoring ReceiverAudioSubscription with mode: ${message.mode}, no change needed.`);
+
+            return;
+        }
+        if (message.mode == ReceiverAudioSubscription.INCLUDE
+            || message.mode == ReceiverAudioSubscription.EXCLUDE) {
+
+            // Ensure the source list is not empty when using INCLUDE or EXCLUDE modes.
```
saghul: Why not? Shouldn't an empty INCLDUE be treated like NONE and an empty EXCLUDE as ALL?

JitsiConference.ts
```diff
@@ -1716,6 +1717,26 @@ export default class JitsiConference extends Listenable {
...
+        this.qualityController.audioController.setAudioSubscriptionMode(message);
+    }
+
+    /**
+     * Sets deafened status for the local user.
+     *
+     * @param {boolean} isDeafened - Whether the user should stop receiving remote audio.
+     * @returns {void}
+     */
+    setIsDeafened(isDeafened: boolean): void {
```
saghul: I'm a bit hesitant to expose this as a public API since it dierges from widespread use. (in Discord). In Discord, deafening also mutes you.

Now, if we are to mute the user, we shall do that at the application layer, since it may need to collaborate with other muting operations (aka the authority).

modules/qualitycontrol/ReceiveAudioController.ts
jallamsetty1: It depends on the implementation in the bridge. @bgrozev ^

JitsiConference.ts
jallamsetty1: Why wouldn't it be? From the LJM perspective, it should be a public API so that the consumers of the library can call it and implement the deafened feature however they want. We could rename it to something else more meaningful perhaps?

saghul: That's what I meant, that with the current name I think the expectation would be that it does what discord does.

I don't have a good name suggestion, why not leave it at setAudioSubscriptionMode and not provide this higher level one just yet?

modules/qualitycontrol/ReceiveAudioController.ts
jallamsetty1: Checked with Boris. Though they are treated that way, there will be difference in certain characteristics. His suggested was to make the client send NONE and ALL in those cases.

JitsiConference.ts
jallamsetty1: Will `muteRemoteAudio` work? I feel this is more intuitive if someone were to implement this functionality in their application rather than figuring the various subscription modes that the bridge supports.

saghul: LJM is supposed to be the low level library, so IMHO we better start with less API surface and enlarge later if necesssary.

Won't oppose landing under the new name though.

jallamsetty1: IMHO this fits right in the next gen LJM effort where we want to make the API more intuitive :) we do plan on adding more APIs so one more wouldn't hurt I guess.

modules/qualitycontrol/ReceiveAudioController.ts
```diff
@@ -0,0 +1,96 @@
...
+        return this._subscriptionMode;
+    }
+
+    /**
+     * Mutes or unmutes the remote audio streams based on the provided parameter.
+     *
+     * @param {boolean} muted - Indicates whether the remote audio should be muted or not.
+     * @returns {void}
+     */
+    muteRemoteAudio(muted: boolean): void {
```
saghul, line 53: Can we have events when the state of this changes?

jallamsetty1: I will merge this PR and will include the events in the dominant speaker calculation PR that I am working on.

PR #2889, opened 2025-08-17
Migrated `modules/xmpp/xmpp.ts`

naman9271:


modules/xmpp/Caps.js
```diff
@@ -195,7 +195,7 @@ export default class Caps extends Listenable {
      * Returns a set with the features for a host.
      * @param {String} jid the jid of the host
      * @param {int} timeout the timeout in ms for reply from the host.
-     * @returns {Promise<Set<String>>}
+     * @returns {Promise<Object>}
```
saghul: Is it not a set of strings what we return?

modules/xmpp/XmppConnection.ts
```diff
@@ -84,15 +87,22 @@ interface IInternalOptions {
...
     private _rawInputTracker: LastSuccessTracker;
     private _resumeTask: ResumeTask;
     private _deferredIQs: IDeferredSendIQ[];
     private _oneSuccessfulConnect: boolean;
     private _status: Strophe.Status;
-    private _wsKeepAlive: ReturnType<typeof setTimeout> | undefined;
+    private _wsKeepAlive: Optional<ReturnType<typeof setTimeout>>;
+
+    /**
+     * @internal
```
saghul, line 100: I think this whole class should be annotated with internal.

modules/xmpp/strophe.jingle.js
```diff
@@ -451,6 +451,7 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
 
     /**
      * Returns the data saved in 'updateLog' in a format to be logged.
+     * @returns {Object} An object containing the data to be logged.
```
saghul: Wouldn't `Record<string, unknown>` be better here?

modules/xmpp/xmpp.ts
```diff
@@ -28,6 +29,130 @@ import initStropheUtil from './strophe.util';
...
+ * Connection options interface
+ */
+interface ICreateConnectionOptions {
+    enableWebsocketResume?: boolean;
+    serviceUrl?: string;
+    shard?: string;
+    token?: string;
+    websocketKeepAlive?: number;
+    websocketKeepAliveUrl?: string;
+    xmppPing?: Record<string, unknown>;
```
saghul, line 42: I see we pass this around, but never actually set it to anything, not here not in JM. Have I missed anything @damencho ? Perhaps we can dump it entirely?

modules/xmpp/xmpp.ts
```diff
@@ -97,37 +222,64 @@ export const DEFAULT_STUN_SERVERS = [
...
  * to avoid additional signaling (disco-info).
  * @type {string}
  */
-export const FEATURE_E2EE = 'https://jitsi.org/meet/e2ee';
+export const FEATURE_E2EE: string = 'https://jitsi.org/meet/e2ee';
 
 /**
  *
  */
 export default class XMPP extends Listenable {
```
saghul, line 253: This class should be marked internal.

modules/xmpp/xmpp.ts
```diff
@@ -143,15 +295,15 @@ export default class XMPP extends Listenable {
...
+    constructor(options: IXMPPOptions, token?: string) {
         super();
 
         if (options.bosh && !options.serviceUrl) {
             throw new Error('The "bosh" option is no longer supported, please use "serviceUrl" instead');
         }
 
         this.connection = null;
-        this.disconnectInProgress = false;
+        this.disconnectInProgress = undefined;
```
saghul: It makes more more sensse for this to be a boolean, the name suggests so

modules/xmpp/xmpp.ts
```diff
@@ -302,164 +453,113 @@ export default class XMPP extends Listenable {
...
-                this.anonymousConnectionFailed = true;
-            } else {
-                this.connectionFailed = true;
-            }
-            this.lastErrorMsg = msg;
-            if (msg === 'giving-up') {
-                this.eventEmitter.emit(
-                    JitsiConnectionEvents.CONNECTION_FAILED,
-                    JitsiConnectionErrors.OTHER_ERROR, msg);
+        if (aprops && Object.keys(aprops).length > 0) {
```
saghul: Use optional chaining here, or a default value above?

modules/xmpp/xmpp.ts
```diff
@@ -302,164 +453,113 @@ export default class XMPP extends Listenable {
...
-                JitsiConnectionErrors.PASSWORD_REQUIRED,
-                msg || this._parseConnectionFailedMessage(lastFailedRawMessage),
-                credentials);
+        this.sendDeploymentInfo = false;
+
+        const { region, shard } = aprops;
+
+        if (region || shard) {
+            // avoids sending empty values
+            this.eventEmitter.emit(JitsiConnectionEvents.PROPERTIES_UPDATED, JSON.parse(JSON.stringify({
```
saghul: Ue `cloneDeep` rather than parse / stringify.

modules/xmpp/xmpp.ts
```diff
@@ -672,30 +1049,19 @@ export default class XMPP extends Listenable {
 
         logger.info('(TIME) Strophe Attaching:\t', now);
         this.connection.attach(options.jid, options.sid,
-            parseInt(options.rid, 10) + 1,
+            (parseInt(options.rid, 10) + 1).toString(),
```
saghul: Consider using Number.parseInt, which won't throw.

saghul:

modules/xmpp/Caps.js
naman9271: I had set the return type to object earlier because I was running into destructuring errors (screenshot attached). But if the correct type is Set<string>, that makes sense, I’ll update it accordingly instead of keeping it as a generic object.

image

Just to confirm for a Set<string>, we can’t destructure it like an object/array. Instead, should we be handling it like this?

```
 identity.name && this.caps.getFeaturesAndIdentities(identity.name, identity.type)
    .then((result: any) => processLobbyFeatures(result.features))
    .catch(e => logger.warn('Error getting features from lobby.', e?.message));
 }

```

naman9271:
@saghul i have make changes please take a look

modules/xmpp/Caps.js
saghul: Check in Caps.js, that's where we intiialize it.

modules/xmpp/Caps.js
```diff
@@ -195,7 +195,7 @@ export default class Caps extends Listenable {
      * Returns a set with the features for a host.
      * @param {String} jid the jid of the host
      * @param {int} timeout the timeout in ms for reply from the host.
-     * @returns {Promise<Object>}
+     * @returns {Promise<Set<String>>}
```
saghul: Ah I see now, we return an object indeed, which has a field called "features" which is a Set of strings, see parseDiscoInfo.

naman9271: yeah i see .. i have fixed it in latest commit please take a look

saghul:
Looks goood but needs a rebase.

naman9271:
> Looks goood but needs a rebase.

Done PTAL.

saghul:

naman9271:
Conlicts Resolved

modules/xmpp/xmpp.ts
damencho: @saghul yep, you can drop it ... we use it only to set the domain to send pings to: 
https://github.com/jitsi/lib-jitsi-meet/blob/b6142c7078637b8c24d0a7e2039c2e1dec96e230/modules/xmpp/xmpp.js#L176

saghul:

PR #2897, opened 2025-08-21
fix(typedoc) fix warnings 

naman9271:
I have leave the warnings of kind `...  but not included in the documentation`as for them they need to be in  as i don't think they are meant to be public
and I also leave the warnings like `Encountered an unknown block tag @static @constructor @noInheritDoc` as i believe they should be there.


saghul:
I think the @ static and constructor can go, since JS/TS already have a way to represent them.

saghul:

naman9271:
cc: @saghul

PR #2900, opened 2025-08-25
Migrated `authenticateAndUpgradeRole.js`

naman9271:


naman9271:
@saghul build error is fixed now, please take a look

JitsiConference.ts
```diff
@@ -548,7 +575,7 @@ export default class JitsiConference extends Listenable {
         this._statsCurrentId = config.statisticsId ?? Settings.callStatsUserName;
         this.room = this._xmpp.createRoom(
             this.options.name, {
-                ...config,
+                customDomain: this.options.customDomain,
```
saghul, line 578: This looks like a very weird change. We were forwarding all other options before.

saghul:

JitsiConference.ts
naman9271: but in the function we are only using customDomain and StatsId for room creation so i don't think we need to pass all options 

```
/**
 * Room creation options interface
 */
interface IRoomCreationOptions {
    customDomain?: string;
    statsId?: string;
}

saghul: Was that before the migration to TS?

saghul: Was that before the migration to TS?

naman9271: no this interface i created during migration of `xmpp`

saghul: https://github.com/jitsi/lib-jitsi-meet/blob/bf3e3a8eda8cd868829f7f05203746a71b4abc9b/JitsiConference.js#L379 but we passed the entire thing here?

saghul: Ok, I see now! Good catch!

damencho: @naman9271 @saghul This broke visitors. Reverting back to ...config make them work again.

PR #2901, opened 2025-08-26
fix: Re-mute audio track after applying constraints

hlvhe:


jallamsetty1:
Jenkins test this please.

PR #2902, opened 2025-08-27
Fixing `xmpp.createroom`

naman9271:


naman9271:
@saghul , @damencho please take a look

saghul:
Don't we call createRoom elsewhere too?

saghul:

naman9271:
> Don't we call createRoom elsewhere too?

Yeah we call it in authenticate method also

saghul:
> > Don't we call createRoom elsewhere too?
> 
> Yeah we call it in authenticate method also

And does it need adjusting there too? It would be good to adjust the type so we don't try to make this mistake in the future.

naman9271:
> And does it need adjusting there too? It would be good to adjust the type so we don't try to make this mistake in the future.

yeah @saghul fix the interface please take a look

saghul:
Ping @damencho

saghul (approved):

JitsiConference.ts
```diff
@@ -354,7 +354,7 @@ export default class JitsiConference extends Listenable {
         this.connection = options.connection;
         this._xmpp = this.connection?.xmpp;
 
-        if (this._xmpp.isRoomCreated(options.name, options.customDomain)) {
+        if (this._xmpp.isRoomCreated(options.name, options.config.customDomain)) {
```
damencho: Why getting customDomain from config? That should be coming from the options... isn't it?

JitsiConference.ts
```diff
@@ -102,6 +102,7 @@ export interface IConferenceOptions {
         channelLastN?: number;
         confID?: string;
         createVADProcessor?: () => IVADProcessor;
+        customDomain?: string;
```
damencho: And this, I don't think we pass that in config. I checked jitsi-meet we have this only in ljm passed directly like name in IConferenceOptions.

JitsiConference.ts
naman9271: Ahh @damencho, my mistake earlier, the call is actually happening in authenticateAndUpgradeRole. I got confused because xmpp.createRoom requires customDomain, and that field was missing in the config. I’ve pushed a fix to add it. please take a look

damencho:
Lint errors

naman9271:
> Lint errors

ahh my bad ... i have fixed now please take a look

damencho:
jenkins test this please.

PR #2903, opened 2025-08-27
Display name updates

damencho:


modules/xmpp/ChatRoom.ts
```diff
@@ -1455,42 +1455,37 @@ export default class ChatRoom extends Listenable {
...
-                const nickEl = $(msg).find('>nick');
-                let nick;
+                let displayName;
+                let originalFrom;
 
-                if (nickEl.length > 0) {
-                    nick = nickEl.text();
-                }
+                if (isVisitorMessage) {
+                    displayName = displayNameEl.text();
```
hristoterezov, line 1466: Why we set `displayName` only for visitor messages?

modules/xmpp/ChatRoom.ts
```diff
@@ -1455,42 +1455,37 @@ export default class ChatRoom extends Listenable {
...
-                }
-
-                // we will fire explicitly that this is a guest(isGuest:true) to the conference
-                // informing that this is probably a message from a guest to the conference (visitor)
+                // we will fire explicitly that this is a visitor(isVisitor:true) to the conference
                 // a message with explicit name set
                 this.eventEmitter.emit(XMPPEvents.MESSAGE_RECEIVED,
-                    from, txt, this.myroomjid, stamp, nick, Boolean(nick), messageId);
+                    from, txt, this.myroomjid, stamp,
+                        displayNameEl.length > 0 ? displayNameEl.text() : undefined, isVisitorMessage, messageId,
```
hristoterezov: I would extract the display name text into variable on top and reuse it where i need it.

modules/xmpp/ChatRoom.ts
```diff
@@ -1455,42 +1455,37 @@ export default class ChatRoom extends Listenable {
...
+                        isVisitorMessage ? undefined : displayNameEl.attr('source'));
```
hristoterezov: would extract the source into variable on top and reuse it where i need it.

JitsiConference.ts
```diff
@@ -575,7 +575,7 @@ export default class JitsiConference extends Listenable {
         this._statsCurrentId = config.statisticsId ?? Settings.callStatsUserName;
         this.room = this._xmpp.createRoom(
             this.options.name, {
-                customDomain: this.options.customDomain,
+                ...config,
```
hristoterezov: Why this change? I don't like it. I was thinking that generally we would like to avoid passing everything and be more specific and pass only what is needed.

hristoterezov (changes requested):
Generally LGTM! Left a few questions, optional suggestions and comments! There are some conflicts to be solved tho!

modules/xmpp/ChatRoom.ts
damencho: Well, these are private messages, and display name is only handled for visitors. 
If you are getting a provate message from non-visitor, that participant is in the meeting we want to ignore this and use anything from the presence.

JitsiConference.ts
damencho: That is a fix, as visitors are broken with current code from master. 
I will drop this commit once https://github.com/jitsi/lib-jitsi-meet/pull/2902 is merged and I rebase to it.

PR #2911, opened 2025-09-03
feat(Utils) Change the log to include log level and component.

jallamsetty1:
This helps us with filtering the logs in RTCStats.

2025-09-03T22:25:48.070Z INFO [RTC:JitsiRemoteTrack] <Gc._playCallback>:  (TIME) TTFM video: 5218.799999982119
2025-09-03T22:25:48.073Z DEBUG [RTC:JitsiRemoteTrack] <Gc._containerEventHandler>:  canplaythrough handler was called for a container with attached RemoteTrack[userID: 73d04aed, type: video, ssrc: 75307853, p2p: false, sourceName: 73d04aed-v0, status: {readyState: live, muted: false, enabled: true}].

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,3 +1,6 @@
+// Initialize enhanced logging transport before any modules are loaded
+require('./modules/util/LoggerInit');
+
 const ljm = require('./JitsiMeetJS').default;
 
 /**
--- a/modules/util/LoggerInit.ts
+++ b/modules/util/LoggerInit.ts
@@ -1,0 +1,11 @@
+/**
+ * Logger initialization module
+ * This module sets up the enhanced logging transport and must be imported
+ * before any other modules that create loggers.
+ */
+
+// Import LoggerTransport which will automatically replace the console transport
+import './LoggerTransport';
+
+// Ensure this runs by exporting a marker
+export const LOGGER_INITIALIZED = true;
--- a/modules/util/LoggerTransport.ts
+++ b/modules/util/LoggerTransport.ts
@@ -1,0 +1,277 @@
+
+import Logger from '@jitsi/logger';
+// Import the Logger constructor directly to access static properties
+import LoggerConstructor from '@jitsi/logger/lib/Logger';
+
+/**
+ * Enhanced console transport that includes log levels in the output format
+ */
+export class EnhancedConsoleTransport {
+    /**
+     * The original console methods
+     */
+    private _originalConsole: Console;
+
+    /**
+     * Constructor
+     */
+    constructor(originalConsole?: Console) {
+        this._originalConsole = originalConsole || console;
+    }
+
+    /**
+     * Extracts the module name.
+     *
+     * @param {string} loggerId - logger ID.
+     * @returns {string} module name.
+     */
+    private _extractModule(loggerId: string): string {
+        if (!loggerId) {
+            return 'unknown';
+        }
+
+        // Handle core components (no modules/ prefix)
+        if (!loggerId.startsWith('modules/')) {
+            // Core lib-jitsi-meet components
+            if (loggerId.startsWith('Jitsi')) {
+                return 'core';
+            }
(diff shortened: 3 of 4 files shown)


index.js
```diff
@@ -1,3 +1,6 @@
+// Initialize enhanced logging transport before any modules are loaded
+require('./modules/util/LoggerInit');
```
saghul, line 2: Nothe that this file is only used in the UMD build, not in the ESM build. Mobile uses the ESM build.

modules/util/LoggerInit.ts
```diff
@@ -0,0 +1,11 @@
+/**
+ * Logger initialization module
+ * This module sets up the enhanced logging transport and must be imported
+ * before any other modules that create loggers.
+ */
+
+// Import LoggerTransport which will automatically replace the console transport
```
saghul, line 7: Not a fan of doing this at import time. it should likely be a function imported in JitsiMeetJS which we can call instead, early in JM?

modules/util/LoggerTransport.ts
```diff
@@ -0,0 +1,277 @@
...
+     *
+     * @param {string} loggerId - logger ID.
+     * @returns {string} module name.
+     */
+    private _extractModule(loggerId: string): string {
+        if (!loggerId) {
+            return 'unknown';
+        }
+
+        // Handle core components (no modules/ prefix)
```
saghul, line 33: Since we no longer use the filename for creating a logger (it's a hardcoded value), how about we rename those instead?

Note we'll need to adjust the default levels in JM to match too.

modules/util/LoggerTransport.ts
```diff
@@ -0,0 +1,277 @@
...
+
+    /**
+     * Format log message with explicit log level and enhanced component info
+     * @param level - Log level (trace, debug, info, log, warn, error)
+     * @param args - Log arguments from @jitsi/logger
+     * @returns Formatted arguments with log level and module info included
+     */
+    private _formatWithLevel(level: string, args: any[]): any[] {
+        // Convert 'log' level to 'INFO' for consistency
+        const normalizedLevel = level.toUpperCase() === 'LOG' ? 'INFO' : level.toUpperCase();
```
saghul, line 86: The logger already has an adapter where we can do that, doesn't it?

modules/util/LoggerTransport.ts
```diff
@@ -0,0 +1,277 @@
+
+import Logger from '@jitsi/logger';
+// Import the Logger constructor directly to access static properties
+import LoggerConstructor from '@jitsi/logger/lib/Logger';
+
+/**
+ * Enhanced console transport that includes log levels in the output format
```
saghul, line 7: We could make this the default in the jitsi logger no?

modules/util/LoggerTransport.ts
```diff
@@ -0,0 +1,277 @@
...
+    info: console.info,
+    log: console.log,
+    trace: console.trace,
+    warn: console.warn,
+};
+
+// Replace console methods with silent versions while preserving our enhanced transport
+console.trace = function(...args: any[]) {
+    // Check if this is being called from our enhanced transport
+    const stack = new Error().stack || '';
```
saghul, line 212: This is a big no-no. Stack traces are expensive, V8 computes the lazily for that reason. Forcing a stack trace collection for each lone we log is going to be very very bad.

Not sure I understand why you are making the checks here.

modules/util/LoggerTransport.ts
jallamsetty1: Yes, that sounds like a better idea, was too lazy to edit each one of them. I can go give more meaningful id's I guess instead of extracting them.

modules/util/LoggerTransport.ts
jallamsetty1: Yes if everyone is ok with it.

PR #2915, opened 2025-09-08
fix(logging) Rename logger ids to facilitate proper filtering of logs.

jallamsetty1:


JitsiConference.ts
```diff
@@ -175,7 +175,7 @@ export interface IStopSessionOptions {
     sendSessionTerminate?: boolean;
 }
 
-const logger = getLogger('JitsiConference');
+const logger = getLogger('Core:JitsiConference');
```
saghul: No strong opinion, question: what capitalization do we like here? lowercase except for acronyms like RTC? Actually all lower sounds ok? Minus the last part, there we use the full component name.

modules/RTC/RTC.js
```diff
@@ -15,7 +15,7 @@ import RTCUtils from './RTCUtils';
 import TraceablePeerConnection from './TraceablePeerConnection';
 
 
-const logger = getLogger('modules/RTC/RTC');
+const logger = getLogger('RTC:RTC');
```
saghul: This reads a bit weird no? rtc/core?

modules/RTC/RTCUtils.js
```diff
@@ -16,7 +16,7 @@ import { isValidNumber } from '../util/MathUtil';
 
 import screenObtainer from './ScreenObtainer';
 
-const logger = getLogger('modules/RTC/RTCUtils');
+const logger = getLogger('RTC:RTCUtils');
```
saghul: If we are using remantic names, how about rtc/utils ?

modules/RTCStats/RTCStats.ts
```diff
@@ -12,7 +12,7 @@ import DefaultLogStorage from './DefaulLogStorage';
 import { RTCStatsEvents } from './RTCStatsEvents';
 import { ITraceOptions } from './interfaces';
 
-const logger = getLogger('modules/RTCStats/RTCStats');
+const logger = getLogger('RTCStats:RTCStats');
```
saghul: rtcstats/core ?

modules/detection/VADAudioAnalyser.ts
```diff
@@ -8,7 +8,7 @@ import EventEmitter from '../util/EventEmitter';
 import { DetectionEvents } from './DetectionEvents';
 import TrackVADEmitter from './TrackVADEmitter';
 
-const logger = getLogger('modules/detection/VADAudioAnalyser');
+const logger = getLogger('Detection:VADAudioAnalyser');
```
saghul: Can we find some better word than "detection" ?

modules/event/Jvb121EventGenerator.ts
```diff
@@ -3,7 +3,7 @@ import { getLogger } from '@jitsi/logger';
 import JitsiConference from '../../JitsiConference';
 import { JitsiConferenceEvents } from '../../JitsiConferenceEvents';
 
-const logger = getLogger('modules/event/Jvb121EventGenerator');
+const logger = getLogger('Event:Jvb121EventGenerator');
```
saghul: Shouldn't this semantically belong together with the TPC?

modules/litemode/LiteModeContext.ts
```diff
@@ -10,7 +10,7 @@ import FeatureFlags from '../flags/FeatureFlags';
 // more than once.
 const kJitsiLiteMode = Symbol('kJitsiLiteMode');
 
-const logger = getLogger('modules/litemode/LiteModeContext');
+const logger = getLogger('Litemode:LiteModeContext');
```
saghul: Since this is only used for testing, how about testing: ?

modules/qualitycontrol/CodecSelection.ts
```diff
@@ -9,7 +9,7 @@ import JitsiLocalTrack from '../RTC/JitsiLocalTrack';
 import browser from '../browser';
 import JingleSessionPC from '../xmpp/JingleSessionPC';
 
-const logger = getLogger('modules/qualitycontrol/CodecSelection');
+const logger = getLogger('Qualitycontrol:CodecSelection');
```
saghul: Something shorter maybe? qc? See my first question too, if we want to capitalize things then I guess this should be QualityControl.

modules/recording/JibriSession.ts
```diff
@@ -209,21 +212,18 @@ export default class JibriSession {
...
+     * @param {string} [options.appData] - Data specific to the app/service that the result file will be uploaded.
+     * @param {string} [options.broadcastId] - The broadcast ID of an associated YouTube stream, used for knowing the
+     * URL from which the stream can be viewed.
+     * @param {string} options.focusMucJid - The JID of the focus participant that controls recording.
+     * @param {streamId} options.streamId - Necessary for live streaming, this is the stream key needed to start a live
+     * streaming session with the streaming service provider.
      * @returns Promise
      */
     start({ appData, broadcastId, focusMucJid, streamId }: IStartOptions): Promise<void> {
+        logger.info('Starting recording session');
```
saghul, line 225: Good! Shall we log some more stuff like who is starting the recording?

modules/recording/JibriSession.ts
```diff
@@ -253,13 +253,13 @@ export default class JibriSession {
...
-     * @param {Object} options - Additional arguments for stopping the
-     * recording.
-     * @param {Object} options.focusMucJid - The JID of the focus participant
-     * that controls recording.
+     * @param {Object} options - Additional arguments for stopping the recording.
+     * @param {Object} options.focusMucJid - The JID of the focus participant that controls recording.
      * @returns Promise
      */
     stop({ focusMucJid }: IStopOptions): Promise<any> {
+        logger.info('Stopping recording session');
```
saghul, line 261: Ditto.

modules/version/ComponentsVersions.ts
```diff
@@ -1,6 +1,6 @@
 import { getLogger } from '@jitsi/logger';
 
-const logger = getLogger('modules/version/ComponentsVersions');
+const logger = getLogger('Version:ComponentsVersions');
```
saghul: I feel like things like this is kinda miscellaneous... "misc" ?

saghul (approved):
Some non-blocking comments / suggestions. Note you'll need a companion JM PR to fix the default log levels for some components.

jallamsetty1:
> Some non-blocking comments / suggestions. Note you'll need a companion JM PR to fix the default log levels for some components.

Initially I thought we should have something meaningful for logger ids but I wasn't sure if it would make sense to others who are debugging so I tried to stick with "dir" structure in LJM. I am ok with all lower case. I will go ahead and make the changes.

jallamsetty1:
@saghul I didn't touch the second part that is the filename, just the first part that will show up as a filter option when viewing the stats.

PR #2916, opened 2025-09-09
feat: Polls using component.

damencho:


modules/xmpp/Polls.ts
```diff
@@ -0,0 +1,110 @@
...
+    }
+
+    /**
+     * Creates and sends a new poll.
+     *
+     * @param pollId
+     * @param question
+     * @param answers
+     */
+    createPoll(pollId: string, question: string, answers: Array<{ name: string; }>) {
```
saghul, line 50: Why not use an array of stringas at this level, rather than have to flatten it here?

modules/xmpp/Polls.ts
```diff
@@ -0,0 +1,110 @@
...
+     * @param {object} payload - Arbitrary data.
+     */
+    _handleMessages(payload) {
+        switch (payload.type) {
+        case COMMAND_NEW_POLL:
+            this._mainRoom.eventEmitter.emit(XMPPEvents.POLLS_RECEIVE_EVENT, payload);
+
+            break;
+        case COMMAND_OLD_POLLS: {
+            if (payload?.polls) {
```
saghul: You don't need this check and can chain it in the next line I think.

modules/xmpp/Polls.ts
```diff
@@ -0,0 +1,110 @@
...
+                payload.polls.forEach((poll: any) => {
```
saghul: Define a type for Poll?

modules/xmpp/xmpp.ts
```diff
@@ -481,6 +482,12 @@ export default class XMPP extends Listenable {
             return true;
         }
 
+        if (from === this.pollsComponentAddress) {
```
saghul: Why is this check different than the other components? Should we be doing things differently overall?

modules/xmpp/Polls.ts
damencho: So we can put it as a type ... I will drop the flattened strings from everywhere so we can have a single strucutre as it now differs and depends on the operation

modules/xmpp/Polls.ts
damencho: It will be best to move the types from jitsi-meet to ljm and use those in jitsi-meet. I will add a todo there, once we can use them that way, I will move them.

modules/xmpp/xmpp.ts
damencho: yeah ... cause we do not have the type in the json messages 
maybe I can add and the type and that will remove this difference.

damencho: oh we already have type and that is new-poll, answer-poll and old-polls ... need to think about it

modules/xmpp/ChatRoom.ts
```diff
@@ -1151,14 +1154,16 @@ export default class ChatRoom extends Listenable {
         const targetJid = useDirectJid ? id : `${this.roomjid}/${id}`;
         const msg = $msg({ to: targetJid,
             type: 'chat' });
+        // when not sending to a direct jid, let's indicate the room jid (used in polls)
+        const roomJid = useDirectJid ? this.roomjid : undefined;
```
saghul: Why is this needed? Isn't the room JID already in the to?

modules/xmpp/Polls.ts
saghul: I forgot where we landed on this one, can you remind me pl?

modules/xmpp/ChatRoom.ts
damencho: Nope, we switched to sending messages to a component and the jid is missing there, it used to be the to of the messages.

damencho:
In the code where it is not voting, the answer is always in a structure with a name, greatly simplifies logic and types.

modules/xmpp/ChatRoom.ts
saghul: Yes, but we can get the target room with:

```
    -- get room name with tenant and find room
    local room = get_room_by_name_and_subdomain(session.jitsi_web_query_room, session.jitsi_web_query_prefix);

```

In Prosody, like we do for breakout rooms, no?

saghul (commented):
I think we can remove the need for passing the roomJid, LGTM otherwise.

PR #2921, opened 2025-09-14
feat: Add XEP-0461 Message Replies protocol support to XMPP ChatRoom

Ankitgkp:
Implement XEP-0461 (Message Replies) support in `modules/xmpp/ChatRoom.ts` to enable threaded conversations in Jitsi Meet chat.

Full XEP-0461 XMPP protocol compliance
sendReplyMessage() for group replies
sendPrivateReplyMessage() for private replies
Automatic fallback format for non-supporting clients
Enhanced message events with `replyToId` parameter

Build passes (TypeScript + Webpack)
Backward compatibility verified
Protocol compliance validated


damencho (changes requested):
Please get rid of the extra changes like formatting and such and just leave the new changes.
I see you have changed ChatRoom, but that is internal to ljm and you cannot use it from jitsi-meet.

Ankitgkp:
@damencho  Since ChatRoom is internal to lib-jitsi-meet and cannot be accessed from jitsi-meet, what is the correct way to implement XEP-0461 Message Replies? Should I remove the public sendReplyMessage() and sendPrivateReplyMessage() methods I added to ChatRoom?

Where should the XEP-0461 parsing logic live if ChatRoom is internal? Should I keep the private parsing methods (_parseReplyMessage, _extractCleanReplyBody) in ChatRoom, or move this functionality elsewhere?

damencho:
I think we should just add an id param to send message: https://github.com/jitsi/lib-jitsi-meet/blob/bc389f3b82ab2964400ee7e9eee1490987b0885c/JitsiConference.ts#L4040C12-L4040C23
to indicate the message we send is a reply to some old message with that id. 

Yep, parsing of messages is done in ChatRoom. 
And when you receive a reply indicate that with a param to which message is that a reply from the message event
https://github.com/jitsi/lib-jitsi-meet/blob/68fc537c25f1a5e52455f3b75a6e46c981f69c47/JitsiConferenceEventManager.ts#L387

Ankitgkp:
@damencho I attempted to implement XEP-0461 Message Replies based on your architectural guidance: I added a `replyToId` parameter to the `sendMessage()` method you referenced, tried to keep all parsing logic contained within ChatRoom using `_parseReplyMessage()` for XEP-0461 stanza handling, and enhanced the existing message events in JitsiConferenceEventManager.ts to include reply information. I focused on avoiding new public methods and maintaining the clean separation between public APIs and internal ChatRoom methods as you suggested.

damencho:
Can you please get rid of unrelated changes, it is hard to review a PR where when you look at the diff it is full of small changes and I cannot find where the new changes are.

Ankitgkp:
The files are mixed up and it’s hard to revert them I will close this PR and open a new one with only the required updates.

Ankitgkp:
@damencho #2922 This PR is a clean implementation of XEP-0461 reply support, with logical, reviewable commits per file. The diff only contains the necessary lines for this feature.

PR #2923, opened 2025-09-16
feat: Add more events to RTCStats

jallamsetty1:


JitsiConferenceEventManager.ts
```diff
@@ -4,9 +4,11 @@ import { Strophe } from 'strophe.js';
 import JitsiConference from './JitsiConference';
 import * as JitsiConferenceErrors from './JitsiConferenceErrors';
 import { JitsiConferenceEvents } from './JitsiConferenceEvents';
+import JitsiMeetJS from './JitsiMeetJS';
```
saghul: Avoid importing this since it's the entrypoint, import from the rtcstats feature directly instead pl.

jallamsetty1: 🤦‍♀️

jallamsetty1: fixed

PR #2926, opened 2025-09-22
fix(RTCStats) Send dominantSpeaker and audio/video mute events correctly

jallamsetty1:


jallamsetty1:
Jenkins test this please.

PR #2940, opened 2025-11-05
fix(ScreenObtainer) Fix echo issues with audio share.

jallamsetty1:
Starting Chromium 137, echo cancellation is getting disabled automatically for desktop audio tracks.
Explicitly enable echo cancellation if "restrictOwnAudio" is not supported (Chrome 140+)

### DIFF
--- a/modules/RTC/ScreenObtainer.ts
+++ b/modules/RTC/ScreenObtainer.ts
@@ -32,10 +32,11 @@ interface IResolutionConfig {
  * Interface for audio quality configuration.
  */
 interface IAudioQuality {
-    autogainControl?: boolean;
+    autoGainControl?: boolean;
     channelCount?: number;
     echoCancellation?: boolean;
     noiseSuppression?: boolean;
+    restrictOwnAudio?: boolean;
     stereo?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


bukharin:
Hi @jallamsetty1 

Using `restrictOwnAudio` has a problem: if you will share Chrome entire window with audio - Chrome cuts all audio producing by all tabs and remote participants will not hear any sharing audio

jallamsetty1:
> Hi @jallamsetty1
> 
> Using `restrictOwnAudio` has a problem: if you will share Chrome entire window with audio - Chrome cuts all audio producing by all tabs and remote participants will not hear any sharing audio

Are you sure about this? I just tested it on Chrome 142.0.7444.60 and cannot reproduce the issue. I am sharing the Chrome window with 2 tabs, one where the conference is running and the other has a youtube audio playing. When I start window capture with audio selected, I only hear the youtube audio, not the remote conference audio being played out in the conference tab.

As per spec of restrictOwnAudio -
 When restrictOwnAudio is enabled, the captured system audio will be filtered to exclude audio originating from the document that performed getDisplayMedia. 
This means it should filter out conference audio only.

bukharin:
@jallamsetty1 yes, you are right. Case described above works in Chrome 141 and seems to be fixed in  Chrome 142

BTW:  in Electron `restrictOwnAudio` does not work and we need to continue to use `echoCancellation: true`

jallamsetty1:
Thanks for testing, `restrictOwnAudio` should have shipped in 141 but maybe not be available in all versions of 141. I will update the browser check to 142 and also exclude electron.

PR #2948, opened 2025-11-18
fix: reverse VP8 simulcast encodings for low-res Chromium (fixes #2939)

### LINKED ISSUE #2939: [BUG] VP8 simulcast causes unexpected downscaling on Chromium at low resolutions
### Describe the bug.

When using VP8 with simulcast enabled, the transmitted video resolution is unexpectedly downscaled on Chromium-based browsers if the requested resolution is low (likely below 1280x720). This issue does not occur on Firefox.

**Conditions:**
- Codec: VP8
- Simulcast: Enabled
- Low resolution input (<1280x720)
- Browser: Chromium-based (Chrome, Edge, etc.)
- Firefox: Not affected

**Observed behavior:**
| Requested resolution | Actual transmitted resolution |
|----------------------|------------------------------|
| 320x240 | 80x60 |
| 640x360 | 320x180 |
| 1280x720 | 1280x720 |

The problem consistently appears on Chromium browsers, while Firefox behaves correctly.

**Regression:**
The issue started occurring after the following commit (Nov 18, 2023):  
https://github.com/jitsi/lib-jitsi-meet/pull/2397/commits/70b6843bc1b475756b57709a75231455967f48cd

### Expected behavior

The transmitted video resolution should match the requested resolution.
For example:
- 320x240 should remain 320x240.
- 640x360 should remain 640x360.

No additional downscaling should be applied when simulcast is enabled with VP8 at low resolutions.

### Screenshots

_No response_

### How to Reproduce

1. I first did this
2. I then did this
3. And so on . . . 

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you checked the Community Forum?

- [x] please make sure you check https://community.jitsi.org

4rayaditya:
## Problem
VP8 simulcast on Chromium at resolutions <640px transmits heavily downscaled video:
- 320×240 camera → 80×60 output (should be 320×240)
- 480×360 camera → 240×180 output (should be 480×360)

**Root cause:** Chromium limits simulcast layers at low resolutions and picks the first encoding. Default order [4x, 2x, 1x] causes it to select the 4× downscaled layer.

## Solution
Reverse encoding order to [1x, 2x, 4x] for VP8 camera tracks <640px on Chromium, ensuring the highest quality layer is selected.

## Changes
- `modules/RTC/TPCUtils.ts`: Added conditional reversal in `_getVideoStreamEncodings()`
- `modules/RTC/TPCUtils.spec.ts`: Added tests for 320px and 480px resolutions

Fixes #2939

### DIFF
--- a/modules/RTC/TPCUtils.ts
+++ b/modules/RTC/TPCUtils.ts
@@ -25,16 +25,12 @@ const logger = getLogger('rtc:TPCUtils');
 const VIDEO_CODECS = [CodecMimeType.AV1, CodecMimeType.H264, CodecMimeType.VP8, CodecMimeType.VP9];
 
 /**
- * Maximum number of simulcast layers supported by the protocol.
- * Used for SIM group sizing and legacy array sizing for backward compatibility.
+ * Maximum number of simulcast layers supported by the WebRTC protocol.
+ * This constant is kept for backward compatibility documentation purposes.
+ * The actual number of layers (1, 2, or 3) is now determined dynamically
+ * based on capture resolution via getEffectiveSimulcastLayers().
  */
 const MAX_SIMULCAST_LAYERS = 3;
-
-/**
- * Number of SSRCs grouped together in a SIM group.
- * Per WebRTC protocol, each SIM group contains up to MAX_SIMULCAST_LAYERS primary SSRCs.
- */
-const SIM_GROUP_SIZE = MAX_SIMULCAST_LAYERS;
 
 /**
  * Gets the default reference height for simulcast configuration.
@@ -301,30 +297,24 @@ export class TPCUtils {
             }];
         }
 
-        // 2. Map bitrate tiers: getEffectiveSimulcastLayers returns layers in canonical order [low→high quality]
-        // Index 0 = lowest quality (scaleFactor 4), gets low bitrate
-        // Index 1 = medium quality (scaleFactor 2), gets standard bitrate  
-        // Index 2 = highest quality (scaleFactor 1), gets high/ssHigh bitrate
-        const bitrateForIndex = (idx: number): number => {
-            if (idx === numLayers - 1) {
-                // Highest quality layer gets ssHigh for screenshare, high for camera
+        // 2. Map bitrate tiers based on scale factor
+        const bitrateForLayer = (layer: { scaleFactor: number }): number => {
+            if (layer.scaleFactor === 1.0) {
                 return videoType === VideoType.DESKTOP ? codecBitrates.ssHigh : codecBitrates.high;
             }
-            if (idx === 0) {
-                // Lowest quality layer gets low bitrate
+            if (layer.scaleFactor === 4.0) {
                 return codecBitrates.low;
             }
-            // Middle layers get standard bitrate
             return codecBitrates.standard;
         };
 
         logger.debug(`TPCUtils._getVideoStreamEncodings: codec=${codecKey}, captureRes=${captureResolution}px, ` +
             `numLayers=${numLayers}, scaleFactors=[${effectiveLayers.map(l => l.scaleFactor).join(',')}]`);
 
-        // 3. Build encodings array from canonical layers (already in correct order)
-        const simulcastEncodings: IRTCRtpEncodingParameters[] = effectiveLayers.map((layer, idx) => ({
+        // 3. Build encodings array from canonical layers
+        const simulcastEncodings: IRTCRtpEncodingParameters[] = effectiveLayers.map(layer => ({
             active: this.pc.videoTransferActive,
-            maxBitrate: bitrateForIndex(idx),
+            maxBitrate: bitrateForLayer(layer),
             rid: layer.rid,
             scaleResolutionDownBy: layer.scaleFactor
         }));
@@ -344,7 +334,7 @@ export class TPCUtils {
             // Use the configuration of the highest quality layer available.
             const topLayerIndex = effectiveLayers.length - 1;
             const topLayer = effectiveLayers[topLayerIndex];
-            const topLayerBitrate = bitrateForIndex(topLayerIndex);
+            const topLayerBitrate = bitrateForLayer(topLayer);
 
             // For SVC, we send one stream at native resolution with spatial scalability.
             // 
@@ -778,13 +768,14 @@ export class TPCUtils {
                 return desc;
             }
 
-            // Add a SIM group for every SIM_GROUP_SIZE FID groups (per WebRTC simulcast protocol).
-            for (let i = 0; i < ssrcs.length; i += SIM_GROUP_SIZE) {
-                const simSsrcs = ssrcs.slice(i, i + SIM_GROUP_SIZE);
-
+            // Only add SIM group if we have more than one SSRC (simulcast is actually enabled).
+            // For single-layer tracks (low resolution), no SIM group should be created.
+            if (ssrcs.length > 1) {
+                // Add a SIM group. Note: With dynamic layers, ssrcs.length can be 1, 2, or 3.
+                // Create one SIM group containing all the simulcast SSRCs.
                 video.ssrcGroups.push({
                     semantics: SSRC_GROUP_SEMANTICS.SIM,
-                    ssrcs: simSsrcs.join(' ')
+                    ssrcs: ssrcs.join(' ')
                 });
             }
         }
--- a/modules/RTC/TPCUtils.spec.ts
+++ b/modules/RTC/TPCUtils.spec.ts
@@ -1496,6 +1496,23 @@ describe('TPCUtils', () => {
                 tpcUtils = null;
             });
 
+            it('and capture resolution is 320 (Fix Check)', () => {
+                const lowResTrack = new MockJitsiLocalTrack(320, MediaType.VIDEO, VideoType.CAMERA);
+                height = 320;
+                
+                activeState = tpcUtils.calculateEncodingsActiveState(lowResTrack, codec, height);
+                expect(activeState.length).toBe(1);
+                expect(activeState[0]).toBe(true);
+
+                maxBitrates = tpcUtils.calculateEncodingsBitrates(lowResTrack, codec, height);
+                expect(maxBitrates.length).toBe(1);
+                expect(maxBitrates[0]).toBe(1500000); 
+
+                scaleFactor = tpcUtils.calculateEncodingsScaleFactor(lowResTrack, codec, height);
+                expect(scaleFactor.length).toBe(1);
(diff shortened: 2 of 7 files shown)


modules/RTC/TPCUtils.ts
```diff
@@ -266,6 +266,18 @@ export class TPCUtils {
             }
         ];
 
+        // Workaround for Chromium bug with VP8 simulcast at low resolutions.
+        // At resolutions below 640px, Chromium limits the number of simulcast layers and picks the first encoding
+        // in the array. By reversing the order for low resolutions, we ensure the highest quality layer is selected.
+        // See: https://github.com/jitsi/lib-jitsi-meet/issues/2939
+        // Chromium source: https://source.chromium.org/chromium/chromium/src/+/main:third_party/webrtc/video/config/simulcast.cc
+        if (codec === CodecMimeType.VP8
+            && !browser.isFirefox()
```
saghul: Shouldn't we be explicit and check for chromium based here?

4rayaditya:
I have made the changes changing the check from !firefox to isChromiumBased()

4rayaditya:
Ok se when we reversed the array, SIM_LAYERS[0] no longer meant 4.0, it meant 1.0! This broke the activation logic. I used the values like 4.0 and 1.0 to solve this error

jallamsetty1:
This will break the default encoding behavior on Chromium which has always been 4:2:1

jallamsetty1:
Have you tested this on older versions of Chromium? What about other codecs that support simulcast, AV1 and VP9?

4rayaditya:
@jallamsetty1 
I think it preserves the encoding as 4:2:1 but for VP8 only. I am working on AV1 and VP9 right now.
As for testing, I have not tried on older versions of chromium yet. But I'll look at it.

saghul:
I'm wondering what the value of simulcast is at those resolutions? Would disabling it be acceptable there @jallamsetty1 ?

jallamsetty1:
The correct way to fix this would be to configure the correct number of encodings based on the captured resolution. If only one layer is needed then we would be effectively disabling simulcast.

4rayaditya:
> The correct way to fix this would be to configure the correct number of encodings based on the captured resolution. If only one layer is needed then we would be effectively disabling simulcast.

can i work on it?

jallamsetty1:
> > The correct way to fix this would be to configure the correct number of encodings based on the captured resolution. If only one layer is needed then we would be effectively disabling simulcast.
> 
> can i work on it?

Yes go ahead and give it a try.

4rayaditya:
@jallamsetty1 So we need to dynamically configure the number of simulcast encodings based on resolution instead of hardcoding

4rayaditya:
The code checks the captured video resolution and determines how many simulcast layers are appropriate (1, 2, or 3).
For VP8 on Chromium at low resolutions (<640px), it reverses the encoding order so the highest quality layer is sent.
For other codecs (VP9, AV1, H264), and higher resolutions, it configures layers and scalability modes according to browser and config settings.
If only one layer is needed, simulcast is effectively disabled.
All these decisions are made dynamically, so the encoding adapts to the user's device, browser, and video settings without hardcoded values.
@saghul @jallamsetty1

jallamsetty1:
> The code checks the captured video resolution and determines how many simulcast layers are appropriate (1, 2, or 3). For VP8 on Chromium at low resolutions (<640px), it reverses the encoding order so the highest quality layer is sent. For other codecs (VP9, AV1, H264), and higher resolutions, it configures layers and scalability modes according to browser and config settings. If only one layer is needed, simulcast is effectively disabled. All these decisions are made dynamically, so the encoding adapts to the user's device, browser, and video settings without hardcoded values. @saghul @jallamsetty1

How did you test if your solution is working as expected? If you already determining the # of simulcast layers based on the capture resolution, why do you have to reverse the encoding order?

4rayaditya:
@jallamsetty1 I checked it on Chrome and tested it using WebRTC internal, although my camera doesnt support 720p or more. I tested it for the low resolution ones and found single encoding in the peer connection.
Although the code determines the number of simulcast layers dynamically but Chromium always selects the fist encoding in the array when one layer is possible at low resolutions. Reversing means highest quality appears first, so it picks best resolution.

4rayaditya:
@jallamsetty1 did you get to check the issues?
Should I try and improve something?

modules/RTC/TPCUtils.ts
```diff
@@ -299,6 +311,69 @@ export class TPCUtils {
         }
 
         return standardSimulcastEncodings;
+        // Dynamically determine number of simulcast layers based on resolution
```
jallamsetty1: How is this code getting executed if it after the return statement? Am I missing something here?

modules/RTC/TPCUtils.ts
```diff
@@ -299,6 +311,69 @@ export class TPCUtils {
...
+        let numLayers = 1;
```
jallamsetty1: Instead of changing the encodings config here based on the captured resolution, you should figure out a way to change the number of simulcast layers needed and configure https://github.com/jitsi/lib-jitsi-meet/blob/f470b5f4fa96c6d5f7601829c966041f6b5a1fd6/service/RTC/StandardVideoSettings.ts#L4 accordingly. That value is used elsewhere to determine the number of simulcast streams and is used for SDP munging and other calculations. This needs to be done only when the video source is configured for the first time it is added to the conference.

4rayaditya: I will definitely look at it. Thanks for the suggestions.

4rayaditya:
@jallamsetty1 please could you rate this approach

- The function ```_getVideoStreamEncodings``` dynamically determines the number of simulcast layers ```numLayers``` based on the video track’s capture resolution:

   - 1 layer for low resolutions (<360p)
   - 2 layers for medium resolutions (360p–719p)
   - 3 layers for high resolutions (≥720p)

- For each layer, it creates a simulcast encoding entry with a unique ```rid``` and a ```scaleFactor``` that determines how much the resolution is downscaled for that layer.
- The function builds arrays for the scale factors and bitrates, matching the number of layers.
- It then constructs the ```simulcastEncodings``` array, where each entry specifies:

   - Whether the layer is active
   - The maximum bitrate for that layer
   - The ```rid``` and ```scaleResolutionDownBy``` (downscale factor)

- For VP8 on Chromium browsers at low resolutions, the array is reversed to ensure the highest quality layer is first (due to a browser quirk).
- If scalability mode is enabled, the function sets the ```scalabilityMode``` property for each encoding as needed.
- The function returns the constructed array, which is then used to configure the RTCRtpSender.

jallamsetty1:
> @jallamsetty1 please could you rate this approach
> 
> *

Did you see my comments above?

4rayaditya:
@jallamsetty1 so I have learnt that this is mostly a Chromium bug and Jitsi is affected by it. So we need a defense mechanism to solve this problem. Would you say we want layers to be counted live in StandardVideoSettings?
Till now what I have worked on is reordering encodings on Chromium for low-res VP8 which is mostly a short term fix. But now that I think about it we could canonicalize simulcast layer count in ```StandardVideoSettings``` / ```getEffectiveSimulcastLayers```.

Although I think the solution I have done till now solves the issue to some extent but the long term approach would be to properly fix the issue. So if you confirm I will work on it.
(I have made some changes but due to my exams I am not able to give my full time.)

4rayaditya:
> > @jallamsetty1 please could you rate this approach
> 
> Did you see my comments above?

yes and I have corrected the bracket error and working on the 2nd one.

4rayaditya:
@jallamsetty1 I have made some improvements, could you check it?
Also, the tpcUtils.spec.ts file is not updated yet if my code changes are right, I will update it. It is showing multiple failures cause of previous hardcoded 3 layers.

4rayaditya:
@saghul @jallamsetty1 Could I please get an update on what to do next?

jallamsetty1:
> @saghul @jallamsetty1 Could I please get an update on what to do next?

This solution is overly complicated and it reverses the encodings which is not needed if the encodings are calculated correctly. A simple solution would be to:
1. Determine the # of simulcast layers needed based on the capture resolution.
2. Update the # of simulcast layers needed during SDP munging.
3. Remove hardcoding of simulcast layers and operate on the current # of encodings.
Once these 3 are addressed, rest of the things will automatically fall into place.

4rayaditya:
@jallamsetty1 I have removed the reverse encoding and sorting as it was complicating things. I have also updated the test file and ran tests, in my opinion now the code should be completely dynamic and no hardcoded part exists.
Could you please review it?

4rayaditya:
Hey guys, I know you have a lot of repos and issues to manage, and I’ve been disturbing you by bringing this issue for quite a while. But if possible, I would love to get another review on this. I really appreciate the time you've spent reviewing this PR and helping me understand the open-source thing.

jallamsetty1:
> Hey guys, I know you have a lot of repos and issues to manage, and I’ve been disturbing you by bringing this issue for quite a while. But if possible, I would love to get another review on this. I really appreciate the time you've spent reviewing this PR and helping me understand the open-source thing.

@4rayaditya I still see the reordering happening here https://github.com/jitsi/lib-jitsi-meet/pull/2948/files#diff-28103bb8c391d124383c4aff137d12198a41376b08af9f71047e61fed1243773R1061. I see a lot of unrelated changes  as well. I am afraid this is too complicated, it affects the core video quality and issue you are trying to fix doesn't warrant overcomplicating the current logic.

4rayaditya:
Let me try again please. I will get back to you again in 2-3 days.
This time hopefully with the solution.

What I will change:
1. Remove re-order logic.
2. Use dynamic layers everywhere.
3. Change SDP munging (change layers from 3 to variable)

Unrelated changes to be reverted:- 

1. encoding order logic in file StandardVideoSettings.ts
2. One in TraceablePeerConnection.ts 
3. Few others in MockClasses.ts

If anything else please guide me.

4rayaditya:
Hey, did you get to check this?

jallamsetty1:
> Hey, did you get to check this?

Sorry this is not going to work. The number of SSRCs being added to the SDP is still 3 like before. We appreciate your enthusiasm though.

4rayaditya:
> > Hey, did you get to check this?
> 
> Sorry this is not going to work. The number of SSRCs being added to the SDP is still 3 like before. We appreciate your enthusiasm though.

Sorry for not being able to solve this. Thanks and can I know what will happen of this issue. Is a maintainer going to solve this or it will stay as it is till another contributor comes and solves it?

PR #2951, opened 2025-12-07
Fix memory leaks and race conditions from improper timeout management

### LINKED ISSUE #2949: [BUG] Memory leaks and race conditions from improper timeout management in ChatRoom, QualityController, and BridgeChannel
### Describe the bug.

Multiple critical modules have improper `setTimeout`/`clearTimeout` patterns that cause memory leaks, race conditions, and zombie timers. This affects long-running conferences and can degrade performance over time.
Impact: Every time a user leaves a conference, an orphaned timeout remains active for 5 seconds even after successful cleanup. In applications with frequent room joins/leaves, this accumulates significantly.

### Expected behavior

When a user successfully leaves a conference room:

1. The `MUC_LEFT` event fires and triggers the `onMucLeft` callback 
2. **The 5-second timeout is NEVER cleared** 
3. Cleanup proceeds (`clean()` is called, promise resolves) 
4. **The timeout remains active and fires 5 seconds after room creation** 
5. When the timeout fires, it calls `onMucLeft(true)` again, potentially:
   - Calling `clean()` on an already-cleaned room
   - Triggering error paths on stale state
   - Keeping closures and references alive in memory

### Screenshots

_No response_

### How to Reproduce

1. Join a conference room
2. Leave the room normally
3. Check active timers (browser dev tools) - orphaned 5s timeout remains
4. Repeat multiple times - timers accumulate

### 🥦 Browser

None

### 👀 Have you checked for similar open issues?

- [x] I checked and didn't find similar issue

### 🏢 Have you checked the Community Forum?

- [x] please make sure you check https://community.jitsi.org

4rayaditya:
- Fix critical 5-second timeout leak in ChatRoom.leave() where timeout was never cleared when MUC_LEFT fired successfully
- Add comprehensive dispose() method to QualityController to clear all active timeouts (_timer, _limitedByCpuTimeout, _lastNRampupTimeout)
- Fix orphaned 5-second keepalive timeout in XmppConnection by storing timeout reference
- Add QualityController disposal to JitsiConference.leave() lifecycle
- Prevent zombie timers, memory leaks, and race conditions in long-running conferences

Fixes #2949

### DIFF
--- a/modules/xmpp/ChatRoom.ts
+++ b/modules/xmpp/ChatRoom.ts
@@ -2288,16 +2288,11 @@ export default class ChatRoom extends Listenable {
         this.lobby?.lobbyRoom && promises.push(this.lobby.leave());
 
         promises.push(new Promise((resolve, reject) => {
-            let timeout: Optional<ReturnType<typeof setTimeout>>;
+            let timeout: number | Timeout = -1;
 
             const onMucLeft = (doReject = false) => {
                 this.eventEmitter.removeListener(XMPPEvents.MUC_LEFT, onMucLeft);
-                
-                // Always clear the timeout to prevent memory leaks and race conditions
-                if (timeout !== undefined) {
-                    clearTimeout(timeout);
-                    timeout = undefined;
-                }
+                clearTimeout(timeout as number);
 
                 // This will reset the joined flag to false. If we reset it earlier any self presence will be
                 // interpreted as muc join. That's why we reset the flag once we have received presence unavalable


modules/xmpp/ChatRoom.ts
```diff
@@ -2288,11 +2288,16 @@ export default class ChatRoom extends Listenable {
...
         promises.push(new Promise((resolve, reject) => {
-            let timeout: number | Timeout = -1;
+            let timeout: Optional<ReturnType<typeof setTimeout>>;
 
             const onMucLeft = (doReject = false) => {
                 this.eventEmitter.removeListener(XMPPEvents.MUC_LEFT, onMucLeft);
-                clearTimeout(timeout as number);
+                
+                // Always clear the timeout to prevent memory leaks and race conditions
+                if (timeout !== undefined) {
```
saghul: You don't need to do this,  this part was working ok before, clearTimeout is resilient to bad input.

modules/xmpp/XmppConnection.ts
```diff
@@ -409,7 +409,8 @@ export default class XmppConnection extends Listenable {
                 } else {
                     // delay it a bit to not interfere with the connection process
                     // and to allow backend to correct any possible split brain issues
-                    setTimeout(() => this._keepAliveAndCheckShard(), 5000);
+                    // Store timeout so it can be cleared if needed
+                    this._wsKeepAlive = setTimeout(() => this._keepAliveAndCheckShard(), 5000);
```
saghul, line 413: Where are you clrearing this?

JitsiConference.ts
```diff
@@ -2559,6 +2559,10 @@ export default class JitsiConference extends Listenable {
             );
         }
 
+        if (this.qualityController) {
+            this.qualityController.dispose();
```
saghul, line 2563: 👍

modules/xmpp/ChatRoom.ts
4rayaditya: ok I have removed the changes.

modules/xmpp/XmppConnection.ts
4rayaditya: So there are 3 instances -
1. around line 423 - When the status becomes DISCONNECTED, it calls ```clearTimeout(this._wsKeepAlive)``` to clean up before attempting to resume the connection.

2. line 469 - The code clears the timeout as part of disconnecting the XMPP connection.

3. around line 537 - ```_maybeStartWSKeepAlive``` clears any existing timeout before scheduling a new keep-alive interval.

4rayaditya:
@saghul I have done the necessary changes. Could you check it again?

modules/qualitycontrol/QualityController.ts
```diff
@@ -115,6 +115,7 @@ export class QualityController {
     private _limitedByCpuTimeout: Optional<number>;
     private _receiveVideoController: ReceiveVideoController;
     private _sendVideoController: SendVideoController;
+    private _timer: Optional<number>;
```
saghul, line 118: Where is this timer set?

4rayaditya: ```_timer``` is set in the _debounce() method (line 184) of QualityController.ts.

The ```_timer``` field stores the timeout ID so it can be properly cleared in the dispose() method

saghul:

saghul (approved):

PR #2953, opened 2025-12-10
fix(e2ee): prevent black screen by improving capability detection and Olm session handling

vishal2005025:
### Fixes: jitsi/jitsi-meet#16735.

This PR resolves the black screen issue triggered when enabling E2EE by improving capability detection and making Olm session handling more reliable.

### Changes
- Added a runtime E2EE capability check (secure context, insertable streams, Olm availability).
- Updated E2EEncryption to refuse enabling E2EE when support is incomplete.
- Added per-participant session state tracking in OlmAdapter to avoid “no session” failures.
- Queued keys and auto-initialized sessions to prevent partial E2EE state.
- Ensured media continues normally when E2EE is unsupported or a session fails.

### Result
E2EE only activates when fully supported. Unsupported or failed E2EE no longer breaks the media pipeline or causes black video.

### DIFF
--- a/JitsiConference.ts
+++ b/JitsiConference.ts
@@ -4314,7 +4314,14 @@ export default class JitsiConference extends Listenable {
      */
     public toggleE2EE(enabled: boolean): void {
         if (!this.isE2EESupported()) {
-            logger.warn('Cannot enable / disable E2EE: platform is not supported.');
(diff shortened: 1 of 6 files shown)


vishal2005025:
@saghul @damencho please take a look.

PR #2956, opened 2025-12-22
fix(JitsiConference): handle mute state updates arriving before track creation

mihhu:
## Description
Fixes race condition where mute state updates can arrive via signaling before the corresponding remote track has been created, causing the initial TRACK_MUTE_CHANGED event to never be emitted.

## How Has This Been Tested?

  - Manual testing with multiple sources per participant
  - Verified events fire on mute state transitions
  - Confirmed aggregate logic works correctly (unmuted if ANY source is unmuted)

## Checklist:

- [x] My change requires a change to the documentation.

### DIFF
--- a/JitsiConference.ts
+++ b/JitsiConference.ts
@@ -3547,9 +3547,15 @@ export default class JitsiConference extends Listenable {
 
         emitter.emit(JitsiConferenceEvents.TRACK_ADDED, track);
 
-        // Emit initial mute state since the track may be created with a mute state
-        // and subsequent setMute() calls won't trigger the event if state doesn't change
-        emitter.emit(JitsiConferenceEvents.TRACK_MUTE_CHANGED, track);
+        // Apply any pending mute state that arrived before the track was created
+        // Always call setMute when there's a pending state, even if it matches current state,
+        // because setMute() is designed to emit on the first call regardless of state change
+        const pendingMutedState = track._getPendingMuteState();
+
+        if (pendingMutedState !== undefined) {
+            track._clearPendingMuteState();
+            track.setMute(pendingMutedState);
+        }
     }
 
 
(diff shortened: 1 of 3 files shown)


saghul (approved):
LGTM. @jallamsetty1 can you PTAL?

saghul:

jallamsetty1:
@mihhu The existing TRACK_MUTE_CHANGED events should fire for remote tracks as well. See https://github.com/jitsi/lib-jitsi-meet/blob/e1078ff20376466efb1934905fc41c5e43d8cfb8/modules/RTC/JitsiRemoteTrack.ts#L464 and https://github.com/jitsi/lib-jitsi-meet/blob/4078190b12a1a162e6e38a9143ab493cd96754af/JitsiConference.ts#L3425. Maybe there is a bug somewhere if that is not happening now.

JitsiConference.ts
```diff
@@ -3546,6 +3546,16 @@ export default class JitsiConference extends Listenable {
         );
 
         emitter.emit(JitsiConferenceEvents.TRACK_ADDED, track);
+
```
jallamsetty1, line 3549:
```suggestion

    // Always apply the final mute state (pending or current).
    const muteState = track._getPendingMuteState() ?? track.isMuted();                                                                                                                                                                                                                                                                                                                        
    track._clearPendingMuteState();                                                                                                                                                                                                                                                                                                                                                           
    track.setMute(muteState);    
```

jallamsetty1: I am wondering if we should always call setMute() track. This should handle the case when presence is available during track creation and there are no subsequent mute events after that initial presence update.

jallamsetty1: nevermind, I think case is already handled.

PR #2960, opened 2026-01-05
fix(xmpp): add React Native DOM compatibility to StropheErrorHandler

MikCam:
## Description

Adds try/catch and null safety for React Native environments where the full DOM API may not be available.

Previously, StropheErrorHandler would crash in React Native when accessing DOM properties like errorEl.children, errorEl.outerHTML, or getElementsByTagName without null checks.

Changes:
- Wrap DOM operations in try/catch block
- Add null coalescing to errorEl.children and getElementsByTagName
- Add conditional check for errorEl.outerHTML before accessing
- Fall back to parse-error on exception with error message captured

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)

## Checklist:

### DIFF
--- a/modules/xmpp/StropheErrorHandler.ts
+++ b/modules/xmpp/StropheErrorHandler.ts
@@ -34,24 +34,32 @@ export function handleStropheError(errResponse: any, context: IStropheErrorConte
     }
 
     if (errorEl) {
-        errorObj.code = errorEl.getAttribute('code') || undefined;
+        try {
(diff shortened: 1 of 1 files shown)


saghul:
LJM requires DOM APIs currently in many places, so we polyfill them in Jitsi Meet: https://github.com/jitsi/jitsi-meet/blob/master/react/features/mobile/polyfills/browser.js

PR #2990, opened 2026-03-07
fix(xmpp) prevent XMPP disconnect when user cancels browser close dialog

raghuramala54:
**PR Description:**
```
## Problem
`handleDisconnect` was registered on both `beforeunload` and `unload` 
events. This caused XMPP to disconnect immediately when the user pressed 
the browser X button, even if they clicked Cancel on the confirmation 
dialog.

## Root Cause
When `disableBeforeUnloadHandlers` is false (default), the events string 
becomes `"beforeunload unload"` and the for loop registers `handleDisconnect` 
on both events. Since `this.disconnect()` was called unconditionally inside 
the handler, XMPP disconnected on `beforeunload` before the user could 
cancel.

## Fix
Added an early return in `handleDisconnect` for `beforeunload` events 
so XMPP only disconnects on `unload`, which fires exclusively when the user 
confirms leaving.

## Test
I tested this with 2 participants and I assured the recording state is achieved back after the user hits cancel on popup.

Related: jitsi/jitsi-meet#16972
Related: jitsi/jitsi-meet#17052

### DIFF
--- a/modules/xmpp/xmpp.ts
+++ b/modules/xmpp/xmpp.ts
@@ -400,9 +400,14 @@ export default class XMPP extends Listenable {
         // they wanted to utilize the connected connection in an unload handler
         // of their own. However, it should be fairly easy for them to do that
         // by registering their unload handler before us.
-        const events = `${this.options.disableBeforeUnloadHandlers ? '' : 'beforeunload '}unload`;
-        const handleDisconnect = ev => {
-            // type-checking added as disconnect returns Promise<void> | boolean
+        const handleDisconnect = (ev: Event) => {
+            // Only disconnect on unload (user confirmed leaving).
+            // Do not disconnect on beforeunload as the user may cancel
+            // the browser close dialog and stay in the meeting.
+            if (ev.type === 'beforeunload') {
+                return;
+            }
+
             const result = this.disconnect(ev);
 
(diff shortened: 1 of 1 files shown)


damencho:
Hum, this is like setting disableBeforeUnloadHandlers to be always true. Basically nothing ever happens on beforeunload?

damencho:
Are the operations in disconnect possible in unload?

raghuramala54:
You're right on both counts. The previous fix was incorrect — early returning on beforeunload was effectively the same as disableBeforeUnloadHandlers=true.

Updated fix: instead of registering handleDisconnect on both events, it now only registers on unload for normal deployments. For WebSocket, the server detects the dropped TCP connection automatically. For BOSH, _cleanupXmppConnection already handles unload correctly via the Beacon API which is designed for this case.

beforeunload is only registered when disableBeforeUnloadHandlers=true (Electron/mobile/iframe), since in those cases the user cannot cancel and it is safe to disconnect there.

damencho:
> For WebSocket, the server detects the dropped TCP connection automatically. 

But this means evertime someone leaves the thumbnail will stay for a minute. As it will always detects it as network interrupt. Unavailable presence for normal leaving will not be sent or bot guaranteed to be sent.

PR #3014, opened 2026-04-09
feat(metadata): Moves reading turn information from metadata.

damencho:


modules/xmpp/strophe.jingle.js
```diff
@@ -364,15 +364,12 @@ export default class JingleConnectionPlugin extends ConnectionPlugin {
     onReceiveStunAndTurnCredentials(res) {
         let iceservers = [];
 
-        findAll(res, ':scope>services>service').forEach(el => {
+        const parseService = ({ type, host, port, username, password, transport }) => {
```
hristoterezov: Can't we extract this function? It doesn't need to be inline I think. Maybe it could return the ice servers and as a result and we can assign them this way. I think this would look simpler. WDYT?

modules/xmpp/xmpp.ts
```diff
@@ -773,14 +775,11 @@ export default class XMPP extends Listenable {
             return;
         }
 
-        this.sendDiscoInfo = false;
-        const foundIceServers = this.connection.jingle.onReceiveStunAndTurnCredentials(msg) || false;
```
hristoterezov, line 777: we don't expect to receive ice servers from system message any more?

damencho: yep, it is removed in https://github.com/jitsi/jitsi-meet/pull/17291

modules/xmpp/strophe.jingle.js
damencho: well ... it is the same ... as we accumulate iceservers ... but I will move it, it will need an extra param

hristoterezov (approved):

PR #3023, opened 2026-04-17
Adds getCaptureTrack() to return the real camera track when an effect is enabled

4rayaditya:
**The Problem:**
When a virtual background or video effect is enabled, `_startStreamEffect()` swaps out `this.track` with a Canvas-based output track, while keeping the actual camera stream hidden away in the private `this._originalStream` field.

This causes issues if an application needs to apply physical hardware constraints to the camera (like Pan/Tilt/Zoom, hardware focus, etc). If you try to call `applyConstraints()` on the current `this.track` while an effect is running, Chrome gets confused applying physical constraints to a Canvas track and throws an `UnknownError`. The UI might look fine, but the camera ignores the command.

**The Fix:**
- Added `getCaptureTrack()` to `JitsiTrack` and `JitsiLocalTrack`. It safely checks if `_originalStream` exists (meaning an effect is running) and returns the real camera track. If there's no effect, it just returns `this.track`.
- Updated `applyConstraints()` in `JitsiLocalTrack` to intercept `MediaTrackConstraints` and proxy them straight to the physical capture track provided by `getCaptureTrack()`. (Audio constraints are handled exactly as before).

**Testing:**
I added a quick test suite for `JitsiLocalTrack` to verify that the accessor handles effect mocks cleanly and that `applyConstraints()` doesn't accidentally touch the Canvas track. Ran it locally through Karma and all tests pass. I have not commited that test suite, but if asked I will commit it.

### DIFF
--- a/modules/RTC/JitsiLocalTrack.ts
+++ b/modules/RTC/JitsiLocalTrack.ts
@@ -754,6 +754,20 @@ export default class JitsiLocalTrack extends JitsiTrack {
     }
 
     /**
+     * Returns the underlying capture MediaStreamTrack.
+     */
+    public getCaptureTrack(): MediaStreamTrack | null {
+        if (this._originalStream) {
+            const kind = this.track?.kind;
+            const tracks = this._originalStream.getTracks().filter(t => t.kind === kind);
+
+            return tracks.length ? tracks[0] : this._originalStream.getTracks()[0] ?? null;
+        }
+
+        return this.track ?? null;
+    }
+
+    /**
      * Returns device id associated with track.
      *
      * @returns {string}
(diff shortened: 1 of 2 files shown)


saghul:
Isn't `getOriginalStream` what you'd want?

4rayaditya:
> Isn't `getOriginalStream` what you'd want?

I wanted to create a function which gives access to the real camera track. Whenever a effect is ON, this.stream returns the canvas track and so there was no function which gave the access to the real track. So adding this getCaptureTrack would give someone working on PTZ cameras the real camera access instead of the effect one. 

In turn we can change the camera angle and viewing properly.

PR #3027, opened 2026-04-23
feat: Removes ignore of visitor redirect for sip-jibri.

damencho:


jallamsetty1 (approved):

modules/xmpp/moderator.js
```diff
@@ -191,7 +191,7 @@ export default class Moderator extends Listenable {
             conferenceRequest.sessionId = sessionId;
         }
 
-        if (!config.iAmRecorder && !config.iAmSipGateway) {
+        if (!config.iAmRecorder || config.iAmSipGateway) {
```
bgrozev, line 194: The previous code suggests that sip-jibri sets `iAmRecorder=false, iAmSipGateway=true`. If that's the case the `|| config.iAmSipGateway` part is superfluous (both here and below)

damencho: Nope, sip-jibri set `iAmRecorder=true, iAmSipGateway=true`

PR #3028, opened 2026-04-27
fix(e2ee): make ExternallyManagedKeyHandler work correctly with the E2EE worker

encedo:
The External API `set-media-encryption-key` command has been present for several
releases, but `ExternallyManagedKeyHandler.setKey()` was effectively broken for
any caller: it forwarded `undefined` as `participantId` (rejected by the worker)
and passed an object wrapper instead of raw bytes (producing an empty key in the
worker).  This PR makes the existing documented API actually work.

**Root cause analysis**

The worker (`Worker.ts`) validates `participantId` strictly:

```ts
if (!participantId || keyIndex === undefined) {
    throw new Error('Missing required data: participantId or keyIndex');
}
```

The handler passed `undefined`, so every call threw silently in the worker
thread (caught only by the generic `onerror` logger).

Additionally, `Context.setKey(key: Uint8Array | ArrayBuffer | false)` expects
raw key material for HKDF derivation.  The handler was passing
`{ encryptionKey: CryptoKey }` — an object — so `new Uint8Array(key)` in the
worker produced an empty buffer.

**Changes**

- `ExternallyManagedKeyHandler`: read `e2ee.externallyManagedSharedKey` from
  config (default `true` for backward compatibility); accept `participantId`
  from `keyInfo`, fall back to `conference.myUserId()`; forward raw bytes to
  `e2eeCtx.setKey()`.
- `JitsiConference.ts`: correct the `setMediaEncryptionKey()` TypeScript
  signature (was `CryptoKey`, is now the correct plain-object shape).
- `ExternallyManagedKeyHandler.spec.ts`: 9 new Jasmine tests covering sharedKey
  propagation, participantId fallback, and the full worker postMessage path.

**Backward compatibility**

`e2ee.externallyManagedSharedKey` defaults to `true`, preserving the existing
single-context behaviour.  No config changes are required for deployments that
do not need per-sender keys.


encedo:

Done, signed.

PR #3031, opened 2026-05-05
fix(RTC): Spec-compliant simulcast on Firefox

jallamsetty1:
Firefox 138+ enforces the WebRTC spec strictly on RTCRtpSender.setParameters: the encodings array length and rid values cannot change after the transceiver is created, otherwise the call is rejected with InvalidModificationError.

### DIFF
--- a/modules/RTC/TPCUtils.ts
+++ b/modules/RTC/TPCUtils.ts
@@ -661,29 +661,30 @@ export class TPCUtils {
      */
     injectSsrcGroupForSimulcast(desc: RTCSessionDescription): RTCSessionDescription {
         const sdp = transform.parse(desc.sdp);
-        const video = sdp.media.find(mline => mline.type === 'video');
+        const videoMlines = sdp.media.filter(mline => mline.type === 'video');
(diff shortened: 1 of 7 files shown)


modules/RTC/TPCUtils.ts
```diff
@@ -661,30 +661,31 @@ export class TPCUtils {
      */
     injectSsrcGroupForSimulcast(desc: RTCSessionDescription): RTCSessionDescription {
         const sdp = transform.parse(desc.sdp);
-        const video = sdp.media.find(mline => mline.type === 'video');
+        const videoMlines = sdp.media.filter(mline => mline.type === 'video');
```
jallamsetty1, line 664: This is needed so that sim group is inserted for all the video sources in the call when signaling sources to JIcofo not just the first video source.

damencho (approved):

paulmenzel:
See related Firefox bug reports regarding older Jitsi installations:

1.  https://bugzilla.mozilla.org/show_bug.cgi?id=2040316

PR #3047, opened 2026-06-02
fix(xmpp): remove window unload listeners on disconnect

damencho:
The beforeunload/unload listeners added in the XMPP constructor were never removed, so each connection created over the lifetime of the page leaked its listeners on the window object. Keep a reference to the handler and remove it in disconnect().


jallamsetty1 (approved):

damencho:
jenkins test this please

PR #3050, opened 2026-06-12
Add optional messageId support for private messages to enable message editing

Alok-work23:
## Description
This PR adds an optional messageId parameter to both sendPrivateMessage and sendTextPrivateMessage.
Previously, private messages were sent without allowing a custom message ID to be specified. This made it difficult to support message editing because both the sender and receiver must reference the same message identifier when an edit operation is performed.

## Related Issue
NA

## Motivation and Context
To support editing of private messages, both participants must be able to identify the same message using a shared message ID. The current implementation generates messages without exposing a way to set a custom ID, making it difficult to reliably reference a previously sent private message for updates.
This change allows applications to provide a consistent messageId when sending private messages, enabling future edit operations to target the correct message on both the sender and receiver sides.

## How Has This Been Tested?
NA

## Types of changes

## Checklist:


Alok-work23:
I have already signed CLA.

mihhu (approved):

PR #3053, opened 2026-06-17
feat(RTC): demux forwarded media by bridge-stamped mid under SSRC rewriting

JonathanLennox:
## Summary

Consume the per-slot mid the bridge signals in the `Audio`/`VideoSourcesMap` messages (and stamps on rewritten RTP) and use it as the receive m-line's mid, so the client demuxes forwarded media by the `sdes:mid` header extension. This disables Chrome's payload-type demuxing fallback — the trigger for the silent audio-demux wedge — and works by spec in every browser.

Part of the cross-repo **bridge-stamped mid** change. Companion `mid-demux-stamping` branches in **jitsi-xmpp-extensions** (the capability constant), **jicofo** (emits the capability), and **jitsi-videobridge** (stamps the mid); meant to be tested together.

## Details

- `FeatureFlags.isRtpMidDemuxSupported()`: new `rtpMidDemuxEnabled` flag, gated on SSRC rewriting. Advertised to jicofo as the `http://jitsi.org/rtp-mid-demux` disco feature.
- `processSourceMap` carries the mid onto the source element; the SDP layer uses it for `a=mid` (and the BUNDLE group) in `addMlineForNewSource` instead of the numeric m-line index. The cloned template m-line keeps the negotiated `sdes:mid` extmap.
- `_stripSdesMid` is skipped when mid demux is enabled: Firefox only mishandled `sdes:mid` when it was negotiated but not sent, which no longer applies once the bridge actually stamps the mid.

Gated off by default (the `rtpMidDemuxEnabled` flag plus the capability handshake); no behavior change until enabled.

## Tests

`SDP.spec` (`addMlineForNewSource` with a supplied mid + the BUNDLE/extmap, and index fallback) and `JingleSessionPC.spec` (mid parsing). `npm run lint` (eslint + tsc) and the affected specs pass.


hristoterezov (approved):

hristoterezov (approved):
LGTM! but will need to be rebased.

jallamsetty1:
Jenkins test this please

PR #3056, opened 2026-06-23
feat(translation) Support realtime translations in conferences

jallamsetty1:


service/RTC/SignalingLayer.ts
```diff
@@ -79,6 +79,18 @@ export function getSourceIndexFromSourceName(sourceName: SourceName): number {
...
+ * Checks whether the given source name belongs to a translated audio source. Translated sources follow the
+ * convention {@code <regularSourceName>.<ISO-639-1-language>}, eg. endpointA-a0.en, whereas regular source
+ * names never contain a dot.
+ *
+ * @param {SourceName} sourceName - The source name to check (callers may pass a null/undefined source name).
+ * @returns {boolean}
+ */
+export function isTranslatedSourceName(sourceName?: SourceName | null): boolean {
+    return typeof sourceName === 'string' && sourceName.includes('.');
+}
```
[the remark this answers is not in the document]
jallamsetty1: source names are always generated using that convention in the client.

JitsiConference.ts
```diff
@@ -3262,6 +3512,87 @@ export default class JitsiConference extends Listenable {
...
+     */
+    public clearTranslation(): void {
+        if (this._receiverTranslationLanguage === null && this._participantTranslationLanguages.size === 0) {
+            return;
+        }
+        this._receiverTranslationLanguage = null;
+        this._participantTranslationLanguages.clear();
+
+        this._updateTranslationRequests();
+        this._updateTranslationSubscription();
```
hristoterezov, line 3593: **[bug] `clearTranslation()` never re-evaluates P2P, so a 1-on-1 call stays pinned to JVB.**

Every other translation mutator (`setReceiverTranslationLanguage`, `setParticipantTranslationLanguage`) calls `this._maybeStartOrStopP2P()`, but `clearTranslation()` doesn't. Enabling translation forces JVB (via the new `_buildDesiredTranslations().size === 0` term in `_shouldBeInP2PMode`), so turning it back off here leaves a 2-party call on the bridge indefinitely — until an unrelated join/leave happens to re-trigger the check. The client dispatches `clearTranslation()` on a moderator room-disable and on `SUBSCRIPTION_LIMIT_REACHED`, so the path is exercised. Add a `this._maybeStartOrStopP2P()` call here.

JitsiConference.ts
```diff
@@ -3262,6 +3512,87 @@ export default class JitsiConference extends Listenable {
...
+     * Enables (or disables, when null) translated audio for a single remote speaker, overriding the default
+     * set via {@link setReceiverTranslationLanguage}. Sends the request-stanza delta and updates the
+     * cumulative audio subscription.
+     *
+     * @param {string} participantId - The speaker's endpoint id.
+     * @param {string|null} language - The target language (2-letter ISO code), or null to disable for this speaker.
+     * @returns {void}
+     */
+    public setParticipantTranslationLanguage(participantId: string, language: string | null): void {
+        const current = this._participantTranslationLanguages.has(participantId)
```
hristoterezov, line 3554: **[bug] resolved-vs-stored asymmetry can silently pin a participant.**

`getParticipantTranslationLanguage()` returns the *resolved* effective language (override ?? default), but this setter's no-op guard compares the *stored* override. A read-modify-write — read `'en'` (inherited from the default), then set `'en'` for the same participant — writes an explicit override (`current` was `undefined`). After that, changing the default via `setReceiverTranslationLanguage('fr')` no longer affects this participant: the stale override wins.

The client hits exactly this: its reducer always writes `participantLanguages[id] = language` and its failure-revert writes `null` (jitsi/jitsi-meet#17538), so a transient failure permanently shadows the default with 'off'. Consider treating a per-participant value equal to the resolved default as 'inherit' (delete the entry), or at least documenting the asymmetry between the getter and the guard.

JitsiConference.ts
```diff
@@ -2087,6 +2147,186 @@ export default class JitsiConference extends Listenable {
...
+
+        if (Object.keys(delta).length === 0) {
+            return;
+        }
+
+        // Only commit the advertised state once the delta has actually been sent. Otherwise a request issued
+        // before the audio-translation component is discovered would be dropped: the desired state would be
+        // recorded as advertised, so subsequent deltas would compare against it and never re-send.
+        if (this._sendTranslationRequestStanza(delta)) {
+            this._translationRequests = desired;
```
hristoterezov, line 2225: **[robustness] the optimistic advertised state is never rolled back on failure.**

`_translationRequests` is committed to `desired` as soon as the stanza is *sent*; `_onTranslationRequestError` doesn't revert it. So a request the component later rejects still counts as 'advertised' — retrying the same language yields an empty delta and never re-sends. In the integrated product this is masked because the client always reverts on failure (clear-all on limit-reached, set-null per endpoint otherwise), which sends a cancel delta and resets this map. But any other lib consumer would be stuck. Consider reverting the affected entries in the error handler so the lib can recover on its own.

JitsiConference.ts
```diff
@@ -2087,6 +2147,186 @@ export default class JitsiConference extends Listenable {
...
+
+        // Tag the request so a (failure-only) error reply from the component can be mapped back to it.
+        const id = `audio-translation-${++this._translationRequestSeq}`;
+
+        this._pendingTranslationRequests.set(id, delta);
+
+        // The component stays silent on success, so drop the pending entry after a grace period.
+        window.setTimeout(() => this._pendingTranslationRequests.delete(id), TRANSLATION_REQUEST_TIMEOUT);
+
+        const message = $msg({ id,
```
hristoterezov, line 2254: **[reuse/altitude] inline component-message construction diverges from the established per-component module pattern.**

`RoomMetadata`, `Polls`, and `FileSharing` each route component-targeted messages through a dedicated `modules/xmpp/*` module; this PR hand-rolls `$msg(...).c('audio-translation', { xmlns }).t(...)` + `connection.send`, the component-address lookup, and the error-handler registration inline in `JitsiConference.ts`. Consider extracting an `AudioTranslation` xmpp module for consistency. Also: those other modules set a `room` attribute on the child element and this one doesn't — worth confirming the component can associate the request with the conference without it (e.g. purely from the sender JID).

JitsiConference.ts
```diff
@@ -2087,6 +2147,186 @@ export default class JitsiConference extends Listenable {
...
+    /**
+     * Sends a translation-request delta to the audio-translation module as a JSON map.
+     *
+     * @param {Object} delta - The changed endpointId -> language entries ('' to cancel).
+     * @returns {boolean} Whether the delta was sent (false when no component has been discovered yet).
+     */
+    private _sendTranslationRequestStanza(delta: { [endpointId: string]: string; }): boolean {
+        const componentAddress = this.xmpp.audioTranslationComponentAddress;
+
+        if (!componentAddress) {
```
hristoterezov, line 2238: **[edge] a request issued before the component is discovered is silently lost with no retry.**

When `componentAddress` is undefined this returns `false` and (correctly) doesn't commit `_translationRequests` — but nothing re-runs `_updateTranslationRequests()` once `audioTranslationComponentAddress` becomes known. If a language is set before disco completes and no roster change follows, the request is never re-sent. Disco usually completes during join so the window is small; still, a re-trigger on component discovery (or piggy-backing the existing bridge-channel-open `resendSubscription` path) would close it.

modules/qualitycontrol/ReceiveAudioController.ts
```diff
@@ -3,94 +3,136 @@ import { getLogger } from '@jitsi/logger';
...
+     * Mutes or unmutes all remote audio by toggling the baseline. The explicit include/exclude overlays (e.g.
+     * opted-in translated sources) are preserved so they are restored on unmute rather than lost across the cycle.
+     *
+     * @param {boolean} muted - Indicates whether the remote audio should be muted.
      * @returns {void}
      */
     muteRemoteAudio(muted: boolean): void {
         this.setAudioSubscriptionMode({
-            mode: muted ? ReceiverAudioSubscription.NONE : ReceiverAudioSubscription.ALL
+            all: !muted,
```
hristoterezov, line 83: **[behavior change] `muteRemoteAudio(true)` no longer guarantees silence.**

It now only flips the `all` baseline and preserves `include`, so any opted-in translated sources keep playing after a caller asks to mute all remote audio (pre-PR `mode: NONE` cleared the list -> full silence). The JSDoc says this is deliberate (restore the overlays on unmute), so flagging for confirmation rather than as a defect: if a 'mute all remote audio' UI is ever wired to this, the user would still hear AI-translated speech. This client PR doesn't appear to call `muteRemoteAudio`, so impact is latent — but the public-API contract change deserves a deliberate decision.

service/RTC/SignalingLayer.ts
```diff
@@ -79,6 +79,18 @@ export function getSourceIndexFromSourceName(sourceName: SourceName): number {
...
+/**
+ * Checks whether the given source name belongs to a translated audio source. Translated sources follow the
+ * convention {@code <regularSourceName>.<ISO-639-1-language>}, eg. endpointA-a0.en, whereas regular source
+ * names never contain a dot.
+ *
+ * @param {SourceName} sourceName - The source name to check (callers may pass a null/undefined source name).
+ * @returns {boolean}
+ */
+export function isTranslatedSourceName(sourceName?: SourceName | null): boolean {
+    return typeof sourceName === 'string' && sourceName.includes('.');
```
hristoterezov, line 91: I'm wondering isn't there a more reliable way to signal this is a translated source?

hristoterezov (approved):

hristoterezov:
Reviewed alongside the client PR jitsi/jitsi-meet#17538. Overall the subscription generalization ({ mode, list } -> { all, include, exclude }) is a clean refactor done at the right altitude, and the failure-event plumbing is reasonable. Findings below are ordered roughly by severity; the clearTranslation() P2P omission is the main correctness one. Several are paired with the client behavior (noted inline).

I would definitely extract the logic currently defined in JitsiConference class into a new class dedicated to the audio translation.

My initial concern that the public setters lack an audioTranslation.enabled guard is mitigated in practice because the client only renders the translation UI when config.audioTranslation.enabled is set — but the lib API itself remains unguarded, so misuse by any other consumer would tear down P2P (via the new _buildDesiredTranslations().size gate) and push a bogus include-subscription with no translation benefit. Worth a defensive early-return if you want the API to be safe standalone.

JitsiConference.ts
jallamsetty1: yes I removed the call to `this._maybeStartOrStopP2P()` on purpose since user turning off translations and enabling it again in a 2 participant call will make the call switch between p2p and jvb which is redundant in my opinion.

PR #3081, opened 2026-08-04
fix(xmpp): strip XML-invalid characters from chat messages

prathamesh04:
Fixes the protocol-side part of https://github.com/jitsi/jitsi-meet/issues/17267, as suggested by @damencho in https://github.com/jitsi/jitsi-meet/pull/17677#issuecomment-5175894528.

### What

Sending a message containing a character that is not allowed in an XML 1.0 document (e.g. the START OF TEXT control character U+0002) produces a malformed stanza. The XMPP server (e.g. Prosody) then terminates the stream, dropping every participant from the meeting.

This sanitizes the message in the XMPP layer, right before building the stanza:

- `sendMessage()` and `sendPrivateMessage()` in `ChatRoom` now run the message through `stripXMLInvalidChars()`.
- The sanitized text is emitted in `SENDING_CHAT_MESSAGE` / `SENDING_PRIVATE_CHAT_MESSAGE` so the UI history never shows the raw control character.
- `SENDING_PRIVATE_CHAT_MESSAGE` also gains a test (it previously had none).

Since the sanitization now happens in lib-jitsi-meet, the client-side strip in https://github.com/jitsi/jitsi-meet/pull/17677 is kept only as defense-in-depth.

### Test

- `npx eslint --max-warnings 0` on the touched source files: pass
- `npx tsc --noEmit`: pass
- Karma: `ChatRoom` spec 42/42 pass (3 new `sendMessage` tests + 2 new `sendPrivateMessage` tests)

cc @damencho

### DIFF
--- a/modules/util/XMLUtils.spec.ts
+++ b/modules/util/XMLUtils.spec.ts
@@ -1,0 +1,37 @@
+import { stripXMLInvalidChars } from './XMLUtils';
+
+describe('stripXMLInvalidChars', () => {
+    it('strips the XML-invalid C0 control characters', () => {
+        // U+0000 - U+0008, U+000B, U+000C, U+000E - U+001F
+        expect(stripXMLInvalidChars('\u0000\u0001\u0002\u0007\u0008\u000B\u000C\u000E\u001F')).toBe('');
+        expect(stripXMLInvalidChars('a\u0000b\u000Bc\u000Cd\u001Fe')).toBe('abcde');
+    });
+
+    it('preserves the XML-valid whitespace characters', () => {
+        expect(stripXMLInvalidChars('\u0009\u000A\u000D')).toBe('\u0009\u000A\u000D');
+        expect(stripXMLInvalidChars('line1\tline2\nline3\rline4')).toBe('line1\tline2\nline3\rline4');
+    });
+
+    it('strips the non-characters U+FFFE and U+FFFF', () => {
+        expect(stripXMLInvalidChars('a\uFFFEb\uFFFFc')).toBe('abc');
+    });
+
+    it('strips lone surrogates', () => {
+        expect(stripXMLInvalidChars('a\uD800b\uDFFFc')).toBe('abc');
+    });
+
+    it('preserves astral characters encoded as surrogate pairs', () => {
+        expect(stripXMLInvalidChars('emoji 😀 is fine')).toBe('emoji 😀 is fine');
+        expect(stripXMLInvalidChars('\u{1F600}')).toBe('\u{1F600}');
+    });
+
+    it('preserves characters in the valid XML ranges', () => {
+        expect(stripXMLInvalidChars('Hello, world! 123')).toBe('Hello, world! 123');
+        // Boundary checks for the allowed ranges: U+20, U+D7FF, U+E000, U+FFFD
+        expect(stripXMLInvalidChars('\u0020\uD7FF\uE000\uFFFD')).toBe('\u0020\uD7FF\uE000\uFFFD');
+    });
+
+    it('handles the empty string', () => {
+        expect(stripXMLInvalidChars('')).toBe('');
+    });
+});
--- a/modules/util/XMLUtils.ts
+++ b/modules/util/XMLUtils.ts
@@ -154,3 +154,33 @@ export function findFirst(element: Element | Document, selector: string): Elemen
         return null;
     }
 }
+
+/**
+ * Matches the characters that are not allowed in an XML 1.0 document. Sending
+ * such characters (e.g. the START OF TEXT control character U+0002) in a
+ * message produces a malformed stanza which makes the server terminate the
+ * stream, dropping everyone from the meeting.
+ *
+ * The pattern is the inverse of the characters allowed by the XML 1.0
+ * specification:
+ *
+ *     #x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
+ *
+ * Encoding the allowed ranges (rather than denying a list of forbidden ones)
+ * keeps every permitted code point intact - including astral characters such
+ * as emoji, which are represented as surrogate pairs - while stripping lone
+ * surrogates (U+D800-U+DFFF), which are invalid in XML 1.0. The unicode flag
+ * makes the pattern operate on whole code points instead of surrogate halves.
+ */
+// eslint-disable-next-line no-control-regex
+const INVALID_XML_CHARS_REGEXP = /[^\u0009\u000A\u000D\u0020-\uD7FF\uE000-\uFFFD\u{10000}-\u{10FFFF}]/gu;
+
+/**
+ * Removes the characters that are not allowed in an XML 1.0 document from the
+ * given string.
+ * @param text - The text to sanitize.
+ * @returns The text without the XML-invalid characters.
+ */
+export function stripXMLInvalidChars(text: string): string {
+    return text.replace(INVALID_XML_CHARS_REGEXP, '');
+}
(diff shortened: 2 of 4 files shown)


damencho (commented):
## Review: `fix(xmpp): strip XML-invalid characters from chat messages`

Good fix for a genuine crash bug. The approach is correct and the scope is well-targeted. A few issues to address before merging:

---

### Issues

**1. Lone surrogates not covered (minor correctness gap)**

The XML 1.0 spec defines valid characters as:

```
#x9 | #xA | #xD | [#x20-#xD7FF] | [#xE000-#xFFFD] | [#x10000-#x10FFFF]
```

U+D800-U+DFFF (lone surrogates) are also invalid but absent from the regex. JavaScript strings can contain unpaired surrogates (e.g. `'\uD800'`), which survive the current filter and still produce malformed XML. Consider adding `\uD800-\uDFFF` to the character class. Alternatively, an allowlist written as an inverse of the spec's valid ranges is more self-documenting than a denylist -- it directly encodes what is permitted rather than what is forbidden.

**2. The json-message test does not actually exercise stripping (`ChatRoom.spec.ts`)**

`JSON.stringify` escapes U+0002 to the literal six-character sequence `\u0002` (backslash + u + digits), so `stripXMLInvalidChars` receives no actual control character and is a no-op here. The test title says "strips XML-invalid control characters from json-message stanzas" but it only verifies that already-safe JSON passes through unchanged. Either rename the test to reflect what it does, or build the malformed input by hand so the raw U+0002 byte is present in the string passed to `sendMessage`.

**3. Runtime type guard is misleading in a TypeScript-typed function (`XMLUtils.ts`)**

The parameter is typed `string`, so TypeScript callers can never legitimately pass a non-string. The `typeof text !== 'string'` guard returns `text` (a non-string) typed as `string` -- a contract violation. The guard should either be removed (TypeScript guarantees the type), or the signature broadened to `(text: unknown): string` with a proper fallback (e.g. `return ''`) if runtime safety for JS callers is desired.

**4. `setSubject()` is not covered**

`setSubject` passes text content directly to `msg.c('subject', valueToProcess)` without sanitization. A subject string containing U+0002 would still crash the stream. Worth closing for a complete fix.

**5. Duplicated inline comments**

The three-line explanation comment is copy-pasted identically into both `sendMessage` and `sendPrivateMessage`. Since `stripXMLInvalidChars` and its JSDoc already explain the why, two identical inline comments are clutter -- one or neither is sufficient.

---

### What's Good

- The regex correctly preserves `\u0009` (tab), `\u000A` (LF), `\u000D` (CR) -- valid XML whitespace -- while stripping the rest of the C0 control block and `\uFFFE`/`\uFFFF`.
- Emitting `cleanMessage` (not the original `message`) on both XMPP events is the right call -- UI history should reflect what was actually transmitted.
- The `sendMessage` tests (strip + preserve-whitespace) are solid.
- Adding tests for `sendPrivateMessage` fills a real gap (it had zero coverage before).
- The `eslint-disable-next-line no-control-regex` placement is correct.

prathamesh04:
Thanks for the thorough review @damencho — all five points addressed in [`892060c`](https://github.com/jitsi/lib-jitsi-meet/pull/3081/commits/892060c).

**1. Lone surrogates → allowlist per XML 1.0 spec**

Switched the denylist for an allowlist that is the inverse of the spec's valid ranges:
```ts
/[^\u0009\u000A\u000D\u0020-\uD7FF\uE000-\uFFFD\u{10000}-\u{10FFFF}]/gu
```
With the `u` flag the pattern operates on whole code points, so astral characters (emoji, surrogate pairs) are preserved while lone surrogates (U+D800–U+DFFF) are stripped. I went with the allowlist rather than adding `\uD800-\uDFFF` to the denylist because a raw `\uD800-\uDFFF` range would have matched each half of a valid surrogate pair and eaten every astral character.

**2. json-message test now actually exercises stripping**

You were right — `JSON.stringify` escapes U+0002 into the literal `\u0002` sequence, so the sanitizer was a no-op. The test now builds the malformed JSON by hand with a raw control char (`'{"text":"hello\u0002world"}'`) and asserts it is removed.

**3. Misleading runtime type guard removed**

Dropped the `typeof text !== 'string'` guard from `stripXMLInvalidChars`. The parameter is typed `string`, so the guard could never fire legitimately, and returning a non-string typed as `string` was a contract violation. This surfaced two legacy tests in `ChatRoom.spec.ts` that passed an object to `sendMessage` via `as any` (relying on Strophe's `Builder.c()` treating a non-string text arg as attributes). Since `JitsiConference.sendMessage` always stringifies objects before they reach `ChatRoom.sendMessage`, and the method is string-typed, I removed those two out-of-contract tests.

**4. `setSubject()` covered**

It now sanitizes via `stripXMLInvalidChars(subject.trim())` before building the stanza, and caches the sent value (`this.subject = valueToProcess`) so the duplicate-subject early return actually works — previously it could never fire because `this.subject` was never assigned. Added tests for trimming, control-char stripping, valid-whitespace preservation, exact dedup, and dedup across sanitized-equivalent inputs.

**5. Duplicated inline comments removed**

Dropped both copies from `sendMessage`/`sendPrivateMessage`; the JSDoc on `stripXMLInvalidChars` already documents the why.

I also added a dedicated `modules/util/XMLUtils.spec.ts` unit suite covering the allowlist boundaries (C0 controls, tab/LF/CR, U+FFFE/U+FFFF, lone surrogates, surrogate pairs, U+20/D7FF/E000/FFFD boundaries, empty string).

Verification: `npm run lint` (incl. `tsc --noEmit`) clean, full Karma suite green — **552 SUCCESS** (native + polyfill).

damencho:
Have you signed the CLA? As I do not see a matching email to the changes provided.

damencho (approved):

prathamesh04:
Hi @damencho, yes — I have just signed the Individual Contributor License Agreement, under the email that authors the commits on this PR. Could you please re-check? Let me know if it still does not show up on your end. Thanks!

damencho:
I see the signed document but the email from it is different from the one from your changes:
https://github.com/prathamesh04/lib-jitsi-meet/commit/8ef593737ac3ed88fa23c96a6acada9fd804dbbf.patch

prathamesh04:
Hi @damencho, thanks for pointing that out. I have rewritten the author email on both commits to the one I used when signing the ICLA (`prathamhukkeri04@gmail.com`) and force-pushed the branch — could you please check again? Thanks!

damencho:
I think something went wrong with the force push

prathamesh04:
You're right, thanks for flagging it. The email rewrite (via git filter-branch) rewrote not only the two feature commits but also the upstream master commits that had been merged into the branch, giving them new hashes. The branch then no longer contained any real upstream master commit as an ancestor, so GitHub started reporting add/add conflicts on files master had added independently (`modules/util/XMLUtils.ts`, `modules/xmpp/ChatRoom.ts`, `modules/xmpp/ChatRoom.spec.ts`).

I've rebuilt the branch on the current `upstream/master` (`0e333927`) with the exact same content (verified byte-for-byte identical to the previously approved tree) as a single commit authored with the ICLA email `prathamhukkeri04@gmail.com`. The PR is back to MERGEABLE with no conflicts. Could you please re-check?

prathamesh04:
For reference: the current head `389a4e9` is byte-for-byte identical to `892060c` (the version you reviewed and approved) on all four changed files (`modules/util/XMLUtils.ts`, `modules/util/XMLUtils.spec.ts`, `modules/xmpp/ChatRoom.ts`, `modules/xmpp/ChatRoom.spec.ts`) — verified via `git diff 892060c HEAD`. The only differences are that it's now based directly on current `upstream/master` and authored with the ICLA email `prathamhukkeri04@gmail.com`.

damencho:
jenkins test this please

PR #3083, opened 2026-08-06
feat: Add in-place ICE restart of the JVB session

bgrozev:
Implements the client side of an in-place ICE restart of the JVB session. Instead of tearing the session down and being re-invited, the client applies a new set of bridge ICE credentials in place, keeping the same DTLS session, SSRCs and RTP state so media keeps flowing throughout.

### Flow

1. `JitsiConference.restartJvbIce(reason)` asks jicofo for a restart with a Jingle `session-info` carrying `<bridge-session ice-restart="true"/>`, and resolves when the IQ is acked.
2. The bridge rotates its credentials and jicofo relays the new transport back as a `transport-info` tagged with an `ice-generation`. `strophe.jingle.js` routes any such transport-info to `JingleSessionPC.onBridgeIceRestartTransport()`.
3. That runs as a single atomic modification-queue task: monotonic generation guard, then a patched offer built from the current remote description with the new ufrag/pwd swapped in and every candidate line stripped, then `setRemoteDescription` / `createAnswer` / `setLocalDescription`, then the bridge's candidates added back via `addIceCandidate()`, then our own new credentials signalled back tagged with the same generation.

The client stays the **answerer** throughout and uses a real `createAnswer()`. Since jicofo already drives renegotiation with offer-shaped Jingle IQs in production, no new offerer machinery is needed, and the whole restart becomes a local operation with no signaling round trip on the critical path. Measured against a real bridge, the queue is held for 85ms.

### Why the candidates are stripped and re-added

Applying the new candidates in the same `setRemoteDescription()` as the new credentials makes libwebrtc stamp them with the new ICE generation (candidate lines carry no credentials of their own), treat them as brand new candidates for the same remote address, and synchronously tear down the selected candidate pair, which defeats the make-before-break this feature exists for. Trickling them in after the offer/answer cycle avoids it.

Reported upstream as https://issues.webrtc.org/issues/543082385

### Gating and fallback

Gated on the `enableIceRestart` config flag. The existing reactive path is unchanged: `_restartJvbIceWithFallback()` still falls back to the legacy full session restart via `IceFailedHandling` when an in-place restart is rejected or does not recover the connection. `APP.conference._room.restartJvbIce()` triggers one manually.

A 3s/100ms media-stats sampler is available behind `config.testing.debugIceRestart`, off by default.

### Tests

15 new tests. `SDPUtil.replaceIceCredentialsAndStripCandidates` is covered for credential replacement, candidate stripping, byte-for-byte preservation of every other line, and both line-separator conventions. `onBridgeIceRestartTransport` is covered for the request element shape, the generation guard (duplicate, older, invalid, and newer-after-a-failure), and for `addIceCandidate` happening strictly after `setLocalDescription`, asserted via jasmine invocation order.

Part of a multi-repo change; all PRs share the `in-place-ice-restart` branch name so CI tests them together:

- jitsi-xmpp-extensions: https://github.com/jitsi/jitsi-xmpp-extensions/pull/148
- jitsi-videobridge: https://github.com/jitsi/jitsi-videobridge/pull/2438
- jicofo: https://github.com/jitsi/jicofo/pull/1299
- jitsi-meet: https://github.com/jitsi/jitsi-meet/pull/17700

### DIFF
--- a/modules/xmpp/JingleSessionPC.ts
+++ b/modules/xmpp/JingleSessionPC.ts
@@ -1873,6 +1873,9 @@ export default class JingleSessionPC extends JingleSession {
                 logger.info(`${this} ${ICE_RESTART_LOG_PREFIX} gen=${generation} t+${dt()}ms: the answer was `
                     + `applied, signalingState=${pc.signalingState}`);
 
+                // The candidates must be added only now, after the offer/answer cycle has completed. Adding them
+                // together with the new credentials tears down the selected pair and breaks make-before-break, see
+                // SDPUtil.replaceIceCredentialsAndStripCandidates and https://issues.webrtc.org/issues/543082385
                 await this._addIceCandidatesToPeerConnection(candidates);
                 logger.info(`${this} ${ICE_RESTART_LOG_PREFIX} gen=${generation} t+${dt()}ms: `
                     + `${candidates.length} of the bridge's candidates were added`);
--- a/modules/xmpp/JingleSessionPC.ts
+++ b/modules/xmpp/JingleSessionPC.ts
@@ -1844,11 +1844,8 @@ export default class JingleSessionPC extends JingleSession {
                     return;
                 }
 
-                // The native peer connection is used on purpose: the SDP we patch is the current remote
-                // description, i.e. the one TraceablePeerConnection has already transformed, so it must not go
-                // through the transformation pipeline a second time.
                 const pc = this.peerconnection.peerconnection;
-                const remoteSdp = pc.currentRemoteDescription?.sdp;
+                const remoteSdp = this.peerconnection.remoteDescription?.sdp;
 
                 if (!remoteSdp) {
                     throw new Error('there is no current remote description');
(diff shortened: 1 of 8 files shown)


modules/xmpp/JingleSessionPC.ts
```diff
@@ -1445,6 +1713,201 @@ export default class JingleSessionPC extends JingleSession {
...
+
+                const patchedOffer = SDPUtil.replaceIceCredentialsAndStripCandidates(remoteSdp, ufrag, pwd);
+
+                await pc.setRemoteDescription({ sdp: patchedOffer, type: 'offer' });
+                logger.info(`${this} ${ICE_RESTART_LOG_PREFIX} gen=${generation} t+${dt()}ms: the patched offer `
+                    + `was applied (candidates stripped), signalingState=${pc.signalingState}`);
+
+                const answer = await this.peerconnection.createAnswer(this.mediaConstraints as RTCOfferOptions);
+
+                await this.peerconnection.setLocalDescription(answer);
```
jallamsetty1: can you call `this._renegotiate({ sdp: patchedOffer, type: 'offer' })` instead of calling sRD/cA/sLD on the native pc? This will also take care of signaling the SSRCs if the browser produces a different set of SSRCs after this operation.

jallamsetty1 (approved):

PR #3086, opened 2026-08-12
feat(RTC): Add camera pan/tilt/zoom control API

jallamsetty1:
## Description

Adds the lib-jitsi-meet foundation for camera pan/tilt/zoom (PTZ) support. This is the shared base that jitsi-meet builds both **local PTZ control** and **far-end camera control (FECC)** on top of — FECC signaling itself lives entirely in jitsi-meet (like `remote-control`), using generic endpoint messages plus the receive-side track API added here.

### Track control (JitsiLocalTrack)

- `getCameraControlCapabilities()` — pan/tilt/zoom `{ min, max, step }` ranges from the live camera `MediaStreamTrack.getCapabilities()`. A control is present only when the camera exposes it.
- `getCameraControlSettings()` — current pan/tilt/zoom values from `getSettings()`.
- `setCameraControl({ pan, tilt, zoom })` — applies the given values via `MediaStreamTrack.applyConstraints({ advanced: [...] })` **directly on the existing track**. No stream re-acquisition, no renegotiation — PTZ does not change the encoded stream or SSRCs. Only the provided controls are applied; the rest are untouched. The caller clamps to the reported ranges. Camera-tracks only (rejects otherwise).

### Capability probing at enumeration (RTCUtils)

- During `enumerateDevices`, each `videoinput` is probed best-effort via `InputDeviceInfo.getCapabilities()` for `pan`/`tilt`/`zoom` presence.
- Exposed through `RTC.getCameraPTZCapabilities(deviceId)` and `JitsiMediaDevices.getCameraPTZCapabilities(deviceId)` so the app can decide whether to offer camera control for a device **without first acquiring a stream from it**.
- Detection is a hint: `pan`/`tilt` may only appear once the `panTiltZoom` permission is granted, so the authoritative per-track capabilities come from `getCameraControlCapabilities()`.

### Permission

Requesting the (separate) `panTiltZoom` permission is left to the application: `createLocalTracks({ devices: ['video'], constraints: { video: { pan: true, tilt: true, zoom: true } } })`. `getConstraints` already passes those through unchanged, so no change is needed here. jitsi-meet drives the mid-call re-acquire + replaceTrack.

## Testing

- `modules/RTC/JitsiLocalTrack.spec.ts` (new): capabilities/settings reads, `setCameraControl` applies only provided controls via advanced constraints, no-op on empty, rejects for non-camera tracks, propagates underlying rejection.
- `modules/RTC/RTCUtils.spec.ts`: enumeration probe for full-PTZ, zoom-only, plain, no-`getCapabilities`, throwing, and non-video devices, plus unknown-device fallback.
- `npm test` 573/573; `npm run type-check` clean.

## Follow-up (jitsi-meet)

Consumes this API to add: local PTZ UI + mid-call permission re-acquire, and the FECC feature (absolute-value endpoint messages both directions, remote video-menu control, receive-side `setCameraControl`), gated to non-mobile Chromium and a config flag. Not part of this PR.

## Types of changes

## Checklist:

- [x] I have added tests to cover my changes.

### DIFF
--- a/JitsiMediaDevices.ts
+++ b/JitsiMediaDevices.ts
@@ -160,6 +160,18 @@ export default class JitsiMediaDevices extends Listenable {
     }
 
     /**
+     * Returns the pan/tilt/zoom support detected for the given camera device during the last device enumeration.
+     * Detection is best-effort and reflects what the browser exposes at enumeration time; pan/tilt may only be
+     * reported once the panTiltZoom permission has been granted for the camera.
+     *
+     * @param {string} deviceId - The id of the 'videoinput' device.
+     * @returns {{ pan: boolean, tilt: boolean, zoom: boolean }}
+     */
+    getCameraPTZCapabilities(deviceId: string): { pan: boolean; tilt: boolean; zoom: boolean; } {
+        return RTC.getCameraPTZCapabilities(deviceId);
+    }
+
+    /**
      * Returns true if changing the input (camera / microphone) or output
      * (audio) device is supported and false if not.
      * @param {string} [deviceType] - type of device to change. Default is
--- a/JitsiMediaDevices.ts
+++ b/JitsiMediaDevices.ts
@@ -2,6 +2,7 @@ import { JitsiMediaDevicesEvents } from './JitsiMediaDevicesEvents';
 import RTC from './modules/RTC/RTC';
 import browser from './modules/browser';
 import Listenable from './modules/util/Listenable';
+import { CameraControlType } from './service/RTC/CameraControlType';
 import { MediaType } from './service/RTC/MediaType';
 import { RTCEvents } from './service/RTC/RTCEvents';
 
@@ -165,9 +166,9 @@ export default class JitsiMediaDevices extends Listenable {
      * reported once the panTiltZoom permission has been granted for the camera.
      *
      * @param {string} deviceId - The id of the 'videoinput' device.
-     * @returns {{ pan: boolean, tilt: boolean, zoom: boolean }}
+     * @returns {Record<CameraControlType, boolean>}
      */
-    getCameraPTZCapabilities(deviceId: string): { pan: boolean; tilt: boolean; zoom: boolean; } {
+    getCameraPTZCapabilities(deviceId: string): Record<CameraControlType, boolean> {
         return RTC.getCameraPTZCapabilities(deviceId);
     }
 
--- a/JitsiMediaDevices.ts
+++ b/JitsiMediaDevices.ts
@@ -7,6 +7,9 @@ import { MediaType } from './service/RTC/MediaType';
 import { RTCEvents } from './service/RTC/RTCEvents';
 
 const AUDIO_PERMISSION_NAME = 'microphone' as PermissionName;
+// The key under which the camera pan/tilt/zoom (panTiltZoom) permission is tracked in the permissions cache
+// and reported in the PERMISSIONS_CHANGED event. It is queried as { name: 'camera', panTiltZoom: true }.
+const CAMERA_PTZ_PERMISSION_KEY = 'panTiltZoom';
 const PERMISSION_GRANTED_STATUS = 'granted';
 const VIDEO_PERMISSION_NAME = 'camera' as PermissionName;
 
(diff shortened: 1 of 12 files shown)


modules/browser/BrowserCapabilities.ts
```diff
@@ -188,6 +188,20 @@ export default class BrowserCapabilities extends BrowserDetection {
...
+     * navigator.mediaDevices.getSupportedConstraints(). This reflects API support only, not whether a connected
+     * camera has PTZ hardware or whether the panTiltZoom permission has been granted - the caller combines this with
+     * the per-camera capabilities and the panTiltZoom permission for those.
+     *
+     * @returns {boolean}
+     */
+    supportsCameraPtz(): boolean {
+        const supported = navigator.mediaDevices?.getSupportedConstraints?.() ?? {};
+
+        return Boolean((supported as any).pan && (supported as any).tilt && (supported as any).zoom);
```
hristoterezov, line 202: Don't we want to support the case where one of those is supported but the other not. for example: tilt is supported but zoom not?

modules/RTC/JitsiLocalTrack.ts
```diff
@@ -342,6 +353,21 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+    /**
+     * Returns the camera MediaStreamTrack that pan/tilt/zoom must be read from and applied to. When a stream effect
+     * (e.g. virtual background) is active, {@link this.track} is the effect's output track (a canvas/insertable-streams
+     * capture) which has no pan/tilt/zoom; the actual camera is the source that feeds the effect and lives in
+     * {@link this._originalStream}. Panning/zooming that source flows through to the composited output.
+     *
+     * @private
+     * @returns {Optional<MediaStreamTrack>} - The camera source track, or undefined if unavailable.
+     */
+    private _getCameraSourceTrack(): Optional<MediaStreamTrack> {
```
hristoterezov, line 365: hmmm I vaguely remember that back in the days there was a similar method. did we remove it at some point?

modules/RTC/JitsiLocalTrack.ts
```diff
@@ -1162,4 +1188,102 @@ export default class JitsiLocalTrack extends JitsiTrack {
...
+
+        const cameraTrack = this._getCameraSourceTrack();
+
+        if (!cameraTrack) {
+            return Promise.reject(new Error('No camera track to apply the control to'));
+        }
+
+        const control: ICameraControls = {};
+
+        for (const key of Object.values(CameraControlType)) {
```
hristoterezov, line 1275: I'm wondering why not iterate over `controls` directly?

modules/RTC/RTCUtils.js
```diff
@@ -85,6 +86,47 @@ if (isAudioOutputDeviceChangeAvailable) {
...
+ *
+ * @param {MediaDeviceInfo[]} devices - The enumerated devices.
+ * @returns {void}
+ */
+function probeCameraPTZSupport(devices) {
+    const support = new Map();
+
+    for (const device of devices) {
+        if (device.kind !== 'videoinput') {
+            continue; // eslint-disable-line no-continue
```
hristoterezov, line 109: why disable the eslint rule?

JitsiMediaDevices.ts
```diff
@@ -55,8 +59,7 @@ export default class JitsiMediaDevices extends Listenable {
      */
     private _handlePermissionsChange(permissions: { [key: string]: boolean; }): void {
         const hasPermissionsChanged
-            = [ MediaType.AUDIO, MediaType.VIDEO ]
-                .some(type => type in permissions && permissions[type] !== this._permissions[type]);
+            = Object.keys(permissions).some(type => permissions[type] !== this._permissions[type]);
```
hristoterezov, line 62: we don't need to go have the audio/video type restriction anymore?

modules/browser/BrowserCapabilities.ts
```diff
@@ -188,6 +188,20 @@ export default class BrowserCapabilities extends BrowserDetection {
...
 
+    /**
+     * Checks whether the browser implements the MediaCapture camera pan/tilt/zoom (PTZ) constraints, via
+     * navigator.mediaDevices.getSupportedConstraints(). This reflects API support only, not whether a connected
+     * camera has PTZ hardware or whether the panTiltZoom permission has been granted - the caller combines this with
+     * the per-camera capabilities and the panTiltZoom permission for those.
+     *
+     * @returns {boolean}
+     */
+    supportsCameraPtz(): boolean {
```
hristoterezov, line 199: where do we use this one?

PR #3091, opened 2026-08-18
feat(xmpp): handle the client-requirements IQ from jicofo.

bgrozev:
Adds support for the `jitsi:client-requirements` IQ, which jicofo uses to signal that this client does not advertise the capabilities that the deployment requires.

- Advertises `http://jitsi.org/client-requirements-1`.
- Handles the IQ and fires a new `CLIENT_REQUIREMENTS_NOT_MET` conference event.
- Advertises the version of the library in presence, so that the server can log it.

A client which the server does not invite stays in the conference. The library does not fire `CONFERENCE_FAILED`, so the features which do not need a media session, for example chat, keep working.

Set of PRs, all on a branch named `client-capability-requirements`:

- jitsi-xmpp-extensions: https://github.com/jitsi/jitsi-xmpp-extensions/pull/149
- jicofo: https://github.com/jitsi/jicofo/pull/1303
- lib-jitsi-meet: https://github.com/jitsi/lib-jitsi-meet/pull/3091
- jitsi-meet: https://github.com/jitsi/jitsi-meet/pull/17730


modules/xmpp/ChatRoom.ts
```diff
@@ -2270,6 +2278,49 @@ export default class ChatRoom extends Listenable {
...
+        if (!requirements) {
+            return;
+        }
+
+        const action = getAttribute(requirements, 'action');
+        const features = findAll(requirements, ':scope>missing-feature').map(feature => {
+            return {
+                details: getAttribute(feature, 'details') ?? undefined,
+                feature: getAttribute(feature, 'var'),
+                level: getAttribute(feature, 'level'),
```
damencho: `getAttribute()` (from `modules/util/XMLUtils.ts`) returns `string | null`, not `string | undefined`. The optional fields all apply `?? undefined` to normalise to `undefined` when absent, but the three required fields do not:

```typescript
const action = getAttribute(requirements, 'action');   // string | null
// ...
feature: getAttribute(feature, 'var'),   // string | null
level: getAttribute(feature, 'level'),   // string | null
```

A malformed or adversarial IQ with a missing `action` attribute will emit `{ action: null, ... }` to the event handler. Consumers that test `action === 'reject'` will silently fail (null !== 'reject') instead of ignoring the stanza. The same applies to `feature` (the capability URI) and `level` (enforcement level).

Consider:
```typescript
const action = getAttribute(requirements, 'action');
if (!action) { return; }   // or throw / log
// ...
feature: getAttribute(feature, 'var') ?? undefined,
level: getAttribute(feature, 'level') ?? undefined,
```
or adjust the `ClientRequirementsReceivedEvent` type to reflect `string | null` explicitly and guard at the consumer.

modules/xmpp/ChatRoom.clientRequirements.spec.ts
```diff
@@ -0,0 +1,132 @@
...
+
+    it('ignores an IQ with no client-requirements element', () => {
+        room.onClientRequirements($iq({
+            from: FOCUS_JID,
+            type: 'set'
+        }).tree());
+
+        expect(emitted.length).toBe(0);
+    });
+});
```
damencho, line 194: The test suite covers well-formed IQs and sender filtering, but not an IQ where a required attribute (`action`, `var`, or `level`) is absent or null. Since `getAttribute()` returns null for missing attributes and the handler currently propagates that null into the emitted event, a test like:

```typescript
it('handles an IQ with a missing action attribute gracefully', () => {
    const iq = buildIq('focus@example.com', [{ 'var': 'http://jitsi.org/ssrc-rewriting-1', level: 'required' }]);
    // patch: remove action attr from the client-requirements element
    iq.querySelector('client-requirements')!.removeAttribute('action');
    room.onClientRequirements(iq);
    // assert: either not emitted, or emitted with action === null (whichever the spec says)
});
```

would pin the intended behavior and catch a future guard if it's added (or removed).

modules/xmpp/ChatRoom.ts
```diff
@@ -2270,6 +2278,49 @@ export default class ChatRoom extends Listenable {
...
+     * session (e.g. chat).
+     *
+     * @param iq The received iq.
+     * @internal
+     */
+    onClientRequirements(iq: Element): void {
+        const from = iq.getAttribute('from');
+
+        if (from !== this.focusMucJid) {
+            logger.warn('Ignored client requirements from non focus peer');
```
damencho: The rejection log omits the sender JID:

```typescript
logger.warn('Ignored client requirements from non focus peer');
```

This follows the pre-existing pattern for `onMute` / `onAudioMuted` / `onVideoMuted` in this file, but the same PR adds an explicit `from !== room.roomjid` guard in `strophe.emuc.ts` where the sender JID is the key diagnostic. Including it here would make rejections actionable without needing packet traces:

```typescript
logger.warn(`Ignored client requirements from non focus peer: ${from}`);
```

damencho (commented):
Three findings from the review. The first two are genuinely new; the third follows a pre-existing pattern in the codebase but is worth fixing for consistency with the new focus-JID guard added in this same PR.

damencho (approved):

modules/xmpp/ChatRoom.ts
```diff
@@ -72,6 +73,56 @@ export interface IChatRoomOptions {
     statsId?: string;
 }
 
+/**
+ * The action of a client-requirements IQ: the client is not invited to the conference.
+ */
+export const CLIENT_REQUIREMENTS_ACTION_REJECT = 'reject';
```
saghul: Does it not make sense to make these 2 an enum?

modules/xmpp/strophe.emuc.ts
```diff
@@ -256,6 +258,12 @@ export default class MucConnectionPlugin extends ConnectionPluginListenable {
             return true;
         }
 
+        if (from !== room.roomjid) {
```
saghul, line 261: Unrelated change?

modules/xmpp/strophe.emuc.ts
```diff
@@ -273,4 +281,23 @@ export default class MucConnectionPlugin extends ConnectionPluginListenable {
...
+     * A client-requirements IQ is received, pass it to the room.
+     *
+     * @param iq The received iq.
+     * @returns {boolean}
+     */
+    onClientRequirements(iq: Element): boolean {
+        const from = iq.getAttribute('from');
+        const room = this.rooms[Strophe.getBareJidFromJid(from)];
+
+        if (!room) {
```
saghul, line 295: How would this happen?

saghul (approved):

modules/xmpp/strophe.emuc.ts
bgrozev: The strophe.emuc listener fires for any IQ to our full JID, so if someone not in the room routes an IQ to us (or if we hit the race of the room being destroyed before we received the IQ)

PR #3093, opened 2026-08-20
fix(RTC): avoid invalid maxBitrate=0 on full-SVC encodings

damencho:
## Summary
- `calculateEncodingsBitrates()` set `maxBitrate` to the literal `0` on the non-primary encodings whenever a video sender runs in full-SVC / single-stream mode (AV1, VP9-SVC). Per the WebRTC spec, `RTCRtpSender.setParameters()` throws a `RangeError` if any encoding's `maxBitrate` is present and `<= 0`, regardless of that encoding's `active` state.
- This caused an unhandled promise rejection out of `_updateVideoSenderEncodings` in `TraceablePeerConnection.ts`: `RangeError: Failed to execute 'setParameters' on 'RTCRtpSender': maxBitrate must be greater than 0`.
- The sibling calculations (`calculateEncodingsScaleFactor`, `calculateEncodingsScalabilityMode`) already use `undefined` for those same non-primary indices; `calculateEncodingsBitrates` was the outlier. This change brings it in line with the same pattern.

## Test plan
- [x] Updated `TPCUtils.spec.ts` expectations for the affected AV1/VP9/H.264 full-SVC test cases from `toBe(0)` to `toBeUndefined()` (23 cases / 46 assertions).
- [x] `npx karma start karma.conf.js --single-run --browsers ChromeHeadless` — 576/576 passing (one pre-existing, unrelated `afterAll` failure due to a missing generated `./version` module in this checkout).

### DIFF
--- a/modules/RTC/TPCUtils.spec.ts
+++ b/modules/RTC/TPCUtils.spec.ts
@@ -245,8 +245,8 @@ describe('TPCUtils', () => {
 
                 maxBitrates = tpcUtils.calculateEncodingsBitrates(track, codec, height);
                 expect(maxBitrates[0]).toBe(1000000);
-                expect(maxBitrates[1]).toBe(0);
-                expect(maxBitrates[2]).toBe(0);
+                expect(maxBitrates[1]).toBeUndefined();
+                expect(maxBitrates[2]).toBeUndefined();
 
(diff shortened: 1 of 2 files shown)


saghul (approved):

damencho:
Unrelated error.

PR #3094, opened 2026-08-24
feat(xmpp): add participant chat shadow-ban support

Alok-work23:
## Description

Adds chat shadow-ban support to `lib-jitsi-meet`.

The change introduces the conference and XMPP-layer support required for Jitsi Meet to enable or disable chat shadow-ban for a participant. The participant remains in the conference, while the shadow-ban request is sent to the server for enforcement by the custom Prosody module.

Changes are limited to the conference and XMPP chat-room layers:

- Added `setChatShadowBan()` to `JitsiConference`.
- Added `setChatShadowBan()` to `ChatRoom`.
- `JitsiConference` resolves the participant ID to the participant JID.
- `ChatRoom` sends a custom `<shadow-ban>` XMPP IQ containing the participant JID and enabled state.
- Added XMPP error handling for the shadow-ban operation.

## Related Issue

N.A.

## Motivation and Context

The Jitsi Meet frontend needs a way to request that a participant's chat messages be shadow-banned without removing the participant from the conference.

This change provides the `lib-jitsi-meet` API and XMPP plumbing required by the Jitsi Meet moderator action and the custom Prosody shadow-ban implementation.

## How Has This Been Tested?

The following checks were completed successfully:

- `npx tsc --noEmit`
- `npm run lint`

Runtime end-to-end verification of the complete shadow-ban flow is still being verified, including the final Prosody-side handling of the custom XMPP IQ and message filtering.

Not applicable. This PR contains no UI changes.

## Types of changes

## Checklist:


damencho (commented):
Reviewed together with jitsi/jitsi-meet#17743, where the bulk of the comments live.

## Blocker: the JID sent to the server is the room's own JID

`JitsiConference.setChatShadowBan()` passes `participant.getJid()`. That is the MUC **occupant** JID (`room@conference.domain/abc123`), not the participant's real JID — see `onMemberJoined()` at `JitsiConference.ts:3827`, where the first argument is the occupant JID and the real JID arrives separately as `fullJid` and is stored via `setConnectionJid()`.

`ChatRoom.setChatShadowBan()` then applies `Strophe.getBareJidFromJid()` to it, which yields `room@conference.domain` — the room, not the participant. Prosody stores that in `room.shadow_banned_jids`, while its `message/bare` handler compares against `jid.bare(stanza.attr.from)`, the sender's real bare JID (`user@domain`). The two can never match, so no message is ever dropped and the feature is a no-op end to end.

If the IQ approach is kept, `participant.getConnectionJid()` is the value you want (real JIDs are disclosed — `muc_room_default_public_jids = true` in the config example). But see below, because I don't think the IQ is the right transport any more.

## The support-detection layer is dead code, and already exists on master

`XMPPEvents.MUC_SHADOW_BAN_SUPPORTED_CHANGED` is emitted but never forwarded to a `JitsiConferenceEvents` member, so nothing outside the lib can listen — compare `JitsiConferenceEventManager.ts:343-346`, which forwards both `MUC_VISITORS_SUPPORTED_CHANGED` and `MUC_MESSAGE_MODERATION_SUPPORTED_CHANGED`. In jitsi-meet the consequence is that `ParticipantContextMenu` reads `conference?.isShadowBanSupported?.()` from inside a `useSelector`, which only ever resolves because some unrelated dispatch happens to follow disco#info.

More to the point: master already has `isMessageModerationSupported()` (`JitsiConference.ts:5005`) fed from `muc#roominfo_messageModerationEnabled`, forwarded as `MESSAGE_MODERATION_SUPPORTED_CHANGED`, and landed in redux as `state['features/chat'].messageModerationSupported`. If shadow-ban moves into `mod_muc_message_moderation` as suggested on the jitsi-meet PR, this whole layer — the new disco feature parse, the new `XMPPEvents` member, `shadowBanSupported`, `isShadowBanSupported()` — can simply be deleted and the existing flag reused.

## API shape

Following on from that: `mod_muc_message_moderation` acts on the `muc-occupant-groupchat` hook, so the client side already has a matching pattern in `moderateMessage()` (`JitsiConference.ts:3237` → `ChatRoom.ts:1460`), which sends a groupchat stanza rather than an IQ. Modelling `setChatShadowBan()` on that gets you the occupant resolved server-side for free and drops the JID problem above entirely.

## Smaller things

- `.t(enabled.toString())` puts the boolean in element text, and the Prosody side treats anything that is not exactly `"true"` as *unban* — so an empty or malformed body silently means the opposite. An attribute (`enabled="true"`) is both the jitsi convention and safer here.
- The trailing `.up()` on that stanza builder is a no-op.
- The error callback only calls `handleStropheError`, which logs. A `forbidden` reply is therefore invisible to the moderator who clicked; something needs to reach the UI.
- Double blank line left behind after the new method in `JitsiConference.ts`.

`npm run lint` (eslint + `tsc --noEmit`) is clean. Karma not run.

PR #3095, opened 2026-08-27
fix(sdp) keep low profile codecs that omit the profile parameter

damencho:
### What

`SDPUtil.stripCodec(mLine, codec, /* highProfile */ true)` is called for VP9 and H264 on every P2P description (`TPCUtils.mungeCodecOrder`) to drop the high profile payload types. The filter treated a *missing* profile parameter as high profile:

```js
codec.toLowerCase() === CodecMimeType.VP9
    ? !item.config.includes('profile-id=0')
    : !item.config.includes('profile-level-id=42');
```

Firefox never signals `profile-id` for VP9 — it emits `a=fmtp:98 max-fs=12288;max-fr=60` — and profile 0 is the default, so its only VP9 payload type was stripped from its own local description. The same applies to H264, where `profile-level-id` is optional and defaults to the baseline profile.

### Why it matters

Firefox 151+ can encode VP9 ([bugzilla 1633876](https://bugzilla.mozilla.org/show_bug.cgi?id=1633876)), so the P2P codec order now actually resolves to VP9 there. The transceiver keeps VP9 as the send codec while the munged answer no longer offers it, and Firefox ends up sending a payload type the peer never negotiated. Captured from a Firefox 154 ↔ Chrome 151 P2P call in the torture tests:

* Firefox answer (raw): `m=video ... 98 99 96 97 108 109 102 103 45 46`, `a=fmtp:98 max-fs=12288;max-fr=60`
* Firefox answer (post munge, set as local description): `m=video ... 96 108 102 45 97 109 103 46` — no VP9
* Firefox `outbound-rtp`: `codecId → video/VP9 payloadType 98`, framesEncoded 470, 505 packets sent
* Chrome `inbound-rtp` for the same ssrc: `packetsReceived 500, packetsLost 0, framesReceived 0, framesDecoded 0, keyFramesDecoded 0, pliCount 0`, no `codecId`

The receiver gets every packet, cannot map the payload type, decodes nothing and falls back to the avatar. It shows up in the torture tests as `Mute › mute before join and screen share after in p2p` timing out on `waitForRemoteVideo`, but it breaks Firefox → Chrome video in P2P in general.

### Fix

Strip a payload type only when it actually declares a profile other than the low one. A payload type with no profile parameter is now treated the same way as one with no `fmtp` line at all, which the method already preserved.

The H264 branch is included because it is the same rule and the same latent bug; browsers do send `profile-level-id` today, but endpoints are not required to.

### Tests

`modules/sdp/SDPUtil.spec.ts` gets a `strip high profile video codec` block over a new `multiProfileVideoSdp` fixture that mixes VP9 without `profile-id`, VP9 `profile-id=2`, H264 without `profile-level-id`, H264 `profile-level-id=42e01f` and H264 `profile-level-id=64001f`. Without the fix:

```
SDPUtil strip high profile video codec should only remove the payload types that signal a high profile FAILED
	Expected [ 45, 46, 96, 97, 108, 109 ] to contain 98.
	Expected [ 45, 46, 96, 97, 108, 109 ] to contain 99.
	Expected [ 45, 46, 96, 97, 108, 109 ] to contain 102.
	Expected [ 45, 46, 96, 97, 108, 109 ] to contain 103.
TOTAL: 1 FAILED, 579 SUCCESS
```

With the fix `TOTAL: 580 SUCCESS`, `eslint` and `tsc --noEmit` are clean.

### DIFF
--- a/modules/sdp/SDPUtil.spec.ts
+++ b/modules/sdp/SDPUtil.spec.ts
@@ -52,6 +52,51 @@ describe('SDPUtil', () => {
         });
     });
 
+    describe('strip high profile video codec', () => {
+        it('should only remove the payload types that signal a high profile', () => {
+            const sdp = SampleSdpStrings.multiProfileVideoSdp;
+            const videoMLine = sdp.media.find(m => m.type === 'video');
+
+            SDPUtil.stripCodec(videoMLine, 'VP9', true /* high profile */);
+            SDPUtil.stripCodec(videoMLine, 'H264', true /* high profile */);
+
+            const payloadTypes = videoMLine.payloads.split(' ').map(Number);
+
+            // Firefox does not signal 'profile-id' for VP9 which implies profile 0, therefore the payload type and
+            // its RTX have to be kept, otherwise Firefox ends up sending a codec that was not negotiated.
+            expect(payloadTypes).toContain(98);
+            expect(payloadTypes).toContain(99);
+
+            // Same for H264 payload types that signal the baseline profile or no profile at all.
+            expect(payloadTypes).toContain(108);
+            expect(payloadTypes).toContain(109);
+            expect(payloadTypes).toContain(102);
+            expect(payloadTypes).toContain(103);
+
+            // The high profile payload types and their RTX are removed.
+            expect(payloadTypes).not.toContain(100);
+            expect(payloadTypes).not.toContain(101);
+            expect(payloadTypes).not.toContain(114);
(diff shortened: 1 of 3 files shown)


damencho:
Test coverage for this on the torture side: https://github.com/jitsi/jitsi-meet/pull/17759 (same branch name, so its PR test job builds this branch).

hristoterezov (approved):

saghul (approved):

PR #3096, opened 2026-08-28
fix(ScreenObtainer): remove the legacy Electron desktop picker path

FuriousTurtle:
## Description

Removes the legacy Electron desktop picker path so Electron always uses `getDisplayMedia`, like every other supported platform.

- `_createObtainStreamMethod()` no longer special-cases `browser.isElectron()`, so Electron falls through to `_obtainScreenFromGetDisplayMedia`
- Removed `_obtainScreenOnElectron()`, including the `window.JitsiMeetScreenObtainer.openDesktopPicker` flow and its `getUserMedia` constraint building
- Removed the `_electronSkipDisplayMedia` latch that selected between the two paths
- Removed `ILegacyVideoConstraints` and `IAudioConstraints`, now unused
- Removed `ELECTRON_DESKTOP_PICKER_ERROR` and `ELECTRON_DESKTOP_PICKER_NOT_FOUND` from `JitsiTrackErrors`, along with their messages in `JitsiTrackError` and their assertions in `JitsiTrackErrors.spec.ts`, since nothing can raise them any more

## Related Issue

Follow-up to #2605 / #2597, which added the gDM path and left this note in the code:

```
// @ts-ignore TODO: legacy flow, remove after the Electron SDK supporting gDM has been out for a while.
```

Discussion that led to this PR: jitsi/jitsi-meet-electron-sdk#512

## Motivation and Context

gDM support landed in `@jitsi/electron-sdk` 7.0.0 (Feb 2025), so the compatibility window the TODO asked for has passed.

Beyond the cleanup, the legacy path is actively broken on Wayland, which is now the default session on Ubuntu, Fedora and RHEL.

There, `desktopCapturer.getSources()` cannot enumerate silently. Every call goes through xdg-desktop-portal and shows a consent dialog. The user picks their source in the portal, then the Jitsi picker opens and asks them to pick again from a list containing only what they already chose. `DesktopPicker` also refreshes its thumbnails on a 2s interval, and each refresh triggers another portal prompt, so the OS dialog keeps reopening for as long as the picker is open. Some of our users concluded screen sharing was broken and stopped using it.

On the gDM path none of that happens. The portal is the only picker, and the selected source shares immediately, which is the same flow browser users already get.

Note the two removed error constants were exported, so this is technically breaking for anyone referencing them. They only ever described failures of the picker flow being removed here, so they cannot be raised any more. Happy to keep them as deprecated no-ops instead if you would rather avoid the break.

## How Has This Been Tested?

`tsc --noEmit` clean, `eslint` clean, 580/580 existing unit tests pass.

Verified end to end with a full local build: this branch packed and installed into `jitsi-meet` master (98de6219c, which pins lib-jitsi-meet at 48c54c20, the commit this branch is based on), then `make compile`, and the resulting `app.bundle.min.js`, `external_api.min.js`, `chunks/` and `lib-jitsi-meet.min.js` deployed together.

Client: Electron 44.0.0 with `@jitsi/electron-sdk` 10.0.4, on Ubuntu 26.04, GNOME Shell 50.1, Wayland, xdg-desktop-portal 1.21.1, PipeWire 1.6.2.

Result: clicking share opens the xdg-desktop-portal dialog once, and the selected source is shared immediately. No second picker, no repeated portal prompts, and re-sharing opens a fresh portal dialog so a different source can be chosen.

For comparison, on a released build (`jitsi-meet-web` 1.0.8384-1) with the legacy path active, both pickers appear, the source has to be selected twice, and the portal dialog reopens every 2 seconds while the Jitsi picker is open. Setting `electronUseGetDisplayMedia` on that same build produces the same good behaviour as this branch, which is what this PR makes unconditional.

Browser clients were unaffected, as expected, since they already used this code path.

N/A

## Types of changes

- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:

### DIFF
--- a/JitsiTrackError.ts
+++ b/JitsiTrackError.ts
@@ -29,8 +29,6 @@ const TRACK_ERROR_TO_MESSAGE_MAP: { [key: string]: string; } = {
     [JitsiTrackErrors.SCREENSHARING_USER_CANCELED]: 'User canceled screen sharing prompt',
     [JitsiTrackErrors.SCREENSHARING_GENERIC_ERROR]: 'Unknown error from screensharing',
     [JitsiTrackErrors.SCREENSHARING_NOT_SUPPORTED_ERROR]: 'Not supported',
-    [JitsiTrackErrors.ELECTRON_DESKTOP_PICKER_ERROR]: 'Unknown error from desktop picker',
-    [JitsiTrackErrors.ELECTRON_DESKTOP_PICKER_NOT_FOUND]: 'Failed to detect desktop picker',
     [JitsiTrackErrors.GENERAL]: 'Generic getUserMedia error',
     [JitsiTrackErrors.PERMISSION_DENIED]: 'User denied permission to use device(s): ',
     [JitsiTrackErrors.NOT_FOUND]: 'Requested device(s) was/were not found: ',
--- a/JitsiTrackErrors.spec.ts
+++ b/JitsiTrackErrors.spec.ts
@@ -5,8 +5,6 @@ import * as exported from "./JitsiTrackErrors";
 describe( "/JitsiTrackErrors members", () => {
     const {
         CONSTRAINT_FAILED,
-        ELECTRON_DESKTOP_PICKER_ERROR,
-        ELECTRON_DESKTOP_PICKER_NOT_FOUND,
         GENERAL,
         NOT_FOUND,
         PERMISSION_DENIED,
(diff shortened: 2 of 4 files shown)


saghul (approved):

FuriousTurtle:

I signed the individual one

jallamsetty1:
> > Hi, thanks for your contribution! If you haven't already done so, could you please make sure you sign our CLA (https://jitsi.org/icla for individuals and https://jitsi.org/ccla for corporations)? We would unfortunately be unable to merge your patch unless we have that piece :(.
> 
> I signed the individual one

Thank you for the PR. Can you share the email ID you used for signing the CLA? I am unable to pull up the signed copy to verify.

FuriousTurtle:
I signed as mourier.tony@outlook.com

damencho:
The commits here are not coming from that email. You need to update one or another.

FuriousTurtle:
I re-signed with the corresponding mail, it should be good now

PR #3105, opened 2026-09-04
fix: Tag the ICE restart answer on React Native, where lastElementChild does not exist

JonathanLennox:
## Problem

On React Native, jitsi-meet runs strophe on top of `@xmldom/xmldom` (see `react/features/mobile/polyfills/browser.js`), which implements `firstChild`/`lastChild`/`childNodes` but none of the element-traversal accessors: `firstElementChild`, `lastElementChild`, `childElementCount`, `nextElementSibling`, `previousElementSibling` are all `undefined` on 0.8.13.

`_sendIceRestartTransportInfo()` (#3083) locates the `<transport>` it just appended via `transportInfo.node?.lastElementChild`. On mobile that is `undefined`, so it logs "could not tag the transport with the ICE generation" and sends the transport-info **without** `ice-generation`. The bridge cannot tell that answer from an ordinary transport update, applies it to the established Agent, and the in-place restart is abandoned when the pending Agent times out. Observed on a `mobile-26.3` build against a bridge with in-place restart: two restarts in a row, both abandoned after 20 s, while jicofo had relayed the tagged offer correctly. (jitsi/jicofo#1313 now tags on the endpoint's behalf as a fallback; this fixes the endpoint side.)

## Fix

- `XMLUtils`: add `getFirstChildElement(node, tagName?)` / `getLastChildElement(node, tagName?)`, which walk `childNodes` and therefore work on both DOMs.
- Use them in `_sendIceRestartTransportInfo`, and at the two other `firstElementChild` call sites: the audio-translation error condition in `JitsiConference`, and `_handleResumeFailed` in stream management.

## Tests

- Unit tests for the new helpers, including one against a bare `childNodes`-only object.
- `test-setup-polyfill.ts` (the React Native simulation used by `npm run test:polyfill`) now also deletes the accessors xmldom lacks from `Element`/`Document`/`DocumentFragment`, so code that relies on them fails there the way it fails on a device. Without the fix, the existing test "signals the new local ICE credentials back tagged with the same generation" fails in that mode with `Expected null to be '7'`; with it, both suites pass 610/610.

### DIFF
--- a/JitsiConference.ts
+++ b/JitsiConference.ts
@@ -45,7 +45,7 @@ import Listenable from './modules/util/Listenable';
 import { isValidNumber, safeSubtract } from './modules/util/MathUtil';
 import RandomUtil from './modules/util/RandomUtil';
 import { getJitterDelay } from './modules/util/Retry';
-import { findAll, findFirst, getAttribute } from './modules/util/XMLUtils';
+import { findAll, findFirst, getAttribute, getFirstChildElement } from './modules/util/XMLUtils';
 import ComponentsVersions from './modules/version/ComponentsVersions';
 import JitsiVideoSIPGWSession from './modules/videosipgw/JitsiVideoSIPGWSession';
 import VideoSIPGW from './modules/videosipgw/VideoSIPGW';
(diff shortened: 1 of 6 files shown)


bgrozev (approved):

saghul (approved):

hristoterezov (approved):
LGTM!

PR #3112, opened 2026-09-14
fix: reduce ICE candidate logging

jallamsetty1:
Two changes to the ICE candidate log output.

Drops the `ip` and `rel-addr` attributes from the `transport-info` candidate log line, and the trailing element dump from the "No ICE candidates to add" error. All other candidate fields are unchanged.

Removes two debug lines that fired per candidate and carried no useful information - `SDPUtil.candidateToJingle`'s `not translating "..."` (three lines per candidate, for `ufrag`, `network-id` and `network-cost`) and `_sendIceCandidates count`.


damencho (approved):

jallamsetty1:
Jenkins test this please.

